PR #6, opened 2020-02-02
add new board - nRF52840 MDK USB Dongle

xiongyihui:
support a new nRF52840 board - nRF52840 MDK USB Dongle #3

The board has a builtin bootloader, so the start address of Tock OS is changed to `0x1000`

- [ ] Tests pass

### DIFF
--- a/patches/tock/05-nrf52840-mdk-usb-dongle.patch
+++ b/patches/tock/05-nrf52840-mdk-usb-dongle.patch
@@ -1,0 +1,381 @@
+diff --git a/boards/nordic/nrf52840_mdk_usb_dongle/Cargo.toml b/boards/nordic/nrf52840_mdk_usb_dongle/Cargo.toml
+new file mode 100644
+index 00000000..5b633ffb
+--- /dev/null
++++ b/boards/nordic/nrf52840_mdk_usb_dongle/Cargo.toml
+@@ -0,0 +1,26 @@
++[package]
++name = "nrf52840_dongle"
++version = "0.1.0"
++authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
++build = "build.rs"
++edition = "2018"
++
++[profile.dev]
++panic = "abort"
++lto = false
++opt-level = "z"
++debug = true
++
++[profile.release]
++panic = "abort"
++lto = true
++opt-level = "z"
++debug = true
++
++[dependencies]
++components = { path = "../../components" }
++cortexm4 = { path = "../../../arch/cortex-m4" }
++capsules = { path = "../../../capsules" }
++kernel = { path = "../../../kernel" }
++nrf52840 = { path = "../../../chips/nrf52840" }
++nrf52dk_base = { path = "../nrf52dk_base" }
+diff --git a/boards/nordic/nrf52840_mdk_usb_dongle/Makefile b/boards/nordic/nrf52840_mdk_usb_dongle/Makefile
+new file mode 100644
+index 00000000..e8359b93
+--- /dev/null
++++ b/boards/nordic/nrf52840_mdk_usb_dongle/Makefile
+@@ -0,0 +1,36 @@
++# Makefile for building the tock kernel for the nRF development kit
++
++TOCK_ARCH=cortex-m4
++TARGET=thumbv7em-none-eabi
++PLATFORM=nrf52840_dongle
++
++include ../../Makefile.common
++
++TOCKLOADER=tockloader
++
++# Where in the nrf52 flash to load the kernel with `tockloader`
++KERNEL_ADDRESS=0x00000
++
++# Upload programs over uart with tockloader
++ifdef PORT
++  TOCKLOADER_GENERAL_FLAGS += --port $(PORT)
++endif
++
++TOCKLOADER_JTAG_FLAGS = --jlink --board nrf52dk
++
++# Upload the kernel
++.PHONY: flash
++flash: target/$(TARGET)/release/$(PLATFORM).bin target/$(TARGET)/release/$(PLATFORM).hex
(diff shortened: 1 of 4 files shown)


xiongyihui:
@googlebot I signed it!

patches/tock/05-nrf52840-mdk-usb-dongle.patch
```diff
@@ -0,0 +1,397 @@
...
++++ b/arch/cortex-m/src/scb.rs
+@@ -54,3 +54,8 @@ pub unsafe fn reset() {
+     let reset = (0x5FA << 16) | (aircr & (0x7 << 8)) | (1 << 2);
+     SCB.aircr.set(reset);
+ }
++
++/// Software reset using the ARM System Control Block
++pub unsafe fn set_vector_table_offset(offset: u32) {
++    SCB.vtor.set(offset);
++}
```
gendx: This will definitely need to be discussed upstream.

patches/tock/05-nrf52840-mdk-usb-dongle.patch
```diff
@@ -0,0 +1,397 @@
...
++    }
++}
++
++#[cfg(not(test))]
++#[no_mangle]
++#[panic_handler]
++/// Panic handler
++pub unsafe extern "C" fn panic_fmt(pi: &PanicInfo) -> ! {
++    // The nRF52840 Dongle LEDs (see back of board)
++    const LED1_PIN: Pin = Pin::P0_06;
```
gendx: This pin doesn't seem to match the one in `../main.rs`

patches/tock/05-nrf52840-mdk-usb-dongle.patch
```diff
@@ -0,0 +1,397 @@
...
++
++use cortexm4::scb;
++use kernel::component::Component;
++#[allow(unused_imports)]
++use kernel::{debug, debug_gpio, debug_verbose, static_init};
++use nrf52840::gpio::Pin;
++use nrf52dk_base::{SpiPins, UartChannel, UartPins};
++
++// The nRF52840 MDK USB Dongle LEDs
++const LED1_PIN: Pin = Pin::P0_08;           // not connected to a LED
```
gendx: If it's not connected to a LED, I think the best is not to mention it at all here.

patches/tock/05-nrf52840-mdk-usb-dongle.patch
```diff
@@ -0,0 +1,397 @@
...
++            capsules::button::GpioMode::LowWhenPressed,
++            kernel::hil::gpio::FloatingState::PullUp
++        )),
++    );
++
++    let led = components::led::LedsComponent::new().finalize(components::led_component_helper!(
++        (
++            &nrf52840::gpio::PORT[LED1_PIN],
++            capsules::led::ActivationMode::ActiveLow
++        ),
```
gendx: Same comment as above. Does everything work when only 3 LED pins are defined?

xiongyihui:
@gendx Thanks for the comments. I also think adding the board to tock is better, will try that.

xiongyihui:
To experiment with OpenSK on the nRF52840 MDK USB Dongle with a builtin bootloader ,  some changes are required:

+ change the start address of tock os to `0x1000` at `OpenSK/third_party/tock/boards/nordic/nrf52840_dongle/layout.ld`
+ relocate interrupt vector table (SCB->VTOR = 0x1000)
+ change the pins of LEDs and buttons

markushx:
I have applied the SCB.vtor as well as the linker script change for the original nordic nrf52840-dongle, because i wanted to install without Segger JLink, but with nRF Connect programmer. 

Thus I created a hex file, out of the elf file:
`arm-none-eabi-objcopy -O ihex  ./third_party/tock/boards/nordic/nrf52840_dongle/target/thumbv7em-none-eabi/release/nrf52840_dongle.elf ./third_party/tock/boards/nordic/nrf52840_dongle/target/thumbv7em-none-eabi/release/nrf52840_dongle.hex`.

Unfortunately it seems the device does not boot-up.

What is the expected behaviour after boot-up? Should the device show up as a ttyUSB?

gendx:
> I have applied the SCB.vtor as well as the linker script change for the original nordic nrf52840-dongle, because i wanted to install without Segger JLink, but with nRF Connect programmer.
> 
> Thus I created a hex file, out of the elf file:
> `arm-none-eabi-objcopy -O ihex ./third_party/tock/boards/nordic/nrf52840_dongle/target/thumbv7em-none-eabi/release/nrf52840_dongle.elf ./third_party/tock/boards/nordic/nrf52840_dongle/target/thumbv7em-none-eabi/release/nrf52840_dongle.hex`.
> 
> Unfortunately it seems the device does not boot-up.

I'd recommend discussing kernel support for more boards directly on Tock's repository (https://github.com/tock/tock) for greater visibility.

That said, another way to check whether your flashing method works would be to:
1. Flash the kernel.
1. Compile the `blink` example from libtock-rs (https://github.com/tock/libtock-rs/blob/master/examples/blink.rs). With this blink app flashed, you should see LEDs blinking as soon as your board boots.

Can you confirm that this works? That'd be helpful to know whether your issue is due to your flashing method (or your board), or specific to OpenSK.

> What is the expected behaviour after boot-up? Should the device show up as a ttyUSB?

When plugged on a Linux machine, the OpenSK key should show up in `lsusb` and you should also see entries in `dmesg` (example: https://github.com/google/OpenSK/issues/3#issuecomment-586142642). There should also be a `/dev/hidrawN` device file created for OpenSK.

jmichelp:
Closing this pull-request as I integrated all your changes as part of a bigger effort to support `nordicdfu`, locally defined boards, etc. in PR#78 

Thanks a lot @xiongyihui for having done the work of upstreaming the required change and thanks Denis for your work on prototyping flashing over DFU. This made PR#78 possible and clearly helped improving OpenSK!

PR #8, opened 2020-02-04
Update arrayref version in libraries/crypto.

gendx:
The `Cargo.toml` at the root directory already uses version 0.3.6, let's update the one in the crypto library as well.

- [x] Tests pass


gendx:
Merging given that the Travis-CI build issue (#9) is unrelated and a fix is in review (#10).

PR #14, opened 2020-02-05
Change the way private key and cert are embedded.

### LINKED ISSUE #12: gen_key_materials script sometimes fails
## Observed Behavior

This Travis-CI build fails with the following error https://travis-ci.org/google/OpenSK/builds/646020026

```
error[E0308]: mismatched types
  --> src/ctap/key_material.rs:30:47
   |
30 |   pub const ATTESTATION_PRIVATE_KEY: [u8; 32] = [
   |  _______________________________________________^
31 | |     0x00, 0xb5, 0xc4, 0xcf, 0x67, 0x32, 0x40, 0x19, 0x0c, 0xcd, 0xaa, 0x04, 0xcf, 0x83, 0x3f, 0x49,
32 | |     0xac, 0xb8, 0x3d, 0xff, 0x4a, 0x77, 0x20, 0x1a, 0xf9, 0x9b, 0x0c, 0xae, 0x9e, 0x8f, 0x2d, 0x31,
33 | |     0x60,
34 | | ];
   | |_^ expected an array with a fixed size of 32 elements, found one with 33 elements
```

The attestation key is generated by the [tools/gen_key_materials.sh](https://github.com/google/OpenSK/blob/master/tools/gen_key_materials.sh) file.

## Expected Behavior

The script should always generate an array of 32 bytes.

jmichelp:
OpenSSL seems to serialize bigints as signed value, which means the ECC
key may end up being 33 bytes instead of the 32 bytes we're expecting,
causing build to fail.
The shell script extraction is now replaced by a build.rs script that
uses OpenSSL to extract the content and do sanity checks.
Forcing generating cryptographic materials now always generate a key and
a certificate (useful to compile/flash multiple keys without them being
considered as clones). The self-signed CA is left untouched.

Fixes #12 

- [x] Tests pass


build.rs
```diff
@@ -0,0 +1,78 @@
...
+        .expect("Failed to load OpenSK certificate");
+
+    // Do some sanity check on the certificate
+    assert!(cert
+        .public_key()
+        .unwrap()
+        .public_eq(&PKey::from_ec_key(pkey).unwrap()));
+    let today = asn1::Asn1Time::days_from_now(0).unwrap();
+    assert!(cert.not_after() > today);
+    assert!(cert.not_before() <= today);
```
gendx: Shouldn't this check also include the time of day?

jmichelp: It's the case. I renamed the variable to `now` to better reflect that.

PR #18, opened 2020-02-06
Don't dereference slices obtained from include_bytes.

gendx:
Some of the slices obtained with `include_bytes` were dereferenced to be stored in arrays, and were then referenced again upon use. This change removes these back and forth conversions.

- [x] Tests pass

### DIFF
--- a/src/ctap/ctap1.rs
+++ b/src/ctap/ctap1.rs
@@ -309,7 +309,7 @@ impl Ctap1Command {
         signature_data.extend(key_handle);
         signature_data.extend_from_slice(&user_pk);
 
-        let attestation_key = crypto::ecdsa::SecKey::from_bytes(&ATTESTATION_PRIVATE_KEY).unwrap();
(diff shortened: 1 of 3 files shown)


ia0:
I don't understand why the back and forth is an issue. Using arrays instead of references to arrays is better because we save one pointer in read-only data. Also, taking a reference to that array is essentially a no-op. This is just the symbol address, so it has no cost. But maybe I'm missing something.

gendx:
The compiler should optimize that away (the constants will be stored in a data section anyway, and any extra indirection via a reference/pointer should also be inline) so the binary should be the same. The difference that I see is that it's clearer from a code point of view, and more consistent with the `ATTESTATION_CERTIFICATE` being a slice as well.

ia0:
Right, the compiler probably does not store the reference in `.rodata`. However, I don't think the "consistency" with certificate is a good argument. The certificate is a reference because a slice has unknown size. Those fields are arrays and have fixed size, so we have a choice. However, I agree this choice is probably just about style, so I don't really care which one we do.

PR #20, opened 2020-02-06
Pass features to crypto bench example.

gendx:
This change forwards enabled features to the crypto bench example, like it's done for the CTAP app. This allows to deploy things like `./deploy.sh --panic-console crypto_bench` and indeed have the panic message being printed to the console.

- [x] Tests pass

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -160,7 +160,7 @@ comma_separated () {
 }
 
 build_app () {
-  local feature_list="$(comma_separated $@)"
+  local feature_list="$(comma_separated "$@")"
   cargo build \
     --release \
     --target=thumbv7em-none-eabi \
@@ -180,7 +180,7 @@ build_app () {
 }
 
 build_crypto_bench () {
-  local feature_list="$(comma_separated $@)"
+  local feature_list="$(comma_separated "$@")"
   cargo build \
     --release \
     --target=thumbv7em-none-eabi \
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -176,9 +176,18 @@ build_app () {
 }
 
 build_crypto_bench () {
+  # Flatten the array
+  # This is equivalent to the following python snippet: ' '.join(arr).replace(' ', ',')
+  local feature_list=$(IFS=$'\n'; echo "$@")
+  if [ "X${feature_list}" != "X" ]
+  then
+    feature_list="${feature_list// /,}"
+  fi
```
ia0: I believe this code should not be duplicated.

```sh
comma_separated () {
  local feature_list=$(IFS=$'\n'; echo "$@")
￼  if [ "X${feature_list}" != "X" ]
￼  then
￼    echo "${feature_list// /,}"
￼  fi
}
```

We would then have `local feature_list="$(comma_separated "$@")"` in both `build_app` and here.

gendx: I had to tweak that a bit but indeed refactored to avoid code duplication.

deploy.sh
```diff
@@ -148,15 +148,19 @@ build_app_padding () {
...
+  if [ "X${list}" != "X" ]
   then
-    feature_list="${feature_list// /,}"
+    feature_list="${list// /,}"
   fi
+  echo ${list}
+}
 
+build_app () {
+  local feature_list="$(comma_separated $@)"
```
ia0: This is usually a bad pattern (`"$@"` should be preferred to `$@`), but in that case we don't expect spaces in feature names. Also the `comma_separated` function does not handle it well neither. And this is hopefully temporary until this becomes a python script.

Issue #24, opened 2020-02-06
Cargo audit warns about RUSTSEC-2019-0031


gendx:
## Expected Behavior

Cargo audit doesn't find any advisory.

## Actual Behavior

[RUSTSEC-2019-0031](https://rustsec.org/advisories/RUSTSEC-2019-0031) is found.

```
Crate:  spin
Title:  spin is no longer actively maintained
Date:   2019-11-21
URL:    https://rustsec.org/advisories/RUSTSEC-2019-0031
Dependency tree: 
spin 0.5.2
├── ring 0.16.11
│   └── crypto 0.1.0
│       └── ctap2 0.1.0
└── linked_list_allocator 0.6.6
    └── libtock 0.1.0
        ├── ctap2 0.1.0
        └── crypto 0.1.0
```

## Next steps

1. I filed an issue for `linked_list_allocator`: https://github.com/phil-opp/linked-list-allocator/issues/22.
  - The libtock-rs version we're based on uses `spin` to lock the allocator ([here](https://github.com/tock/libtock-rs/blob/ab2c945184b98ecae3e70ac678e9f5231deef73b/src/entry_point.rs#L360)), even though there is only one thread of execution so the lock isn't necessary and I don't expect much issue given that only one thread acquires the lock only once.
  - Upstream libtock-rs doesn't lock the allocator anymore (https://github.com/tock/libtock-rs/pull/107), so we can pick that up as a patch in the short term.
2. The `ring` dependency is only used for unit tests, and the advisory is tracked here: https://github.com/briansmith/ring/issues/921.

gendx:
The `linked_list_allocator` warning is now removed, due to not depending on the `spin` feature.

kaczmarczyck:
After #662, cargo audit is happy.

PR #26, opened 2020-02-07
Don't lock the allocator with spin.

gendx:
This change uses the heap of `linked_list_allocator` directly instead of locking it via a mutex. Since the app is single threaded this shouldn't be a problem.

This also allows to remove the `spin` dependency of `linked_list_allocator`, which isn't maintained anymore since RUSTSEC-2019-0031, thereby addressing one half of #24.

After this change, `cargo audit` should only mention `spin` via the `ring` dev-dependency.

- [x] Tests pass

### DIFF
--- a/patches/libtock-rs/06-no_spin_allocator.patch
+++ b/patches/libtock-rs/06-no_spin_allocator.patch
@@ -1,0 +1,77 @@
+diff --git a/Cargo.toml b/Cargo.toml
+index 844e424..386a9ed 100644
+--- a/Cargo.toml
++++ b/Cargo.toml
+@@ -6,7 +6,7 @@ license = "MIT/Apache-2.0"
+ edition = "2018"
+ 
+ [dependencies]
+-linked_list_allocator = "0.6.6"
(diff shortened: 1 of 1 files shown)


ia0:
I don't think #24 is an issue that should trigger code changes on our side. The fact that a crate is not maintained since a few months is not a short-term issue.

Do we believe adding one more patch to maintain is worth it?

gendx:
Although there's no urgency, I think keeping a clean `cargo audit` is worth it in the long term for better code hygiene.

As for the additional maintenance, it's indeed an extra patch, but not that many more lines of code and also following what was done upstream in https://github.com/tock/libtock-rs/pull/107. If it ever turns out to be an issue, we can also coordinate that with upstream libtock-rs.

gendx:
> I agree having a clean cargo audit is good, but it's just a tool and its output has to be interpreted. In that case, I think the output can be dismissed.

Since we chose to integrate this tool in our CI, I don't think it makes sense to dismiss the output when we can reasonably fix it.

PR #46, opened 2020-02-20
Better handling of OpenSSL generated keys.

jmichelp:
It seems that ECC keys generated by OpenSSL can be between 30 and 33
bytes long whereas OpenSK code expects a fixed size of 32 bytes.
This variation could cause Travis CI to fail but also invalid ECC keys
to be flashed, causing the authenticator to not work.
Also improves debugging of such issue in the future by printing more actionnable messages

- [x] Tests pass


build.rs
```diff
@@ -43,15 +43,37 @@ fn main() {
...
+        key_len <= 33,
+        "Invalid private key (too big): {} ({:#?})",
+        priv_key_hex,
+        priv_key_vec,
+    );
+
+    // Copy OpenSSL generated key to our vec, starting from the end
+    let mut output_vec = [0u8; 32];
+    let min_key_len = std::cmp::min(key_len, 32);
+    output_vec[32usize.saturating_sub(min_key_len)..]
```
gendx: `min_key_len <= 32` by definition so we can use a regular subtraction. Same for the next line.

build.rs
```diff
@@ -43,15 +43,37 @@ fn main() {
...
+    let priv_key_hex = pkey.private_key().to_hex_str().unwrap();
+    let priv_key_vec = pkey.private_key().to_vec();
+    let key_len = priv_key_vec.len();
+
+    assert!(
+        key_len >= 30,
+        "Invalid private key (too small): {} ({:#?})",
+        priv_key_hex,
+        priv_key_vec,
+    );
```
gendx: There's no reason to ban keys with length < 30 bytes. As far as I understand it, these are just keys with first bytes equal to zero - which openssl serializes as a variable length array.

Rejecting keys with many zeros at the beginning would introduce bias in the generation of private keys, which would be wrong from a cryptographic perspective.

If that's indeed openssl's process, in principle the probability to observe a key of length <= N bytes is `256^(N-32)` - so if you generate enough keys you should observe one with `N = 29` bytes for example.

build.rs
```diff
@@ -43,15 +43,37 @@ fn main() {
...
+    assert!(
+        key_len <= 33,
+        "Invalid private key (too big): {} ({:#?})",
+        priv_key_hex,
+        priv_key_vec,
+    );
```
ia0: Not very important, but it's obvious from the size if it's too big or too small. Those 2 asserts could be merged into one.

```rust
assert!(
    30 <= key_len && key_len <= 33,
    "Invalid private key length ({}): {} ({:#?})",
    key_len,
    priv_key_hex,
    priv_key_vec,
);
```

build.rs
ia0: We actually don't need `saturating_sub` if we use the min as discussed offline :-)

```rust
32 - min_key_len
```

Same for `key_len - min_key_len` below.

build.rs
jmichelp: I thought we would expect a minimum entropy on the private key, that's why I put this test. But I'll remove it.

build.rs
jmichelp: The minimum length will go away as per Guillaume's comment

build.rs
jmichelp: Done

jmichelp: Done

build.rs
gendx: That's not the definition of entropy. If a cryptographic algorithm specifies sampling an integer with a uniform distribution in `[1, N]`, then `1` is a valid sample and cannot be ruled out as "low-entropy".
- If one does that, then it's the distribution `[2, N]` instead - this actually *decreases* the entropy.
- `1` is not more special than any other specific value.

The same applies if instead of a single value like `1` one takes an arbitrary sub-range of integers, such as those which start with 3 (or more) zeros.

Note that the fact that an all-zeroes private key is forbidden is different (as zero behaves differently than other values in this case from a mathematical point-of-view) and this fact is specified by the cryptographic algorithm.

PR #49, opened 2020-02-21
Fix errors in deploy script.

jmichelp:
- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -25,6 +25,7 @@ import os
 import shutil
 import subprocess
 import sys
+from tockloader.exceptions import TockLoaderException
 from tockloader import tab, tbfh, tockloader
 
 # This structure allows us in the future to also support out-of-tree boards.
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -104,7 +105,7 @@ def __call__(self, parser, namespace, values, option_string=None):
     else:
       items = copy.copy(items)
     if self.const in items:
-      self.remove(self.const)
+      items.remove(self.const)
```
ia0: Looks like it should have been a type error to me. Happy I don't write Python :-)

Also, this looks weird to have to check for existence before removal.

jmichelp: I saw it because of pylint warning me that `self` has no `remove` methods actually :)

Removing an item from a list in python raises an exception is the item doesn't exist. Hence the check.

ia0: > I saw it because of pylint warning me that self has no remove methods actually :)

Nice! Do we plan to run pylint in Travis as part of #50 ?

> Removing an item from a list in python raises an exception is the item doesn't exist. Hence the check.

I see. It looks like a bad library design choice because it removes control from the user. I prefer Rust's choice of letting the user decide what they want to do by returning whether the element was present.

jmichelp: I'll add pylint and formatting check for python and markdown. that's a good idea.

Issue #50, opened 2020-02-21
Ensure that the deploy script is also covered by Travis runs


jmichelp:
## Expected Behavior

The `deploy.py` script should be covered by Travis CI to catch errors.

This will imply several things:
- add a `--dry-run` parameter to the script to run all the actions except execution of `JLinkExe`
- split compilation and flash operations (also useful to support more flashing methods such as USB bootloader, OpenOCD, etc.)
- assess if `tockloader` also has such a dry-run flag
- modify Travis configuration to run deploy script with the different parameters

## Actual Behavior

Script is not run.

## Steps to Reproduce the Problem

N/A

## Specifications

- Version: N/A
- Platform: N/A

jmichelp:
As suggested during the review of PR #49, we should also:

- Add linter for python and markdown
- Check for proper formatting of python and markdown files

gendx:
The `workflows/boards_build.yml` already tests `./deploy.py --board=<board> --no-app --programmer=none`. I guess the `--programmer=none` corresponds to the `--dry-run` parameter proposed in this issue.

Should we add anything else?
I'd say at the very least not only testing `--no-app`.

kaczmarczyck:
Several `--programmer=none` are added.

PR #51, opened 2020-02-21
Configure pylint for Google's Python style.

jmichelp:
Adding pylint configuration, VSCode configuration to lint on save and
make the deploy.py script compliant.

Related to issue #50 

- [x] Tests pass (`deploy.py` is not tested yet)


.vscode/settings.json
```diff
@@ -14,7 +14,10 @@
...
     "--style=chromium"
   ],
+  "python.linting.enabled": true,
+  "python.linting.lintOnSave": true,
+  "python.linting.pylintEnabled": true,
   "[python]": {
     "editor.tabSize": 2
   },
-}
+}
```
ia0: Looks like a typo. It's usually a good idea to end files with a newline.

deploy.py
```diff
@@ -19,12 +19,13 @@
...
 
 import argparse
-import colorama
 import copy
 import os
 import shutil
 import subprocess
 import sys
+
+import colorama
```
ia0: Out of curiosity (don't know Python), why do we need to reorder this import?

jmichelp: Just a style guide thing: system imports first. And colorama is a separate package so it belongs to the 2nd set of imports.

.vscode/settings.json
jmichelp: The newline is being removed by the auto-formater in VSCode :/

deploy.py
ia0: Oh I see. Makes sense. Thanks!

.vscode/settings.json
ia0: I see. So it's fine since this file is meant to be edited by VSCode users anyway :-)

jmichelp: Found the culprit! Configuration mismatch between my workstation and my laptop :)

ia0: Cool thanks!

PR #57, opened 2020-02-27
Synchronize with upstream Tock to remove 04-rtt.patch

gendx:
This pull request removes the need for `04-rtt.patch` now that https://github.com/tock/tock/pull/1556 is merged upstream.

Instead, `03-usb-debugging.patch` now enables USB debugging via the RTT protocol and decreases application memory accordingly. This removes the old `03-app-memory.patch`, which was tied to USB debugging. Applying only patches 01 and 02 should work as well, with traditional UART debugging instead of USB debugging.

- [x] Tests pass


patches/tock/01-persistent-storage.patch
```diff
@@ -18,9 +18,9 @@ index ddac9dbd..7e2a3298 100644
              kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
              _ => f(None),
          }
-@@ -408,6 +410,14 @@ pub unsafe fn setup_board<I: nrf52::interrupt_service::InterruptService>(
-         None
-     };
+@@ -405,6 +407,14 @@ pub unsafe fn setup_board<I: nrf52::interrupt_service::InterruptService>(
+     );
+     nrf52::acomp::ACOMP.set_client(analog_comparator);
```
jmichelp: This line surely has nothing to do with persistent storage.
Isn't this line supposed to be added anyway by bumping the commit on `third_party/tock`?

PR #63, opened 2020-03-02
Add feature to track allocations in libtock-rs

gendx:
This PR adds a `--debug-allocations` parameter to the `deploy.py` script. This parameter enables a Cargo feature to add tracking of allocations in libtock-rs, and printing general statistics (total memory usage, number of allocated pointers) to the console.

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -367,6 +367,14 @@ if __name__ == '__main__':
             "board."),
   )
   app_commands.add_argument(
+      "--debug-allocations",
+      action="append_const",
+      const="debug_allocations",
+      dest="features",
+      help=("The console will be used to output allocator statistics every "
+            "yime an allocation/deallocation happens."),
+  )
+  app_commands.add_argument(
       "--no-u2f",
       action=RemoveConstAction,
       const="with_ctap1",
--- a/deploy.py
+++ b/deploy.py
@@ -372,7 +372,7 @@ if __name__ == '__main__':
       const="debug_allocations",
       dest="features",
       help=("The console will be used to output allocator statistics every "
-            "yime an allocation/deallocation happens."),
+            "time an allocation/deallocation happens."),
   )
   app_commands.add_argument(
       "--no-u2f",
--- a/patches/libtock-rs/07-debug_allocations.patch
+++ b/patches/libtock-rs/07-debug_allocations.patch
@@ -11,10 +11,10 @@ index 386a9ed..af3c5db 100644
  [dev-dependencies]
  corepack = { version = "0.4.0", default-features = false, features = ["alloc"] }
 diff --git a/src/entry_point.rs b/src/entry_point.rs
-index 2fe5c40..c978ee5 100644
+index 2fe5c40..545d163 100644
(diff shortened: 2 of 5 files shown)


deploy.py
```diff
@@ -366,6 +366,14 @@ def main(args):
             "output messages before starting blinking the LEDs on the "
             "board."),
   )
+  app_commands.add_argument(
+      "--debug-allocations",
+      action="append_const",
+      const="debug_allocations",
+      dest="features",
+      help=("The console will be used to output allocator statistics every "
+            "yime an allocation/deallocation happens."),
```
jmichelp: typo

gendx:
> Should we also add at least 1 `cargo check` with this feature as part of the workflows?

Good idea! I added more combinations for this feature as well as `panic_console` (each feature individually, and all features together).

patches/libtock-rs/07-debug_allocations.patch
```diff
@@ -0,0 +1,98 @@
...
++            self.count.fetch_add(1, atomic::Ordering::SeqCst);
++            self.size.fetch_add(layout.size(), atomic::Ordering::SeqCst);
++            writeln!(
++                crate::console::Console::new(),
++                "alloc[{}, {}] = {:?} ({} ptrs, {} bytes)",
++                layout.size(),
++                layout.align(),
++                ptr,
++                self.count.load(atomic::Ordering::SeqCst),
++                self.size.load(atomic::Ordering::SeqCst)
```
ia0: I think it's preferable to use the result of `fetch_add` (and `fetch_sub` for deallocation below) and either:
- Recompute the new count and size locally.
- Just print the old values (which is equivalent anyway since we also print the size).

This way we avoid race conditions.

gendx: Libtock apps are single-threaded and not preemptible, except at `yield` points where control-flow can jump to a callback. So there's no risk of race condition here and the displayed results are correct.

I used `AtomicUsize` because it has interior mutability while being `Sync`, therefore we can have a `static TockAllocator` object not marked as `mut` and avoid using `unsafe` (accessing `static mut` objects is unsafe in general, notably due to the possibility of multiple threads). If I recall correctly, that wouldn't be possible with a `Cell`, because the Rust compiler forbids `static` objects that are not `Sync` (due to the possibility of multi-threaded access, even though libtock isn't multi-threaded).

If libtock became multi-threaded in the future:
- Fetching the values once wouldn't remove the risk of a race condition between fetching `count` and `size`.
- Even though some printed values may be incorrect, the overall count would still converge to a correct value due to the use of atomics, in the sense that non-racy (de)allocations would print correct values.

Overall, I find it clearer to perform each arithmetic operation only once, rather than fetching fewer times. Given the trade-offs, it's mostly a matter of style, and the scope is limited to some debugging scenarios anyway.

ia0: I see. Then I think it's worth documenting why `AtomicUsize` is used (next to `TockAllocator`).

PR #66, opened 2020-03-04
Support storing in RAM instead of flash

ia0:
This permits to run without persistent storage. The benefit is that the board doesn't implement a the syscall API in Tock. The disadvantage is that rebooting the key will reset the storage.

This uses 512 bytes of RAM and can store a residential keys of about 150 bytes.


src/embedded_flash/buffer.rs
```diff
@@ -13,6 +13,10 @@
 // limitations under the License.
 
 use super::{Index, Storage, StorageError, StorageResult};
+#[cfg(not(feature = "std"))]
+use alloc::boxed::Box;
+#[cfg(not(feature = "std"))]
+use core as std;
```
gendx: I find this confusing. The code should rather use `core`, and making sure that `core` is recognized in a std context for the tests.

ia0: Used `core`. I don't know why I thought I should use `std` with `features = "std"`.

.github/workflows/cargo_check.yml
```diff
@@ -70,6 +70,12 @@ jobs:
           command: check
           args: --target thumbv7em-none-eabi --release --features debug_ctap,with_ctap1,panic_console,debug_allocations
 
+      - name: Check OpenSK ram_storage
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --target thumbv7em-none-eabi --release --features ram_storage
+
```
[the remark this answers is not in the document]
ia0: Done.

PR #67, opened 2020-03-05
Remove patch

### LINKED ISSUE #58: Debug output sometimes freezes the OpenSK application
## Expected Behavior

When deploying the OpenSK app with `--debug` enabled, the key works and debug output is visible in the console. The debug output starts with the following:

```
Initialization complete. Entering main loop
NRF52 HW INFO: Variant: AAD0, Part: N52840, Package: QI, Ram: K256, Flash: K1024
usbc::start() - State=PoweredOn
Your chip is NRF52840 revision 3. Although this USB implementation should be compatible, your chip hasn't been tested.
usbc::start() - subscribing to interrupts.
enable_out_endpoint_(0) - State=Started
enable_in_out_endpoint_(1) - State=Started
attach() - State=Started
enable_pullup() - State=Started
Enabling USB pullups
New state is Attached
usbc::ready not implemented
usbc::suspend not implemented
usbc::resume not implemented
Bus reset
Cancelling USB receive due to timeout
Cancelling USB receive due to timeout
Set Address = 21
Cancelling USB receive due to timeout
Cancelling USB receive due to timeout
...
```

## Actual Behavior

Sometimes, the log output stops after the following. In that case, the control flow seems blocked:  the key doesn't process any USB request, no LED blinks.

```
Initialization complete. Entering main loop
NRF52 HW INFO: Variant: AAD0, Part: N52840, Package: QI, Ram: K256, Flash: K1024
usbc::start() - State=PoweredOn
Your chip is NRF52840 revision 3. Although this USB implementation should be compatible, your chip hasn't been tested.
usbc::start() - subscribing to interrupts.
enable_out_endpoint_(0) - State=Started
enable_in_out_endpoint_(1) - State=Started
attach() - State=Started
enable_pullup() - State=Started
Enabling USB pullups
New state is Attached
usbc::ready not implemented
```

**Note**: without `--debug`, the log output sometimes also stops after `usbc::ready not implemented`. But in that case the application is still functional afterwards.

On an nRF52840-DK.

1. Deploy OpenSK with `$ ./deploy.py os --board nrf52840_dk` and `$ ./deploy.py app --debug --panic-console --opensk`.
1. Run `$ JLinkExe -device nrf52 -if swd -speed 1000 -autoconnect 1` and `$ JLinkRTTClient` on a separate terminal.
1. Press the reset button multiple times and observe the console output each time.

## Specifications

- Version: https://github.com/gendx/OpenSK/tree/3831872585fa544661b6449d9f6642f6c9e4776e
- Platform: `NRF52 HW INFO: Variant: AAD0, Part: N52840, Package: QI, Ram: K256, Flash: K1024`, with a Linux host.

gendx:
Fixes #58. This synchronizes with upstream Tock to remove the patch added in #60.

- [x] Tests pass


run_desktop_tests.sh
```diff
@@ -40,6 +40,16 @@ cargo check --release --target=thumbv7em-none-eabi --examples
 echo "Checking that CTAP2 builds and links properly (1 set of features)..."
 cargo build --release --target=thumbv7em-none-eabi --features with_ctap1
 
+echo "Checking that supported boards build properly..."
+cd third_party/tock/boards
+cd nordic/nrf52840dk
+make
+cd ../..
+cd nordic/nrf52840_dongle
+make
```
ia0: We can as well do `make -C third_party/tock/boards/nordic/nrf52840dk`.

jmichelp: +1 on this modification

.github/workflows/boards_build.yml
```diff
@@ -0,0 +1,30 @@
...
+          target: thumbv7em-none-eabi
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install Python dependencies
+        run: python -m pip install --upgrade pip setuptools wheel
+      - name: Set up OpenSK
+        run: ./setup.sh
+
+      - name: Building board nrf52840dk
```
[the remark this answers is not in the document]
gendx: Yes that was the idea. I'm not sure though why paths use `**/*` vs. `**` vs. `*`. The workflow rules are inconsistent.

.github/workflows/boards_build.yml
```diff
@@ -0,0 +1,36 @@
+---
+name: Build supported boards
+on:
+  push:
+    paths:
+      - 'patches/tock/**'
+      - 'third_party/**'
```
gendx: @jmichelp will this pick up changes in git submodules?

gendx:
There's something weird happening on my desktop.

My guess is that because my build folder is longer than on Travis, the binary is bigger. Indeed, due to Rust including debug paths in binaries, the longer the folder, the larger the binary.

```
$ strings boards/nordic/nrf52840dk/target/thumbv7em-none-eabi/release/nrf52840dk | grep <my build folder> | wc -l
579
```

This means that I reach the limit of 128K currently set in the [layout.ld](https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/layout.ld#L3), but not Travis and GitHub workflows (yet).

We should be able to increase the limit to 192K (the `prog` section starts at address `0x00030000`), but I want to make sure that regressions are caught by our CI.

---

On Travis:
```
   Compiling nrf52dk_base v0.1.0 (/home/travis/build/google/OpenSK/third_party/tock/boards/nordic/nrf52dk_base)
    Finished release [optimized + debuginfo] target(s) in 26.07s
   text	   data	    bss	    dec	    hex	filename
 128512	   1720	 260424	 390656	  5f600	target/thumbv7em-none-eabi/release/nrf52840dk

=> text + data = 130232 = 0x1FCB8
```

For me the `text` section takes 1024 more bytes:
```
    Finished release [optimized + debuginfo] target(s) in 0.02s
   text	   data	    bss	    dec	    hex	filename
 129536	   1720	 260424	 391680	  5fa00	target/thumbv7em-none-eabi/release/nrf52840dk

=> text + data = 131256 = 0x200B8
```

ia0:
Sounds good to me to use 192K. Also note that we don't use 0x30000-0x40000. So we could also push that limit if needed.

gendx:
> Sounds good to me to use 192K. Also note that we don't use 0x30000-0x40000. So we could also push that limit if needed.

That's good to know. Although I don't expect reaching the next limit soon. We could also tune the board configuration to remove all the drivers that we don't use, which would definitely free up some space but also be one more patch to maintain.

gendx:
I also tested the app on the nRF52840-DK. Works.

```
$ ./deploy.py os --board nrf52840_dk
$ ./deploy.py app --opensk
$ ./deploy.py app --panic-console --debug --opensk
```

.github/workflows/boards_build.yml
gendx: Also, should we add `deploy.py` or more generally any `.sh` or `.py` script, because they may affect the list of supported boards?

.github/workflows/boards_build.yml
```diff
@@ -0,0 +1,38 @@
+---
+name: Build supported boards
+on:
+  push:
+    paths:
+      - 'patches/tock/**'
```
jmichelp: We don't have subdirectory in `patches/` so you can remove a star in the pattern here.

.github/workflows/boards_build.yml
jmichelp: Building a kernel for a given board is unrelated to `deploy.py`. The only script that is required is `setup.sh` because it applied kernel patches.
But we can add a flag to `deploy.py` to output the supported boards and iterate on them instead.

It also doesn't depend on `libtock-rs` to we should modify the pattern to be `third_party/tock/*` instead

.github/workflows/boards_build.yml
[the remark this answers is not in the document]
jmichelp: It's not inconsistent, these patterns simply have different meanings.

.github/workflows/boards_build.yml
gendx: So what are the rules for these patterns?

I'm also afraid that by over-optimizing the patterns, we may miss regressions because the wrong rule prevented the right test to run.

Except for the crypto tests, all the workflows run in a reasonable time (and GitHub is faster than Travis in that regard), so we should rather remove most of the rules to focus on a few rules for the cases where it's *really* slow, than having complex rules that let regressions through.

.github/workflows/boards_build.yml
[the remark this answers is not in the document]
gendx: You wrote:

> We don't have subdirectory in patches/ so you can remove a star in the pattern here.

So if `**` matches more than `*`, we should use `**` everywhere to simplify the patterns (unless one really doesn't want to match files in nested folders in a specific case).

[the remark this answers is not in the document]
gendx: > It's not inconsistent, these patterns simply have different meanings.

I wouldn't say it's "simple". I haven't seen any link to the relevant documentation of their meanings in our workflows, so new contributors can be confused as well.

.github/workflows/boards_build.yml
gendx: > It also doesn't depend on `libtock-rs` to we should modify the pattern to be `third_party/tock/*` instead

Why a single star in that case? You mentioned:

> We don't have subdirectory in patches/ so you can remove a star in the pattern here.

but isn't there sub-directories in third_party/tock? Or is the submodule treated as a single file?

gendx:
Removed the path rules in the new workflow, to keep this pull-request focused on removing the patch.

I don't think the `build_boards` is slow enough to justify specialized path rules, but if so we can do that in a separate pull-request.

.github/workflows/boards_build.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: Build supported boards
+on:
+  push:
```
jmichelp: With this, the build will run on every single commit. Hence the path protection.

gendx: Yes, see my previous comment. I think it's premature optimization to do this in this pull-request.

> Removed the path rules in the new workflow, to keep this pull-request focused on removing the patch.
> 
> I don't think the `build_boards` is slow enough to justify specialized path rules, but if so we can do that in a separate pull-request.

gendx: Done.

PR #72, opened 2020-03-09
Add patch for bidirectional USB endpoints to fix panics in case of cancellation.

### LINKED ISSUE #71: Cancellation doesn't work
## Expected Behavior

When cancelling user presence checks on Google Chrome, the transaction is cancelled and the key is ready to accept more requests.

## Actual Behavior

When compiled with `--panic-console`, the kernel panics.

```
Kernel panic at /tock/chips/nrf52/src/usbd.rs:1702:
	"assertion failed: `(left == right)`
  left: `OutDelay`,
 right: `Init`"
	Kernel version release-1.4-714-gfbc863fa
```

1. Visit https://webauthn.io/ with Google Chrome, with OpenSK plugged in.
1. Click "register".
1. Click "cancel" on Chrome's popup when asked to touch the key.

## Specifications

- Version: 47bd524c40de73f058c6a6a4b37cf98d808c8217
- Platform: Linux

gendx:
Fixes #71.

Although CTAP HID uses a bidirectional USB endpoint, there was currently only one state per endpoint. However, in order to accept cancellation packets we need to split the endpoint state into 2 independent directions.

This additional patch fixes that by splitting the endpoint state into an IN state and an OUT state, each having its own buffer. This fixes the observed issue.

I'll also send a pull-request upstream for the changes in the chip and HIL, together with the relevant changes in `patches/tock/02-usb.patch`.

- [x] Tests pass

### DIFF
--- a/patches/tock/05-usb-cancel.patch
+++ b/patches/tock/05-usb-cancel.patch
@@ -1,0 +1,747 @@
+diff --git a/capsules/src/usb/usbc_client.rs b/capsules/src/usb/usbc_client.rs
+index b0678c23..9fb43781 100644
+--- a/capsules/src/usb/usbc_client.rs
++++ b/capsules/src/usb/usbc_client.rs
+@@ -115,11 +115,11 @@ impl<'a, C: hil::usb::UsbController<'a>> hil::usb::Client<'a> for Client<'a, C>
+         self.client_ctrl.enable();
+ 
+         // Set up a bulk-in endpoint for debugging
+-        self.controller().endpoint_set_buffer(1, self.buffer(1));
++        self.controller().endpoint_set_in_buffer(1, self.buffer(1));
+         self.controller().endpoint_in_enable(TransferType::Bulk, 1);
+ 
+         // Set up a bulk-out endpoint for debugging
+-        self.controller().endpoint_set_buffer(2, self.buffer(2));
++        self.controller().endpoint_set_out_buffer(2, self.buffer(2));
+         self.controller().endpoint_out_enable(TransferType::Bulk, 2);
+     }
+ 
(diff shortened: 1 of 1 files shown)


patches/tock/05-usb-cancel.patch
```diff
@@ -0,0 +1,747 @@
...
+diff --git a/chips/nrf52/src/usbd.rs b/chips/nrf52/src/usbd.rs
+index 8c1992cc..972871d0 100644
+--- a/chips/nrf52/src/usbd.rs
++++ b/chips/nrf52/src/usbd.rs
+@@ -623,7 +623,7 @@ pub enum UsbState {
+ pub enum EndpointState {
+     Disabled,
+     Ctrl(CtrlState),
+-    Bulk(TransferType, EndpointDirection, BulkState),
++    Bulk(TransferType, Option<BulkInState>, Option<BulkOutState>),
```
ia0: Is there an invariant that at least one of the option is not none? If yes, please document.

patches/tock/05-usb-cancel.patch
```diff
@@ -0,0 +1,747 @@
...
+@@ -1497,29 +1480,27 @@ impl<'a> Usbd<'a> {
+         // Endpoint 8 (isochronous) doesn't receive any EPDATA event.
+         for endpoint in 1..NUM_ENDPOINTS {
+             if epdatastatus.is_set(status_epin(endpoint)) {
+-                let (transfer_type, direction, state) =
++                let (transfer_type, in_state, out_state) =
+                     self.descriptors[endpoint].state.get().bulk_state();
+-                match state {
+-                    BulkState::InData => {
++                assert!(in_state.is_some());
```
ia0: There are many assert/panic in this code. Are they only for internal invariants? Or also for preconditions (if this API is callable by some user)?

PR #73, opened 2020-03-09
Wipe sensitive data on entry deletion

ia0:
When inserting (or replacing) entries in the store, the data may be marked as sensitive. When that entry is deleted, the data is wiped by overwriting it with zeroes. This may cost a few bytes of overhead per entry with sensitive data to satisfy the constraint that words may only be written twice.

This has been fuzzed for 4 days without crash. See https://github.com/ia0/OpenSK/tree/wipe_fuzz for fuzzing code. Current coverage is `cov: 7482 ft: 49438` and current histograms are:
```
                          0    1    2    4    8   16   32   64  128  256 512 1024 2048 4096 8192 16384 32768 65536 count
Input length (bytes):          5   21  825  207  131   97   67   16                                                 1369
   Word size (bytes):                  813  556                                                                     1369
   Page size (bytes):                                                    698  337  217  117                         1369
     Number of pages:             806  130  104   98   52  138   41                                                 1369
  Maximum page erase:        676  143  103   72   52   61   43   27   56  23   25   17   27   21    13     6     4  1369
 Initial erase count: 12921 4580  673  882  751  849 1093  959  463 2731  12   10  132   59  166    24    14       26319
     Initial inserts:   388   55   82  127  223  157  241   96                                                      1369
     Initial deletes:   741   53  117  139  147  149   13   10                                                      1369
    Initial replaces:   719   75   91  139  165  162   13    5                                                      1369
    Initial compacts:  1054   90   85   67   59   14                                                                1369
      Initial errors:  1004   56   52   54   59   67   68    9                                                      1369
Initial data (bytes): 11178 4204 2805 2243 2741 1858 1502 1053 2495 1726 600  411  182                             32998
   Initial sensitive:   645   68   66  120  140  170  138   22                                                      1369
    Number of insert:   684  311  136   93   91   40   13    1                                                      1369
    Number of delete:  1060  162   85   46   10    5    1                                                           1369
   Number of replace:  1066  178   64   47   10    3    1                                                           1369
   Number of compact:  1212   81   47   22    7                                                                     1369
     Number of error:  1171  110   37   21   14    9    6    1                                                      1369
 Data length (bytes):  1741  330  270  357  940  133  181  100  148  121  30   18    3                              4372
   Sensitive entries:  1008  106   81   86   60   26    2                                                           1369
```


src/embedded_flash/store/mod.rs
```diff
@@ -449,6 +463,31 @@ impl<S: Storage, C: StoreConfig> Store<S, C> {
...
+    fn wipe_sensitive_data(&mut self, mut index: Index) {
+        let entry = self.read_entry(index);
+        debug_assert!(self.format.is_present(entry));
+        debug_assert!(self.format.is_deleted(entry));
+        if self.format.is_internal(entry) || !self.format.is_sensitive(entry) {
+            // No need to wipe the data.
+            return;
+        }
+        let gap = self.format.entry_gap(entry);
+        let data = &entry[gap.start..][..gap.length];
```
gendx: Since the next line iterates on the bytes anyway, I think it'd be more readable as `.iter().skip(gap.start).take(gap.length)`. I think that the "double slicing" `entry[gap.start..][..gap.length]` is more confusing.

src/embedded_flash/store/mod.rs
```diff
@@ -728,6 +770,22 @@ impl<C: StoreConfig> Store<BufferStorage, C> {
     pub fn set_erase_count(&mut self, page: usize, erase_count: usize) {
         self.initialize_page(page, erase_count);
     }
+
+    /// Checks whether all deleted sensitive entries have been wiped.
+    pub fn check_wipe(&self) {
```
gendx: `#cfg(test)`? Nit: I'd also rename it `check_wiped`.

ia0: This `impl` is under `cfg(feature = "std")` and documented as not for production. This is used both in unit tests and in fuzzing. I don't think you can build a library such that `cfg(test)` is enabled.

Renamed to `check_wiped`.

src/embedded_flash/store/mod.rs
ia0: I created a helper function for that in `ByteGap` since it's used twice. And I used `[start .. start + length]` which is fine because `start` is a variable. I find it more readable than chaining iterator transformations.

src/embedded_flash/store/mod.rs
```diff
@@ -728,6 +770,22 @@ impl<C: StoreConfig> Store<BufferStorage, C> {
     pub fn set_erase_count(&mut self, page: usize, erase_count: usize) {
         self.initialize_page(page, erase_count);
     }
+
+    /// Checks whether all deleted sensitive entries have been wiped.
+    pub fn check_wiped(&self) {
```
gendx: I'd return the number of entries checked, so that the caller can check that indeed N entries were wiped.

ia0: The caller doesn't know how many entries should be wiped, because some entries could have been erased (through compaction). What is the goal you are trying to achieve? It would be simpler to suggest solutions if I know the goal.

gendx: - Although I agree that in general it's hard to model compaction - and that it could be an implementation detail - in the simple case (`delete_sensitive_ok` test) with (1) insert 1 entry and (2) delete & wipe it, there shouldn't be compaction, so 1 entry should be wiped.
- Likewise, is there any compaction if you insert N entries and then delete them linearly?
- Another potential test would be to insert some entries, wipe them, and then search for them in the whole memory (or even substrings of say at least 4 bytes) to rule out that no sensitive fragments are left behind. I'm not sure how that overlaps with the work-in-progress fuzzing.

ia0: I'm not sure we should check implementation details. We should check the properties we are interested in:
- The store behaves as a data structure: insert, replace, delete, and find behave as expected.
- The store operations may be interrupted at any word boundary (this additional constraint will be lifted with checksum) and completed or rolled-back at next initialization and so recursively (the initialization may be interrupted).
- The store wipes data from sensitive deleted entries.

Those properties are checked for simple scenarios in unit tests and for arbitrary scenarios in fuzzing. What would be the additional property you want to check?

Regarding your points:
- What would we gain by checking that? If the deleted entry is not present in the storage because erased, it's even a better outcome. We shouldn't need to check that the deleted entry is still in the storage.
- With the current implementation (which is not proactive and only compacts when needed), you should be able to insert N times then delete entries without compaction.
- Yes I thought about that, but I think it's too complicated for what it's worth. This could happen in case of corruption. We should deal with that when dealing with corruption.

gendx: > * What would we gain by checking that? If the deleted entry is not present in the storage because erased, it's even a better outcome. We shouldn't need to check that the deleted entry is still in the storage.

The expectation is that sensitive data is indeed wiped. The current check accepts an implementation that just clears the metadata without actually wiping the sensitive data. So by checking that, we would gain to rule out implementation that just clear the metadata.

> * With the current implementation (which is not proactive and only compacts when needed), you should be able to insert N times then delete entries without compaction.

If the current implementation behaves like this, I'd add a test for that, with a comment that indeed it tests an implementation detail (therefore the check could be removed/relaxed in the future if the implementation details become incompatible).

> * Yes I thought about that, but I think it's too complicated for what it's worth. This could happen in case of corruption. We should deal with that when dealing with corruption.

If sensitive data is not wiped in case of corruption, I consider it breaking the contract that sensitive data is wiped when erased. If that's not the case, the contract of what a "sensitive" entry does should be clarified (e.g. "wiped upon deletion unless corruption case X happens"), and basic cases within the contract should definitely be unit tested. Otherwise, what is the value in having sensitive entries in the first place?

ia0: I think all those remarks get the same answer: Trying to get something perfect is not beneficial. It's all about compromises between flash lifetime, CPU usage, RAM usage, latency, code complexity, code size, probability of events, usage scenarios, etc. Focusing on the wrong things will only get the more important things be less studied.

In particular when I compare with other flash storage implementations, we protect (and plan to protect) against way more scenarios, some of which are not really realistic. Also, if we want to increase this coverage, we should focus on the scenarios that represent the highest cumulative probability in the distribution function of scenarios.

Applied to the current question: Trying to check whether deleted sensitive data leaks is not worth the effort. Instead, checking whether deleted sensitive data is wiped is good enough. First, because it is not well-defined: what about if one byte leaks? What if it happens to appear somewhere else by accident? For example the user has a duplicate entry and deletes one. Second, because it will never be always true (assuming we have some definition). So we would need either a probabilistic model to check against or we would need to define the domain on which it should hold. For the first one, we would need to write tooling to measure it, it's not simple fuzzing. For the second one, we need to understand how flash works, which we don't: what about a power loss during a word write? Is reading that word giving deterministic results?

The space of possible issues is too big and we don't have good enough approximation of the distribution function of those issues. So it matters to focus on what is realistic as much as we can and not try to over-specify corner cases. This is not productive.

src/embedded_flash/store/mod.rs
```diff
@@ -728,6 +770,22 @@ impl<C: StoreConfig> Store<BufferStorage, C> {
...
+            if !self.format.is_present(entry)
+                || !self.format.is_deleted(entry)
+                || self.format.is_internal(entry)
+                || !self.format.is_sensitive(entry)
+            {
+                continue;
+            }
+            let gap = self.format.entry_gap(entry);
+            let data = gap.slice(entry);
+            assert!(data.iter().all(|&byte| byte == 0x00));
```
jmichelp: Do we want to panic here?
Because the device will potentially be bricked at that point, right: at each boot it may try to wipe data and panic.

gendx: This function is only used in unit tests on the desktop.

src/embedded_flash/store/format.rs
```diff
@@ -59,6 +59,14 @@ pub struct Format {
     /// - 1 for insert entries.
     replace_bit: usize,
 
+    /// Whether a user entry has sensitive data.
+    ///
+    /// When a user entry with sensitive data is deleted, the data is overwritten with zeroes.
```
gendx: Could you be more specific about the contract here? In which cases is data indeed guaranteed to be overwritten at the moment? In which (corruption) cases is it not guaranteed? When moving a sensitive entry for compaction, is the old entry wiped out at an appropriate time? What happens when recovering while some sensitive entry was being copied around?

The contract doesn't have to be very strict for now, but there must at least be a disclaimer about what cases are not covered.

Of course, this contract can evolve as the implementation evolves (better corruption recovery for example).

src/embedded_flash/store/mod.rs
gendx: > Trying to get something perfect is not beneficial.

I agree. I'm not asking for a perfect solution.

What I'm proposing is something *pragmatic*: the current `delete_sensitive_ok` scenario is so simple that checking that indeed one entry was found and wiped should be checked. If somehow in the future the implementation evolves such that in this simple scenario there isn't any sensitive entry left at all, then we can remove this check. But this is highly hypothetical at this point - I don't have any pragmatic implementation in mind that would behave like that.

I just want to rule out the case where the metadata would be cleared of the sensitive bit while no sensitive data is wiped. Yes this assumes that no data is moved around in this simple scenario, but I think this is a pragmatic assumption. I don't think such "wipe metadata instead of data" is a corner case, because it would be easy to write or introduce such a bug.

ia0: > the current delete_sensitive_ok scenario is so simple that checking that indeed one entry was found and wiped should be checked

It's not because something is simple that we should do it. The negative effect of returning the number of wiped entries (besides checking implementation details) is that it will confuse readers of the code. There is no such semantics notion as the number of wiped entries. Same for the number of deleted entries, we never count the number of deleted entries. Those notions are not part of the model.

When exhaustive unit testing is not feasible, unit tests are reserved for simple scenarios. For the rest, there is fuzzing.

That said, we may decide that this feature is not useful, which kind of makes sense. It's mostly a sanity feature than any kind of protection. The main case where the user would be more concerned about deleted data than non-deleted data would be when there are no non-deleted data, so when resetting the key and planning to not reuse it (e.g. giving it to someone else or throwing to the trash). If this is indeed the only scenario where wiping data matters, we could do something specific for reset only.

src/embedded_flash/store/mod.rs
ia0: I changed the function to return a `bool` such that the assert only appears in tests and fuzzing.

src/embedded_flash/store/format.rs
ia0: The contract is the same for all operations and documented in the module. Currently, the only failure scenario is power loss with atomic word writes.

src/embedded_flash/store/mod.rs
gendx: Leaving the decision to @jmichelp to resolve this point.

src/embedded_flash/store/format.rs
gendx: It would be useful to repeat the assumptions in the documentation of this field and/or at least write a sentence pointing to the module's documentation.

ia0: Done.

src/embedded_flash/store/mod.rs
```diff
@@ -43,6 +43,28 @@
 //! The data-structure can be configured with the `StoreConfig` trait. By implementing this trait,
 //! the number of possible tags and the association between keys and entries are defined.
 //!
+//! # Properties
+//!
+//! The data-structure provides the following properties:
+//! - When an operation returns success, then the represented multi-set is updated accordingly. For
+//!   example, an inserted entry can be find with alteration until replaced or deleted.
```
gendx: "find" -> "found"

src/embedded_flash/store/mod.rs
```diff
@@ -43,6 +43,28 @@
...
+//!
+//! Those properties rely on the following assumptions:
+//! - Writing a word to flash is atomic. When power is lost, the word is either fully written or not
+//!   written at all.
+//! - Reading a word from flash is deterministic. When power is lost while writing or erasing a word
+//!   (erasing a page containing that word), reading that word repeatedly returns the same result
+//!   (until it is written or its page is erased).
+//! - To decide whether a page has been erased, it is enough to test if all its bits are equal to 1.
+//!
+//! The properties still hold outside those assumptions but with weaker probabilities as the usage
```
gendx: I'd be less assertive here, i.e. "The properties still hold" -> "Some of the properties may still hold"

src/embedded_flash/store/mod.rs
ia0: Done. Actually there was another typo for "with" -> "without".

src/embedded_flash/store/mod.rs
ia0: Done.

src/embedded_flash/store/mod.rs
```diff
@@ -43,6 +43,28 @@
...
+//! - When power is lost before an operation returns, the operation will either succeed or be
+//!   rolled-back on the next initialization. So the multi-set would be either left unchanged or
+//!   updated accordingly.
+//!
+//! Those properties rely on the following assumptions:
+//! - Writing a word to flash is atomic. When power is lost, the word is either fully written or not
+//!   written at all.
+//! - Reading a word from flash is deterministic. When power is lost while writing or erasing a word
+//!   (erasing a page containing that word), reading that word repeatedly returns the same result
+//!   (until it is written or its page is erased).
```
jmichelp: Side note: this property will stop being true when the flash cell reaches its end-of-life (i.e. beyond max erase/write cycles)

src/embedded_flash/store/mod.rs
jmichelp: I would either return nothing or a bool. But not a number. Reading and writing to flash consumes power and may create side-channels. If at some point we start trying to reduce the leakage we have on OpenSK, the fewer points we have to fix, the better.

src/embedded_flash/store/mod.rs
ia0: Yes indeed. But we can note that the store will reject updates once we reach the flash lifetime, such that it is still possible to read the key content successfully to migrate to a replacement key. However, the user can tweak this behavior by setting up the store with a maximum number of erase cycles that is higher than the one specified by the hardware. It would then be possible to use the store outside the flash lifetime.

ia0:
I'll keep this around until the design of checksum is working and stable, because supporting this from the start would complicate things a little. And it's more important to have checksum than this.

ia0:
I'm merging this. There shouldn't be any problem when adding checksum and there would be too much merge conflicts otherwise.

PR #78, opened 2020-03-11
Revamp deploy.py

### LINKED ISSUE #3: NRF52840 boards?
Does the key need to be the specific Nordic NRF52840 devoper dongle or does the key only need to have the NRF52840 SOC on the dongle?

### LINKED ISSUE #70: Reproducible builds
As mentioned in https://github.com/tock/tock/issues/1666, binaries built with Rust are not reproducible by default. One main reason is panic messages that embed an absolute path to the file containing them (so building with `--panic-console` isn't reproducible by default).

We should apply the same techniques as in https://github.com/tock/tock/issues/1666 to make OpenSK builds reproducible.

This would avoid issues like spurious linker errors when using a long build path, as discovered in #67.

jmichelp:
Now the script supports more flashing methods:
- JLink (with tockloader)
- OpenOCD (with tockloader)
- pyOCD
- Nordic DFU
- none (will produce an IntelHex file)

Also merged the contributions from:
- Yihui Xiong to support the Makerdiary USB dongle board
- Dennis Geurts to support Nordic DFU

Doc updated accordingly.

Imported 2 patches for Tock kernel:
- 06-add-set_vector_table_offset.patch (upstream tock/tock#1579)
- 07-nrf52-bootloader.patch (upstream tock/tock#1681)

Fixes #3 
Fixes #70 

- [x] Tests pass


README.md
```diff
@@ -19,8 +19,9 @@ successfully tested on the following boards:
 
 ## Disclaimer
 
-This project is proof-of-concept and a research platform. It's still under
-development and as such comes with a few limitations:
+This project is **proof-of-concept and a research platform**. It is **NOT**
+meant for a day to day usage. It's still under development and as such comes
```
gendx: "day to day" -> "daily"?

README.md
```diff
@@ -53,25 +54,17 @@ For a more detailed guide, please refer to our
...
-2.  If Tock OS is already installed on your board, move to the next step.
-    Otherwise, just run one of the following commands, depending on the board
-    you have:
+2.  Next step is to install Tock OS as well as the OpenSK application on your
+    board (**Warning**: it will erase the locally stored credentials). Run:
 
     ```shell
     # Nordic nRF52840-DK board
-    ./deploy.py os --board=nrf52840_dk
+    ./deploy.py --board=nrf52840dk --opensk
```
gendx: I found it useful to have the possibility of flashing only os or app (to avoid flakiness of the flashing protocol), at least for boards that support it. Only supporting a "flash all" would be a regression.

boards/nrf52840_mdk_dfu/src/io.rs
```diff
@@ -0,0 +1,65 @@
+use core::fmt::Write;
```
gendx: What's specific about this file that doesn't exist in one of the `third_party/tock/boards/nordic/foo/src/io.rs`? We shouldn't be duplicating code.

If the change is a one-liner over the `nrf52840_dongle` board, we should keep it as a patch applied by `./setup.sh`.

Same question for the `main.rs`.

deploy.py
```diff
@@ -166,38 +292,104 @@ def update_rustc_if_needed(self):
...
 
-  def build_and_install_tockos(self):
+  def search_binary(self, name, start_directory="."):
+    for root, _, files in os.walk(start_directory):
+      for fname in files:
+        if fname == name:
+          return os.path.join(root, fname)
+    return None
+
+  def build_tockos(self):
```
gendx: I don't think we should manually build things instead of relying on Tock's Makefiles. That's too much to maintain for us and there are pending changes that affect it (tock/tock#1660, tock/tock#1668, tock/tock#1669).

We should rather have some patches to `third_party/tock` that add the relevant boards.

deploy.py
```diff
@@ -284,84 +513,219 @@ def verify_flashed_app(self, expected_app):
       app_found = expected_app in apps
     return app_found
 
+  def create_hex_file(self, dest_file):
+    # We produce an intelhex file with everything in it
```
gendx: Could you link to the intelhex documentation?

deploy.py
```diff
@@ -284,84 +513,219 @@ def verify_flashed_app(self, expected_app):
...
-      help=("Indicates which part of the firmware should be compiled and "
-            "flashed to the connected board."))
-
-  os_commands = commands.add_parser(
-      "os",
-      parents=[shared_parser],
-      help=("Compiles and installs Tock OS. The target board must be "
-            "specified by setting the --board argument."),
+  main_parser.add_argument(
+      "--no-tockos",
```
gendx: This partially replies to my previous comment. Is there a `--no-app` option as well?

README.md
jmichelp: Done

README.md
jmichelp: Flashing the app without the OS is possible by passing `--no-tockos` flag.
The scenario that is not possible anymore is flashing only the OS because of DFU

boards/nrf52840_mdk_dfu/src/io.rs
jmichelp: It's a different board. Buttons and LEDs are different.

deploy.py
jmichelp: But then it means we need to have a Makefile for every out-of-tree board that we want to support.

deploy.py
jmichelp: I can add it.

deploy.py
jmichelp: Done

deploy.py
gendx: Will these Makefiles be any different from one another? The `./setup.sh` script should take care of that.

deploy.py
gendx: This saves time when fixing bugs in the kernel. On the other hand, calling `make flash` directly in the relevant `board/` folder of Tock allows to do the same. So it's not of high priority.

deploy.py
jmichelp: Moved back to using make.

deploy.py
jmichelp: Added the `--no-app` flag.

deploy.py
jmichelp: Done

deploy.py
```diff
@@ -284,84 +468,227 @@ def verify_flashed_app(self, expected_app):
...
+      fatal("Listing {} is not implemented.".format(self.args.listing))
+
+    self.check_prerequisites()
     self.update_rustc_if_needed()
 
-    if self.args.action == "os":
-      info("Installing Tock on board {}".format(self.args.board))
-      self.build_and_install_tockos()
+    if not self.args.tockos and not self.args.application:
+      info("Nothing to do.")
```
gendx: I'd return a non-zero error code in that case so that people are aware they need to provide something to do in their command line?

boards/nrf52840_mdk_dfu/src/io.rs
gendx: Resolving for now. Ideally we would just have some diff patch to apply, but I don't want to block this pull-request for that.

README.md
```diff
@@ -19,8 +19,9 @@ successfully tested on the following boards:
```
gendx: Do you want to mention the DFU-based boards in this README? And the DFU limitations regarding re-flashing Tock?

patches/tock/07-nrf52-bootloader.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/chips/nrf52/src/crt1.rs b/chips/nrf52/src/crt1.rs
```
gendx: This is now patch number 05.

deploy.py
```diff
@@ -20,22 +20,129 @@
...
 
-# This structure allows us in the future to also support out-of-tree boards.
+PROGRAMMERS = frozenset(("jlink", "openocd", "pyocd", "nordicdfu", "none"))
+
+# This structure allows us to support out-of-tree boards as well as (in the
+# future) more achitectures.
+OpenSKBoard = collections.namedtuple(
+    "OpenSKBoard",
+    [
+        # Location of the Tock board (where the Makefile file is
```
gendx: Missing parenthesis at end of line?

deploy.py
```diff
@@ -20,22 +20,129 @@
...
+        "path",
+        # Targeted architecture (e.g. thumbv7em-none-eabi)
```
gendx: I think "target" is a more common spelling in this context.

deploy.py
```diff
@@ -20,22 +20,129 @@
...
+        "arch",
+        # Size of 1 page of flash memory
+        "page_size",
+        # Flash address at which the kernel will be written
+        "kernel_address",
+        # Set to None is padding is not required for the board
```
gendx: Might be worth being more specific in what padding we're talking about. Padding starts in flash from this address, which must be the same as `prog` in the board's `layout.ld`, and until `app_address`.

deploy.py
```diff
@@ -166,38 +293,58 @@ def update_rustc_if_needed(self):
       # Need to update
       self.checked_command_output(
           ["rustup", "install", target_toolchain_fullstring])
-      self.checked_command_output(
-          ["rustup", "target", "add", "thumbv7em-none-eabi"])
+    targets = {x.arch for x in SUPPORTED_BOARDS.values()}
+    for arch in targets:
+      self.checked_command_output(["rustup", "target", "add", arch])
```
gendx: Shouldn't we only update the selected board's arch?

deploy.py
```diff
@@ -244,16 +403,38 @@ def install_elf_file(self, elf_path):
...
+    with open(kernel_file, "rb") as f:
+      kernel = f.read()
+    args = copy.copy(self.tockloader_default_args)
+    setattr(args, "address", board_props.app_address)
+    tock = loader.TockLoader(args)
+    tock.open(args)
+    try:
+      tock.flash_binary(kernel, board_props.kernel_address)
+    except TockLoaderException as e:
+      fatal("Couldn't install padding: {}".format(str(e)))
```
gendx: Is it "padding" here?

deploy.py
jmichelp: Indeed.

deploy.py
jmichelp: Done

deploy.py
jmichelp: Done

deploy.py
jmichelp: Done.
I started updating all because I wanted to build the app for all boards at one (the TAB file is supposed to handle that). But then I faced some limitations on what tockloader can do. We can revisit that once we can build a single TAB file.

deploy.py
jmichelp: Oops.

patches/tock/07-nrf52-bootloader.patch
jmichelp: Actually 04

PR #80, opened 2020-03-12
Sync upstream kernel to get more reproducible builds.

gendx:
To move towards reproducible builds (#70), this pull-request syncs the kernel with upstream now that https://github.com/tock/tock/pull/1668 and https://github.com/tock/tock/pull/1669 have been merged. The `sha256sum` tool is also used to compute the sum of the built binary on CI.

- [x] Tests pass

### DIFF
--- a/.github/workflows/opensk_build.yml
+++ b/.github/workflows/opensk_build.yml
@@ -24,8 +24,16 @@ jobs:
       - name: Set up OpenSK
         run: ./setup.sh
 
+      - name: Building sha256sum tool
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --manifest-path third_party/tock/tools/sha256sum/Cargo.toml
+
       - name: Building OpenSK
         uses: actions-rs/cargo@v1
         with:
           command: build
           args: --release --target=thumbv7em-none-eabi --features with_ctap1
+      - name: Compute SHA-256 sum
+        run: ./third_party/tock/tools/sha256sum/target/debug/sha256sum target/thumbv7em-none-eabi/release/ctap2
--- a/run_desktop_tests.sh
+++ b/run_desktop_tests.sh
@@ -48,6 +48,10 @@ echo "Checking that supported boards build properly..."
 make -C third_party/tock/boards/nordic/nrf52840dk
 make -C third_party/tock/boards/nordic/nrf52840_dongle
 
+echo "Checking that other boards build properly..."
+make -C boards/nrf52840_dongle_dfu
+make -C boards/nrf52840_mdk_dfu
+
 if [ -z "${TRAVIS_OS_NAME}" -o "${TRAVIS_OS_NAME}" = "linux" ]
 then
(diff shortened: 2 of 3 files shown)


.github/workflows/opensk_build.yml
```diff
@@ -24,8 +24,16 @@ jobs:
...
+          command: build
+          args: --manifest-path third_party/tock/tools/sha256sum/Cargo.toml
+
       - name: Building OpenSK
         uses: actions-rs/cargo@v1
         with:
           command: build
           args: --release --target=thumbv7em-none-eabi --features with_ctap1
+      - name: Compute SHA-256 sum
+        run: ./third_party/tock/tools/sha256sum/target/debug/sha256sum target/thumbv7em-none-eabi/release/ctap2
```
ia0: Don't we generate new crypto materials at each build?

gendx: Good point! We'd need to account for that. Computing the checksum now can help track reproducibility progress though.

ia0: We could also dump the key materials to the standard output. That way we can reproduce a build and check the hash.

gendx: We can add some example crypto material as a stub, and use that in the workflow where we want a reproducible build.

jmichelp: IIRC there's a way on Github workflows to flag files and save them as produced artefact. Normally this is used to create release packages. Maybe we can experiment around that.

run_desktop_tests.sh
```diff
@@ -24,6 +24,9 @@ cd libraries/crypto
 cargo fmt --all -- --check
 cd ../..
 
+echo "Building sha256sum tool..."
```
jmichelp: I wouldn't bother updating the travis file anymore. It's going away soon (workflows seem more reliable and OS X build on travis proved it was flaky).

gendx: I find it useful to actually run the tests on a desktop (hence the script's name) to iterate fast locally instead having to wait for CI to complete.
- Regarding Travis, we can just remove the `.travis.yml` if that's the issue.
- This script should ideally replicate what the GitHub workflows do.
  - One possibility is to move away from the action dependencies (e.g. `uses: actions-rs/cargo@v1`) which are hard to replicate locally, and instead just run plain Bash scripts in the workflow. This may have some drawbacks by removing the additional features that actions provide, but would allow consistency between local tests and workflows.
  - Another possibility is to opportunistically synchronize this script with the workflows (as is done in this pull-request). Although things can go out of sync, it's easy to re-sync, so I'm also fine with it.

.github/workflows/opensk_build.yml
gendx: Let's do it in a separate pull-request, as it will require more experimentation.

run_desktop_tests.sh
jmichelp: My bad, I thought it was the `travis.yml` that was updated. It's ok then

PR #82, opened 2020-03-13
Check nrfutil version.

jmichelp:
Ensure that the version of Nordic's `nrfutil` tool is one we can work with.

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -528,7 +528,7 @@ class OpenSKInstaller:
       assert_mandatory_binary("nrfutil")
       assert_python_library("intelhex")
       assert_python_library("nordicsemi.lister")
-      nrfutil_version = __import__("nordicsemi").version.NRFUTIL_VERSION
+      nrfutil_version = __import__("nordicsemi.version").version.NRFUTIL_VERSION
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -528,6 +528,10 @@ def check_prerequisites(self):
       assert_mandatory_binary("nrfutil")
       assert_python_library("intelhex")
       assert_python_library("nordicsemi.lister")
+      nrfutil_version = __import__("nordicsemi").version.NRFUTIL_VERSION
+      if not nrfutil_version.startswith("6."):
+        fatal(("You need to install nrfutil python3 package v6.0 or above. "
+               "Found: {}".format(nrfutil_version)))
```
gendx: What if no version is found? Shouldn't the message say that it's not found?

jmichelp: I haven't checked earlier versions but at least since v1.2.4 of the package this fields exists.
And there's a check earlier that the package "nordicsemi" exists and is imported successfully.

gendx: > And there's a check earlier that the package "nordicsemi" exists and is imported successfully.

Sounds good then!

PR #87, opened 2020-03-17
Reduce the number of syscalls made by the console in libtock-rs.

### LINKED ISSUE #89: The --verbose mode is impractical
## Expected Behavior

OpenSK on an nRF52840-DK with the `--verbose` option enabled can be registered on https://webauthn.io/.

## Actual Behavior

Currently, the `--verbose` option triggers many syscalls to write to the console. This slows down the whole app, making it impractical. For example, USB packets are dropped when trying to register on https://webauthn.io/ in `--verbose` mode on an nRF52840-DK, so registration fails altogether in this scenario.

1. Flash OpenSK on a nRF52840-DK with `--verbose` enabled `$ ./deploy.py --board nrf52840dk --opensk --panic-console --verbose`.
1. Visit https://webauthn.io/ and start the registration flow.
1. Look at the console output and observe unexpected USB packets.
1. Look at the webpage and observe that the registration flow hangs.

## Specifications

- Version: 284b1426ecaf71d7a9b2e1b445ed55953efbd370
- Platform: Linux

gendx:
Fixes #89.

Currently, the `--verbose` option triggers many syscalls to write to the console. This slows down the whole app, making it impractical. For example, USB packets are dropped when trying to register on https://webauthn.io/ in `--verbose` mode on an nRF52840-DK, so registration fails altogether in this scenario.

This pull-request changes the console behavior to only flush the buffer to the kernel when it's full (or when the console object is dropped). Tested on a nRF52840-DK with `$ ./deploy.py --board nrf52840dk --opensk --verbose --panic-console`, the app is much faster, and registration & login work smoothly on https://webauthn.io/.

- [x] Tests pass


patches/libtock-rs/08-buffered-console.patch
```diff
@@ -0,0 +1,95 @@
+diff --git a/src/console.rs b/src/console.rs
+index ecd7ad1..785de01 100644
+--- a/src/console.rs
++++ b/src/console.rs
+@@ -16,33 +16,58 @@ mod allow_nr {
+     pub const SHARE_BUFFER: usize = 1;
+ }
+ 
++const BUFFER_SIZE: usize = 64;
```
gendx: **Note:** this buffer size seems to work fine.

If I use a larger buffer such as 1024 bytes on an nRF52840-DK with Segger RTT debugging, the output gets somehow scrambled.

patches/libtock-rs/08-buffered-console.patch
```diff
@@ -0,0 +1,95 @@
...
+         if result_code < 0 {
+             return;
+         }
+@@ -69,6 +93,14 @@ impl Console {
+     }
+ }
+ 
++impl Drop for Console {
++    fn drop(&mut self) {
++        if !self.is_empty() {
```
jmichelp: Shouldn't the check of the buffer being empty be part of `flush`?
I haven't checked the full logic but I don't necessarily expect calling `flush` on an empty buffer to issue syscalls or return errors.

patches/libtock-rs/08-buffered-console.patch
```diff
@@ -0,0 +1,95 @@
...
++            if self.is_full() {
++                self.flush();
++            }
++
+             not_written_yet = &not_written_yet[num_bytes_to_print..];
+         }
+     }
+ 
+-    fn flush(&mut self, num_bytes_to_print: usize) {
++    fn flush(&mut self) {
```
jmichelp: In this situation, `flush` should become public so that the user can be explicit about when to flush. Otherwise it may be hard to debug what's happening if everything is buffered.

patches/libtock-rs/08-buffered-console.patch
gendx: I could reproduce this bug more generally as https://github.com/tock/tock/issues/1697.

patches/libtock-rs/08-buffered-console.patch
gendx: The other call to flush is when `self.is_full`, in which case checking that it's not empty would be redundant. However, if flush becomes public, this could be useful indeed.

gendx:
Waiting a couple more days to see whether https://github.com/tock/tock/issues/1697 will be easy to fix.

gendx:
Marking as blocked until https://github.com/tock/tock/issues/1697 is resolved.

gendx:
Now that https://github.com/tock/tock/issues/1697 is fixed upstream, I back-ported the fix and tested the --verbose mode on https://webauthn.io/. This worked well.

PR #91, opened 2020-04-08
Sync with upstream Tock to have reproducible builds.

gendx:
This pull request is the next step towards reproducible builds (#70). It syncs the Tock kernel to the latest upstream, removing the need for `04-nrf52-bootloader.patch`, and benefiting from faster and reproducible builds of the kernel.

Now that builds are reproducible, a follow-up could be to have a CI pipeline checking the SHA-256 hashes of all the binaries, against the expected ones obtained by any developer.

- [x] Tests pass


gendx:
@jmichelp, the Cargo format workflow complains about it not being installed for the current toolchain. I'm not sure what to do about that.

jmichelp:
Weird that it starts to fail now.

Try to modify the lines 20 to 22 in `.github/workflows/cargo_fmt.yml` like this:
```yaml
      - uses: actions-rs/toolchain@v1
        with:
          components: rustfmt
```

gendx:
One thing to note is that our `rust-toolchain` doesn't match the Tock kernel's. Shall they be the same?

gendx:
Updated the formatting workflow. It now works :)

Issue #79, opened 2020-03-12
App object has no attribute get_binary


mweinelt:
## Expected Behavior

`deploy.py` should build the image to be flashed.

## Actual Behavior

It fails with an AttributeError, because there is no function defintion for `get_binary()`. Looks to me like the issue was introduced in #78 

```
% ./deploy.py --board=nrf52840_mdk_dfu --opensk --programmer=none
info: Updating rust toolchain to nightly-2020-02-03
info: syncing channel updates for 'nightly-2020-02-03-x86_64-unknown-linux-gnu'
info: component 'rust-std' for target 'thumbv7em-none-eabi' is up to date
info: Rust toolchain up-to-date
info: Building Tock OS for board nrf52840_mdk_dfu
    Finished release [optimized + debuginfo] target(s) in 0.49s
info: Building OpenSK application
    Finished release [optimized] target(s) in 0.64s
info: Generating Tock TAB file for application/example ctap2
Traceback (most recent call last):
  File "./deploy.py", line 772, in <module>
    main(main_parser.parse_args())
  File "./deploy.py", line 645, in main
    OpenSKInstaller(args).run()
  File "./deploy.py", line 595, in run
    self.create_hex_file(dest_file)
  File "./deploy.py", line 508, in create_hex_file
    app_tab.extract_app(board_props.arch).get_binary(),
AttributeError: 'App' object has no attribute 'get_binary'
```

## Steps to Reproduce the Problem

1. ./setup.sh
2. ./deploy.py --board=nrf52540_mdk_dfu --opensk --programmer=none

## Specifications

- Version: ec1ce661d3dde69ea2139d32def21a2fdf9c9362
- Platform: Linux x86_64

jmichelp:
This is due to tockloader project. The last version available on PyPI is v1.3 and we need v1.4 which isn't released yet.
Brad mentioned he's working on releasing v1.4

diegobes:
As a workaround:
1- uninstall tockloader: pip3 uninstall tockloader
2- clone tockloader repository
3- from the tockloader repository root: `python3 setup.py install`
This will install the latest version from master with App.get_binary support

PR #94, opened 2020-04-14
Add GitHub workflow to check that binaries are reproducible.

### LINKED ISSUE #70: Reproducible builds

gendx:
Fixes #70.

This pull request is the last step for reproducible builds.
- A new `reproduce_hashes.sh` tool generates reference SHA-256 sums for the OpenSK binaries. These hashes must then be checked in git in the `reference_binaries.sha256sum` file.
- A new GitHub workflow then compares the cryptographic hashes obtained on continuous integration with those checked into git by the developer. If the hashes don't match, then it means that the builds were not reproducible (although the converse is not necessarily true, this should catch most causes of non-reproducibility).
- Anyone can compare the SHA-256 sums obtained on their machine with those published in git and computed by continuous integration.

---

- [x] Tests pass

---

Currently blocked on:

- [x] A new release of Tockloader on pip3: https://github.com/tock/tockloader/issues/58 (related to #79).
- [x] ~Cargo doesn't yield reproducible builds if the host is different: https://github.com/rust-lang/cargo/issues/8140 (also https://github.com/rust-lang/cargo/pull/7873).~ The upstream Cargo issue won't be fixed soon, so for now 2 sets of hashes are provided for reference and checked by GitHub Actions: Linux and MacOS.
- [x] `elf2tab` is currently not reproducible (https://github.com/tock/elf2tab/pull/16). It stores files into a TAR file, which contains, for each file: a uid, file permissions, a timestamp. The `metadata.toml` also contains a build timestamp.


gendx:
For now, this fails for two reasons.
- As mentioned in https://github.com/google/OpenSK/pull/80#discussion_r391801426, we'll have to use a fixed crypto material for the CTAP2 app to be reproducible.
- For some reason, the binary hashes of the Tock kernel boards are still different on Linux and on MacOS.

gendx:
This is currently blocked on `./deploy.py --board=$board --opensk --no-tockos --programmer=none`, itself blocked on #79 aka https://github.com/tock/tockloader/issues/58.

gendx:
There is currently a discrepancy between Linux and OSX builds of Tock, despite the `--remap-path-prefix` parameter. For example, the `tock_cells` library is built with a different `metadata` parameter on both platforms (whereas I've observed it to be the same on the workflow's Linux machine and on my own Linux machine).

- On Linux (https://github.com/google/OpenSK/pull/94/checks?check_run_id=602439620): `rustc --crate-name tock_cells libraries/tock-cells/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C opt-level=z -C panic=abort -C debuginfo=2 -C metadata=92487154152022d3 -C extra-filename=-92487154152022d3 --out-dir /home/runner/work/OpenSK/OpenSK/third_party/tock/target/thumbv7em-none-eabi/release/deps --target thumbv7em-none-eabi -L dependency=/home/runner/work/OpenSK/OpenSK/third_party/tock/target/thumbv7em-none-eabi/release/deps -L dependency=/home/runner/work/OpenSK/OpenSK/third_party/tock/target/release/deps -C link-arg=-Tlayout.ld -C linker=rust-lld -C linker-flavor=ld.lld -C relocation-model=dynamic-no-pic -C link-arg=-zmax-page-size=512 -C link-arg=-icf=all --remap-path-prefix=/home/runner/work/OpenSK/OpenSK/third_party/tock/= -D warnings`
- On OSX (https://github.com/google/OpenSK/pull/94/checks?check_run_id=602439631): `rustc --crate-name tock_cells libraries/tock-cells/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts --crate-type lib --emit=dep-info,metadata,link -C opt-level=z -C panic=abort -C debuginfo=2 -C metadata=9fc982d890c0358d -C extra-filename=-9fc982d890c0358d --out-dir /Users/runner/runners/2.169.0/work/OpenSK/OpenSK/third_party/tock/target/thumbv7em-none-eabi/release/deps --target thumbv7em-none-eabi -L dependency=/Users/runner/runners/2.169.0/work/OpenSK/OpenSK/third_party/tock/target/thumbv7em-none-eabi/release/deps -L dependency=/Users/runner/runners/2.169.0/work/OpenSK/OpenSK/third_party/tock/target/release/deps -C link-arg=-Tlayout.ld -C linker=rust-lld -C linker-flavor=ld.lld -C relocation-model=dynamic-no-pic -C link-arg=-zmax-page-size=512 -C link-arg=-icf=all --remap-path-prefix=/Users/runner/runners/2.169.0/work/OpenSK/OpenSK/third_party/tock/= -D warnings`

I'll report it upstream to Tock and Rust.

jmichelp:
Should we do like in the tock-on-titan project and symlink the `rust-toolchain` from the kernel at the app level so that we use the same version?

gendx:
> Should we do like in the tock-on-titan project and symlink the `rust-toolchain` from the kernel at the app level so that we use the same version?

Definitely! This would also reduce the time wasted by the workflows to install various versions of the compiler, and therefore reduce the likelihood of being cancelled by the GitHub runner.

gendx:
With elf2tab 0.5.0 published on crates.io, this pull request is now unblocked.

deploy.py
```diff
@@ -392,19 +392,17 @@ def create_tab_file(self, binaries):
...
-    package_parameter = "-n"
     elf2tab_ver = self.checked_command_output(["elf2tab", "--version"]).split(
-        " ", maxsplit=1)[1]
-    # Starting from v0.5.0-dev the parameter changed.
-    # Current pyblished crate is 0.4.0 but we don't want developers
-    # running the HEAD from github to be stuck
-    if "0.5.0-dev" in elf2tab_ver:
-      package_parameter = "--package-name"
+        "\n", maxsplit=1)[0]
+    if elf2tab_ver != "elf2tab 0.5.0":
```
jmichelp, line 397: I would either:
* remove version check
* issue a fatal error if version is less than 0.5.0
* issue a warning if version is different/above 0.5.0

I don't see a reason to block people using 0.6.0-dev from Github.

gendx: My concern is that doing the comparison (less than / greater than) of versions would be a much more complex logic.

I could replace the `fatal` with an `error` message, or add a `--allow-unknown-elf2tab-version` parameter to the `deploy.py` script which turns this `fatal` error into a `warning`. WDYT?

gendx: Also, using `cargo install --vers 0.5.0 elf2tab` would guarantee that the cargo-installed elf2tab will have the tested version (https://doc.rust-lang.org/cargo/commands/cargo-install.html), so we shouldn't need to support any other version.

Another alternative would be to pin the `elf2tab` repository as a submodule in `third_party`, and build it locally instead of relying on an external binary from Cargo or from a custom directory in the user's $PATH.

jmichelp: Ok then let's modify `setup.sh` to install the correct version and either remove the version check or only output a warning.

jmichelp:
What's going to be the process to keep these hashes up-to-date?
They're going to change on pretty much every merged PR and waiting for the workflow to fail in order to grab the correct hashes for both platforms and then update the PR with it is not the most friendly workflow IMHO.

gendx:
> What's going to be the process to keep these hashes up-to-date?
> They're going to change on pretty much every merged PR and waiting for the workflow to fail in order to grab the correct hashes for both platforms and then update the PR with it is not the most friendly workflow IMHO.

One should be able to run the `./reproduce_hashes.sh` script locally to update the hashes on the platform they use. Unfortunately, one still needs to check the workflow for other platforms until https://github.com/rust-lang/cargo/issues/8140 is fixed.

Maybe we can leave this workflow as non required until then? And for example only update hashes after all the reviews have been resolved on a pull request.

jmichelp:
Ok, let's merge it like this and revisit later when we need to run the workflow.
Maybe we need to move to a stable vs. dev branches where reproducibility needs to be run only on stable branch. The other option I have in mind is to modify the workflow: instead of checking hashes, it can publish them.

gendx:
> The other option I have in mind is to modify the workflow: instead of checking hashes, it can publish them.

This would be a good idea too (although wouldn't catch regressions in reproducibility). The current workflow already uploads the reproduced binaries, so it wouldn't be hard to upload the expected hashes in there, or to print them in the workflow's console output.

PR #96, opened 2020-04-20
add 2.1 features to GetInfo

kaczmarczyck:
This is the start of a series of pull requests for 2.1 features. First command: GetInfo.
Also, the data type PublicKeyCredentialParameter is now used in 2 places and is refactored into a new struct.

Changes should all be backwards compatible. We only add more optional commands and outputs.

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,11 @@ few limitations:
 Although we tested and implemented our firmware based on the published
 [CTAP2.0 specifications](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html),
 our implementation was not reviewed nor officially tested and doesn't claim to
-be FIDO Certified. With the upcoming next version of the
-[CTAP2.1 specifications](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html),
-we started adding features, so master is currently between version 2.0 and 2.1.
+be FIDO Certified.
+We started adding features of the upcoming next version of the
+[CTAP2.1 specifications](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html).
+The development is currently between 2.0 and 2.1, with updates hidden behind a feature flag.
+Please add the flag `shell --ctap2-1` to the deploy command to include them.
 
 ### Cryptography
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ be FIDO Certified.
 We started adding features of the upcoming next version of the
 [CTAP2.1 specifications](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html).
 The development is currently between 2.0 and 2.1, with updates hidden behind a feature flag.
-Please add the flag `shell --ctap2-1` to the deploy command to include them.
+Please add the flag `--ctap2.1` to the deploy command to include them.
 
 ### Cryptography
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,9 @@ few limitations:
 Although we tested and implemented our firmware based on the published
 [CTAP2.0 specifications](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html),
 our implementation was not reviewed nor officially tested and doesn't claim to
-be FIDO Certified.
+be FIDO Certified. With the upcoming next version of the
+[CTAP2.1 specifications](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html),
+we started adding features, so master is currently between version 2.0 and 2.1.
 
 ### Cryptography
 
--- a/deploy.py
+++ b/deploy.py
@@ -742,6 +742,14 @@ if __name__ == "__main__":
             "support for U2F/CTAP1 protocol."),
   )
   main_parser.add_argument(
+      "--ctap2-1",
+      action=RemoveConstAction,
+      const="with_ctap2_1",
+      dest="features",
+      help=("Compiles the OpenSK application with backward compatible "
+            "support for CTAP2.1 protocol."),
+  )
+  main_parser.add_argument(
       "--regen-keys",
       action="store_true",
       default=False,
--- a/deploy.py
+++ b/deploy.py
@@ -755,7 +755,7 @@ if __name__ == "__main__":
             "support for U2F/CTAP1 protocol."),
   )
   main_parser.add_argument(
-      "--ctap2-1",
+      "--ctap2.1",
       action=RemoveConstAction,
       const="with_ctap2_1",
       dest="features",
--- a/run_desktop_tests.sh
+++ b/run_desktop_tests.sh
@@ -30,6 +30,7 @@ cargo build --manifest-path third_party/tock/tools/sha256sum/Cargo.toml
 echo "Checking that CTAP2 builds properly..."
 cargo check --release --target=thumbv7em-none-eabi
 cargo check --release --target=thumbv7em-none-eabi --features with_ctap1
+cargo check --release --target=thumbv7em-none-eabi --features with_ctap2_1
 cargo check --release --target=thumbv7em-none-eabi --features debug_ctap
 cargo check --release --target=thumbv7em-none-eabi --features panic_console
 cargo check --release --target=thumbv7em-none-eabi --features debug_allocations
@@ -86,4 +87,10 @@ then
 
   echo "Running unit tests on the desktop (debug mode + CTAP1)..."
   cargo test --features std,with_ctap1
+
+  echo "Running unit tests on the desktop (release mode + CTAP2.1)..."
+  cargo test --release --features std,with_ctap2_1
+
+  echo "Running unit tests on the desktop (debug mode + CTAP2.1)..."
+  cargo test --features std,with_ctap2_1
 fi
--- a/run_desktop_tests.sh
+++ b/run_desktop_tests.sh
@@ -93,4 +93,10 @@ then
 
   echo "Running unit tests on the desktop (debug mode + CTAP2.1)..."
   cargo test --features std,with_ctap2_1
+
+  echo "Running unit tests on the desktop (release mode + CTAP1 + CTAP2.1)..."
+  cargo test --release --features std,with_ctap1,with_ctap2_1
+
+  echo "Running unit tests on the desktop (debug mode + CTAP1 + CTAP2.1)..."
+  cargo test --features std,with_ctap1,with_ctap2_1
 fi
--- a/src/ctap/command.rs
+++ b/src/ctap/command.rs
@@ -23,9 +23,10 @@ use alloc::string::String;
 use alloc::vec::Vec;
 use core::convert::TryFrom;
 
-// Depending on your memory, you can use Some(n) to limit request sizes.
+// Depending on your memory, you can use Some(n) to limit request sizes in
+// MakeCredential and GetAssertion. This affects allowList and excludeList.
 // You might also want to set the max credential size in process_get_info then.
-pub const MAX_CREDENTIAL_COUNT_IN_LIST: Option<u64> = None;
+pub const MAX_CREDENTIAL_COUNT_IN_LIST: Option<usize> = None;
 
 // CTAP specification (version 20190130) section 6.1
 #[cfg_attr(any(test, feature = "debug_ctap"), derive(Debug, PartialEq))]
@@ -136,25 +137,19 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
         )?)?;
 
         let cred_param_vec = read_array(ok_or_missing(param_map.get(&cbor_unsigned!(4)))?)?;
-        let mut pub_key_cred_params = vec![];
-        for cred_param_map_value in cred_param_vec {
-            if let Ok(cred_param) = PublicKeyCredentialParameter::try_from(cred_param_map_value) {
-                pub_key_cred_params.push(cred_param);
-            }
-        }
+        let pub_key_cred_params = cred_param_vec
+            .iter()
+            .map(PublicKeyCredentialParameter::try_from)
+            .collect::<Result<Vec<PublicKeyCredentialParameter>, Ctap2StatusCode>>()?;
 
         let exclude_list = match param_map.get(&cbor_unsigned!(5)) {
             Some(entry) => {
                 let exclude_list_vec = read_array(entry)?;
-                let mut exclude_list = vec![];
-                for exclude_list_value in exclude_list_vec {
-                    if let Some(count) = MAX_CREDENTIAL_COUNT_IN_LIST {
-                        if exclude_list.len() as u64 >= count {
-                            break;
-                        }
-                    }
-                    exclude_list.push(PublicKeyCredentialDescriptor::try_from(exclude_list_value)?);
-                }
+                let exclude_list = exclude_list_vec
+                    .iter()
+                    .take(MAX_CREDENTIAL_COUNT_IN_LIST.unwrap_or(exclude_list_vec.len()))
+                    .map(PublicKeyCredentialDescriptor::try_from)
+                    .collect::<Result<Vec<PublicKeyCredentialDescriptor>, Ctap2StatusCode>>()?;
                 Some(exclude_list)
             }
             None => None,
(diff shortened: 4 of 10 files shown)


jmichelp:
Considering that with this PR the authenticator would be between 2 specifications, do we expect some interop issues here?
If yes, then maybe we should develop the CTAP2.1 in a separate branch and we merge it into master when we're confident it's stable.

kaczmarczyck:
We can use a different branch. There shouldn't be any interop issues, though. Adding new commands goes unnoticed until you call them, naturally. Sometimes, we will output a few more optionals than before or store more stuff on flash.

For MakeCredential and GetAssertion, we are a little bit between versions already. The state automaton changed in early draft versions and I implemented the new one right away. Also, a few of the planned features could also benefit 2.0, i.e. the credProtect extension. And that might be harder to maintain on 2 branches since it changes the MakeCredential implementation logic.

gendx:
Regarding branches, why not adding a "ctap-2.1" feature to OpenSK's Cargo.toml instead? That is, by default OpenSK targets CTAP 2.0 but when the feature is enabled, the (work in progress) additional 2.1 commands are available.

Then we can later remove the feature and make OpenSK support CTAP 2.1 by default.

README.md
```diff
@@ -28,7 +28,9 @@ few limitations:
 Although we tested and implemented our firmware based on the published
 [CTAP2.0 specifications](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html),
 our implementation was not reviewed nor officially tested and doesn't claim to
-be FIDO Certified.
+be FIDO Certified. With the upcoming next version of the
+[CTAP2.1 specifications](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html),
+we started adding features, so master is currently between version 2.0 and 2.1.
```
gendx, line 33: Cf. what I mentioned in the main thread. Now or a separate pull request could be the occasion to add a new "ctap-2.1" feature.

src/ctap/command.rs
```diff
@@ -13,16 +13,20 @@
...
+    ClientPinSubCommand, CoseKey, Extensions, GetAssertionOptions, MakeCredentialOptions,
+    PublicKeyCredentialDescriptor, PublicKeyCredentialParameter, PublicKeyCredentialRpEntity,
+    PublicKeyCredentialUserEntity,
 };
 use super::status_code::Ctap2StatusCode;
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::convert::TryFrom;
 
+// Depending on your memory, you can use Some(n) to limit request sizes.
```
gendx, line 26: You should clarify here which "requests" are affected by this. It's unclear without context.

src/ctap/command.rs
```diff
@@ -134,19 +138,21 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
         let cred_param_vec = read_array(ok_or_missing(param_map.get(&cbor_unsigned!(4)))?)?;
         let mut pub_key_cred_params = vec![];
         for cred_param_map_value in cred_param_vec {
-            let cred_param_map = read_map(cred_param_map_value)?;
-            let cred_type = PublicKeyCredentialType::try_from(ok_or_missing(
-                cred_param_map.get(&cbor_text!("type")),
-            )?)?;
-            let alg = read_integer(ok_or_missing(cred_param_map.get(&cbor_text!("alg")))?)?;
-            pub_key_cred_params.push((cred_type, alg));
+            if let Ok(cred_param) = PublicKeyCredentialParameter::try_from(cred_param_map_value) {
```
gendx, line 141: What's the rationale for ignoring the error code instead of propagating it to the caller? I'm not used to `if let Ok(foo) = ...` pattern, as the error should usually be dealt with.

src/ctap/command.rs
```diff
@@ -134,19 +138,21 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
...
+            }
         }
 
         let exclude_list = match param_map.get(&cbor_unsigned!(5)) {
             Some(entry) => {
                 let exclude_list_vec = read_array(entry)?;
                 let mut exclude_list = vec![];
                 for exclude_list_value in exclude_list_vec {
+                    if let Some(count) = MAX_CREDENTIAL_COUNT_IN_LIST {
+                        if exclude_list.len() as u64 >= count {
```
gendx, line 152: `MAX_CREDENTIAL_COUNT_IN_LIST` should be a `usize` instead.

As far as I understand it, this parameter is local to this code, i.e. isn't used in any protocol, so a native size should be fine. Also, is there a use case where this would overflow an `u32`, to warrant using a `u64`? Unless there is more than 4GB of RAM, I don't see how that would ever happen.

src/ctap/command.rs
```diff
@@ -134,19 +138,21 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
...
-            pub_key_cred_params.push((cred_type, alg));
+            if let Ok(cred_param) = PublicKeyCredentialParameter::try_from(cred_param_map_value) {
+                pub_key_cred_params.push(cred_param);
+            }
         }
 
         let exclude_list = match param_map.get(&cbor_unsigned!(5)) {
             Some(entry) => {
                 let exclude_list_vec = read_array(entry)?;
                 let mut exclude_list = vec![];
```
gendx, line 149: Nit: this vector should `reserve()` some space here, given that the length is known (`let count = std::cmp::min(exclude_list_vec.len(), MAX_CREDENTIAL_COUNT_IN_LIST.unwrap_or(usize::MAX))`).

src/ctap/command.rs
```diff
@@ -134,19 +138,21 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
...
                 for exclude_list_value in exclude_list_vec {
```
gendx, line 150: You can use the `for exclude_list_value in exclude_list_vec.iter().take(count)` pattern for a more concise loop.

src/ctap/data_formats.rs
```diff
@@ -121,6 +121,36 @@ impl TryFrom<&cbor::Value> for PublicKeyCredentialType {
...
 }
 
+#[derive(PartialEq)]
+#[cfg_attr(any(test, feature = "debug_ctap"), derive(Debug))]
+pub struct PublicKeyCredentialParameter {
+    pub cred_type: PublicKeyCredentialType,
+    pub alg: SignatureAlgorithm,
+}
+
+impl TryFrom<&cbor::Value> for PublicKeyCredentialParameter {
```
gendx, line 131: Can you add unit tests at the end of this file?

src/ctap/mod.rs
```diff
@@ -99,6 +100,11 @@ pub const FIDO2_VERSION_STRING: &str = "FIDO_2_0";
 #[cfg(feature = "with_ctap1")]
 pub const U2F_VERSION_STRING: &str = "U2F_V2";
 
+pub const CREDENTIAL_PARAMETER: PublicKeyCredentialParameter = PublicKeyCredentialParameter {
```
gendx, line 103: I don't think `CREDENTIAL_PARAMETER` is descriptive enough of what this constant contains, which makes it hard to understand in the other contexts. What parameter (public key, ES256)? Why is it fixed? A line of documentation would also be welcome.

src/ctap/response.rs
```diff
@@ -228,6 +243,11 @@ mod test {
             options: None,
             max_msg_size: None,
             pin_protocols: None,
+            max_credential_count_in_list: None,
+            max_credential_id_length: None,
+            transports: None,
+            algorithms: None,
+            firmware_version: None,
```
gendx, line 250: IIRC a `None` option is not serialized in the CBOR map. It would be useful to have tests with `Some(dummy_value)` as well, otherwise we miss testing coverage in the parsing side.

src/ctap/command.rs
kaczmarczyck: This code is only filtering unknown credential parameters. Maybe the conversion should just return an Option, not a Result then? I wanted to leave it to the application how to deal with different scenarios. If we later support more than one algorithm, it has to be on mod to decide which one to choose. Empty lists could throw an error right away, if you like that better?

src/ctap/command.rs
kaczmarczyck: It is surfaced to the client in GetInfo. The Option is necessary since you can also decide to not filter at all.

gendx: I didn't mean to remove the `Option` part of it. But representing as a `usize` internally and converting it to a `u64` in GetInfo would be simpler IMO.

src/ctap/command.rs
gendx: `PublicKeyCredentialParameter::try_from` will fail if the CBOR value is not a map, or doesn't contain the required fields ("type", "alg"). We usually surface back these errors to the caller. Why should we silently ignore such CBOR values?

The `SignatureAlgorithm::try_from` failing is another class of error. If we don't want to reject unknown algorithms, then maybe `PublicKeyCredentialParameter::try_from` should return `Result<Option<T>, Err>`, where `Ok(None)` is returned in case of unknown algorithm? Or this should be some kind of `RawPublicKeyCredentialParameter` type in which the algorithm is simply encoded as an integer, and then maybe there is a separate conversion function from `RawPublicKeyCredentialParameter` to `Option<PublicKeyCredentialParameter>` which does the validation of the algorithm.

src/ctap/command.rs
kaczmarczyck: For now we only accept 1 algorithm in the current implementation, so the vector has length 0 or 1 for specification-conform calls. Shouldn't hurt, given that we need the count for take anyway.

README.md
kaczmarczyck: Feature is added and mentioned.

src/ctap/command.rs
kaczmarczyck: Comments added.

src/ctap/command.rs
kaczmarczyck: Done.

src/ctap/command.rs
kaczmarczyck: Neat, done.

src/ctap/mod.rs
kaczmarczyck: Renamed to ES256_CRED_PARAM. This also coincides with the specification, where the types are always spelled out like CredentialParameter, and the variables are shortened like cred_param.

src/ctap/response.rs
kaczmarczyck: Test added.

src/ctap/data_formats.rs
kaczmarczyck: Done, and also gave the other tests in this file a check and extended some.

src/ctap/command.rs
kaczmarczyck: Update: I changed it to a collect construction, that solves the problem more elegantly.

src/ctap/command.rs
kaczmarczyck: Solved by extending SignatureAlgorithm and PublicKeyCredentialType to have an Unknown value.

README.md
```diff
@@ -29,6 +29,10 @@ Although we tested and implemented our firmware based on the published
 [CTAP2.0 specifications](https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html),
 our implementation was not reviewed nor officially tested and doesn't claim to
 be FIDO Certified.
+We started adding features of the upcoming next version of the
+[CTAP2.1 specifications](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html).
+The development is currently between 2.0 and 2.1, with updates hidden behind a feature flag.
+Please add the flag `shell --ctap2-1` to the deploy command to include them.
```
gendx, line 35: Where does one need to type `shell`?

deploy.py
```diff
@@ -754,6 +754,14 @@ def main(args):
       help=("Compiles the OpenSK application without backward compatible "
             "support for U2F/CTAP1 protocol."),
   )
+  main_parser.add_argument(
+      "--ctap2-1",
```
gendx, line 758: `--ctap2.1` could be more readable?

run_desktop_tests.sh
```diff
@@ -86,4 +87,10 @@ then
 
   echo "Running unit tests on the desktop (debug mode + CTAP1)..."
   cargo test --features std,with_ctap1
+
+  echo "Running unit tests on the desktop (release mode + CTAP2.1)..."
+  cargo test --release --features std,with_ctap2_1
+
+  echo "Running unit tests on the desktop (debug mode + CTAP2.1)..."
+  cargo test --features std,with_ctap2_1
```
gendx, line 95: So we never test `with_ctap1` + `with_ctap2_1`? Having both is a relevant use case, and may cause issues like exceeding available memory, so I'd test that too.

run_desktop_tests.sh
```diff
@@ -30,6 +30,7 @@ cargo build --manifest-path third_party/tock/tools/sha256sum/Cargo.toml
 echo "Checking that CTAP2 builds properly..."
 cargo check --release --target=thumbv7em-none-eabi
 cargo check --release --target=thumbv7em-none-eabi --features with_ctap1
+cargo check --release --target=thumbv7em-none-eabi --features with_ctap2_1
```
gendx, line 33: Can you add these tests to the relevant GitHub workflows?

src/ctap/command.rs
```diff
@@ -132,23 +137,19 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
...
-            let cred_type = PublicKeyCredentialType::try_from(ok_or_missing(
-                cred_param_map.get(&cbor_text!("type")),
-            )?)?;
-            let alg = read_integer(ok_or_missing(cred_param_map.get(&cbor_text!("alg")))?)?;
-            pub_key_cred_params.push((cred_type, alg));
-        }
+        let pub_key_cred_params = cred_param_vec
+            .iter()
+            .map(PublicKeyCredentialParameter::try_from)
+            .collect::<Result<Vec<PublicKeyCredentialParameter>, Ctap2StatusCode>>()?;
```
gendx, line 143: I'm surprised collect returns a result of vec, rather than a vec of results.

src/ctap/mod.rs
```diff
@@ -751,14 +754,37 @@ where
...
+                pin_protocols: Some(vec![
+                    CtapState::<R, CheckUserPresence>::PIN_PROTOCOL_VERSION,
+                ]),
+                max_credential_count_in_list: MAX_CREDENTIAL_COUNT_IN_LIST.map(|c| c as u64),
+                // You can use ENCRYPTED_CREDENTIAL_ID_SIZE here, but if your
+                // browser passes that value, it might be used to fingerprint.
+                max_credential_id_length: None,
+                transports: Some(vec![AuthenticatorTransport::Usb]),
+                algorithms: Some(vec![ES256_CRED_PARAM]),
+                firmware_version: None,
```
gendx, line 785: As an alternative to duplicating all the other fields, these ones could be individually enabled by `cfg`. This would also be consistent with the definition of the struct in the other file.

src/ctap/data_formats.rs
```diff
@@ -94,16 +96,22 @@ impl From<PublicKeyCredentialUserEntity> for cbor::Value {
...
+    // Unknown types should be ignored, instead of returning errors.
+    Unknown,
 }
 
 impl From<PublicKeyCredentialType> for cbor::Value {
     fn from(cred_type: PublicKeyCredentialType) -> Self {
         match cred_type {
             PublicKeyCredentialType::PublicKey => "public-key",
+            // We should never create this credential type.
+            PublicKeyCredentialType::Unknown => unreachable!(),
```
gendx, line 114: Even though we indeed shouldn't create such unknown credential types, using `unreachable` here sounds like a code smell. It's only unreachable as long as no caller of this function passes the `Unknown` variant. Usually, `unreachable` can be checked locally within the current function.

So we should probably remove the `Unknown` variant of `PublicKeyCredentialType`, and either provide `impl TryFrom<&cbor::Value> for Option<PublicKeyCredentialType>`, or have another enum to layer that:

```rust
enum CredentialType {
    PublicKeyCredentialType,
    Unknown,
}
```

I'm actually surprised that the `PublicKeyCredentialType` enum contains `PublicKey` only. Shouldn't it be named `CredentialType` if it could in principle contain something else than `PublicKey`?

src/ctap/command.rs
ia0: This is because `Result<V, E>` implements `FromIterator<Result<A, E>>` when `V` implements `FromIterator<A>`. This is also documented in https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect:

> A collection of Result<T, E> can be thought of as single Result<Collection<T>, E>.

gendx: Nice!

README.md
kaczmarczyck: That was unintentional. Removed.

deploy.py
kaczmarczyck: Happy to change that.

run_desktop_tests.sh
kaczmarczyck: Tests are added.

run_desktop_tests.sh
kaczmarczyck: Added tests in cargo_check.yml and opensk_test.yml.

src/ctap/mod.rs
kaczmarczyck: Done.

src/ctap/data_formats.rs
kaczmarczyck: Neither CTAP2 nor WebAuthn share a vision for this enum, it is simply called an "extension point". Not sure if it's worth raising a naming issue after it is already established.

Why would we need a new custom enum when `Option` solves the same problem? I see how `unreachable` is not really what we want here. To me, the next best solution would be to just return the string `"unknown"` in that case. That way, we behave similarly to the enum `SignatureAlgorithm` that assign a plain 0 to the `Unknown` variant.

The alternative is to remove `PublicKeyCredentialType` and `SignatureAlgorithm` and let `PublicKeyCredentialParameter` have 2 Options instead. The point of Unknown was to simplify, but I'll change it if you don't like it?

kaczmarczyck:
Feature for #106

PR #103, opened 2020-04-29
Do not use writeable flash regions for persistent storage

ia0:
Writable flash regions don't play well with DFU.

- [ ] Erase the storage in deploy script?
- [ ] Tests pass

### DIFF
--- a/nrf52840dk_layout.ld
+++ b/nrf52840dk_layout.ld
@@ -3,8 +3,10 @@
  */
 
 MEMORY {
-  /* The application region is 64 bytes (0x40) */
-  FLASH (rx) : ORIGIN = 0x00040040, LENGTH = 0x000BFFC0
+  /* The application region is 64 bytes (0x40) and we reserve 0x40000 at the end
+   * of the flash for the persistent storage.
+   */
+  FLASH (rx) : ORIGIN = 0x00040040, LENGTH = 0x0007FFC0
   SRAM (rwx) : ORIGIN = 0x20020000, LENGTH = 128K
 }
 
--- a/patches/tock/01-persistent-storage.patch
+++ b/patches/tock/01-persistent-storage.patch
@@ -106,7 +106,7 @@ index 5abd2d84..5a726fdb 100644
              let word: u32 = (data[i + 0] as u32) << 0
                  | (data[i + 1] as u32) << 8
                  | (data[i + 2] as u32) << 16
-@@ -374,3 +386,178 @@ impl hil::flash::Flash for Nvmc {
+@@ -374,3 +386,170 @@ impl hil::flash::Flash for Nvmc {
          self.erase_page(page_number)
      }
  }
@@ -189,11 +189,7 @@ index 5abd2d84..5a726fdb 100644
 +    /// Fails with `EINVAL` if any of the following conditions does not hold:
 +    /// - `ptr` must be word-aligned.
 +    /// - `slice.len()` must be word-aligned.
-+    /// - The slice starting at `ptr` of length `slice.len()` must fit in a writeable flash region.
-+    fn write_slice(&self, appid: AppId, ptr: usize, slice: &[u8]) -> ReturnCode {
-+        if !appid.in_writeable_flash_region(ptr, slice.len()) {
-+            return ReturnCode::EINVAL;
-+        }
++    fn write_slice(&self, ptr: usize, slice: &[u8]) -> ReturnCode {
 +        if ptr & WORD_MASK != 0 || slice.len() & WORD_MASK != 0 {
 +            return ReturnCode::EINVAL;
 +        }
@@ -217,11 +213,7 @@ index 5abd2d84..5a726fdb 100644
 +    ///
 +    /// Fails with `EINVAL` if any of the following conditions does not hold:
 +    /// - `ptr` must be page-aligned.
-+    /// - The slice starting at `ptr` of length `PAGE_SIZE` must fit in a writeable flash region.
-+    fn erase_page(&self, appid: AppId, ptr: usize) -> ReturnCode {
-+        if !appid.in_writeable_flash_region(ptr, PAGE_SIZE) {
-+            return ReturnCode::EINVAL;
-+        }
++    fn erase_page(&self, ptr: usize) -> ReturnCode {
 +        if ptr & PAGE_MASK != 0 {
 +            return ReturnCode::EINVAL;
 +        }
@@ -257,11 +249,11 @@ index 5abd2d84..5a726fdb 100644
 +                        None => return ReturnCode::EINVAL,
 +                        Some(slice) => slice,
 +                    };
-+                    self.write_slice(appid, ptr, slice.as_ref())
++                    self.write_slice(ptr, slice.as_ref())
 +                })
 +                .unwrap_or_else(|err| err.into()),
 +
-+            (3, ptr) => self.erase_page(appid, ptr),
++            (3, ptr) => self.erase_page(ptr),
 +
 +            _ => ReturnCode::ENOSUPPORT,
 +        }
@@ -285,39 +277,3 @@ index 5abd2d84..5a726fdb 100644
 +        }
 +    }
 +}
-diff --git a/kernel/src/callback.rs b/kernel/src/callback.rs
-index c812e0bf..bd1613b3 100644
---- a/kernel/src/callback.rs
-+++ b/kernel/src/callback.rs
-@@ -130,6 +130,31 @@ impl AppId {
-             (start, end)
-         })
-     }
-+
-+    pub fn in_writeable_flash_region(&self, ptr: usize, len: usize) -> bool {
-+        self.kernel.process_map_or(false, *self, |process| {
-+            let ptr = match ptr.checked_sub(process.flash_start() as usize) {
-+                None => return false,
-+                Some(ptr) => ptr,
-+            };
-+            (0..process.number_writeable_flash_regions()).any(|i| {
-+                let (region_ptr, region_len) = process.get_writeable_flash_region(i);
-+                let region_ptr = region_ptr as usize;
-+                let region_len = region_len as usize;
-+                // We want to check the 2 following inequalities:
-+                // (1) `region_ptr <= ptr`
-+                // (2) `ptr + len <= region_ptr + region_len`
-+                // However, the second one may overflow written as is. We introduce a third
-+                // inequality to solve this issue:
-+                // (3) `len <= region_len`
-+                // Using this third inequality, we can rewrite the second one as:
-+                // (4) `ptr - region_ptr <= region_len - len`
-+                // This fourth inequality is equivalent to the second one but doesn't overflow when
-+                // the first and third inequalities hold.
-+                region_ptr <= ptr && len <= region_len && ptr - region_ptr <= region_len - len
-+            })
-+        })
-+    }
- }
- 
- /// Type to uniquely identify a callback subscription across all drivers.
--- a/src/ctap/storage.rs
+++ b/src/ctap/storage.rs
@@ -138,11 +138,14 @@ const PAGE_SIZE: usize = 0x100;
 #[cfg(not(feature = "ram_storage"))]
 const PAGE_SIZE: usize = 0x1000;
 
+// We have the following layout:
+// 0x00000-0x2ffff: Tock
+// 0x30000-0x3ffff: Padding
+// 0x40000-0xbffff: App
+// 0xc0000-0xfffff: Store
+const STORE_ADDR: usize = 0xC0000;
+const STORE_SIZE_LIMIT: usize = 0x40000;
 const STORE_SIZE: usize = NUM_PAGES * PAGE_SIZE;
-
-#[cfg(not(any(test, feature = "ram_storage")))]
-#[link_section = ".app_state"]
-static STORE: [u8; STORE_SIZE] = [0xff; STORE_SIZE];
 
 impl PersistentStore {
     /// Gives access to the persistent store.
@@ -164,9 +167,11 @@ impl PersistentStore {
 
     #[cfg(not(any(test, feature = "ram_storage")))]
     fn new_prod_storage() -> Storage {
+        // This should ideally be a compile-time assert, but Rust doesn't have native support.
+        assert!(STORE_SIZE <= STORE_SIZE_LIMIT);
         let store = unsafe {
             // Safety: The store cannot alias because this function is called only once.
-            core::slice::from_raw_parts_mut(STORE.as_ptr() as *mut u8, STORE_SIZE)
+            core::slice::from_raw_parts_mut(STORE_ADDR as *mut u8, STORE_SIZE)
         };
         unsafe {
             // Safety: The store is in a writeable flash region.


src/ctap/storage.rs
```diff
@@ -151,6 +155,8 @@ impl PersistentStore {
     ///
     /// This should be at most one instance of persistent store per program lifetime.
     pub fn new(rng: &mut impl Rng256) -> PersistentStore {
+        // This should ideally be a compile-time assert, but Rust doesn't have native support.
+        assert!(STORE_SIZE <= STORE_SIZE_LIMIT);
```
gendx, line 159: Can we do this in a `#[test]` to avoid the runtime overhead? And if this predicate would still be false at runtime, would it lead to a memory safety issue, or just the kernel stopping the app due to the MPU catching a memory violation?

ia0: It would be an issue to run without this check because the driver doesn't check that it's in bound (and the MPU won't help when talking to the NVMC). But actually this should be considered a bug of the driver. I think it's better to do this check in the driver rather than here, so I'm moved the test.

Regarding the runtime overhead, the current one is totally negligible because it's once per process lifetime. The one when the driver does the check is also negligible but occurs at each syscall. Note that we already to this check. Now instead of checking against each writeable region of the app we check against the storage.

nrf52840dk_layout.ld
```diff
@@ -3,8 +3,10 @@
  */
 
 MEMORY {
-  /* The application region is 64 bytes (0x40) */
-  FLASH (rx) : ORIGIN = 0x00040040, LENGTH = 0x000BFFC0
+  /* The application region is 64 bytes (0x40) and we reserve 0x40000 at the end
+   * of the flash for the persistent storage.
+   */
```
gendx, line 8: What about the layout for the other boards we support? Is this layout file used for all of them (in which case we should probably rename it `nrf52840_layout`, as a property of the chip rather than the board)? How are dfu-based boards affected?

cc @jmichelp

patches/tock/01-persistent-storage.patch
```diff
@@ -42,7 +42,7 @@ index fe493727..105f7120 100644
  
      platform.pconsole.start();
 diff --git a/chips/nrf52/src/nvmc.rs b/chips/nrf52/src/nvmc.rs
-index 5abd2d84..5a726fdb 100644
+index 60fc2da8..77e7423d 100644
```
gendx, line 45: Is there a plan to upstream (part of) this patch to mainline Tock? Or is it too much WIP?

ia0: The initial plan was https://github.com/tock/tock/pull/1457 now obsoleted in favor of https://github.com/tock/tock/pull/1467 but currently blocked by design considerations regarding chips implementing syscalls. There's also some chance for this to progress given https://groups.google.com/g/tock-dev/c/aduN7fHWXdI/m/Lk3bmMVxBQAJ. The first step to provide applications access to a persistent storage is to check that access is permitted. Maybe once this first step is resolved, progress will unblock on the second step.

nrf52840dk_layout.ld
jmichelp: Indeed the layout depends on the chip, not the board it's soldered on. And because we only support the nrf52840 chip so far, that's the only layout file we need (it's used to compile the app, not the kernel).

If we rename it, we need to also change `deploy.py` accordingly.

ia0: I created #105 for that.

ia0:
The last commit permits to save 1 page in the binary by removing the alignment constraint of the writable flash regions. (It also reduces the diff with libtock-rs which doesn't support writable flash regions.)

src/ctap/storage.rs
```diff
@@ -138,11 +138,14 @@ const PAGE_SIZE: usize = 0x100;
...
-
+// We have the following layout:
+// 0x00000-0x2ffff: Tock
+// 0x30000-0x3ffff: Padding
+// 0x40000-0xbffff: App
+// 0xc0000-0xfffff: Store
 #[cfg(not(any(test, feature = "ram_storage")))]
-#[link_section = ".app_state"]
-static STORE: [u8; STORE_SIZE] = [0xff; STORE_SIZE];
+const STORE_ADDR: usize = 0xC0000;
```
jmichelp, line 147: Not a huge fan of having flash addresses that are hard-coded.
Maybe we should move this in the ldscript file and import the symbol instead?

ia0: I'm also not a fan of hard-coded stuff but we can't have it in the linker script of the application because this should be in sync with the kernel. However, I moved the truth value to the kernel only. The application reads it from the kernel and is completely abstract over this information. The information is currently in 2 places:
- In the flash driver
- In the MPU configuration

When Tock supports persistent storage, we will be able to remove those 2 hard-coded values:
- The flash driver should ask the process whether the region is valid (like it was the case with writable flash regions)
- The kernel should configure the MPU properly when spawning a process (this was automatic with writable flash regions)

jmichelp: It was briefly mentioned in the TockOS Slack that using the existing `nonvolatile_storage_driver.rs` capsule in combination with a [syscall filter](https://github.com/tock/tock/blob/3242eac532d999cd6fe15bb508bc84ba62ce4a5b/kernel/src/platform/mod.rs#L55) might allow us to implement our storage system that will only be available for the intended app.

ia0: Those APIs don't support erasing a page. I did an analysis a [long time ago](https://github.com/tock/tock/pull/1457#issuecomment-551897059) and the situation did not change since then (except for this email thread that might end up somewhere in the next year or two).

We know we want to improve on the current state of this PR, but it can't be done until Tock supports persistent storage correctly. I don't see this happening in the next 6 months, because those are design decisions that take time. So if we believe we need this PR before that, we should submit, otherwise we can wait.

patches/tock/01-persistent-storage.patch
```diff
@@ -285,39 +313,41 @@ index 5abd2d84..5a726fdb 100644
...
++++ b/kernel/src/process.rs
+@@ -1604,6 +1604,33 @@ impl<C: 'static + Chip> Process<'a, C> {
+             return Ok((None, 0));
+         }
  
- /// Type to uniquely identify a callback subscription across all drivers.
++        // Allocate MPU region for storage. The storage is currently static and readable by all
++        // processes. This will be fixed once Tock supports persistent storage outside the
++        // application flash (i.e. not the current writeable flash regions support).
++        const STORAGE_PTR: usize = 0xc0000;
```
jmichelp, line 327: The change is much better with the address and size only being put in the kernel.
The main problem now is that we can't even compile/run a Tock kernel that is not for the nRF52840 chip because of this address. I still think this would be better to have it in the `nrf52840_layout.ld`. Just like there is a `FLASH` and a `SRAM` portion, we should be able to define a `STORAGE` area and use the symbols in the kernel directly. The tricky part (I'm not familiar enough with ldscript files) is to handle the cases where the area doesn't exist.
Another approach that is being used in `libtock-rs` is to use environment variables at compile time. This might be less complicated and would allow to handle the case where the kernel doesn't need a storage area. WDYT?

ia0: Oh I see the problem. I didn't anticipate that the MPU part would be an issue for other boards. The part in NVMC is fine because it's only for nrf52.

Indeed, trying to use the linker script through the TBF could be a solution. I need to look into it. I might have seen something relevant in Tock on Titan about that. I need to check again.

I guess the environment variable should be a last resort solution right?

jmichelp: As you said, it's for `nrf52` which is a big family and `nrf52840` is the biggest one in this chip family in terms of flash. So having a hard-coded address is likely to cause a kernel panic at runtime if the address doesn't exist because the flash is smaller.

Changing the linker script is one option. As I said, it might be tricky to handle a non-existing symbol. This may require us to define the symbol for all the boards. But on the other hand, it may allow us to do some sanity checks that the defined area for storage will work with the MPU (it has the [constraint](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dai0179b/CHDFDFIG.html) that the base address of the region must be aligned to a multiple of its size).

The env variable gives us more flexibility and we fully control the environment with the `deploy.py` script. That's probably a quickest option that will let us investigate the linker script road at a later stage.

ia0: I don't see a way to use the linker script without having to change elf2tab. I see those 2 solutions (assuming we want to define the storage in the linker script):
- We change the semantics of writable flash regions to actually represent the out-of-app persistent storage. We probably don't need to change much on kernel side because writable flash regions don't have extensive support (but we would still need to review where it's used). We need to change elf2tab to read the storage pointer and size in the ELF and output them as a writable flash region.
- We add a new TbfHeaderType for persistent storage. We implement the parsing of this header in the Kernel. We implement the generation of this header in elf2tab and populate it from the ELF.

In both cases, we need a way to output the storage address and size in the ELF. I guess we could use a section for that. We would filter out this section in elf2tab and only use it to know how to write the TBF headers.

I think the env variable looks actually like a much simpler approach, unless we find a way that doesn't need to modify elf2tab.

ia0: The last commit defines the storage locations in the board, which doesn't need to modify elf2tab as it would be the case with the linker script. So the storage locations are now defined in a single place which is board-specific.

I also added support for cases where the persistent storage spans multiple non-contiguous locations, like it could be the case if the flash had multiple banks and one would want to use the last pages of each bank for the persistent storage.

The diff between this state and possible future Tock states is:
- The persistent storage is readable by all apps. For that we would need stable app ids or define the persistent storage in the app (which needs to modify elf2tab).
- The persistent storage is statically defined by the board. It could instead be defined by metadata in the list of apps and thus dynamically edited by tockloader. It could also be defined inside each app if support for multiple apps accessing the same persistent storage is not needed.

Remaining to do: Use the syscall filter to prevent the application from mutating the flash outside the persistent storage. This test is currently done by the driver.

ia0: It's actually not possible to use the syscall filter because we can't access the driver state from the syscall filter. We need to know the length of the allowed slice. I'll try to see if there is some issue regarding syscall filter to mention this problem.

jmichelp: I think this hook is a first step at filtering syscall but there are more PRs planned around that. We can revisit the syscall filter at a later stage

ia0: By the way, I just asked [here](https://github.com/tock/tock/issues/1350#issuecomment-628023606) whether this would be supported in the future.

ia0: To move the storage test from the driver to the syscall driver, I had to change the driver API in the following ways:
- Move the storage locations accessors from the driver to `memop` syscalls.
- Add a length argument to the write slice and erase page commands. This is redundant but works around the current syscall filter limitation.

Now that the NVMC driver only uses the app to access the grants and not to access the writable flash regions, I'll try to resuscitate https://github.com/tock/tock/pull/1467. If accepted it would simplify a bit the persistent storage patch we have to only address the future persistent storage support of Tock.

deploy.py
```diff
@@ -493,6 +505,40 @@ def clear_apps(self):
...
+    # pylint: disable=g-import-not-at-top,import-outside-toplevel
+    import intelhex
+    storage_hex = intelhex.IntelHex()
+    storage_hex.frombytes(storage, offset=board_props.storage_address)
+    os.makedirs("target", exist_ok=True)
+    storage_file = "target/{}_storage.hex".format(self.args.board)
+    storage_hex.tofile(storage_file, format="hex")
+    # Flash the intelhex file
+    info("Flashing the persistent storage HEX file")
+    if self.args.programmer == "pyocd":
```
jmichelp, line 534: pycod supports the `erase` command.
So in this case, it's probably better to have:
```python
self.checked_command([
  "pycod", "erase", "--target={}".format(board_props.pycod_target),
  "--sector {}+{}".format(board_props.storage_address, board_props.storage_size)
])
```

deploy.py
```diff
@@ -493,6 +505,40 @@ def clear_apps(self):
...
+        fatal("Couldn't erase the persistent storage: {}".format(str(e)))
+      return 0
+    # Create an intelhex file otherwise
+    info("Creating the persistent storage HEX file")
+    # pylint: disable=g-import-not-at-top,import-outside-toplevel
+    import intelhex
+    storage_hex = intelhex.IntelHex()
+    storage_hex.frombytes(storage, offset=board_props.storage_address)
+    os.makedirs("target", exist_ok=True)
+    storage_file = "target/{}_storage.hex".format(self.args.board)
```
jmichelp, line 530: This should rather be a tempfile:
```python
import tempfile
storage_file = tempfile.NamedTemporaryFile(delete=False)
storage_hex.tofile(storage_file, format="hex")
storage_file.close()
[...]
os.remove(storage_file.name)
```

deploy.py
ia0: Nice. Updated (assuming pycod -> pyocd).

deploy.py
ia0: I wanted to do that first like in tockloader, but I've seen that they do a lot of weird stuff to support windows. However in your version you're fine keeping the file (`delete=False`) so I guess it would work around some issues. Anyway we don't need the file anymore.

deploy.py
```diff
@@ -508,33 +508,23 @@ def clear_apps(self):
...
-    storage_file = "target/{}_storage.hex".format(self.args.board)
-    storage_hex.tofile(storage_file, format="hex")
-    # Flash the intelhex file
-    info("Flashing the persistent storage HEX file")
     if self.args.programmer == "pyocd":
       self.checked_command([
-          "pyocd", "flash", "--target={}".format(board_props.pyocd_target),
-          "--format=hex", "--erase=auto", storage_file
+          "pyocd", "erase", "--target={}".format(board_props.pyocd_target),
+          "--sector {}+{}".format(board_props.storage_address,
```
jmichelp, line 526: Oops, forgot to split the arguments :)
`"--sector", "{}+{}".format(...)`

deploy.py
jmichelp: In normal mode, a tempfile is deleted when you close the file. Here, to avoid complex logic, we don't automatically delete the file but we delete it manually when we don't need it anymore (`os.remove()`)

deploy.py
ia0: Oh right, I missed that too. I used the more conventional notation for arguments which is also used for `--target`.

jmichelp: Except that this would be wrong.
`--sector` is a parameter setting the erase mode (can also be `--chip`, `--mass`).
The `{start}+{size}` parameter is the first element of the list.
So we really need to split the parameters here.

ia0: I see, the flag name is confusing, thanks for the explanation. Fixed.

ia0:
> pyocd fails after erasing because it can't resume the CPU

Not sure if related, but when initially testing JLink, I had a JLink session open while running the deploy script. I used the session to write data before running the deploy script and reading the data after to make sure it was erased. But apparently this creates some undetected interactions between the JLink session created from tockloader and the local one I have in another shell. As a consequence the data was not correctly erased. However, when closing the local JLink session before running the deploy script and restarting it after, it all worked well. Maybe there is something similar with pyocd. Note that for JLink, there is no error in any of the sessions (the local one and the tockloader one). It's just that it's not correctly erased. Took me some time to figure it out.

ia0:
> I have to do more tests on my side

Should I wait for them before submitting?

jmichelp:
Testing will take me a while because ideally we would like to catching errors related to page erasure or device connection but resuming the halted CPU is not critical. Not sure pyocd is able to do that :)

ia0:
Ok, I guess it means I can submit while this is being worked out.

I tested the latest commit:
```
./reset.sh
./setup.sh
./deploy.py --board=nrf52840dk --opensk --panic-console --clear-storage
# Run the test suite -> all green
# Dump the storage: savebin before c0000 14000
./deploy.py --board=nrf52840dk --dont-clear-apps --no-app --no-tockos --clear-storage
# Dump the storage: savebin after c0000 14000
# Compare before and after -> expected diff
# - after has only master keys, pin retries, and storage page metadata
# - it's not erased because the app has booted and initialized the storage
```

Issue #106, opened 2020-05-06
Implement CTAP 2.1


kaczmarczyck:
The review draft for CTAP 2.1 adds new commands and requires seom changes in the old ones.
https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html
We want to support new features like deleting resident keys.

Our access to a more recent version of CTAP 2.1 allows us to prototype those new features and give feedback on the command design. This issue tracks our progress. Please be aware that these commands are still exprimental and the implementation can change in the future. Do not rely on them!

kaczmarczyck:
The current implementation in the `develop` branch follows [review draft 2 of CTAP 2.1](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html).

PRs merged for finishing support:

- #254
- #256 
- #263 
- #265 
- #267 
- #273 

Current WIP:
#274 

Missing work:

- updating all existing command algorithms
- `alwaysUv` for `MakeCredential` and `GetAssertion`
- `toggleAlwaysUv` in `authenticatorConfig`
- enterprise attestation
- `enableEnterpriseAttestation` in `authenticatorConfig`
- PIN Protocol V2
- all potential changes after review draft 2

kaczmarczyck:
Features around `alwaysUv`, enterprise attestation and PIN protocol V2 (PRs for the latter: #293, #295, #296) are implemented. Missing work:
- remaining updates for existing command algorithms
- missing changes from [review draft 3](https://fidoalliance.org/specs/fido-v2.1-rd-20210309/fido-client-to-authenticator-protocol-v2.1-rd-20210309.html)

kaczmarczyck:
CTAP2.1 changes are complete with #297 and #299 (to be reviewed as of now). Not all features are thoroughly tested yet.

Sami32:
Any update on this?

It seem that features have been tested and that CTAP 2.1 get its final version:
https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-errata-20220621.html

kaczmarczyck:
Hi! We haven't noticed any problems with our implementation of CTAP 2.1 yet. It will take more time for an official release, but you can try out the `develop` branch if you are curious about the current state. Is there anything specific you are interested in?

Sami32:
Hi!
Sorry for my late answer.
[...]

PR #107, opened 2020-05-06
Credential management

kaczmarczyck:
New command for #106 . This PR contains all changes for reference, and will be split into smaller chunks. The commit `6724acf` moves all changes to a separate file. While they belong to `mod` conceptually, this file has grown too large already and the commit is for opening discussion about if we want to split `mod` up like that.

### DIFF
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -62,8 +62,10 @@ impl SecKey {
         // - https://www.secg.org/sec1-v2.pdf
     }
 
-    // DH key agreement method defined in the FIDO2 specification, Section 5.5.4. "Getting
-    // sharedSecret from Authenticator"
+    /// Creates a shared key using the Diffie Hellman key agreement.
+    ///
+    /// The key agreement is defined in the FIDO2 specification,
+    /// Section 6.5.5.4. "Obtaining the Shared Secret"
     pub fn exchange_x_sha256(&self, other: &PubKey) -> [u8; 32] {
         let p = self.exchange_raw(other);
         let mut x: [u8; 32] = [Default::default(); 32];
@@ -83,12 +85,13 @@ impl PubKey {
         self.p.to_bytes_uncompressed(bytes);
     }
 
+    /// Creates a new PubKey from their coordinates.
     pub fn from_coordinates(x: &[u8; NBYTES], y: &[u8; NBYTES]) -> Option<PubKey> {
         PointP256::new_checked_vartime(Int256::from_bin(x), Int256::from_bin(y))
             .map(|p| PubKey { p })
     }
 
-    // Writes the coordinates into the passed in arrays.
+    /// Writes the coordinates into the passed in arrays.
     pub fn to_coordinates(&self, x: &mut [u8; NBYTES], y: &mut [u8; NBYTES]) {
         self.p.getx().to_int().to_bin(x);
         self.p.gety().to_int().to_bin(y);
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -88,6 +88,7 @@ impl PubKey {
             .map(|p| PubKey { p })
     }
 
+    // Writes the coordinates into the passed in arrays.
     pub fn to_coordinates(&self, x: &mut [u8; NBYTES], y: &mut [u8; NBYTES]) {
         self.p.getx().to_int().to_bin(x);
         self.p.gety().to_int().to_bin(y);
--- a/libraries/crypto/src/ecdsa.rs
+++ b/libraries/crypto/src/ecdsa.rs
@@ -21,11 +21,14 @@ use super::rng256::Rng256;
 use super::{Hash256, HashBlockSize64Bytes};
 use alloc::vec;
 use alloc::vec::Vec;
+#[cfg(test)]
+use arrayref::array_mut_ref;
 #[cfg(feature = "std")]
 use arrayref::array_ref;
-use arrayref::{array_mut_ref, mut_array_refs};
-use cbor::{cbor_bytes, cbor_map_options};
+use arrayref::mut_array_refs;
 use core::marker::PhantomData;
+
+pub const NBYTES: usize = int256::NBYTES;
 
 #[derive(Clone, PartialEq)]
 #[cfg_attr(feature = "derive_debug", derive(Debug))]
@@ -38,6 +41,7 @@ pub struct Signature {
     s: NonZeroExponentP256,
 }
 
+#[cfg_attr(feature = "derive_debug", derive(Clone))]
 pub struct PubKey {
     p: PointP256,
 }
@@ -242,35 +246,10 @@ impl PubKey {
         representation
     }
 
-    // Encodes the key according to CBOR Object Signing and Encryption, defined in RFC 8152.
-    pub fn to_cose_key(&self) -> Option<Vec<u8>> {
-        const EC2_KEY_TYPE: i64 = 2;
-        const P_256_CURVE: i64 = 1;
-        let mut x_bytes = vec![0; int256::NBYTES];
-        self.p
-            .getx()
-            .to_int()
-            .to_bin(array_mut_ref![x_bytes.as_mut_slice(), 0, int256::NBYTES]);
-        let x_byte_cbor: cbor::Value = cbor_bytes!(x_bytes);
-        let mut y_bytes = vec![0; int256::NBYTES];
-        self.p
-            .gety()
-            .to_int()
-            .to_bin(array_mut_ref![y_bytes.as_mut_slice(), 0, int256::NBYTES]);
-        let y_byte_cbor: cbor::Value = cbor_bytes!(y_bytes);
-        let cbor_value = cbor_map_options! {
-            1 => EC2_KEY_TYPE,
-            3 => PubKey::ES256_ALGORITHM,
-            -1 => P_256_CURVE,
-            -2 => x_byte_cbor,
-            -3 => y_byte_cbor,
-        };
-        let mut encoded_key = Vec::new();
-        if cbor::write(cbor_value, &mut encoded_key) {
-            Some(encoded_key)
-        } else {
-            None
-        }
+    // Writes the coordinates into the passed in arrays.
+    pub fn to_coordinates(&self, x: &mut [u8; NBYTES], y: &mut [u8; NBYTES]) {
+        self.p.getx().to_int().to_bin(x);
+        self.p.gety().to_int().to_bin(y);
     }
 
     #[cfg(feature = "std")]
--- a/libraries/crypto/src/ecdsa.rs
+++ b/libraries/crypto/src/ecdsa.rs
@@ -62,10 +62,11 @@ impl SecKey {
         }
     }
 
-    // ECDSA signature based on a RNG to generate a suitable randomization parameter.
-    // Under the hood, rejection sampling is used to make sure that the randomization parameter is
-    // uniformly distributed.
-    // The provided RNG must be cryptographically secure; otherwise this method is insecure.
+    /// Creates an ECDSA signature based on a RNG.
+    ///
+    /// Under the hood, rejection sampling is used to make sure that the
+    /// randomization parameter is uniformly distributed. The provided RNG must
+    /// be cryptographically secure; otherwise this method is insecure.
     pub fn sign_rng<H, R>(&self, msg: &[u8], rng: &mut R) -> Signature
     where
         H: Hash256,
@@ -81,8 +82,7 @@ impl SecKey {
         }
     }
 
-    // Deterministic ECDSA signature based on RFC 6979 to generate a suitable randomization
-    // parameter.
+    /// Creates a deterministic ECDSA signature based on RFC 6979.
     pub fn sign_rfc6979<H>(&self, msg: &[u8]) -> Signature
     where
         H: Hash256 + HashBlockSize64Bytes,
@@ -105,8 +105,10 @@ impl SecKey {
         }
     }
 
-    // Try signing a curve element given a randomization parameter k. If no signature can be
-    // obtained from this k, None is returned and the caller should try again with another value.
+    /// Try signing a curve element given a randomization parameter k.
+    ///
+    /// If no signature can be obtained from this k, None is returned and the
+    /// caller should try again with another value.
     fn try_sign(&self, k: &NonZeroExponentP256, msg: &ExponentP256) -> Option<Signature> {
         let r = ExponentP256::modn(PointP256::base_point_mul(k.as_exponent()).getx().to_int());
         // The branching here is fine because all this reveals is that k generated an unsuitable r.
@@ -246,7 +248,7 @@ impl PubKey {
         representation
     }
 
-    // Writes the coordinates into the passed in arrays.
+    /// Writes the coordinates into the passed in arrays.
     pub fn to_coordinates(&self, x: &mut [u8; NBYTES], y: &mut [u8; NBYTES]) {
         self.p.getx().to_int().to_bin(x);
         self.p.gety().to_int().to_bin(y);
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -1,0 +1,900 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::command::AuthenticatorCredentialManagementParameters;
+use super::data_formats::{
+    CoseKey, CredentialManagementSubCommand, CredentialManagementSubCommandParameters,
+    PublicKeyCredentialDescriptor, PublicKeyCredentialRpEntity, PublicKeyCredentialSource,
+    PublicKeyCredentialUserEntity,
+};
+use super::pin_protocol_v1::{PinPermission, PinProtocolV1};
+use super::response::{AuthenticatorCredentialManagementResponse, ResponseData};
+use super::status_code::Ctap2StatusCode;
+use super::storage::PersistentStore;
+use super::timed_permission::TimedPermission;
+use super::{check_command_permission, StatefulCommand, STATEFUL_COMMAND_TIMEOUT_DURATION};
+use alloc::collections::BTreeSet;
+use alloc::string::String;
+use alloc::vec;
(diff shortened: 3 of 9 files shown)


src/ctap/storage.rs
```diff
@@ -285,6 +319,11 @@ impl PersistentStore {
             .count()
     }
 
+    #[cfg(feature = "with_ctap2_1")]
+    pub fn remaining_credentials(&self) -> usize {
+        MAX_SUPPORTED_RESIDENTIAL_KEYS - self.count_credentials()
```
ia0: We might want to use `MAX_SUPPORTED_RESIDENTIAL_KEYS.checked_sub(self.count_credentials()).ok_or(XXX)` at some point. But for that we need to update all usages of the persistent storage to support failure. I'll do that in a separate PR once we know which error code we want to return for the different kinds of errors.

kaczmarczyck: Fixed after the rebase (more like rework).

kaczmarczyck:
The implementation is heavily reworked after the rebase!

libraries/crypto/src/ecdh.rs
```diff
@@ -88,6 +88,7 @@ impl PubKey {
             .map(|p| PubKey { p })
     }
 
+    // Writes the coordinates into the passed in arrays.
```
ia0: I wonder if it's worth already starting to use `///` for  documentation. If the file was self-consistent, using `//` would make sense, but there's even usage of `/** **/` in this file.

libraries/crypto/src/ecdsa.rs
```diff
@@ -38,6 +41,7 @@ pub struct Signature {
     s: NonZeroExponentP256,
 }
 
+#[cfg_attr(feature = "derive_debug", derive(Clone))]
```
ia0, line 44: Why should `derive(Clone)` be protected by `derive_debug` feature?

Also, I don't think gating derives has any benefit. The compiler is smart enough to remove dead code. This just adds noise.

src/ctap/storage.rs
```diff
@@ -201,6 +203,48 @@ impl PersistentStore {
...
+        iter_result?;
+        if credentials.len() > 1 {
+            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        }
+        credentials
+            .pop()
+            .ok_or(Ctap2StatusCode::CTAP2_ERR_NO_CREDENTIALS)
+    }
+
+    /// Deletes a credential. Returns an error if the credential was not found.
```
ia0: The usual style for documentation (not only in Rust) is one short line of description followed by a detailed description. So I would suggest here and below to have something like:

```rust
/// Deletes a credential.
///
/// # Errors
///
/// Returns `CTAP2_ERR_NO_CREDENTIALS` if the credential is not found.
```

src/ctap/storage.rs
```diff
@@ -201,6 +203,48 @@ impl PersistentStore {
...
+    #[cfg(feature = "with_ctap2_1")]
+    pub fn delete_credential(&mut self, credential_id: &[u8]) -> Result<(), Ctap2StatusCode> {
+        let (key, _) = self.find_credential_item(credential_id)?;
+        Ok(self.store.remove(key)?)
+    }
+
+    /// Update a credential's user information. Returns an error if the credential was not found.
```
ia0: The rest of the documentation uses third person present tense (like Google style):
https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#appendix-a-full-conventions-text

src/ctap/storage.rs
```diff
@@ -225,8 +269,18 @@ impl PersistentStore {
         Ok(result)
     }
 
+    /// Returns the list of all credentials.
+    #[cfg(feature = "with_ctap2_1")]
+    pub fn all_credentials(&self) -> Result<Vec<PublicKeyCredentialSource>, Ctap2StatusCode> {
```
ia0: Unless we have good reasons to believe that all credentials fit in RAM, I would make `iter_credentials` public and use it instead. The same concern may apply to `filter_credential` above but I expect the resulting vector to be very often smaller than 10 credentials. While here it could easily be 100 credentials.

src/ctap/storage.rs
```diff
@@ -640,6 +702,74 @@ mod test {
         assert!(persistent_store.count_credentials().unwrap() > 0);
     }
 
+    #[cfg(feature = "with_ctap2_1")]
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_delete() {
```
ia0: test_delete_credential?

src/ctap/storage.rs
```diff
@@ -640,6 +702,74 @@ mod test {
         assert!(persistent_store.count_credentials().unwrap() > 0);
     }
 
+    #[cfg(feature = "with_ctap2_1")]
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
```
ia0: If we can move this to the problematic line it would be better to reduce the scope.

src/ctap/storage.rs
```diff
@@ -640,6 +702,74 @@ mod test {
...
+        let mut rng = ThreadRng256 {};
+        let mut persistent_store = PersistentStore::new(&mut rng);
+        assert_eq!(persistent_store.count_credentials().unwrap(), 0);
+
+        // To make this test work for bigger storages, implement better int -> Vec conversion.
+        assert!(MAX_SUPPORTED_RESIDENTIAL_KEYS < 256);
+        let mut credential_ids = vec![];
+        for i in 0..MAX_SUPPORTED_RESIDENTIAL_KEYS {
+            let credential_source =
+                create_credential_source(&mut rng, "example.com", vec![i as u8]);
```
ia0: We could as well do `i.as_ne_bytes().to_vec()` and have `i: usize` so we don't have to check for 256.

src/ctap/storage.rs
```diff
@@ -640,6 +702,74 @@ mod test {
...
+        let stored_credential = persistent_store
+            .find_credential("example.com", &credential_id, false)
+            .unwrap()
+            .unwrap();
+        assert_eq!(stored_credential.user_name, Some("name".to_string()));
+        assert_eq!(
+            stored_credential.user_display_name,
+            Some("display_name".to_string())
+        );
+        assert_eq!(stored_credential.user_icon, Some("icon".to_string()));
```
ia0: What about cloning `user` before and using `user.user_icon` here? (same for name and display name) This avoids constant duplication.

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,900 @@
...
+    persistent_store: &PersistentStore,
+    stateful_command_permission: &mut TimedPermission,
+    stateful_command_type: &mut Option<StatefulCommand>,
+    now: ClockValue,
+) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
+    let credentials = persistent_store.all_credentials()?;
+    let mut rp_set = BTreeSet::new();
+    for cred in credentials {
+        rp_set.insert(cred.rp_id);
+    }
```
ia0, line 140: This loop could be:

```rust
let mut iter_result = Ok(());
for (_, cred) in persistent_store.iter_credentials()? {
    rp_set.insert(cred.rp_id);
}
iter_result?;
```

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,900 @@
...
+    stateful_command_type: &mut Option<StatefulCommand>,
+    now: ClockValue,
+) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
+    check_command_permission(stateful_command_permission, now)?;
+    let rp_id = if let Some(StatefulCommand::EnumerateRps(rp_ids)) = stateful_command_type {
+        rp_ids.pop().ok_or(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED)?
+    } else {
+        return Err(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED);
+    };
+    enumerate_rps_response(Some(rp_id), None)
```
ia0: I don't understand what it should do, but it looks like this should be inside the then-branch and the `None` should be `Some(rp_ids)`.

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,900 @@
...
+    sub_command_params: CredentialManagementSubCommandParameters,
+    now: ClockValue,
+) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
+    let rp_id_hash = sub_command_params
+        .rp_id_hash
+        .ok_or(Ctap2StatusCode::CTAP2_ERR_NO_CREDENTIALS)?;
+    // TODO(kaczmarczyck) filter by rp_id_hash inside of storage?
+    let mut rp_credentials: Vec<PublicKeyCredentialSource> = persistent_store
+        .all_credentials()?
+        .into_iter()
```
ia0: Same, we could use `iter_credentials` here.

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,900 @@
...
+        .collect();
+    let total_credentials = rp_credentials.len();
+
+    let credential = rp_credentials
+        .pop()
+        .ok_or(Ctap2StatusCode::CTAP2_ERR_NO_CREDENTIALS)?;
+    if total_credentials > 1 {
+        *stateful_command_permission =
+            TimedPermission::granted(now, STATEFUL_COMMAND_TIMEOUT_DURATION);
+        *stateful_command_type = Some(StatefulCommand::EnumerateCredentials(rp_credentials));
```
ia0, line 202: Mmmm... this RAM usage is kind of "persistent" so might be more costly than the `filter_credential` case. We might want to consider only handling "keys" instead of credentials when we deal with credential sets, and only read the given key if needed.

We could have helper functions like:
```rust
fn get_credential_keys(&self, filter: impl Fn(&PublicKeyCredentialSource) -> bool) -> Result<Vec<usize>, Ctap2StatusCode>;
fn read_credential(&self, key: usize) -> Result<PublicKeyCredentialSource, Ctap2StatusCode>;
```

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,900 @@
...
+    let credential = if let Some(StatefulCommand::EnumerateCredentials(rp_credentials)) =
+        &mut stateful_command_type
+    {
+        rp_credentials
+            .pop()
+            .ok_or(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED)?
+    } else {
+        return Err(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED);
+    };
+    enumerate_credentials_response(credential, None)
```
ia0: Same as before, I'm surprised we return `None` even if `rp_credentials` might be longer than 1.

libraries/crypto/src/ecdsa.rs
kaczmarczyck: I think we noticed that at least guarding `Debug` is important to reduce the binary size significantly. I didn't check if this is also true for `Clone` etc, but I'd say we should keep them?

ia0: Who noticed and how did they notice? If the binary size is bigger when adding dead-code, it should be reported to the compiler/linker.

Here is how I test it:

```
% git rev-parse HEAD
deeabe026f07e714a118ab365914b5f9b7d6de34
% ./deploy.py --board=nrf52840dk --opensk --panic-console --programmer=none
[... omitted ...]
% stat -c%s target/thumbv7em-none-eabi/release/ctap2                       
411632
% edit Cargo.toml
- panic_console = ["lang_items/panic_console"]
+ panic_console = ["lang_items/panic_console", "crypto/derive_debug"]
% ./deploy.py --board=nrf52840dk --opensk --panic-console --programmer=none
[... omitted ...]
% stat -c%s target/thumbv7em-none-eabi/release/ctap2                       
411448
```

The size is actually smaller, not sure why. But I would not be surprised if trying to manually optimize stuff gives worse results than letting the compiler do its job.

src/ctap/credential_management.rs
kaczmarczyck: I can move it inside the if!

The `None` here is for the response parameter with the number of total credentials. This response parameter is only answered in the `Begin` command, not in `Next`.

src/ctap/credential_management.rs
kaczmarczyck: Same as above, I'll move the function call inside the `if`. The `None` is intended.

See [the spec](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authenticatorCredentialManagement), 6.8.3 and 6.8.4.

src/ctap/credential_management.rs
ia0: Oh I see. The remaining list is actually taken as `&mut` and mutated in place. Ok it's alright then.

src/ctap/credential_management.rs
ia0: Yeah, I confused this parameter with the remaining items to return. All good.

libraries/crypto/src/ecdh.rs
kaczmarczyck: I added `///` in the library files I touched and the new file `credential_management.rs`.

src/ctap/storage.rs
kaczmarczyck: Done

src/ctap/storage.rs
kaczmarczyck: Outside of the other changes I made, is this fixed by changing `Update` to `Updates`?

src/ctap/storage.rs
kaczmarczyck: I removed `all_credentials` in both occasions and made the iterator public.

src/ctap/storage.rs
kaczmarczyck: Renamed.

src/ctap/storage.rs
kaczmarczyck: Resolved by the below change to correctly choose a `user_handle` everywhere.

src/ctap/storage.rs
kaczmarczyck: Done

src/ctap/storage.rs
kaczmarczyck: That's nicer, done!

src/ctap/credential_management.rs
kaczmarczyck: As mentioned above, `iter_credentials` is now public.

src/ctap/credential_management.rs
kaczmarczyck: Done.

src/ctap/credential_management.rs
kaczmarczyck: (Note: This comment wasn't worked on. To be resolved.)

libraries/crypto/src/ecdsa.rs
kaczmarczyck: I think that was @gendx , i.e. see his comment in [libtock](https://github.com/google/OpenSK/blob/67a94c8067f12e1c642ee338b81722495240659c/third_party/libtock-drivers/src/result.rs#L7). The discussion is pretty old I think, so maybe we have to re-evaluate?

This feels like it's worth a new issue / PR if we generally move to no `cfg` for `derive`?

ia0: Yes, this question is orthogonal to the PR. I'll create a separate issue.

Also, the comment you point out is unrelated to what I'm saying. That comment makes sense, because it's about avoiding the need of error types implementing the `Debug` trait. What I'm saying is that we don't need to feature-gate generated dead-code, like deriving the `Debug` trait.

ia0 (commented):
Overall looks good to me, but it's quite big. I have the feeling we need stronger testing strategy than review, unit tests, and test suite to get some confidence about the correctness.

src/ctap/credential_management.rs
kaczmarczyck: I generally think handling keys only is reasonable. A few comments on that:

1. The same memory usage is currently required by `Get(Next)Assertion`, also storing all `PublicKeyCredentialSource` for a given `RP ID`. It doesn't seem to be problematic so far, given the testing during certification.
2. Having a general filter function could be nicer, but we currently only need to filter by `RP ID` and `RP ID hash` (and this should probably be fixed in the spec to be the same).
3. The credentials have to fit into memory for sorting them like this:

> applicable_credentials.sort_unstable_by_key(|c| c.creation_order);

If the only issue is the almost-persistent RAM usage, then that's fine, but it would be nice to address the issue completely? Also some other code in `GetAssertion` would need changes. I'll propose a plan in a general comment below.

libraries/crypto/src/ecdsa.rs
kaczmarczyck: #251 to continue this discussion.

kaczmarczyck:
You mentioned a potential improvement / optimization of the storage API and RAM usage, PR size and testing. What do you think of the following plan:

For storage: I keep the storage in this PR as is for reference (see below), and work on improving RAM usage in a new PR.

For PR size: This PR is pretty old, and was blocked by the 2.1 spec to land. Therefore, it has some ballast. I need heavier code changes with #252 anyway, and wanted to split out some of the changes made here for some time. But, I wanted to finish the storage review with you first so you don't have to review it again. I'll keep this PR open for reference, and send out a few smaller PRs for crypto, storage etc.

For testing: I tested this PR with the latest Chrome, since it supports credential management. It works, when you add one little change: You remove the check for PIN permissions (not the PIN check, just the permissions). This is a new 2.1 feature that doesn't seem to be supported yet. Everything else (the crypto, the command and subcommands) works. If you have any other ideas, please let me know.

src/ctap/credential_management.rs
ia0: 1. Yeah, maybe it's not problematic, but it would need some maths (how much RAM is available for this usage and how much can be used at most, we might need to say that at most 10 credentials per RPid are supported). Maybe we should start an issue to analyze and plan for RAM usage.
2. SGTM to keep as is. Actually even the more general filter function might not always be more readable.
3. This can be worked around with `Vec<(u64, usize)>` where `u64` is creation order and `usize` is the storage key.

I'll follow up on the general comment below for the rest.

ia0:
For storage: SGTM

For PR size: No problem, I wasn't saying we should split it. Just that I've most probably missed some stuff.

For testing: Ideally (is it even feasible?), we would have our implementation be a library with a thin wrapper for Tock and other side-effect dependency code, and we would have a Rust specification of CTAP (which should be very small and straight-forward code because there's no embedded-programming constraints or other performance constraints). We would then be able to fuzz our implementation against this specification (like for the persistent storage). The Rust specification could also be a reference specification for the community (including the configuration and scenario generator). We might at least want to start an issue to make our CTAP code a library, since it would enable simpler fuzzing (just crash-detection or other basic soundness properties, not correctness). And it might be easier to adapt to other backends as Tock.

src/ctap/credential_management.rs
kaczmarczyck: Solved in #254, storage code from this PR can be used in credential management.

kaczmarczyck:
Closing this PR as #256 is the new, rebased command, minus #253 and #254.

PR #108, opened 2020-05-11
Support parsing old credentials

ia0:
- [x] Test suite passes


libraries/cbor/src/macros.rs
```diff
@@ -41,18 +41,30 @@ macro_rules! cbor_map_options {
     };
 
     ( $( $key:expr => $value:expr ),* ) => {
+        cbor_extend_map_options! (
```
gendx, line 44: What's the purpose of this new macro?

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,91 @@ pub struct PublicKeyCredentialSource {
...
+    /// Contains the unknown fields when parsing a CBOR value.
+    ///
+    /// Those fields could either be deleted fields from older versions (they should have reserved
+    /// tags) or fields from newer versions (the tags should not be reserved). If this is empty,
+    /// then the parsed credential is probably from the same version (but not necessarily).
+    pub unknown_fields: BTreeMap<cbor::KeyType, cbor::Value>,
+}
+
+// We simulate protocol buffers in CBOR with maps. Each field of a message is associated with a
+// unique tag, implemented with a CBOR unsigned key.
```
gendx, line 458: There's nothing protocol-buffer-specific in this. FIDO2 already encodes many structs as maps, with string or integer keys, for example:
https://github.com/google/OpenSK/blob/8bbf42623f57f09c2c17ed5acd26f77eb49d6545/src/ctap/data_formats.rs#L1127-L1131

Protocol buffers refers to a [specific encoding of tag-value pairs](https://developers.google.com/protocol-buffers/docs/encoding), which this change doesn't implement at all (the CBOR encoding is used instead). So referring to "protocol buffers" here is misleading/confusing.

I'd simply mention "Encode credentials as a map, for versioning" or something like that.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,91 @@ pub struct PublicKeyCredentialSource {
...
+            PrivateKey => Some(private_key.to_vec()),
+            RpId => Some(credential.rp_id),
+            UserHandle => Some(credential.user_handle),
+            OtherUi => credential.other_ui,
+            CredRandom => credential.cred_random
         }
     }
 }
 
-impl TryFrom<cbor::Value> for PublicKeyCredentialSource {
```
gendx, line 473: All data formats implement the `TryFrom<cbor::Value>` to convert from CBOR values. What's the purpose of implementing a novel `parse_cbor` function instead, which isn't tied to any trait?

src/ctap/data_formats.rs
```diff
@@ -1218,11 +1243,12 @@ mod test {
...
             other_ui: None,
             cred_random: None,
+            unknown_fields: BTreeMap::new(),
         };
 
         assert_eq!(
-            PublicKeyCredentialSource::try_from(cbor::Value::from(credential.clone())),
-            Ok(credential.clone())
+            PublicKeyCredentialSource::parse_cbor(cbor::Value::from(credential.clone())),
+            Some(credential.clone())
```
gendx, line 1251: These changes shouldn't be necessary.

libraries/cbor/src/macros.rs
ia0: It permits to append elements to an existing map instead of creating one from scratch.

src/ctap/data_formats.rs
ia0: It's mostly the ideas:
- Do not reuse an tag that existed in any past versions.
- Provide access to possibly older/newer fields (to detect possible issues).

Anyone familiar with proto-buf would recognize those patterns. If we believe those are common knowledge, I can remove the mention to proto-buf and just keep the ideas.

src/ctap/data_formats.rs
ia0: It's because it looked to me that the others struct are specified by FIDO. In particular which error should be returned for which kind of issue. In the case of credentials, the kind of issue is irrelevant because it is not a user input but something coming from the persistent storage. So all errors should be something like "storage format error". We could as well use `()` as the error type. But it will add a bit of boilerplate to convert all errors to `()`.

And actually it could make sense to want different implementations for the `TryFrom<cbor::Value>` and this `parse_cbor` one. Because the first would be the FIDO version while this one would be the internal representation we use for the persistent storage. We should try to not conflate both since it will introduce future issues.

src/ctap/data_formats.rs
ia0: As discussed above.

src/ctap/data_formats.rs
gendx: I don't think this pattern (tag + value) is specific to protobufs (https://en.wikipedia.org/wiki/Type-length-value). Sure, protobufs use it (so "anyone familiar with proto-buf would recognize those patterns"), but so do many encoding formats, and I don't think knowledge of protobufs should be required to understand what's being done here.

Just saying that we replace an array by a mapping from (versioned) tags to values would be enough.

libraries/cbor/src/macros.rs
gendx: I find the naming confusing. For example `Vec::extend_from_slice` takes a Vec by `&mut ref`, whereas this *copies* the initial map.
If a copy happens anyway, is this new macro necessary?
How about simply using existing macros/functions:

```rust
let mut map = cbor_map_options! { /* key-value pairs */ };
map.extend(initial_map.iter());
```

src/ctap/data_formats.rs
ia0: It's not only the tag+value, it's alse the [reserved fields](https://developers.google.com/protocol-buffers/docs/proto#reserved) and [unknown fields](https://developers.google.com/protocol-buffers/docs/proto#updating) support. All of those are supported by default in proto. Here we have to encode the tags ourself. There is no description of the message.

libraries/cbor/src/macros.rs
ia0: Where did you see that this copies the map? This would be a bug. The intent was to move the initial value.

src/ctap/data_formats.rs
gendx: Returning `Ctap2StatusCode::CTAP2_ERR_INVALID_CBOR` and then calling `.ok()` to convert the `Result` to an `Option` would be fine. Otherwise I think that this change introduces to much fragmentation.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,91 @@ pub struct PublicKeyCredentialSource {
...
+        };
 
-        let fields = read_array(&cbor_value)?;
-        if fields.len() != 6 {
-            return Err(Ctap2StatusCode::CTAP2_ERR_INVALID_CBOR);
-        }
-        let credential_id = read_byte_string(&fields[0])?;
-        let private_key = read_byte_string(&fields[1])?;
+        let credential_id = read_byte_string(&map.remove(&CredentialId.into())?).ok()?;
+        let private_key = read_byte_string(&map.remove(&PrivateKey.into())?).ok()?;
```
gendx, line 504: All the `.ok()` inside of this function are unnecessarily verbose. A single `.ok()` at the call site would be simpler.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,91 @@ pub struct PublicKeyCredentialSource {
...
-
-    fn try_from(cbor_value: cbor::Value) -> Result<PublicKeyCredentialSource, Ctap2StatusCode> {
-        use cbor::{SimpleValue, Value};
+impl PublicKeyCredentialSource {
+    pub fn parse_cbor(cbor_value: cbor::Value) -> Option<PublicKeyCredentialSource> {
+        use PublicKeyCredentialSourceField::*;
+        let mut map = match cbor_value {
+            cbor::Value::Map(x) => x,
+            _ => return None,
+        };
```
gendx, line 501: This is unnecessarily verbose when we have the `read_map` function. Another argument in favor of having a consistent `TryFrom` interface.
https://github.com/google/OpenSK/blob/bd6870845a853263e195652089eb223542c58b92/src/ctap/data_formats.rs#L671-L678

src/ctap/data_formats.rs
ia0: Before doing the change I want to be sure I understand what you mean. What I understand is:
- We don't expect to have to parse credentials from user input in the future.
- Because of that, this is OK to reuse the same API as user input parsing.
- As such we can use the representation we want because it won't conflate with future use.

Is that correct?

src/ctap/data_formats.rs
ia0: The `read_map` function takes a reference. Here we need a value. But indeed, if the design decision is that we will never have to parse user input, then we can conflate the TryFrom API.

libraries/cbor/src/macros.rs
gendx: Then, the calling convention should be `extend_cbor_map_options!{ &mut map, /* key-value pairs */}`, with an explicit `&mut`.

And I think `extend_cbor_map_options` is more readable than `cbor_extend_map_options`. The function "extends" a "cbor-map", rather than "cbor-extending" a "map".

src/ctap/data_formats.rs
gendx: Reserved/unknown fields are also not specific to protobufs. Plenty of formats have those. For example PNG images (https://en.wikipedia.org/wiki/Portable_Network_Graphics#File_format) have a concept of "critical" (i.e. mandatory) and "ancillary" (i.e. vendor-specific, potentially unknown to a specific decoder) fields.

Again, protobufs mostly refers to a specific encoding on the wire, which we don't use here at all, since we use the CBOR encoding on the wire.

> Here we have to encode the tags ourself. There is no description of the message.

So another difference with protos is that we don't even have such a description of the message, right?

I think mentioning protobufs at all is just confusing - we can just explicitly mention the concepts of reserved/unknown fields, or simply the concept of "versioning".

libraries/cbor/src/macros.rs
ia0: Done.

src/ctap/data_formats.rs
ia0: > So another difference with protos is that we don't even have such a description of the message, right?

This is exactly why it's useful to mention it, because the concept of description is missing. And it's an important concept to understand to be able to understand the code. But I just removed mentions to protobufs since it looks like it's more problematic than anything else.

src/ctap/data_formats.rs
gendx: > * We don't expect to have to parse credentials from user input in the future.
> * Because of that, this is OK to reuse the same API as user input parsing.
> * As such we can use the representation we want because it won't conflate with future use.

My comments have nothing to do with representation. The representation is the same (some CBOR encoding).

This is simply about having a **consistent parsing API** from CBOR to structs. That is, implementing `impl TryFrom<cbor::Value> for T` instead of having each `T` calling its parsing function however it likes. Otherwise we'll end up with fragmentation between `T::from_cbor`, `T::parse_cbor`, `T::from`, `cbor::Value::into::<T>`, and what not.

If `TryFrom<cbor::Value>` isn't the right API, we can revisit this and use another trait, as long as it fits all data formats in OpenSK. But so far it has worked well.

In any case, `PublicKeyCredentialSource` was already implementing `TryFrom<cbor::Value>`, so changing this API is independent from the core change of this pull request (using a map instead of an array).

So this pull request should focus on addressing one change (the versioned map) - if we want to change the parsing API for `PublicKeyCredentialSource` this should be discussed in a separate pull request.

libraries/cbor/src/macros.rs
```diff
@@ -41,20 +41,31 @@ macro_rules! cbor_map_options {
...
+}
+
+#[macro_export]
+macro_rules! extend_cbor_map_options {
+    // Add trailing comma if missing.
+    ( &mut $initial:expr, $( $key:expr => $value:expr ),+ ) => {
+        extend_cbor_map_options! ( &mut $initial, $($key => $value,)+ )
+    };
+
+    ( &mut $initial:expr, $( $key:expr => $value:expr, )* ) => {
```
gendx, line 59: This should be named `$map` instead of `$initial`. This only contains the initial map at the beginning, then it's mutated.

libraries/cbor/src/macros.rs
```diff
@@ -41,20 +41,31 @@ macro_rules! cbor_map_options {
...
         {
             // The import is unused if the list is empty.
             #[allow(unused_imports)]
             use $crate::values::{IntoCborKey, IntoCborValueOption};
-            let mut _map = ::alloc::collections::BTreeMap::<_, $crate::values::Value>::new();
             $(
-            {
-                let opt: Option<$crate::values::Value> = $value.into_cbor_value_option();
```
gendx, line 51: IIRC it was necessary to write the type here so that "into" knows into what type to convert it. In any case, the explicit type doesn't hurt readability of the macro's code.

src/ctap/data_formats.rs
ia0: > My comments have nothing to do with representation. The representation is the same (some CBOR encoding).

My answer was about the fact that you say the representation is "some CBOR encoding". As far as I understand, the other structs in this file have a CBOR representation specified by FIDO. Is that correct?

If it is correct, then it would be a good design decision to not use the same API for representations that are specified by FIDO from the representations that are up to us. This would avoid confusion by having a consistent design. But maybe your point is that we don't care. Is that the case?

Also, note that using CBOR for credentials is accidental (since not specified). If we ever decide to serialize in a different way, it would break the API. Note also that we would actually need to do the same for the encoding depending on the decision.

> In any case, PublicKeyCredentialSource was already implementing TryFrom<cbor::Value>, so changing this API is independent from the core change of this pull request (using a map instead of an array).
> So this pull request should focus on addressing one change (the versioned map) - if we want to change the parsing API for PublicKeyCredentialSource this should be discussed in a separate pull request.

Indeed, I can make a first PR doing just this change. It would be cleaner. I didn't think it would be an issue.

Note that there is also a technical reason to change the API for credentials (beside the design reason), which is to avoid the copy of unknown fields. But this is minor since we expect those to be often empty. However we would still need a constant set of all valid fields, to be able to build the unknown fields by filtering the map.

libraries/cbor/src/macros.rs
ia0: Done.

libraries/cbor/src/macros.rs
ia0: It might have been in the past, but now at least it's not. The return type is specified by `into_cbor_value_option`.

src/ctap/data_formats.rs
gendx: cc @kaczmarczyck

kaczmarczyck: I don't think we need to have different traits for CBOR for CTAP-defined data structures coming from USB and self-defined data structures going to our storage.
PublicKeyCredentialSource is in our control entirely and we can always rename it. As long as we are close to the WebAuthn version, I am happy with the current name.
I vote for TryFrom for easier code and make up for the lack of self-documenting with a comment.

src/ctap/data_formats.rs
kaczmarczyck: If we go back to returning Result, this should resolve? Concerning the error code: We should define and return vendor-specific error codes. In the case of read_* , you could choose to change the error code that read_* returns to this vendor specific code, either is fine I think.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,93 @@ pub struct PublicKeyCredentialSource {
     pub user_handle: Vec<u8>, // not optional, but nullable
     pub other_ui: Option<String>,
     pub cred_random: Option<Vec<u8>>,
+
+    /// Contains the unknown fields when parsing a CBOR value.
+    ///
+    /// Those fields could either be deleted fields from older versions (they should have reserved
+    /// tags) or fields from newer versions (the tags should not be reserved). If this is empty,
+    /// then the parsed credential is probably from the same version (but not necessarily).
+    pub unknown_fields: BTreeMap<cbor::KeyType, cbor::Value>,
```
kaczmarczyck, line 454: I don't yet see a use case for unknown_fields, we can remove it for simplicity and revise that decision later if we think we need it.

kaczmarczyck:
And in general, I agree that "protocol buffer" in the title threw me off first.

libraries/cbor/src/macros.rs
```diff
@@ -41,20 +41,31 @@ macro_rules! cbor_map_options {
...
+            extend_cbor_map_options! ( &mut _map, $( $key => $value, )* );
+            $crate::values::Value::Map(_map)
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! extend_cbor_map_options {
+    // Add trailing comma if missing.
+    ( &mut $map:expr, $( $key:expr => $value:expr ),+ ) => {
```
gendx, line 55: The syntax shouldn't require a `&mut` token. The caller could obtain a mutable reference without directly using `&mut`, for example by calling a function which returns a `&mut T`.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,93 @@ pub struct PublicKeyCredentialSource {
...
+}
+
+// We serialize credentials for the persistent storage using CBOR maps. Each field of a credential
+// is associated with a unique tag, implemented with a CBOR unsigned key.
+#[repr(u64)]
```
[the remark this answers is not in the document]
gendx: I don't see the use case of specifying the representation here. CBOR can dynamically represent integers of up to 64 bits via varint encoding. And once parsed from CBOR into an enum, there's no need to reserve 64 bits of memory. It's unlikely that there will ever be 2^32 variants in this enum, so forcing 64 bits is wasting space (and our typical target CPU architecture has words of 32 bits).

src/ctap/data_formats.rs
ia0: Done.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,93 @@ pub struct PublicKeyCredentialSource {
...
 }
 
-impl TryFrom<cbor::Value> for PublicKeyCredentialSource {
-    type Error = Ctap2StatusCode;
-
-    fn try_from(cbor_value: cbor::Value) -> Result<PublicKeyCredentialSource, Ctap2StatusCode> {
-        use cbor::{SimpleValue, Value};
+impl PublicKeyCredentialSource {
+    pub fn parse_cbor(cbor_value: cbor::Value) -> Option<PublicKeyCredentialSource> {
+        use PublicKeyCredentialSourceField::*;
```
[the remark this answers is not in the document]
ia0: Not sure what you mean, but it doesn't look related to this PR.

src/ctap/data_formats.rs
ia0: Only the next version of the store will return errors for parsing. Currently credentials that don't parse are just ignored.

src/ctap/data_formats.rs
ia0: Obsolete.

src/ctap/data_formats.rs
kaczmarczyck: Ack.

src/ctap/data_formats.rs
```diff
@@ -445,66 +445,93 @@ pub struct PublicKeyCredentialSource {
...
+    OtherUi = 4,
+    CredRandom = 5,
+    // When a field is removed, its tag should be reserved and not used for new fields. We document
+    // those reserved tags below.
+    // Reserved tags: none.
+}
+
+impl From<PublicKeyCredentialSourceField> for cbor::KeyType {
+    fn from(field: PublicKeyCredentialSourceField) -> cbor::KeyType {
+        (field as u64).into()
```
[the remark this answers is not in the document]
kaczmarczyck: Currently, the enums like `SignatureAlgorithm` do not implement
`impl From<SignatureAlgorithm> for cbor::Value`
or
`impl From<SignatureAlgorithm> for cbor::KeyType`
since the cast is a simple `as i64` or `as u64`.

My comment is about consistency. I could these kinds of conversions for other enums with integer representations if we start adding them here.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
kaczmarczyck: We could add smaller integer types to `KeyType`. Our CBOR implementation is efficient in output, but not necessarily in memory. I was wondering if your `repr` statement addresses something unique to your storage. If not, it adds more confusion than documentation, at least for me.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
kaczmarczyck: Again consistency. Is it common practise to `use` all enum variants? Then I will add that to all other TryFrom implementations.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
ia0: Syntactical consistency should not break design (and should also not break engineering and small PRs). Would you use `void *` for all pointers just for consistency? Things should be implemented if they make sense, i.e. they are part of a design and not because they look consistent.

It looks to me that those consistency changes can be done in another PR if needed (and I'm not doing them because they don't look correct and well motivated). So I'm only addressing those for added/modified lines in this PR that are problematic. Please flag those explicitly.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
ia0: The current CBOR implementation seems to be "canonical" in the sense that there is a bijection between the binary data and the Rust structure (modulo map ordering I guess). Adding smaller integer types to `KeyType` would break this property and also would not fix any memory issue. The problem is that `KeyType` shares variant of different sizes. One solution would be to use one level of indirection (like in functional languages), but this comes with the price of an indirection so there's a compromise to be done. I don't think it's worth it. If there are memory issues, they should be analyzed and the bigger problems should be addressed first.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
ia0: If syntactical consistency matters, I would prefer to change those lines rather than the whole file.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
kaczmarczyck: I didn't want to stop you from implementing this trait. Just mentioned that I did it differently in this file so far. I think implementing the `From` trait for all of my enums is the way to go. This comment is not actionable, but more informative (like the one below about consistency). Feel free to resolve.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
kaczmarczyck: Agreed we should change the CBOR implementation. Since the line is gone in in your current commit, I consider this resolved.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
kaczmarczyck: Your choice! Whatever is more Rust-y.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
ia0: I see. It wasn't clear to me. Thanks for clarifying!

Note that I'm reluctant modifying the CTAP structs because they are specified by FIDO and as such need a good understanding of the specification which I don't have. As clarified before, the source credential is not specified and is thus much safer to touch.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
ia0: I slightly prefer the current version for the following reasons:
- The scope is limited to the current function which is short.
- The amount of boilerplate saved is non-negligible, which improves readability.

src/ctap/data_formats.rs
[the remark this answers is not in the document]
ia0: Yes, I removed the `repr` because it was meant for documentation and apparently was confusing, so it was obviously not a good documentation.

libraries/cbor/src/macros.rs
gendx: Forget about my other comment. I didn't mean `&mut` as literally it in the source code, but rather as an expression of type `&mut T`. It doesn't make sense to put `&mut` in the macro's syntax.

gendx: Obsolete.

src/ctap/data_formats.rs
```diff
@@ -447,56 +450,72 @@ pub struct PublicKeyCredentialSource {
...
+    }
+}
+
 impl From<PublicKeyCredentialSource> for cbor::Value {
     fn from(credential: PublicKeyCredentialSource) -> cbor::Value {
-        let mut private_key = [0u8; 32];
+        use PublicKeyCredentialSourceField::*;
+        let mut private_key = [0; 32];
```
gendx, line 476: I find the previous expression `0u8` more readable, as it makes it clear that we have 32 **bytes** and not potentially some other integer type.

src/ctap/storage.rs
```diff
@@ -420,8 +420,7 @@ impl From<StoreError> for Ctap2StatusCode {
 }
 
 fn deserialize_credential(data: &[u8]) -> Option<PublicKeyCredentialSource> {
-    let cbor = cbor::read(data).ok()?;
-    cbor.try_into().ok()
+    PublicKeyCredentialSource::try_from(&cbor::read(data).ok()?).ok()
```
gendx, line 423: What is the rationale for changing this test?

src/ctap/data_formats.rs
```diff
@@ -447,56 +450,72 @@ pub struct PublicKeyCredentialSource {
...
+            RpId => Some(credential.rp_id),
+            UserHandle => Some(credential.user_handle),
+            OtherUi => credential.other_ui,
+            CredRandom => credential.cred_random
         }
     }
 }
 
-impl TryFrom<cbor::Value> for PublicKeyCredentialSource {
+impl TryFrom<&cbor::Value> for PublicKeyCredentialSource {
```
gendx, line 489: What is the rationale for taking the CBOR value by reference? In principle, the CBOR value is a short-lived object, out of which we could move elements in this `try_from`. All the other types parse it by value.

src/ctap/data_formats.rs
ia0: Done. (Note that the next line hints to this fact.)

src/ctap/data_formats.rs
ia0: I was anticipating the "consistency" argument to avoid too many ping-pong since all other types parse a `cbor::Value` by reference (I don't know where you saw that they do it by value). I'll be happy to change it back to call-by-value if you think it's fine to not be "consistent".

src/ctap/storage.rs
ia0: This is because when `TryFrom` is on a reference, the `.try_into()` notation doesn't take the reference implicitly, so we would either need to write `(&x).try_into()` or `T::try_from(&x)`. I chose the second to avoid ping-pong due to the possibly confusing `(&x).try_into()` notation.

src/ctap/data_formats.rs
kaczmarczyck: The reason is that most other data structures appear as values in maps. We then get references of those maps entries. This data structure is not used as a parameter, but for our own sake. So I don't mind the "inconsistency" here.

ia0: Done.

Note that _the fact that the other data structures appear as values in maps_ is not the real reason why it's not call-by-value. It's possible to extract the value from a map (as done for the credential source in the latest commit). If it's not possible to use call-by-value, it would be because the initial CBOR value is used more than once (breaking the linearity required by call-by-value). I don't know if this is the case. If it's not the case, then the other functions could as well use call-by-value and avoid unnecessary allocations. But that would be another PR and only useful if we believe allocations are an issue.

PR #109, opened 2020-05-13
credProtect extension

kaczmarczyck:
Introduces the credProtect extension. This might need changes after #108 is merged.

- [x] Tests pass


README.md
```diff
@@ -99,6 +99,9 @@ a few things you can personalize:
 4.  Depending on your available flash storage, choose an appropriate maximum
     number of supported residential keys and number of pages in
     `ctap/storage.rs`.
+5.  Change the default level for the credProtect extension. Resident credentials
+    become undiscoverable without user verification. This helps privacy, but
+    can make usage less comfortable for less important credentials.
```
jmichelp, line 104: You probably want to say that the setting is in `ctap/mod.rs` :)

src/ctap/mod.rs
```diff
@@ -109,6 +110,10 @@ pub const ES256_CRED_PARAM: PublicKeyCredentialParameter = PublicKeyCredentialPa
     cred_type: PublicKeyCredentialType::PublicKey,
     alg: SignatureAlgorithm::ES256,
 };
+// You can change this value to one of the following for more privacy.
+// - Some(CredentialProtectionPolicy::UserVerificationOptionalWithCredentialIdList)
+// - Some(CredentialProtectionPolicy::UserVerificationRequired)
+const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
```
jmichelp, line 116: Unrelated to this PR but I think we need to move these configuration constants from compile-time to one-time-programmable area in the storage.

src/ctap/mod.rs
```diff
@@ -521,9 +546,11 @@ where
...
         };
         auth_data.extend(cose_key);
-        if use_hmac_extension {
-            let extensions = cbor_map! {
-                "hmac-secret" => true,
+        if has_extension_output {
+            let hmac_secret_output = if use_hmac_extension { Some(true) } else { None };
+            let extensions = cbor_map_options! {
+                "hmac-secret" => hmac_secret_output,
+                "credProtect" => cred_protect_policy.map(|policy| policy as i64),
```
jmichelp, line 553: What is the implication of having credProtect on CTAP2.0?
Currently CTAP2.1 is gated by a rust feature and credProtect is part of it and wasn't there for 2.0, right?

src/ctap/mod.rs
kaczmarczyck: So if I manually update, I don't override my custom settings? Sounds good.

src/ctap/mod.rs
kaczmarczyck: From what I heard, this extension only barely didn't make it into 2.0 (without being an extension). I don't know if extensions need to be officially mentioned in a specification to be valid. Or if I could just invent an extension and use it. This one is described in 2.1, so we could hide it behind the feature flag, if you want to. That being said, it is perfectly compatible with 2.0 and I didn't see why I wouldn't allow 2.0 users to use that extension. With the current default policy, it does not change any behavior.

Your call.

src/ctap/data_formats.rs
```diff
@@ -432,6 +440,27 @@ impl TryFrom<&cbor::Value> for SignatureAlgorithm {
     }
 }
 
+#[derive(Clone, PartialEq, PartialOrd)]
```
gendx, line 443: You can make this enum `Copy` as well. This doesn't take a lot of bytes in memory, and allows simpler code elsewhere (see my other comment).

src/ctap/mod.rs
```diff
@@ -428,37 +434,55 @@ where
...
-            // The extension is actually supported, but we need resident keys.
-            return Err(Ctap2StatusCode::CTAP2_ERR_UNSUPPORTED_EXTENSION);
-        }
+        let (use_hmac_extension, cred_protect_policy) = if let Some(extensions) = extensions {
+            let mut cred_protect = extensions
+                .make_credential_cred_protect_policy()
+                .transpose()?;
+            if cred_protect
+                .as_ref()
+                .unwrap_or(&CredentialProtectionPolicy::UserVerificationOptional)
```
gendx, line 443: Here if the enum type is `Copy`, you don't need the `as_ref` part, nor the reference inside the `unwrap_or`. This will be more readable.

src/ctap/mod.rs
jmichelp: I would say it depends how far we are at having all CTAP2.1 implemented (assuming that once fully implemented, we'll remove the feature and only be CTAP2.1) and whether supporting this extension would break OpenSK when compiled without CTAP2.1 in the meantime.

kaczmarczyck: 2.1 is not finalized yet, and I still expect some commands to change in the next months. We are still missing implementations for bigger work items like enterprise attestation. credProtect should be stable and is independent of other new features. It only interacts with old commands (MakeCredential, GetAssertion, GetInfo).

src/ctap/mod.rs
ia0: Note that with the current storage, "persistent" entries can be implemented by not deleting them when calling [reset](https://github.com/google/OpenSK/blob/af79d92a888fb398bd1e630f1d821c4ed4128f95/src/ctap/storage.rs#L397) (which currently deletes all entries). With the next storage, "persistent" entries would simply have a key below the persistent/reset threshold (and the reset operation would be atomic, once started it cannot be stopped).

Note also that the test suite might need to support the fact that the persistent storage don't get fully reset when reset is called (if those persistent entries may alter the test suite behavior). We may provide an additional custom CTAP command to fully reset the storage (including persistent entries).

If we don't want to wait the next storage, I can implement a facility to simplify adding "persistent" (that survive reset) entries in the current one (so that we only need to modify a predicate on entries whether they are persistent). Tell me.

kaczmarczyck: No worries, I think we can worry about our compile-time constants later.

src/ctap/mod.rs
kaczmarczyck: Done, thanks.

src/ctap/data_formats.rs
kaczmarczyck: Done.

README.md
kaczmarczyck: Done.

kaczmarczyck:
I moved the credProtect checks into storage.rs to let it check discoverability itself. And addressed your comments. Please let me know if you want to add the feature flag.

src/ctap/mod.rs
```diff
@@ -657,11 +682,14 @@ where
...
             for allowed_credential in allow_list {
-                match self
-                    .persistent_store
-                    .find_credential(&rp_id, &allowed_credential.key_id)
-                {
-                    Some(credential) => found_credentials.push(credential),
+                match self.persistent_store.find_credential(
+                    &rp_id,
+                    &allowed_credential.key_id,
+                    !has_uv,
```
ia0, line 688: I guess CTAP doesn't require the authenticator to return a different error whether the credential does not exist and whether it is protected. But maybe this question doesn't make sense, because it looks like asking for user verification does not depend on which credential we need (it's always done upfront).

kaczmarczyck: External user Verification is done with another command before the actual request. Internal user verification like a fingerprint reader on the security key happens while processing a command, like waiting for user presence.

In this case, it shouldn't really matter since the point of credProtect is to make it indistinguishable whether the credential is protected or does not exist.

src/ctap/data_formats.rs
```diff
@@ -477,7 +512,7 @@ impl TryFrom<cbor::Value> for PublicKeyCredentialSource {
         use cbor::{SimpleValue, Value};
 
         let fields = read_array(&cbor_value)?;
-        if fields.len() != 6 {
+        if fields.len() != 7 {
```
jmichelp, line 515: What happens if you advertise credProtect but the other side decides to not use the option? Isn't the array only going to be with 6 elements?

ia0: The array always has 7 elements (for a given version of OpenSK). If credProtect is not used, the last element is the CBOR `null` value. The issue would be using a more recent OpenSK with an old storage. But a fix has already been merged. So after syncing to master this test disappears since we would use a map instead of an array.

ia0: Actually, it is currently not possible to use a recent OpenSK with an old storage since we still use writable flash regions. So this test (length of array) cannot fail for 2 reasons:
- the fix to support old storage with new code is submitted
- the storage is still inside the app so a new app cannot use an old storage

kaczmarczyck: Please check the code again after the merge with master.

ia0: Did you forget to push or is the merge with master still in progress? I don't see any new commits.

kaczmarczyck:
Just merged, added some more conversion functions. Will refactor to moving more values in a later PR.

src/ctap/data_formats.rs
```diff
@@ -481,7 +525,8 @@ impl From<PublicKeyCredentialSource> for cbor::Value {
             RpId => Some(credential.rp_id),
             UserHandle => Some(credential.user_handle),
             OtherUi => credential.other_ui,
-            CredRandom => credential.cred_random
+            CredRandom => credential.cred_random,
+            CredProtectPolicy => credential.cred_protect_policy.map(|p| p as i64),
```
ia0, line 529: Is there a reason not to implement `From<CredentialProtectionPolicy> for cbor::Value`? We could remove the `map(|p| p as i64)` in that case.

src/ctap/data_formats.rs
```diff
@@ -432,6 +440,40 @@ impl TryFrom<&cbor::Value> for SignatureAlgorithm {
...
 
+#[derive(Clone, Copy, PartialEq, PartialOrd)]
+#[cfg_attr(any(test, feature = "debug_ctap"), derive(Debug))]
+pub enum CredentialProtectionPolicy {
+    UserVerificationOptional = 0x01,
+    UserVerificationOptionalWithCredentialIdList = 0x02,
+    UserVerificationRequired = 0x03,
+}
+
+impl TryFrom<cbor::Value> for CredentialProtectionPolicy {
```
ia0, line 451: If it is a short term plan to use move semantics for parsing, then you can disregard everything below.

If we want to avoid this code duplication, we can either:
- Remove this implementation and change line 559
    - from `.map(CredentialProtectionPolicy::try_from)`
    - to `.map(|x| CredentialProtectionPolicy::try_from(&x))`
- Implement it using the call-by-reference version:
    ```
        Self::try_from(&cbor_value)
    }
    ```

We can also remove the `extract_integer` function because CBOR integers are essentially `Copy` so it's fine to take them as reference and return a copy of the contained integer. The `extract_` functions are only useful for non-`Copy` types (i.e. move-only types). However, if one day we only use move semantics for parsing, then we can as well use `extract_` and call-by-value for all methods and have a single implementation for each.

src/ctap/mod.rs
```diff
@@ -521,9 +543,11 @@ where
...
         };
         auth_data.extend(cose_key);
-        if use_hmac_extension {
-            let extensions = cbor_map! {
-                "hmac-secret" => true,
+        if has_extension_output {
+            let hmac_secret_output = if use_hmac_extension { Some(true) } else { None };
+            let extensions = cbor_map_options! {
+                "hmac-secret" => hmac_secret_output,
+                "credProtect" => cred_protect_policy.map(|policy| policy as i64),
```
ia0, line 550: We should be able to remove this `map` if `From<Policy> for cbor::Value` is implemented.

src/ctap/mod.rs
```diff
@@ -501,6 +522,7 @@ where
                     .user_display_name
                     .map(|s| truncate_to_char_boundary(&s, 64).to_string()),
                 cred_random,
+                cred_protect_policy: cred_protect_policy.clone(),
```
ia0, line 525: I'm surprised we need this `clone`. Rust should deduce `Option<Policy>: Copy` from `Policy: Copy`.

src/ctap/data_formats.rs
kaczmarczyck: Plan is to make everything

- extract_*
- try_from without reference

as you suggested and started. So I leave it as is for now?

src/ctap/mod.rs
kaczmarczyck: I think we needed before it implemented Copy, and I forgot to remove the clone. Fixed.

src/ctap/data_formats.rs
kaczmarczyck: The other enum that is a glorified i64 also didn't, I changed both.

src/ctap/mod.rs
kaczmarczyck: Removed.

src/ctap/data_formats.rs
ia0: Yes, if it's planned to complete the migration before the formats are modified there's no problem. The risk with code duplication would be that another change needs to update some struct or enum and forgets to update all duplicated parts.

reproducible/binaries.sha256sum
```diff
@@ -0,0 +1,9 @@
+c182bb4902fff51b2f56810fc2a27df3646cd66ba21359162354d53445623ab8	third_party/tock/target/thumbv7em-none-eabi/release/nrf52840dk.bin
```
ia0, line 1: I don't think this file is meant to be checked in the repository.

ia0: @gendx Should we keep this file in the repository?

kaczmarczyck: That was unintended, like the binary. I'll remove them ASAP.

src/ctap/data_formats.rs
```diff
@@ -1064,7 +1149,7 @@ mod test {
         let signature_algorithm = SignatureAlgorithm::try_from(&cbor_signature_algorithm);
         let expected_signature_algorithm = SignatureAlgorithm::ES256;
         assert_eq!(signature_algorithm, Ok(expected_signature_algorithm));
-        let created_cbor: cbor::Value = cbor_int!(signature_algorithm.unwrap() as i64);
+        let created_cbor = cbor::Value::from(signature_algorithm.unwrap());
```
gendx, line 1152: Nit: it would be more consistent with the rest of the code to do `let created_cbor: cbor::Value = signature_algorithm.unwrap().into();` rather than using `cbor::Value::from`.

kaczmarczyck: I can include that in another PR, if you don't like cbor::Value::from.

kaczmarczyck:
Feature for #106

PR #110, opened 2020-05-14
Fix nrf52840_mdk buttons

ataffanel:
This PR fixes the button pin assignment for the nrf52840_mdk board according to schematic: https://github.com/makerdiary/nrf52840-mdk/blob/master/docs/hardware/nrf52840-mdk-schematic_v1_0.pdf

Change tested on hardware.

- [x] Tests pass

### DIFF
--- a/boards/nordic/nrf52840_mdk_dfu/src/main.rs
+++ b/boards/nordic/nrf52840_mdk_dfu/src/main.rs
@@ -18,9 +18,9 @@ const LED1_R_PIN: Pin = Pin::P0_23;
 const LED1_G_PIN: Pin = Pin::P0_22;
 const LED1_B_PIN: Pin = Pin::P0_24;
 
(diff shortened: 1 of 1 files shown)


jmichelp:
It seems there are 2 platforms:
* nrf52840-mdk
* nrf52840-mdk-usb-dongle

The board that you modify in your PR is actually for the USB dongle, not the IoT kit. And for this one, the pinout is correct.

ataffanel:
I see, this explains the strange pin-out. Sorry about the confusion I guess I should close this PR then.

Supporting the nrf52840-mdk board might be interesting since this is an good small dev board with debugger and serial port integrated. If it is supported, the folder and target name should likely be changed (the source of my confusion is that the target is called "nrf52840_mdk_dfu"). I can open a ticket about it.

PR #112, opened 2020-05-30
Access the persistent keys through the store

ia0:
This enables the possibility to not embed their value in the application.

- [x] Test suite passes


src/ctap/mod.rs
```diff
@@ -534,10 +533,11 @@ where
         signature_data.extend(client_data_hash);
         let (signature, x5c) = if USE_BATCH_ATTESTATION {
```
kaczmarczyck, line 534: Maybe we can use this opportunity to remove the boolean flag and use the correct supported behavior:
If certificate exists:
  batch attestation
if not:
  self attestation

If do we think that deciding on batch attestation on compile time is a feature, and this way, you can't forget to write a certificate before usage?

src/ctap/status_code.rs
```diff
@@ -67,5 +67,11 @@ pub enum Ctap2StatusCode {
     // CTAP2_ERR_VENDOR_FIRST = 0xF0,
     CTAP2_ERR_VENDOR_RESPONSE_TOO_LONG = 0xF0,
     CTAP2_ERR_VENDOR_RESPONSE_CANNOT_WRITE_CBOR = 0xF1,
+
+    /// An internal invariant is broken.
```
kaczmarczyck, line 71: Three /// ?

src/ctap/storage.rs
```diff
@@ -211,6 +223,33 @@ impl PersistentStore {
                 })
                 .unwrap();
         }
+        if self.store.find_one(&Key::AttestationPrivateKey).is_none() {
```
kaczmarczyck, line 226: I introduced `USE_BATCH_ATTESTATION` for users to have a choice. Batch attestation can be a privacy concern, and some relying parties might not require it. If we want to replace the boolean flag `USE_BATCH_ATTESTATION` with "exists in the persistent store", we could add that here. This is more of an open question which is cleaner.

src/ctap/mod.rs
ia0: I modified the decision on batch attestation as suggested (presence of attestation) and only write the attestation to the storage if compiled with batch attestation. We can tweak later those decisions as needed (e.g. batch attestation per RP id).

src/ctap/status_code.rs
ia0: This means documentation. It's similar to doxygen but with language support, in particular `cargo doc` generates HTML files for the documentation. When writing `///` the documentation is attached to the following item. When using `//!` it is attached to the containing item (usually used for modules).

src/ctap/storage.rs
ia0: I think it makes sense to do that for now (only copy the attestation to the storage if compiled with batch attestation and use batch attestation only if the attestation is present in the storage). It should not be hard to improve on that later (e.g. per-rpid batch attestation).

src/ctap/status_code.rs
kaczmarczyck: Thanks!

src/ctap/storage.rs
```diff
@@ -56,9 +56,14 @@ const GLOBAL_SIGNATURE_COUNTER: usize = 1;
...
 const PIN_HASH: usize = 3;
 const PIN_RETRIES: usize = 4;
-const NUM_TAGS: usize = 5;
+const ATTESTATION_PRIVATE_KEY: usize = 5;
+const ATTESTATION_CERTIFICATE: usize = 6;
+const AAGUID: usize = 7;
+const NUM_TAGS: usize = 8;
 
 const MAX_PIN_RETRIES: u8 = 6;
+const ATTESTATION_PRIVATE_KEY_LENGTH: usize = 32;
```
kaczmarczyck, line 65: This constant is from the crypto library in ec/int256.rs called BITS_PER_DIGIT. Since you also use this to double-check, maybe it's intended to be its own constant?

ia0: I'm not sure what you mean. The BITS_PER_DIGIT constant is private so I can't use it here. Do you mean to make it public?

The reason I added the constant here was that it is hard-coded in the key_material.rs file. And I need it multiple times in this file so I just factored it out. I'm fine using a shared constant if there's one.

kaczmarczyck: Yeah, I meant making it public. Not sure if just having your own constant isn't better though. So feel free to keep it like that, just wanted to pointed out.

ia0: If it makes sense for the constant (actually it looks like it should be `NBYTES` and not `BITS_PER_DIGIT` according to the signature of `Int256::from_bin`) to be part of the crypto public API, then I should use it here. As I understand it, it's OK to make it public so I did it and used the constant here.

kaczmarczyck: Yeah sorry, definitely NBYTES. LGTM

libraries/crypto/src/ecdsa.rs
```diff
@@ -140,7 +142,7 @@ impl SecKey {
         }
     }
 
-    pub fn from_bytes(bytes: &[u8; 32]) -> Option<SecKey> {
+    pub fn from_bytes(bytes: &[u8; NBYTES]) -> Option<SecKey> {
```
gendx, line 145: I don't think this change makes it clearer on the public API of this module. And changes in this file are unrelated to this pull request anyway.

ia0: I can revert the last commit. Please agree with @kaczmarczyck .

kaczmarczyck: Go ahead and merge with the 32, and we can do library changes outside of this PR if we still think we want them.

ia0: SGTM. Reverted.

kaczmarczyck:
I'll leave it to @gendx to approve, since my comments were addressed.

src/ctap/storage.rs
```diff
@@ -420,6 +523,13 @@ impl From<StoreError> for Ctap2StatusCode {
     }
 }
 
+fn should_reset<'a>(entry: &StoreEntry<'a>) -> bool {
```
gendx, line 526: Nit: IIRC we can use an anonymous lifetime for this kind of use case, i.e. `fn should_reset(entry: &StoreEntry<'_>) -> bool`.

ia0: Indeed, the name is not needed. Done.

gendx:
I don't know what's causing Travis-CI to fail, but it's been like this for the last few pull requests. Maybe we can rely solely on GitHub actions?

PR #113, opened 2020-06-05
reads CBOR maps by moving instead of references

kaczmarczyck:
As started by ia0 in #108 and used later, this PR changes all calls to use moves for parsing CBOR values.
Also accidentially submitted files with hashes are removed.

- [x] Tests pass


src/ctap/data_formats.rs
```diff
@@ -807,7 +758,7 @@ fn extract_map(
     }
 }
 
-pub(super) fn read_bool(cbor_value: &cbor::Value) -> Result<bool, Ctap2StatusCode> {
+pub(super) fn extract_bool(cbor_value: cbor::Value) -> Result<bool, Ctap2StatusCode> {
```
ia0, line 761: You can make this function (and all functions that return a type implementing `Copy`) take `impl Borrow<cbor::Value>` then use `match cbor_value.borrow()`. This would permit to avoid some `.clone()` I've seen above.

src/ctap/data_formats.rs
```diff
@@ -281,23 +281,23 @@ impl Extensions {
     pub fn has_make_credential_hmac_secret(&self) -> Result<bool, Ctap2StatusCode> {
         self.0
             .get("hmac-secret")
-            .map(read_bool)
+            .map(|b| extract_bool(b.clone()))
```
ia0, line 284: I've added a comment on `extract_bool`. If it works, you should be able to simply write `extract_bool` removing the need for `clone`.

src/ctap/data_formats.rs
```diff
@@ -281,23 +281,23 @@ impl Extensions {
...
             .unwrap_or(Ok(false))
     }
 
     pub fn get_assertion_hmac_secret(
-        &self,
+        mut self,
```
ia0, line 289: This is a semantics change. If we really expect to call this function exactly once per `Extensions` we should rename this function to `extract_` instead of `get_`. Otherwise we should simply clone the output: `self.0.get("hmac").map(|x| try_from(x.clone())`.

Same for the function below.

src/ctap/data_formats.rs
```diff
@@ -717,28 +707,14 @@ impl TryFrom<&cbor::Value> for ClientPinSubCommand {
     }
 }
 
-pub(super) fn read_unsigned(cbor_value: &cbor::Value) -> Result<u64, Ctap2StatusCode> {
+pub(super) fn extract_unsigned(cbor_value: cbor::Value) -> Result<u64, Ctap2StatusCode> {
```
ia0, line 710: The same comment as for `extract_bool` applies here.

src/ctap/data_formats.rs
```diff
@@ -717,28 +707,14 @@ impl TryFrom<&cbor::Value> for ClientPinSubCommand {
...
-                Err(Ctap2StatusCode::CTAP2_ERR_CBOR_UNEXPECTED_TYPE)
-            }
-        }
-        cbor::Value::KeyValue(cbor::KeyType::Negative(signed)) => Ok(*signed),
-        _ => Err(Ctap2StatusCode::CTAP2_ERR_CBOR_UNEXPECTED_TYPE),
-    }
-}
-
-fn extract_integer(cbor_value: cbor::Value) -> Result<i64, Ctap2StatusCode> {
+pub(super) fn extract_integer(cbor_value: cbor::Value) -> Result<i64, Ctap2StatusCode> {
```
ia0, line 717: The same comment as for `extract_bool` applies here.

src/ctap/data_formats.rs
```diff
@@ -1177,23 +1128,23 @@ mod test {
...
-        let policy = CredentialProtectionPolicy::try_from(&cbor_policy);
+        let policy = CredentialProtectionPolicy::try_from(cbor_policy.clone());
         let expected_policy = CredentialProtectionPolicy::UserVerificationOptional;
         assert_eq!(policy, Ok(expected_policy));
         let created_cbor = cbor::Value::from(policy.unwrap());
         assert_eq!(created_cbor, cbor_policy);
 
-        let cbor_policy_error = cbor_int!(-1);
-        let policy_error = CredentialProtectionPolicy::try_from(&cbor_policy_error);
+        let cbor_policy_error: cbor::Value = cbor_int!(-1);
```
ia0, line 1137: Out of curiosity, why do we need to specify the type?

src/ctap/command.rs
```diff
@@ -124,30 +124,32 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
     type Error = Ctap2StatusCode;
 
     fn try_from(cbor_value: cbor::Value) -> Result<Self, Ctap2StatusCode> {
-        let param_map = read_map(&cbor_value)?;
+        let mut param_map = extract_map(cbor_value)?;
 
-        let client_data_hash = read_byte_string(ok_or_missing(param_map.get(&cbor_unsigned!(1)))?)?;
+        let client_data_hash =
+            extract_byte_string(ok_or_missing(param_map.remove(&cbor_unsigned!(1)))?)?;
```
gendx, line 130: Rather than removing elements one by one, which doesn't have ideal performance and by itself could copy values around inside the map, would it be possible to loop over `map.into_iter()` and rely on the fact that keys are sorted in a `BTreeMap`?

I'm thinking of something like the following (very rough draft), but maybe a macro would make it easier to use (a "mirror" of `cbor_map!` that creates CBOR values).

```rust
let mut map = extract_map(cbor_value)?;
let mut it = map.into_iter();

let client_data_hash = loop {
    let (k, v) = it.next()?;
    match k.cmp(cbor_unsigned!(1)) {
        Ordering::Less => /* ignore unknown key */ continue,
        Ordering::Equal => extract_byte_string(v),
        Ordering::Greater => /* key "1" is not in the map */ return Err(...),
    }
};

// Same with cbor_unsigned!(2).
// etc.
```

kaczmarczyck: Do you think that the proposed implementation is slower than the former one with `get`?

src/ctap/data_formats.rs
kaczmarczyck: The compiler complained, so I was assuming the macro does not uniquely specifiy the type.

ia0: Oh right, having to specify the type is actually a feature: https://github.com/google/OpenSK/blob/642ee69a9bd6939d4f05c298fcfea101a7dc265a/libraries/cbor/src/macros.rs#L129-L130. Good to know.

src/ctap/command.rs
ia0: I think that readability is more important than performance. In particular the performance benefit is very small here. The maps are small (around 10 elements on average). Re-balancing is thus cheap (it's just rewriting 5 or so pointers) and on par with the cost of moving the value (that we have to do anyway). So we won't gain anything significant.

gendx: If performance is not so important, what is the rationale for moving rather than copying? The current implementation with `get` is readable enough and the few copies don't have that much overhead.

And on the readability front, my idea would be to embed this logic in a macro, so that it's readable. For example:

```rust
// This internally has `let client_data_hash = ...` statements creating the `client_data_hash`,
// `rp`, ... variables.
parse_cbor_map!{param_map,
    1 => client_data_hash,
    2 => rp,
    3 => user,
    ...
};

let client_data_hash = extract_byte_string(client_data_hash)?;
let rp = PublicKeyCredentialRpEntity::try_from(rp)?;
...
```

Then we can easily swap the internal implementation of this `parse_cbor_map!` macro, to experiment with `get`, `remove` or `into_iter` implementations.

ia0: Yes, such a macro would definitely help and we should probably add it (this is the kind of stuff you get for free with protocol buffers, serde, or other libraries based on schema).

The rationale I see for this change is that parsing is something that is essentially linear and reflecting that in the types is natural. This is surprising to see a copy-based parsing libraries. So the change is not about performance but about API.

kaczmarczyck: What this PR is supposed to do is clean up from former PRs where Julian used moves in data_formats.rs, and have a consistent way of parsing the data structures. Can we postpone any potential performance improvements to a later PR? Your latest thumbs up lets me assume that we like the remove better than the get for now?

gendx: My take is mostly that there is an extra cost to refactoring all the use cases from get to remove, and then from remove to the macro, as opposed to refactoring directly from get to the macro, and then changing how the macro is implemented.

That said, now that the get -> remove refactoring is implemented, the cost to moving to the macro will be essentially the same, whether this pull request is merged or not.

So I'm pretty neutral on this pull request. It's not a perfect performance improvement due to the remove being non-trivial, but it at least clarifies that parsing is linear (consuming types rather than taking references).

src/ctap/data_formats.rs
ia0: After discussion, the change is not worth it. See the rationale in the `extract_bool` comment.

src/ctap/data_formats.rs
ia0: Obsolete.

src/ctap/data_formats.rs
ia0: Obsolete.

src/ctap/data_formats.rs
ia0: The benefit of doing so is small (1) and introduces an inconvenience in tests (2).

1. We should not have to `extract_bool` on a reference. The particular existing case may disappear depending on the chosen API on `Extensions`. In particular if `Extensions` become a real Rust struct like others in this file.

2. The pattern `extract_bool(cbor_int!(x))` doesn't work well because `cbor_int!` is polymorphic in its return type while `extract_bool` would be polymorphic in its argument. So the type system can't decide which type to use. One would need to use a monomorph version  of `extract_bool` in the test.

src/ctap/command.rs
kaczmarczyck: I'll go ahead with it then, so we at least have only one of the two: `get`, `remove`

src/ctap/data_formats.rs
```diff
@@ -243,66 +243,51 @@ impl From<PublicKeyCredentialDescriptor> for cbor::Value {
...
-            } else {
-                return Err(Ctap2StatusCode::CTAP2_ERR_CBOR_UNEXPECTED_TYPE);
-            }
-        }
-        Ok(Extensions(extensions))
+        let mut extensions_map = extract_map(cbor_value)?;
+        let hmac_secret = extensions_map
+            .remove(&cbor_text!("hmac-secret"))
+            .map(extract_bool)
+            .unwrap_or(Ok(false))?;
```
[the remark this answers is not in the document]
kaczmarczyck: Yeah, slightly better, thanks.

src/ctap/mod.rs
```diff
@@ -1521,25 +1517,26 @@ mod test {
...
         make_credential_params.extensions = make_extensions;
         assert!(ctap_state
             .process_make_credential(make_credential_params, DUMMY_CHANNEL_ID)
             .is_ok());
 
         let pk = sk.genpk();
-        let hmac_secret_parameters = cbor_map! {
-            1 => cbor::Value::Map(CoseKey::from(pk).0),
-            2 => vec![0; 32],
-            3 => vec![0; 16],
```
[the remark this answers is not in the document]
kaczmarczyck: Not a bug necessarily, because it should be able work with any byte vector, but the current auth is 16 bytes long. The 32 was copied from data_formats.rs and I changed all to a consistent 16 bytes length.

kaczmarczyck:
I included the nit that @gendx mentioned on the closed PR #109

PR #116, opened 2020-06-09
Add a read_cbor_map macro to avoid removing values one-by-one in BTreeMaps

gendx:
`This pull request fixes what was discussed in https://github.com/google/OpenSK/pull/113#discussion_r435975483.

A new `read_cbor_map` macro is created, which takes as input a `BTreeMap<cbor::KeyType, cbor::Value>`, as well as a mapping from identifiers to `cbor::KeyType`. This generates a code block that extracts the corresponding values from the `BTreeMap` and assign them to the identifiers as new variables.

Essentially:

```rust
read_cbor_map! {
    map,
    x @ 1,
    y @ 2,
};
```

generates code semantically equivalent to the following, but more optimized:

```rust
let x: Option<cbor::Value> = map.remove(&cbor_unsigned!(1));
let y: Option<cbor::Value> = map.remove(&cbor_unsigned!(2));
```

**Important:** Keys passed to the `read_cbor_map!` macro **must be sorted** in increasing order. If not, the algorithm can yield incorrect results (such a assigning `None` to a variable even if the corresponding key existed in the map). **No runtime checks** are made for this in the `read_cbor_map!` macro, in order to avoid overhead at runtime. However, assertions that keys are sorted are added in `cfg(test)` mode, so that unit tests can verify ahead of time that the keys are indeed sorted. This macro is therefore **not suitable for dynamic keys** that can change at runtime.

- [ ] Tests pass


libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,127 @@
...
+///
+/// ```rust
+/// # extern crate alloc;
+/// # #[macro_use]
+/// # extern crate cbor;
+/// #
+/// # fn main() {
+/// #     let mut map = alloc::collections::BTreeMap::<cbor::KeyType, _>::new();
+/// let x: Option<cbor::Value> = map.remove(&cbor_unsigned!(1));
+/// let y: Option<cbor::Value> = map.remove(&cbor_unsigned!(2));
```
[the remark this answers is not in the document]
kaczmarczyck: I wouldn't include a wrong example, but you could change the `cbor_unsigned!(2)` into a `cbor_text!("key")` if you want to show that it works in general.

libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,127 @@
...
+
+        let mut it = $map.into_iter().peekable();
+        $(
+        let $variable: Option<Value> = read_cbor_map_peek_value(&mut it, $key.into_cbor_key());
+        )+
+    };
+}
+
+/// This function is an internal detail of the `read_cbor_map!` macro, but has public visibility so
+/// that users of the macro can use it.
```
kaczmarczyck, line 84: What would be a usage scenario for this function?

libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,127 @@
...
+    it: &mut Peekable<btree_map::IntoIter<KeyType, Value>>,
+    needle: KeyType,
+) -> Option<Value> {
+    loop {
+        match it.peek() {
+            None => return None,
+            Some(item) => {
+                let key: &KeyType = &item.0;
+                match key.cmp(&needle) {
+                    Ordering::Less => {
```
kaczmarczyck, line 99: Is this continue redundant?

libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,127 @@
...
+}
+
+/// This function is an internal detail of the `read_cbor_map!` macro, but has public visibility so
+/// that users of the macro can use it.
+///
+/// The logic is separated into its own function to reduce binary size, as otherwise the logic
+/// would be inlined for every use case. As of June 2020, this saves ~40KB of binary size for the
+/// CTAP2 application of OpenSK.
+pub fn read_cbor_map_peek_value(
+    it: &mut Peekable<btree_map::IntoIter<KeyType, Value>>,
```
kaczmarczyck, line 90: Since this is a performance optimization, did you also measure how much faster an `iter` in a `BTreeMap` is compared to the `remove`? Both have O(log(n)) worst case, right?

libraries/cbor/src/macros.rs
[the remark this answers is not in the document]
ia0: It's not a wrong example (I just didn't test if it needed adjustments, one just needs to add `# use cbor::values::IntoCborKey;` which is done by the macro). The idea is to show that anything that converts to a key can be used, not only base types.

libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,127 @@
...
+/// # #[macro_use]
+/// # extern crate cbor;
+/// #
+/// # fn main() {
+/// #     let map = alloc::collections::BTreeMap::new();
+/// read_cbor_map! {
+///     map,
+///     x @ 1,
+///     y @ 2,
+/// };
```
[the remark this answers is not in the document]
gendx: Good point on the `let { key => value } = map;` syntax. I somehow considered using the arrow, and then got confused, but it's clear with the `let` part. I'd vote for `extract_cbor_map`.

libraries/cbor/src/macros.rs
gendx: The function is only used inside of the macro. But because when using the macro the macro contents are expanded into some code, it's required that the functions introduced by this expansion are visible in the current scope. I'm not aware of a way to restrict that, but maybe @ia0 has an idea?

libraries/cbor/src/macros.rs
gendx: Iterating through all the items in the map without doing any re-balancing, and then dropping the whole map at once, should overall be `O(n)`, whereas doing `n` removes amounts to `O(n log(n))`. Also, removing an item means some moving/copying of data around to re-balance the other items, which is avoided here. That's the primary benefit of this optimization.

libraries/cbor/src/macros.rs
ia0: I don't know any solution. What I do in those cases is to put `internal` in the name of the variable (function name, macro name, etc) and add `#[doc(hidden)]`.

```
#[doc(hidden)]
pub fn internal_read_cbor_map_peek_value(...) ...
// or
pub fn read_cbor_map_peek_value_internal(...) ...
```

libraries/cbor/src/macros.rs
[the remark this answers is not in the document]
gendx: Applied the `let { key => value } = map;` suggestion. Keeping the macro name for now to let others weigh in to reach a consensus. Changing the name will be a simple search + replace once we agree on it.

libraries/cbor/src/macros.rs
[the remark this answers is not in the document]
gendx: Added a test key example and use `cbor::values::IntoCborKey`.

libraries/cbor/src/macros.rs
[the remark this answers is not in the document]
kaczmarczyck: Problem with extract_ is that is name-clashes with all the other extract_x functions that return an x. Whereas this macro does not return a cbor_map, so we shouldn't call it extract_cbor_map. Also, the macro does not take a cbor::Value(MapValue) as input, but an actual BTreeMap, right? If that is correct, it shouldn't be cbor_map, but just map. Or is this supposed to hint on the fact that it is a map of CBOR?

So anything like `(parse|destructure)_map` potentially with a `_of_cbor`?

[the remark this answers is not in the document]
gendx: - The macro is definitely CBOR-specific (due to the keys), so the name should include that.
- The problem with taking a `cbor::Value` as input is that it's unclear how the macro should handle the case where the input is not a `MapValue`. It's up to the caller to decide that. In almost all cases with do `extract_map(map)?`, but there is also `cose_key.0`, so the input should be BTreeMap. But I don't think it's useful to give a complex name to the macro including "of CBOR" or "BTreeMap".

So I'd definitely vouch for `[verb]_cbor_map`. The question is which verb.
- "Destructure" is likely the best semantically, but also the more verbose.
- There's no real parsing here (the `[u8]` => `cbor::Value` conversion happens elsewhere).
- "Read" is concise although not the most precise.

Given that, I'd suggest `destructure_cbor_map`.

WDYT?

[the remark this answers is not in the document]
kaczmarczyck: The macro should take a map, not a cbor::Value, I agree.
`destructure` is good. So`destructure_cbor_map` is good enogh for me.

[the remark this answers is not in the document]
gendx: I take it as a unanimous vote for `destructure_cbor_map` from @ia0 and @kaczmarczyck.

gendx:
> It might be worth splitting the PR:
> 
> 1. A semantics-preserving change introducing and using the macro.
> 2. An optimization of the macro implementation.
> 
> The first change is an obvious benefit since it doesn't change the behavior and removes boilerplate.
> 
> The second change is questionable since the benefit doesn't seem worth the drawbacks (inconvenient usage, code complexity).

- The usage is not more or less inconvenient, thanks to the macro.
- The extra code complexity is the `destructure_cbor_map_peek_value` function. That's 25 lines of code. I can add some more comments for clarity.

Overall, the split only provides little benefit. It could be useful to benchmark both strategies before/after, however we don't have such a benchmarking framework available at the moment.

libraries/cbor/src/macros.rs
gendx: Ack. We don't currently generate any docs, but in any case the documentation for this function clearly states that it's an internal detail. I guess this sounds good then?

ia0: Yes, the `#[doc(hidden)]` is only useful if the library is used by others (such that they are not tempted to use it). However, the `internal` in the name is a pattern useful even for us: a reviewer would see from the function name that some additional check has to be done (in a similar way to `unsafe` and `unchecked`). But I guess for this particular case, there's low chance someone would call it (by mistake or intentionally). This is mostly to give rationale of why those patterns can be useful.

ia0:
> * The usage is not more or less inconvenient, thanks to the macro.

I think it is, because one has to pay attention to the keys order. It may involve some ping-pong with the tests.

> * The extra code complexity is the `destructure_cbor_map_peek_value` function. That's 25 lines of code. I can add some more comments for clarity.

I agree that the complexity is well encapsulated (modulo the leak to the usability point above).

> Overall, the split only provides little benefit. It could be useful to benchmark both strategies before/after, however we don't have such a benchmarking framework available at the moment.

That's my main concern, that this would be premature optimization. But I actually would not be surprised if the number of fields of credentials (and maybe other structs) would one day grow above 100. In that case this optimization would probably be useful. So I'm fine to do it now.

libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,142 @@
...
+        // in strictly increasing order of keys. Then, the repeated calls to the "peek value"
+        // helper function will consume this iterator and yield values (or `None`) when reaching
+        // the keys to extract.
+        //
+        // This is where the pre-requisite that keys to extract are sorted is important: the
+        // algorithm does a single linear scan over the iterator and therefore keys to extract have
+        // to come in the same order (i.e. sorted).
+        let mut it = $map.into_iter().peekable();
+        $(
+        let $variable: Option<Value> = destructure_cbor_map_peek_value(&mut it, $key.into_cbor_key());
```
ia0, line 89: `$key.into_cbor_key()` is constant. The compiler probably optimizes it (common subexpression elimination) but it's not guaranteed if it cannot detect that `into_cbor_key` is pure. If you care about performance, it would be useful to factor it out:

```
let key = $key.into_cbor_key();
$(
let $variable ... (&mut it, &key);
)+
```

You also need to change the signature of the helper function to take the key by reference since `KeyType` is not `Copy`.

src/ctap/data_formats.rs
```diff
@@ -31,16 +31,18 @@ impl TryFrom<cbor::Value> for PublicKeyCredentialRpEntity {
...
-            .remove(&cbor_text!("icon"))
-            .map(extract_text_string)
-            .transpose()?;
+        destructure_cbor_map! {
+            let {
+                "id" => rp_id,
+                "icon" => rp_icon,
+                "name" => rp_name,
+            } = extract_map(cbor_value)?;
+        };
```
ia0, line 40: I'm surprised we need a semi-colon here. Do you get an error/warning/clippy if you remove it?

libraries/cbor/src/macros.rs
ia0: Sorry, I mixed it up :-) $key is *not* constant. So we have to compute it each time.

gendx: Yes, it's `$($key)+`

src/ctap/data_formats.rs
gendx: I didn't check. But to me it was more in the spirit of the `let {foo} = bar;` syntax to have a full statement, followed by a semi-colon, inside of the macro.

ia0: Ok, it's not important. It's just that I recently got a warning (or error) that macro calls are either $foo(...); or $foo { ... }. But maybe it actually depends on whether the macro result is an expression or a toplevel item. I'm not sure.

ia0: I got time to play a bit in the playground. Apparently what I said is only enforced for items, not for expressions. However, it's essentially similar to if and match expressions. In Rust, braces are not followed by semi-colons (struct, enum, if, match, etc.) but for expressions (if, match, etc.) it doesn't hurt to add it. I don't mind, I was just surprised to see a semi-colon there.

gendx: Oh, I thought you meant the semi-colon inside of the macro. The one outside of the macro is indeed not necessary with this new syntax.

ia0: Yes, the one outside which is the one relevant to the compiler and clippy (the one inside is up to us and I think we should have it to be coherent with the let syntax). If we have a clippy test and it doesn't complain I guess it's fine (and I hope it should be the same result if we have a semi-colon after an if/match otherwise I would consider it a clippy bug).

gendx:
Regarding heap usage of these changes, I made some simple measurements with `./deploy.py --board nrf52840dk --opensk --verbose --panic-console --debug-allocations`, now that the `--verbose` more is fixed (#87).

On a simple https://webauthn.io/ register + login flow, I get the following heap usage for the three strategies:
- `map.get().cloned()`: peak heap usage is ~12.7 KB
- `map.remove()`: peak heap usage is ~7.2 KB
- This pull request: peak heap usage is ~7.2 KB

This is generally expected, as `map.remove()` shouldn't reallocate values of the map. It could create/remove nodes inside of the B-tree, although if there are only few key-value pairs, the B-tree likely collapses to a single root node. However, `map.remove()` can memcpy keys/values inside of B-tree nodes (with is avoided by the peekable iterator strategy).

So overall, #113 reduced heap usage by ~43% in this simple register + login scenario, while this pull request improves the ergonomics with the new macro and potentially removes some unneeded memcopies.

libraries/cbor/src/macros.rs
```diff
@@ -12,6 +12,142 @@
...
+/// # extern crate cbor;
+/// #
+/// # fn main() {
+/// #     let map = alloc::collections::BTreeMap::new();
+/// destructure_cbor_map! {
+///     let {
+///         1 => x,
+///         "key" => y,
+///     } = map;
+/// };
```
ia0, line 51: There is still this one.

src/ctap/data_formats.rs
gendx: > If we have a clippy test and it doesn't complain [...]

We currently don't have a Clippy test. IIRC this was due to some dependencies somehow triggering plenty of lints.

jmichelp: It's due to Tock and clippy doesn't allow whitelisting.
But maybe the situation improved and we can add clippy. I'll have a look at it.

PR #117, opened 2020-06-09
Add elf2tab output to the reproducible checks

gendx:
This pull request adds a dump of elf2tab to the reproducible workflow. This allows to easily check for regressions in binary size, notably thanks to the following lines.

```
Number of writeable flash regions: 0
  Adding .crt0_header section. Offset: 64 (0x40). Length: 64 (0x40) bytes.
Entry point is in .text section
  Adding .text section. Offset: 128 (0x80). Length: 176436 (0x2b134) bytes.
  Adding .stack section. Offset: 176564 (0x2b1b4). Length: 16384 (0x4000) bytes.
```

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -428,7 +428,12 @@ class OpenSKInstaller:
         "--stack={}".format(STACK_SIZE), "--app-heap={}".format(APP_HEAP_SIZE),
         "--kernel-heap=1024", "--protected-region-size=64"
     ])
-    self.checked_command(elf2tab_args)
+    if self.args.elf2tab_output:
+      output = self.checked_command_output(elf2tab_args)
+      with open(self.args.elf2tab_output, 'a') as f:
+        f.write(output)
+    else:
+      self.checked_command(elf2tab_args)
 
   def install_tab_file(self, tab_filename):
     assert self.args.application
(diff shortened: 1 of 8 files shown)


deploy.py
```diff
@@ -861,6 +866,14 @@ def main(args):
             "storage (i.e. unplugging the key will reset the key)."),
   )
 
+  main_parser.add_argument(
+      "--elf2tab-output",
+      metavar="FILE",
+      dest="elf2tab_output",
+      default=None,
```
jmichelp, line 873: I would specify this argument with a type:
```python
main_parser.add_argument(
  "--elf2tab-output",
  metavar="FILE",
  type=argparse.FileType("a"),
  dest="elf2tab_output",
  default=None,
```

This way, opening/closing the file is left to the argparse module and you don't have to handle errors.
The code above that logs the error will then be:
```python
if self.args.elf2tab_output:
  self.args.elf2tab_output.write(self.checked_command_output(elf2tab_args))
else:
  self.checked_command(elf2tab_args)
```

reproduce_board.sh
```diff
@@ -16,10 +16,14 @@
...
+echo "========================================" >> reproducible/elf2tab.txt
+echo "Board: $BOARD" >> reproducible/elf2tab.txt
+echo "----------------------------------------" >> reproducible/elf2tab.txt
+
 ./deploy.py --verbose-build --board=$BOARD --no-app --programmer=none
 ./third_party/tock/tools/sha256sum/target/debug/sha256sum third_party/tock/target/thumbv7em-none-eabi/release/$BOARD.bin >> reproducible/binaries.sha256sum
 tar -rvf reproducible/reproduced.tar third_party/tock/target/thumbv7em-none-eabi/release/$BOARD.bin
 
-./deploy.py --verbose-build --board=$BOARD --opensk --programmer=none
+./deploy.py --verbose-build --board=$BOARD --opensk --programmer=none --elf2tab-output reproducible/elf2tab.txt
```
jmichelp, line 27: nit: for consistency, add a equal sign

PR #120, opened 2020-06-11
implements the AuthenticatorSelection command

kaczmarczyck:
Tracked by #106 .

Implements the new command AuthenticatorSelection and handles unknown commands more gracefully (error code instead of panic).

- [x] Tests pass


src/ctap/mod.rs
```diff
@@ -1094,6 +1096,16 @@ where
...
 
+    #[cfg(feature = "with_ctap2_1")]
+    fn process_selection(&self, cid: ChannelID) -> Result<ResponseData, Ctap2StatusCode> {
+        (self.check_user_presence)(cid)?;
+        Ok(ResponseData::AuthenticatorSelection)
+    }
+
+    fn process_unknown_command(&self) -> Result<ResponseData, Ctap2StatusCode> {
+        Err(Ctap2StatusCode::CTAP1_ERR_INVALID_COMMAND)
+    }
```
gendx, line 1107: Can you add a unit test for this case?

kaczmarczyck: Sure, done.

PR #121, opened 2020-06-15
Always run git-diff in reproducible workflow, even if the job failed so far.

gendx:
This change allows to see both the binary size comparison and the hashes comparison, even if the former failed. Otherwise, the job would be cancelled after the first failure, preventing us from seeing all the diffs.

Note: It's unclear from the documentation how GitHub will aggregate the status of the whole job in case one step fails but not the next. We might have to adjust that after seeing the effect in practice on pull requests. On the other hand, if the sizes don't match, the cryptographic hashes shouldn't match either.

- [x] Tests pass

### DIFF
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -37,6 +37,8 @@ jobs:
           path: reproducible/reproduced.tar
 
       - name: Comparing binary sizes
+        if: always()
         run: git diff --no-index reproducible/reference_elf2tab_${{ matrix.os }}.txt reproducible/elf2tab.txt
       - name: Comparing cryptographic hashes
+        if: always()
         run: git diff --no-index reproducible/reference_binaries_${{ matrix.os }}.sha256sum reproducible/binaries.sha256sum


.github/workflows/reproducible.yml
```diff
@@ -37,6 +37,8 @@ jobs:
           path: reproducible/reproduced.tar
 
       - name: Comparing binary sizes
+        if: always()
```
jmichelp, line 40: Comparing hashes doesn't make sense if the script `reproduce_hashes.sh` failed as the required files won't be created.
But I don't think Github workflows will allows us to run 2 steps if and only if the text file has been produced. I guess it's doable for the 1st check but running the second comparison even if the 1st one failed would be tricky.

gendx: The use case I see is:
1. Someone writes a PR. They can only test locally on either of Ubuntu or MacOS (or worse, maybe neither).
2. They see the size difference mismatch. But not the hash differences, especially on the platform they don't develop on.
3. They add a commit for the size difference.
4. They now see the hash differences.
5. They add a commit for the hash differences.

The proposal is to merge 2+4 and 3+5 to increase developer velocity (2 and 4 have a high latency).

If things broke before and there is no available files, git-diff will just fail but it won't be the end of the world, nor will it drain CI resources.

Maybe there's a way to optimize this further than `always()`, but unless we have a clear way of doing it, I think it's fine to just go with this non-optimal solution.

gendx:
Travis-CI failure seems unrelated. Merging as-is.

PR #123, opened 2020-06-22
Update libtock dependency to libtock-core

### LINKED ISSUE #95: Update libtock dependency to libtock-rs-core
Due to the revamp of the runtime in libtock-rs to use futures, we are currently pinned at an old version of libtock-rs (https://github.com/tock/libtock-rs/tree/ab2c945184b98ecae3e70ac678e9f5231deef73b).

However, the core part of libtock-rs has now been split out of the drivers: https://github.com/tock/libtock-rs/tree/master/core. We can update to depend on that to keep in sync with upstream, which should also allow to reduce the number of our [custom patches](https://github.com/google/OpenSK/tree/master/patches/libtock-rs).

However, this means that for now we'll have to import all drivers in OpenSK, given that libtock-rs's drivers are now tailored for the futures-based runtime. But we can re-use and adapt the old ones. For now our uses of libtock are:

```shell
$ git grep libtock:: | cut -d' ' -f2 | cut -d':' -f3 | cut -d';' -f1 | sort -u
buttons
console
led
result
rng
syscalls
timer
```

Given that `libtock::{result, syscalls}` are now in libtock core, this is only 5 drivers to maintain - until (1) libtock supports again synchronous drivers or (2) we migrate OpenSK to the futures-based runtime.

gendx:
Fixes #95.

This pull request updates the libtock-rs dependency:
- Depend on the latest libtock-rs/core crate.
- Remove the 8 patches in libtock-rs. Adds a new patch for the custom allocator, until https://github.com/tock/libtock-rs/pull/205 is merged upstream.
- Adds the relevant drivers in a new crate. These drivers are mostly derived from those in the [previously pinned commit](https://github.com/tock/libtock-rs/tree/ab2c945184b98ecae3e70ac678e9f5231deef73b/src) of libtock-rs, with some adaptations to changes in libtock-rs/core since that commit. This code duplication is needed because the drivers in libtock-rs are currently exclusively asynchronous, but OpenSK hasn't been ported to that asynchronous runtime.
- Add two examples to more specifically test integration of the panic handler and allocator.
  - `panic_test`, which immediately panics.
  - `oom_test`, which allocates powers-of-2-sized chunks, until it eventually triggers an out-of-memory error. This allows to check the size available in the heap, as well as the alloc error handler.

Tested on an nRF52840-DK board, with the register/login flow of https://webauthn.io/.

- [ ] Tests pass
  - [x] `./deploy.py --board nrf52840dk --crypto_bench`
  - [x] `./deploy.py --board nrf52840dk --crypto_bench --panic-console`
  - [x] `./deploy.py --board nrf52840dk --crypto_bench --panic-console --debug-allocations`
  - [x] `./deploy.py --board nrf52840dk --panic_test`
  - [x] `./deploy.py --board nrf52840dk --panic_test --panic-console`
  - [x] `./deploy.py --board nrf52840dk --oom_test`
  - [x] `./deploy.py --board nrf52840dk --oom_test --panic-console`
  - [x] `./deploy.py --board nrf52840dk --oom_test --panic-console --debug-allocations`
  - [x] `./deploy.py --board nrf52840dk --opensk`
  - [x] `./deploy.py --board nrf52840dk --opensk --panic-console`
  - [x] `./deploy.py --board nrf52840dk --opensk --debug --panic-console`
  - [x] `./deploy.py --board nrf52840dk --opensk --verbose --panic-console`
  - [x] `./deploy.py --board nrf52840dk --opensk --verbose --panic-console --debug-allocations`


gendx:
> Overall comment: we can't mix code we authored and code we took from third_party projects. So we need to keep code borrowed from libtock-rs under `third_party` directory.

Done.

rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2020-02-03
+nightly-2020-06-10
```
jmichelp, line 1: Should we symlink this file to the one in `third_party/tock/` so that we're always in sync?

src/embedded_flash/syscall.rs
```diff
@@ -41,16 +41,14 @@ mod memop_nr {
...
+    let code = syscalls::command(DRIVER_NUMBER, command_nr::GET_INFO, nr, arg);
+    code.map_err(|e| StorageError::KernelError {
+        code: e.return_code,
+    })
 }
 
 fn memop(nr: u32, arg: usize) -> StorageResult<usize> {
-    let code = unsafe { syscalls::memop(nr, arg) };
+    let code = unsafe { syscalls::raw::memop(nr, arg) };
     if code < 0 {
```
jmichelp, line 52: Why not doing the same as above with `code.map_err()` pattern?

third_party/libtock-drivers/src/result.rs
```diff
@@ -0,0 +1,133 @@
...
+// We sometimes need to handle errors in a `TockResult` by calling `unwrap`. However,
+// `Result::unwrap` requires that the error type implements `core::fmt::Debug`. Under the hood,
+// this requires dynamic dispatch, which has non-negligible overhead on code size. Therefore errors
+// don't derive from `Debug` in libtock-rs.
+//
+// Instead one can call `.ok().unwrap()` which relies on `Option::unwrap` and doesn't require any
+// debugging of the error type.
+//
+// This trait allows to flexibly use `Result::unwrap` or `Option::unwrap` and is configured to do
+// so depending on the `debug_ctap` feature.
```
jmichelp, line 16: Now I'm confused.
Who authored `third_party/libtock-drivers`? According to the TOML it's Tock team bug `debug_ctap` is OpenSK specific. Is it Google's code, Tock's code or a mix of both?

third_party/libtock-drivers/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "libtock_drivers"
+version = "0.1.0"
+authors = ["Tock Project Developers <tock-dev@googlegroups.com>"]
+license = "MIT/Apache-2.0"
```
jmichelp, line 5: To be valid, the Apache license requires the LICENSE file to be present as well as a copyright+license header on each file. But if it's Tock's code, they have to fix it upstream.

src/embedded_flash/syscall.rs
gendx: Because `syscalls::raw::memop` returns a `usize` whereas `syscalls::command` returns a `Result`.

third_party/libtock-drivers/src/result.rs
gendx: It's a mix of both. It is derived from the libtock-rs developed by "Tock Project Developers", with some additional code (in this pull request) to make it work with the new syscall API.

In a sense "Tock Project Developers" already includes some Google employees that contributed to (lib)tock, but if that's too unclear, I can add explicit names in the `authors` section.

third_party/libtock-drivers/Cargo.toml
gendx: > libtock-rs is licensed under either of
> 
> - Apache License, Version 2.0 (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0)
> - MIT license (LICENSE-MIT or http://opensource.org/licenses/MIT)
> 
> at your option.

If Apache is too complicated, can we pick MIT for this derived work (i.e. the `third_party/libtock-drivers` crate)? Then ultimately this `third_party/libtock-drivers` should hopefully disappear if the relevant API is available in libtock-rs directly.

rust-toolchain
gendx: There are 3 `rust-toolchain` files:

- `third_party/tock/rust-toolchain` to build the boards.
- `third_party/libtock-rs/rust-toolchain` to build provide userspace libtock.
- `rust-toolchain` to build OpenSK. This should be in sync with userspace's libtock, as building the OpenSK app depends on that.

The kernel and userspace don't have to be in sync, and in fact `third_party/tock/rust-toolchain` and `third_party/libtock-rs/rust-toolchain` are currently not in sync (anyway, as long as there are these `rust-toolchain` files, there will be times when they are out-of-sync, given that separate repositories are not updated exactly at the same time).

Therefore, what would make the most sense is to symlink `third_party/libtock-rs/rust-toolchain`, not the boards.

WDYT?

jmichelp: SGTM

ia0:
Tested webauthn.io: register successful + login successful + unplug/replug + login successful.
Tested 80% of the test suite: all green up to when I hit enter too fast after replugging the key and got an error because the test suite did not find the key. Those errors should retry to avoid having to start the test suite from the beginning.

LGTM with regard to testing on a dongle + DFU.

Note: I had to change `"nrfutil", "dfu", "usb-serial"` into `"nrfutil", "dfu", "usb_serial"` in deploy.py though, but that was already the case last time I tried.

gendx:
I rebased and updated the reference binaries. This should now be good to go.

PR #127, opened 2020-07-02
Client pin features

kaczmarczyck:
This PR adds new subcommands from CTAP 2.1 in clientPin.

CTAP 2.1 introduces an interface for PIN protocol versions, since a v2 is currently being drafted. I decided to move all clientPin code for the existing PIN protocol v1 to a separate module. For better code sharing with a potential v2, we might introduce the common interface later when needed. I decided to not proactively refactor v1 into that interface, and wait until it is stable.

This has the added benefit of moving some complexity out of `ctap/mod.rs` which will have to accommodate more commands as they are introduced.

We will need to do more testing of changes for 2.1 in the wild when platforms and relying parties support it.

This PR contains some unused functions, this is intentional. They will later be used by the "Minimum PIN Length Extension" to allow some relying parties to see the minimum PIN length required by this authenticator, and include it in its attestation during MakeCredential. This feature establishes trust with relying parties with strict PIN policies.

Other open TODOs are (1) enforcing PIN length in code points and (2) places where the specification is ambiguous and discussed in the [FIDO Alliance](https://github.com/fido-alliance/fido-2-specs/issues/901) (you might lack permissions).

- [ ] Tests pass


src/ctap/storage.rs
```diff
@@ -892,4 +994,68 @@ mod test {
...
+            _DEFAULT_MIN_PIN_LENGTH_RP_IDS
+        );
+
+        // Changes by the setter are reflected by the getter..
+        let rp_ids = vec![String::from("example.com")];
+        assert_eq!(
+            persistent_store._set_min_pin_length_rp_ids(rp_ids.clone()),
+            Ok(())
+        );
+        assert_eq!(persistent_store._min_pin_length_rp_ids(), rp_ids);
```
ia0, line 1031: The expected value is not really `rp_ids` but `merge(_DEFAULT, rp_ids)`. It doesn't matter for now because `_DEFAULT` is empty. But if we think the default won't be empty, we should probably write the correct expectation.

src/ctap/storage.rs
```diff
@@ -59,11 +62,25 @@ const PIN_RETRIES: usize = 4;
...
+#[cfg(feature = "with_ctap2_1")]
+const NUM_TAGS: usize = 10;
 
 const MAX_PIN_RETRIES: u8 = 6;
 const ATTESTATION_PRIVATE_KEY_LENGTH: usize = 32;
 const AAGUID_LENGTH: usize = 16;
+#[cfg(feature = "with_ctap2_1")]
+const DEFAULT_MIN_PIN_LENGTH: u8 = 4;
+#[cfg(feature = "with_ctap2_1")]
+const _DEFAULT_MIN_PIN_LENGTH_RP_IDS: Vec<String> = Vec::new();
```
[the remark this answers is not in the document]
gendx: In a `no_std` context, `lazy_static` can be used with the `spin_no_std` feature, but that depends on `spin` which triggers a Cargo audit warning for us (https://github.com/google/OpenSK/issues/24).

src/ctap/storage.rs
kaczmarczyck: Fixed. How do you think should be approach `MAX_RP_IDS_LENGTH`? The specification simply says: If you can't store the list of RP IDs, return `CTAP2_ERR_KEY_STORE_FULL`. And I believe storage v2 will have that feature? I thought we probably should have another mechanism to block too large requests in v1 and v2, just to have at least some guarantees for how many credentials we can store.

src/ctap/command.rs
```diff
@@ -304,6 +328,30 @@ impl TryFrom<cbor::Value> for AuthenticatorClientPinParameters {
         let pin_auth = pin_auth.map(extract_byte_string).transpose()?;
         let new_pin_enc = new_pin_enc.map(extract_byte_string).transpose()?;
         let pin_hash_enc = pin_hash_enc.map(extract_byte_string).transpose()?;
+        #[cfg(feature = "with_ctap2_1")]
+        let min_pin_length = min_pin_length
+            .map(extract_unsigned)
+            .transpose()?
+            .map(|m| m as u8);
```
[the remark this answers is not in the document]
ia0: Ok I see. As you want, I just wanted to make sure the behavior was expected.

src/ctap/data_formats.rs
```diff
@@ -1166,6 +1166,13 @@ mod test {
         let policy_error = CredentialProtectionPolicy::try_from(cbor_policy_error);
         let expected_error = Err(Ctap2StatusCode::CTAP2_ERR_CBOR_UNEXPECTED_TYPE);
         assert_eq!(policy_error, expected_error);
+
+        for policy_literal in 1..=3 {
```
ia0, line 1170: We still have room for errors by hardcoding the last index. One solution would be to add a last enum value (always at the end without explicit value):

```rust
pub enum CredentialProtectionPolicy {
    UserVerificationOptional = 0x01,
    UserVerificationOptionalWithCredentialIdList = 0x02,
    UserVerificationRequired = 0x03,
    DoNotUseAndKeepLast,  // we can find a better name
}
```

Then we can loop with `for policy_literal in 1..CredentialProtectionPolicy::DoNotUseAndKeepLast as i64`.

src/ctap/data_formats.rs
```diff
@@ -1180,6 +1187,15 @@ mod test {
         );
         let created_cbor: cbor::Value = authenticator_transport.unwrap().into();
         assert_eq!(created_cbor, cbor_authenticator_transport);
+
+        let transports = ["usb", "nfc", "ble", "internal"];
+        for transport in transports.iter() {
```
ia0, line 1192: This is not future-proof, but here I really don't see how to do better without procedural macros.

src/ctap/data_formats.rs
```diff
@@ -1322,6 +1338,17 @@ mod test {
         assert_eq!(sub_command, Ok(expected_sub_command));
         let created_cbor: cbor::Value = sub_command.unwrap().into();
         assert_eq!(created_cbor, cbor_sub_command);
+
+        #[cfg(not(feature = "with_ctap2_1"))]
+        let last_literal = 0x05;
+        #[cfg(feature = "with_ctap2_1")]
+        let last_literal = 0x09;
```
ia0, line 1345: We would not need this using the `DoNotUseAndKeepLast` enum value. It would automatically get the right value.

src/ctap/storage.rs
```diff
@@ -415,17 +417,19 @@ impl PersistentStore {
...
-                    .replace(
-                        index,
-                        StoreEntry {
-                            tag: PIN_RETRIES,
-                            data: &[new_value],
-                            sensitive: false,
-                        },
-                    )
-                    .unwrap();
+                if entry.data[0] > 0 {
```
ia0, line 420: To avoid extra-indentation, you can use:

```rust
if entry.data[0] == 0 {
    return;
}
```

src/ctap/storage.rs
ia0: Both v1 and v2 reject too large entries. However only v1 returns "StoreFull" error, v2 returns "InvalidArgument" since the error is on the client. Both v1 and v2 reject with "StoreFull" when there is not enough capacity.

The `MAX_RP_IDS_LENGTH` is CTAP-specific and unrelated to the persistent store. The logic should be implemented here in the CTAP persistent storage. Regarding provisioning, it also has to be done in CTAP since it is CTAP-specific. For v2, the way I did it is to limit the number of instances per object type and account for the maximum size if all instances are used with maximal size. [Here](https://github.com/ia0/OpenSK/blob/d41a8fc10a9bfc1697d1f5674c31f1e54746d76a/src/ctap/storage/key.rs#L54-L87) is the mapping from keys to objects (which defines the maximum number of instances per object type). To compute the minimum required capacity, one has to know the maximum size for each object type.

src/ctap/response.rs
```diff
@@ -301,13 +305,22 @@ mod test {
...
         let response_cbor: Option<cbor::Value> =
             ResponseData::AuthenticatorGetInfo(get_info_response).into();
+        #[cfg(not(feature = "with_ctap2_1"))]
+        let expected_cbor = cbor_map_options! {
+            0x01 => cbor_array_vec![vec!["FIDO_2_0"]],
+            0x03 => vec![0x00; 16],
+        };
+        #[cfg(feature = "with_ctap2_1")]
         let expected_cbor = cbor_map_options! {
             0x01 => cbor_array_vec![vec!["FIDO_2_0"]],
```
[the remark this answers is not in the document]
ia0: Maybe using a variable for the versions, so that it's clear it's the same value.

```rust
let versions = vec!["FIDO_2_0".to_string()];
let get_info_response = AuthenticatorGetInfoResponse {
    versions: versions.clone(),
    extensions: None,
    ...
};
let response_cbor = ...;
let expected_cbor = cbor_map_options! {
    0x01 => cbor_array_vec![versions],
    0x03 => vec![0x00; 16],
    0x0D => 4
};
```

Ping @gendx to also add feature support for the `cbor_map_options` macro.

src/ctap/storage.rs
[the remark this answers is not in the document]
ia0: Is this "default min pin length RP ids" something specified in the standard? Do we really want to bake it in the binary? Or should we instead do as for the attestation certificate, aaguid, and friends and add a command to set it? We would have a store entry just for that default value (independent of the actual final value).

src/ctap/command.rs
```diff
@@ -292,6 +312,10 @@ impl TryFrom<cbor::Value> for AuthenticatorClientPinParameters {
                 4 => pin_auth,
                 5 => new_pin_enc,
                 6 => pin_hash_enc,
+                7 => min_pin_length,
+                8 => min_pin_length_rp_ids,
+                9 => permissions,
+                10 => permissions_rp_id,
```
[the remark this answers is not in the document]
kaczmarczyck: I'll close this as work for another PR.

src/ctap/data_formats.rs
gendx: It looks like the `enum_iterator` crate achieves exactly what we need here: https://docs.rs/enum-iterator/0.6.0/enum_iterator/trait.IntoEnumIterator.html

Not sure whether we want to pull in another dependency, but the code boils down to the following macro: https://docs.rs/crate/enum-iterator-derive/0.6.0/source/src/lib.rs

src/ctap/storage.rs
```diff
@@ -62,13 +62,11 @@ const PIN_RETRIES: usize = 4;
...
 const ATTESTATION_CERTIFICATE: usize = 6;
 const AAGUID: usize = 7;
-#[cfg(not(feature = "with_ctap2_1"))]
-const NUM_TAGS: usize = 8;
 #[cfg(feature = "with_ctap2_1")]
 const MIN_PIN_LENGTH: usize = 8;
 #[cfg(feature = "with_ctap2_1")]
 const MIN_PIN_LENGTH_RP_IDS: usize = 9;
-#[cfg(feature = "with_ctap2_1")]
+// Different NUM_TAGS make the storage incompatible, so we use max(8,10).
```
gendx, line 69: Nit: "Different NUM_TAGS depending on the CTAP version make the storage [...]"

src/ctap/storage.rs
```diff
@@ -397,9 +395,13 @@ impl PersistentStore {
     }
 
     pub fn pin_retries(&self) -> u8 {
-        self.store
-            .find_one(&Key::PinRetries)
-            .map_or(MAX_PIN_RETRIES, |(_, entry)| entry.data[0])
+        match self.store.find_one(&Key::PinRetries) {
+            None => MAX_PIN_RETRIES,
+            Some((_, entry)) => {
+                debug_assert_eq!(entry.data.len(), 1);
```
gendx, line 401: Does this extra line require moving away from `map_or`?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1108 @@
+// Copyright 2019 Google LLC
```
gendx, line 1: 2020

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1108 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crypto::rng256::ThreadRng256;
+
+    fn set_standard_pin(persistent_store: &mut PersistentStore) {
+        let mut pin = [0x00; 64];
```
gendx, line 606: In `0x00` the type is not obvious. It would be clearer to write is as `0u8` or `0u32` depending on what it actually is.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1108 @@
...
+        pin[0] = 0x31;
+        pin[1] = 0x32;
+        pin[2] = 0x33;
+        pin[3] = 0x34;
+        let mut pin_hash = [0; 16];
```
gendx, line 611: Same here.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1108 @@
...
+    }
+
+    fn encrypt_standard_pin(shared_secret: &[u8; 32]) -> Vec<u8> {
+        let aes_enc_key = crypto::aes256::EncryptionKey::new(shared_secret);
+        let mut blocks = [[0u8; 16]; 4];
+        blocks[0][0] = 0x31;
+        blocks[0][1] = 0x32;
+        blocks[0][2] = 0x33;
+        blocks[0][3] = 0x34;
+        let iv = [0; 16];
```
gendx, line 623: Same here and for the following arrays in this file.

src/ctap/data_formats.rs
ia0: Right, the fact that we only need it in test makes it less problematic to add a dependency. This would also fix the issue with the transport enum.

src/ctap/storage.rs
ia0: I think the current version is better. Because what makes the storage incompatible is different NUM_TAGS. It doesn't matter why we have different NUM_TAGS. In particular, if we add a new tag that is present in all CTAP versions it would still be a breaking change storage-wise.

Note that this issue is only present with v1. The number of keys is fixed in v2.

src/ctap/command.rs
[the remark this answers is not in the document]
kaczmarczyck: I made it more explicit.

src/ctap/command.rs
```diff
@@ -434,6 +490,9 @@ mod test {
 
     #[test]
     fn test_from_cbor_client_pin_parameters() {
+        // TODO(kaczmarczyck) inline the #cfg when the ICE is resolved:
+        // https://github.com/rust-lang/rust/issues/73663
+        #[cfg(not(feature = "with_ctap2_1"))]
```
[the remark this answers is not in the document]
kaczmarczyck: This request seems to not only affect this PR, so we can treat it independently?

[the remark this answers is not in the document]
kaczmarczyck: I created #128 to track that.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1103 @@
...
+            0x8D, 0x7A, 0xA3, 0x9F, 0x7F, 0xC6, 0x08, 0x13, 0x9A, 0xC8, 0x56, 0x97, 0x70, 0x74,
+            0x99, 0x66,
+        ];
+        pin_protocol_v1.consecutive_pin_mismatches = 3;
+        assert_eq!(
+            pin_protocol_v1.check_pin_hash_enc(
+                &mut rng,
+                &mut persistent_store,
+                &aes_dec_key,
+                pin_hash_enc.clone()
```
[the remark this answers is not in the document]
kaczmarczyck: Removed.

src/ctap/response.rs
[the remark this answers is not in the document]
kaczmarczyck: Variable is added, issue is created.

kaczmarczyck:
This is a CTAP2.1 feature, tracked in #106 .

src/ctap/storage.rs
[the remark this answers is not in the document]
kaczmarczyck: It's probably the same use case, I agree. Since it is unused until I add the extension dealing with this value, can we agree on implementing that as part of that extension?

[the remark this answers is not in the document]
ia0: Yes, I'm fine with that.

src/ctap/storage.rs
kaczmarczyck: Let's chat offline and resolve this in the extension that is actually using `MAX_RP_IDS_LENGTH`.

src/ctap/data_formats.rs
kaczmarczyck: This seems like a complicated workaround for testing an enum. `DoNotUseAndKeepLast` should at least be behind a `#[cfg(test)]` ? We'd have to add this dummy entry into all enums that use integer indices. Because the risk is someone implementing an incorrect extension, and also not implementing a correct unit test for it?

src/ctap/data_formats.rs
kaczmarczyck: Agreed for not being future proof. We can discuss in the comment above about if this is tolerable.

src/ctap/storage.rs
kaczmarczyck: Done.

src/ctap/storage.rs
kaczmarczyck: Done, and changed to just say maximum for easier maintenance.

src/ctap/storage.rs
kaczmarczyck: I didn't see a way to include the debug_assert_eq without a match.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done.

src/ctap/data_formats.rs
ia0: Yes you would need to add `#[cfg(test)]` and have it for all enums. There should not be more risks with this compared to hard-coding the last index.

However, I didn't realize that we only need the crate for testing, so I think this would be the best way to go. Because it also fixes the problem with the enums mapping to strings.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1103 @@
...
+        match cred_random {
+            Some(cr) => encrypt_hmac_secret_output(&shared_secret, &salt_enc[..], cr),
+            // This is the case if the credential was not created with HMAC-secret.
+            None => Err(Ctap2StatusCode::CTAP2_ERR_UNSUPPORTED_EXTENSION),
+        }
+    }
+
+    #[cfg(feature = "with_ctap2_1")]
+    pub fn has_make_credential_permission(&mut self, rp_id: &str) -> Result<(), Ctap2StatusCode> {
+        self.has_permission(0x01, rp_id)
```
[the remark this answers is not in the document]
kaczmarczyck: enum is added, with `allow(dead_code)` for unused variants.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Resolved throughout the file.

src/ctap/pin_protocol_v1.rs
```diff
@@ -136,12 +136,31 @@ fn check_and_store_new_pin(
...
+    CredentialManagement = 0x04,
+    BioEnrollment = 0x08,
+    PlatformConfiguration = 0x10,
+    AuthenticatorConfiguration = 0x20,
+}
+
+#[cfg(feature = "with_ctap2_1")]
+impl PinPermission {
+    pub fn check(self, stored_bits: u8) -> bool {
+        self as u8 & stored_bits != 0
```
ia0, line 160: Just making sure. This relies on the fact that PinPermission is exactly a bitflag (i.e. each enum value has exactly one bit set). Otherwise we would need `self as u8 & stored_bits == self as u8` or something like that.

src/ctap/pin_protocol_v1.rs
```diff
@@ -136,12 +136,31 @@ fn check_and_store_new_pin(
...
+    GetAssertion = 0x02,
+    CredentialManagement = 0x04,
+    BioEnrollment = 0x08,
+    PlatformConfiguration = 0x10,
+    AuthenticatorConfiguration = 0x20,
+}
+
+#[cfg(feature = "with_ctap2_1")]
+impl PinPermission {
+    pub fn check(self, stored_bits: u8) -> bool {
```
ia0, line 159: I'm not sure why we can't inline this function. It seems to be used only once. And the name of the function where it is used is much clearer than `check`. Otherwise I would add at least some documentation on the expected semantics here.

src/ctap/data_formats.rs
kaczmarczyck: That should work, yes. I added the crate for `enum-iterator` and use it conditioned on `#[cfg(test)]`.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Yes, and I don't see any signs of a AND-style permission in the future that would require your proposed comparison. So should be good.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: We can, if we want that. I thought it belongs to permission more than it belongs to the protocol, but I am happy to change that.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: I'll add documentation to clearly state that intention.

Cargo.toml
```diff
@@ -16,6 +16,7 @@ crypto = { path = "libraries/crypto" }
 byteorder = { version = "1", default-features = false }
 arrayref = "0.3.6"
 subtle = { version = "2.2", default-features = false, features = ["nightly"] }
+enum-iterator = "0.6.0"
```
ia0, line 19: Shouldn't this go in dev-dependencies?

kaczmarczyck: It should, thanks for checking!

src/ctap/storage.rs
gendx: Doesn't the following work?

```rust
foo.map_or(bar, || {
    debug_assert_eq!(a, b);
    c
});
```

README.md
```diff
@@ -103,6 +103,9 @@ a few things you can personalize:
     When changing the default, resident credentials become undiscoverable without
     user verification. This helps privacy, but can make usage less comfortable
     for credentials that need less protection.
+6.  Increase the default minimum length for PINs in `ctap/storage.rs`.
+    The current minimum is 4. Values from 4 to 63 are allowed.
```
gendx, line 107: Maybe you can add a rationale as to why one would want to increase this default minimum? (I guess to avoid users setting too easy PINs?)

src/ctap/command.rs
```diff
@@ -278,12 +278,32 @@ pub struct AuthenticatorClientPinParameters {
...
+    pub permissions: Option<u8>,
+    #[cfg(feature = "with_ctap2_1")]
+    pub permissions_rp_id: Option<String>,
 }
 
 impl TryFrom<cbor::Value> for AuthenticatorClientPinParameters {
     type Error = Ctap2StatusCode;
 
     fn try_from(cbor_value: cbor::Value) -> Result<Self, Ctap2StatusCode> {
+        #[cfg(not(feature = "with_ctap2_1"))]
```
gendx, line 295: So, shouldn't we fix https://github.com/google/OpenSK/issues/128 first?

src/ctap/command.rs
```diff
@@ -434,6 +492,9 @@ mod test {
 
     #[test]
     fn test_from_cbor_client_pin_parameters() {
+        // TODO(kaczmarczyck) inline the #cfg when the ICE is resolved:
+        // https://github.com/rust-lang/rust/issues/73663
```
gendx, line 496: So, we can link to #128 instead, right?

src/ctap/data_formats.rs
```diff
@@ -176,11 +179,12 @@ pub enum AuthenticatorTransport {
...
+        use AuthenticatorTransport::*;
         match transport {
-            AuthenticatorTransport::Usb => "usb",
-            AuthenticatorTransport::Nfc => "nfc",
-            AuthenticatorTransport::Ble => "ble",
-            AuthenticatorTransport::Internal => "internal",
+            Usb => "usb",
+            Nfc => "nfc",
+            Ble => "ble",
+            Internal => "internal",
```
gendx, line 187: I'm not a big fan of this wildcard import. The raw words make it seem like they are identifiers to variables, when in fact they are compile-time constant values. This confused me when implementing the CBOR map destructuring macro.

Also, the type of the enum gets lost.

The import should at the very least be explicit, but in general prefixing enum variants by the enum type is clearer, and more in line with the general Rust style I think.

src/ctap/storage.rs
```diff
@@ -59,11 +62,24 @@ const PIN_RETRIES: usize = 4;
...
+// so we use the maximum.
+const NUM_TAGS: usize = 10;
 
 const MAX_PIN_RETRIES: u8 = 6;
 const ATTESTATION_PRIVATE_KEY_LENGTH: usize = 32;
 const AAGUID_LENGTH: usize = 16;
+#[cfg(feature = "with_ctap2_1")]
+const DEFAULT_MIN_PIN_LENGTH: u8 = 4;
+#[cfg(feature = "with_ctap2_1")]
+const _DEFAULT_MIN_PIN_LENGTH_RP_IDS: Vec<String> = Vec::new();
```
gendx, line 79: Any reason to prefix this with an underscore? If it's unused, shouldn't it be `#[cfg(test)]` instead?

src/ctap/storage.rs
```diff
@@ -380,44 +395,117 @@ impl PersistentStore {
...
+                self.store.insert(entry).unwrap();
+            }
+            Some((index, _)) => {
+                self.store.replace(index, entry).unwrap();
+            }
+        }
+    }
+
+    #[cfg(feature = "with_ctap2_1")]
+    pub fn _min_pin_length_rp_ids(&self) -> Vec<String> {
```
gendx, line 470: Same here, is this unused, or used for tests only?
Same for the next function.

src/ctap/storage.rs
```diff
@@ -892,4 +1000,73 @@ mod test {
...
+        assert_eq!(persistent_store.min_pin_length(), new_min_pin_length);
+    }
+
+    #[cfg(feature = "with_ctap2_1")]
+    #[test]
+    fn test_min_pin_length_rp_ids() {
+        let mut rng = ThreadRng256 {};
+        let mut persistent_store = PersistentStore::new(&mut rng);
+
+        // The minimum PIN lenght is initially at the default.
```
gendx, line 1025: Typo: "lenght"

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+                    if self.consecutive_pin_mismatches >= 3 {
+                        return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_BLOCKED);
+                    }
+                    let mut message = vec![0xFF; 32];
+                    message.extend(&[0x06, 0x08]);
+                    message.extend(&[min_pin_length as u8, 0x00, 0x00, 0x00]);
+                    // TODO(kaczmarczyck) commented code is useful for the extension
+                    // if !cbor::write(cbor_array_vec!(min_pin_length_rp_ids), &mut message) {
+                    //     return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_RESPONSE_CANNOT_WRITE_CBOR);
+                    // }
```
gendx, line 408: When do you expect to resolve these TODO? Linking to a tracking GitHub issue would be better.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+                None => return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID),
+            };
+        }
+        if min_pin_length < persistent_store.min_pin_length() {
+            return Err(Ctap2StatusCode::CTAP2_ERR_PIN_POLICY_VIOLATION);
+        }
+        persistent_store.set_min_pin_length(min_pin_length);
+        // if let Some(min_pin_length_rp_ids) = min_pin_length_rp_ids {
+        //     persistent_store.set_min_pin_length_rp_ids(min_pin_length_rp_ids)?;
+        // }
```
gendx, line 422: Is this also a TODO?

README.md
kaczmarczyck: Provided.

src/ctap/command.rs
kaczmarczyck: If you don't mind, I'd rather fix it later, since I have some dependencies on the new PIN features lined up.

src/ctap/command.rs
kaczmarczyck: Done.

src/ctap/data_formats.rs
kaczmarczyck: Making these changes is consistent to the implementation of `PublicKeyCredentialSource`, line 530.

What do you mean by

> The import should at the very least be explicit

The import is 2 lines above the match, so I guess this is not what you mean? Does explicit mean to list all variants in the use clause? I'd rather just revert the change then. And also change `PublicKeyCredentialSource` to the new coding style.

src/ctap/storage.rs
kaczmarczyck: The reason there is an unused variable is because there is the minPinLength extension that is spread between comments. In this case, the clientPin command contains instructions about storing relying parties. This is not conditioned on the extension actually existing, but storing the data is only useful if you really support the extension. But, those extra flash writes are unncessary, so I disabled them for until the extension is implemented. The new clientPin command is implemented as described, usage is postponed to a later PR that includes the extension.

There is no clear separation between extension and protocol in the specification. This feature is not really only for testing, the test exists for the actual feature that just doesn't make sense without the extension to read the data. I could also move all that unused code to a later PR, and document the gap in the implementation?

src/ctap/storage.rs
kaczmarczyck: As mentioned above, it's a future feature that belongs to this change when you read the specification.

src/ctap/storage.rs
kaczmarczyck: Fixed.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: The above comment was meant to cover that, too. Copied for clarity.

src/ctap/storage.rs
kaczmarczyck: That works, if it's clearer. I changed the below use of map_or to the same idiom.

src/ctap/storage.rs
kaczmarczyck: I have the commits for the extension ready, let me know if you want them to be in a separate PR or not.

README.md
```diff
@@ -103,6 +103,13 @@ a few things you can personalize:
     When changing the default, resident credentials become undiscoverable without
     user verification. This helps privacy, but can make usage less comfortable
     for credentials that need less protection.
+6.  Increase the default minimum length for PINs in `ctap/storage.rs`.
+    The current minimum is 4. Values from 4 to 63 are allowed. Requiring longer
+    PINs can help establish trust between users and relying parties. It makes
+    user verification harder to break, but less convenient.
+    NIST recommends 6 at least digit PINs in section 5.1.9.1:
```
gendx, line 110: "recommends 6 at least digit PINs" -> "recommends at least 6-digit PINs"

src/ctap/command.rs
gendx: As you wish. This would avoid some code duplication to fix that early, but I don't mind either.

src/ctap/storage.rs
gendx: Let's do a separate PR once this one is merged.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+    let aes_dec_key = crypto::aes256::DecryptionKey::new(&aes_enc_key);
+    // The specification specifically asks for a zero IV.
+    let iv = [0u8; 16];
+
+    let mut cred_random_secret = [0u8; 32];
+    cred_random_secret.clone_from_slice(cred_random);
+
+    // Initialization of 4 blocks in any case makes this function more readable.
+    let mut blocks = [[0u8; 16]; 4];
+    let block_len = salt_enc.len() / 16;
```
gendx, line 75: What happens if the length of `salt_enc` is not a multiple of 16? Might be worth documenting the expected length of this slice.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+        match persistent_store.pin_hash() {
+            Some(pin_hash) => {
+                if self.consecutive_pin_mismatches >= 3 {
+                    return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_BLOCKED);
+                }
+                // We need to copy the pin hash, because decrementing the pin retries below may
+                // invalidate the reference (if the page containing the pin hash is compacted).
+                let pin_hash = pin_hash.to_vec();
+                persistent_store.decr_pin_retries();
+                if pin_hash_enc.len() != PIN_AUTH_LENGTH {
```
gendx, line 202: Should we check the length of `pin_hash` as well?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+            return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID);
+        }
+
+        let aes_enc_key = crypto::aes256::EncryptionKey::new(&shared_secret);
+        let aes_dec_key = crypto::aes256::DecryptionKey::new(&aes_enc_key);
+        self.check_pin_hash_enc(rng, persistent_store, &aes_dec_key, pin_hash_enc)?;
+
+        if !check_and_store_new_pin(persistent_store, &aes_dec_key, new_pin_enc) {
+            return Err(Ctap2StatusCode::CTAP2_ERR_PIN_POLICY_VIOLATION);
+        }
```
gendx, line 305: This code has a lot of overlap with the previous function. It would be worth factoring it into a utility function to avoid code duplication.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+    ) -> Result<AuthenticatorClientPinResponse, Ctap2StatusCode> {
+        if persistent_store.pin_retries() == 0 {
+            return Err(Ctap2StatusCode::CTAP2_ERR_PIN_BLOCKED);
+        }
+        let pk: crypto::ecdh::PubKey = CoseKey::try_into(key_agreement)?;
+        let shared_secret = self.key_agreement_key.exchange_x_sha256(&pk);
+
+        let aes_enc_key = crypto::aes256::EncryptionKey::new(&shared_secret);
+        let aes_dec_key = crypto::aes256::DecryptionKey::new(&aes_enc_key);
+        self.check_pin_hash_enc(rng, persistent_store, &aes_dec_key, pin_hash_enc)?;
```
gendx, line 325: This part is similar as well (though not exactly the same). Maybe some utility function to "exchange and check shared secret"?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+            key_agreement: None,
+            pin_token: Some(pin_token),
+            retries: None,
+        })
+    }
+
+    #[cfg(feature = "with_ctap2_1")]
+    fn process_get_pin_uv_auth_token_using_uv_with_permissions(
+        &self,
+        _: CoseKey,
```
gendx, line 355: Will this be used in the future PR?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+        }
+        if persistent_store.pin_hash().is_some() {
+            match pin_auth {
+                Some(pin_auth) => {
+                    if self.consecutive_pin_mismatches >= 3 {
+                        return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_BLOCKED);
+                    }
+                    let mut message = vec![0xFF; 32];
+                    message.extend(&[0x06, 0x08]);
+                    message.extend(&[min_pin_length as u8, 0x00, 0x00, 0x00]);
```
gendx, line 400: Can you document what these magic values mean (maybe with some reference to the specifications), given that they don't seem to use CBOR encoding but some custom format instead?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1137 @@
...
+        let response =
+            self.process_get_pin_token(rng, persistent_store, key_agreement, pin_hash_enc)?;
+
+        self.permissions = permissions;
+        self.permissions_rp_id = permissions_rp_id;
+
+        Ok(response)
+    }
+
+    pub fn process(
```
gendx, line 452: I think `process_command` or `process_subcommand` would be clearer.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: It's documented in line 58:
`if salt_enc.len() != 32 && salt_enc.len() != 64 {`
`    return Err(Ctap2StatusCode::CTAP2_ERR_UNSUPPORTED_EXTENSION);`
`}`

I'l add an additional comment.

README.md
kaczmarczyck: Done.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: It is loaded from persistent storage (where it should only ever get with 16 bytes), and later input into `array_ref` and `ct_eq`. So we would change behaviour from panic to outputting an OpenSK-specific error code. Is this change in behaviour what you were asking for?

src/ctap/pin_protocol_v1.rs
kaczmarczyck: It's quite entangled. The lower of `change_pin` needs the `self.check_pin_hash_enc` between the shared code. The `shared_secret` is used multiple times in the function, too. I could have a function that does

`        let pk: crypto::ecdh::PubKey = CoseKey::try_into(key_agreement)?;`
`        let shared_secret = self.key_agreement_key.exchange_x_sha256(&pk);`
`        if !check_pin_auth(&shared_secret, &message, &pin_auth) {`
`            return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID);`
`        }`
`        return shared_secret;`

But I don't like the side effect of generating shared_secret and passing it around. What was your idea?

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Part of the problem is that I am often not allowed to reorder instructions - for code logic or protocol reasons. I could make this a function

`        let aes_enc_key = crypto::aes256::EncryptionKey::new(&shared_secret);`
`        let aes_dec_key = crypto::aes256::DecryptionKey::new(&aes_enc_key);`
`        self.check_pin_hash_enc(rng, persistent_store, &aes_dec_key, pin_hash_enc)?;`

That outputs the AES keys is a result, but doesn't seem worth the extra level of indirection.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Not for the hardware we support right now. We'd need internal user verification like

- fingerprint reader
- PIN pad
- iris scanner ;)

And wanted to include this function with the correct parameters as an entry point in case someone wants to add this feature. GetInfo should already report we don't support internal UV, so we are free to just respond with `CTAP2_ERR_INVALID_*COMMAND` for now.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: This is tricky, since the currently public CTAP 2.1 revision
https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html
is from December 2019, and `SetMinPinLength` is in the internal-only version.

I added a TODO to improve the documentation when possible.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: I went with `process_subcommand` to distinguish between this function and the one in `mod.rs`.

kaczmarczyck:
Thanks @gendx for the review and the Rust comments.

Concerning reviewing for details of the protocol, the old comments are thoroughly tested and mostly taken from `mod.rs`, so those should be good. The new commands, on the other hand, or not even in the publicly visible specification, so we don't really have any real life applications for testing. They are hidden behind the `ctap2_1` feature flag for now for that reason.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1140 @@
...
+    let mut pin = vec![];
+    'pin_block_loop: for block in blocks.iter().take(PIN_PADDED_LENGTH / 16) {
+        for cur_char in block.iter() {
+            if *cur_char != 0 {
+                pin.push(*cur_char);
+            } else {
+                break 'pin_block_loop;
+            }
+        }
+    }
```
gendx, line 132: This whole thing could be its own `cbc_decrypt_pin` function. Very similar to `cbc_decrypt` except that it takes an `encrypted_pin: &[u8; PIN_PADDED_LENGTH]` as input and returns a `Vec<u8>` as output.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1140 @@
...
+    let iv = [0u8; 16];
+    // Assuming PIN_PADDED_LENGTH % block_size == 0 here.
+    let mut blocks = [[0u8; 16]; PIN_PADDED_LENGTH / 16];
+    for i in 0..PIN_PADDED_LENGTH / 16 {
+        blocks[i].copy_from_slice(&new_pin_enc[i * 16..(i + 1) * 16]);
+    }
+    cbc_decrypt(aes_dec_key, iv, &mut blocks);
+    let mut pin = vec![];
+    'pin_block_loop: for block in blocks.iter().take(PIN_PADDED_LENGTH / 16) {
```
gendx, line 124: This loop can be replaced by `let pin: Vec<u8> = blocks.iter().take(n).flatten().collect();`

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Ah, now I think I see the problem. The `to_vec()` should be a `clone()`. The `persistent_store` already returns an `Option<&[u8; PIN_AUTH_LENGTH]>`. I also removed the `array_ref` here. I think returning results in the storage  is planned and I'll not touch that here.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: The [specification](https://fidoalliance.org/specs/fido2/fido-client-to-authenticator-protocol-v2.1-rd-20191217.html#gettingPinUvAuthToken) under `Getting pinUvAuthToken using built-in user verification methods` describes this command. It's parameters include `key_agreement`. Anyone who would want to implement internal user verification needs to also implement this function. The new versions of the specification is a bit clearer in describing this subcommand.

In general, the structure of this files follows the naming scheme `process_*`, like `process_get_pin_uv_auth_token_using_uv_with_permissions`. Here, `get_pin_uv_auth_token_using_uv_with_permissions` is the name of a subcommand (however trivial). All of them are mapped with correct parameter set from the specification. I'd find it confusing to have a command that is lacking a parameter only because it is unused so far.

Does that make sense?

src/ctap/pin_protocol_v1.rs
kaczmarczyck: No, this is a cryptographic trick to make `pin_auth` codes not reusable between commands.

The longer explanation: CTAP 2.1 introduces the concept of permissions for tokens. So when you enter your PIN, you still get a token, but it only valid for a specific use case. To enforce that, reply attacks are supposed to be prevented by making the hashed message start with a unique prefix per command.

Since this cryptographic feature did not exist before, they start with `[0xFF; 32]` to distinguish form old commands. Then, they add one byte for the command (0x06 == clientPin) and one byte for the subcommand (0x08 == minPinLength).

In summary I think this is a backwards compatible attempt to defend against an attack that wasn't considered before. This level of explanation is not even given in the command description, so I assumed this source code isn't the place to discuss cryptographic protocol details as well?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1140 @@
...
+    for b in &blocks[..block_len] {
+        encrypted_output.extend(b);
+    }
+    Ok(encrypted_output)
+}
+
+fn check_and_store_new_pin(
+    persistent_store: &mut PersistentStore,
+    aes_dec_key: &crypto::aes256::DecryptionKey,
+    new_pin_enc: Vec<u8>,
```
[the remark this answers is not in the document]
kaczmarczyck: This variable names allows you to search the specification for its counterpart (`newPinEnc`) easier. I thought not introducing duplicate names is more important than the style guide. So before changing this name, please confirm you think this should be fixed.

src/ctap/data_formats.rs
gendx: I meant listing all variants rather than using the wildcard `*` (because with the wildcard it's unclear where the values are coming from).

src/ctap/pin_protocol_v1.rs
gendx: Then I think the name should at least be `_foo` with a comment explaining why it's currently not used.

src/ctap/pin_protocol_v1.rs
gendx: Hum "cryptographic trick" triggers my reviewer scrutiny ^^ Though after all I don't see much cryptography in your explanation.

So the format is `[0xFF; 32]` + a regular encoded command/subcommand, right? Can we somehow use the command/subcommand serializer to create these bytes then (and using constants/enums that explicitly mention "clientPin" and "minPinLength" rather than the unexaplained `0x06` and `0x08`)?

src/ctap/pin_protocol_v1.rs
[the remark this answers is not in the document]
gendx: Ok, if it's in the specs then fine. Though a reference to the specs would be preferable to track that.

src/ctap/pin_protocol_v1.rs
gendx: Why would you need to clone it? Isn't a reference enough? I only see it used in `array_ref` again.

src/ctap/pin_protocol_v1.rs
gendx: I missed that part. You can chain with `.take_while(|&c| c != 0)` before the `.collect()` and this should be good to go.

src/ctap/data_formats.rs
kaczmarczyck: If we list all variants, then there is no benefit to just keeping the old style everywhere. (And also changing all other wildcard uses throughout the file to being explicit.) `PublicKeyCredentialSourceField` uses both styles at the moment.

So if wildcard import is not okay, I vote for removing all `use` clauses entirely.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: The clone is necessary as explained in the comment above the line. We need the following order of events:

1. get `pin_hash` from persistent store
2. decrease PIN retries in persistent store
3. use `pin_hash`

The 2. step could invalidate the reference returned in step 1. I could change the match on `pin_hash` into an early returning assignment, to make the flow clearer?

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Updated, renamed and commented.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Is there any code reuse I am missing? Or is this only about splitting the function into 2 parts?

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Okay, I added a comment that we use the implementation that is is compatible with 2.0. The behavior is different for padded PINs like

> ['1', '2', '3', '4', '\0', '6', '\0', ...]

In 2.0, the PIN is read as "1234". In 2.1 instead, two more bytes are included. IMO we should never receive padded PINs like that anyway.

src/ctap/pin_protocol_v1.rs
gendx: Here it was only about splitting out the logic into its own function (given the non-trivial logic).

src/ctap/pin_protocol_v1.rs
gendx: That's IMO a good reason to separate the logic into a `cbc_decrypt_pin` function. Then it's easy to make unit tests like `assert_eq!(cbc_decrypt_pin(encrypt(['1', '2', '3', '4', 0, '6', 0, ...])), ['1', '2', '3', '4'])`

src/ctap/pin_protocol_v1.rs
gendx: See my other comment, a separate function also makes it easier to write unit tests like `assert_eq!(cbc_decrypt_pin(encrypt(['1', '2', '3', '4', 0, '6', 0, ...])), ['1', '2', '3', '4'])`.

src/ctap/pin_protocol_v1.rs
gendx: Hum, it wasn't clear to me that it was needed due to a borrow checker violation. The comment could make it clearer, e.g. `because decrementing the pin retries below mutably borrows from the same persistent_store reference, and therefore invalidates the pin_hash reference`.

It's not exactly that it "*may* invalidate the reference", but rather that the reference is automatically and *always* invalidated by Rust's borrow checker.

In any case, we shouldn't call `.to_vec()` to clone a `[u8; PIN_AUTH_LENGTH]`. With a `.clone()` it should be fine.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1140 @@
...
+use enum_iterator::IntoEnumIterator;
+use subtle::ConstantTimeEq;
+
+// Those constants have to be multiples of 16, the AES block size.
+pub const PIN_AUTH_LENGTH: usize = 16;
+const PIN_PADDED_LENGTH: usize = 64;
+const PIN_TOKEN_LENGTH: usize = 32;
+
+/// Checks the given pin_auth against the truncated output of HMAC-SHA256.
+fn check_pin_auth(hmac_key: &[u8], hmac_contents: &[u8], pin_auth: &[u8]) -> bool {
```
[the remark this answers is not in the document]
kaczmarczyck: Done.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1140 @@
...
+        // We are strict here. We need at least 32 byte, but expect exactly 32.
+        return Err(Ctap2StatusCode::CTAP2_ERR_UNSUPPORTED_EXTENSION);
+    }
+    let aes_enc_key = crypto::aes256::EncryptionKey::new(shared_secret);
+    let aes_dec_key = crypto::aes256::DecryptionKey::new(&aes_enc_key);
+    // The specification specifically asks for a zero IV.
+    let iv = [0u8; 16];
+
+    let mut cred_random_secret = [0u8; 32];
+    cred_random_secret.clone_from_slice(cred_random);
```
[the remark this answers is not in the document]
kaczmarczyck: Changed to `copy_from_slice` everywhere.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: I added the recommended unit test. The function was pretty heavily modified and I am not sure if the logic split still applies.

src/ctap/pin_protocol_v1.rs
gendx: I still don't see why we need to check the length. Don't we get back a statically sized slice/array, as written in https://github.com/google/OpenSK/pull/127/files#r463190693?

gendx: Never mind, I misread it as the other "pin hash".

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1173 @@
...
+                // pin_hash reference.
+                let pin_hash = pin_hash.clone();
+                persistent_store.decr_pin_retries();
+                if pin_hash_enc.len() != PIN_AUTH_LENGTH {
+                    return Err(Ctap2StatusCode::CTAP2_ERR_PIN_INVALID);
+                }
+
+                let iv = [0u8; 16];
+                let mut blocks = [[0u8; 16]; 1];
+                blocks[0].copy_from_slice(&pin_hash_enc[0..PIN_AUTH_LENGTH]);
```
gendx, line 217: If the length has already been checked, this indexing `0..PIN_AUTH_LENGTH` is unnecessary. Just make sure that there is a test where `pin_hash_enc` has too many (or too few) bytes.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1173 @@
...
+            None => return Err(Ctap2StatusCode::CTAP2_ERR_PIN_REQUIRED),
+        }
+        persistent_store.reset_pin_retries();
+        self.consecutive_pin_mismatches = 0;
+        Ok(())
+    }
+
+    /// Uses the self-owned and passed halves of the key agreement to generate the
+    /// shared secret for checking pin_auth and generating a decryption key.
+    fn create_decryption_key_verified(
```
gendx, line 242: Nit:
- `exchange_decryption_key` would be clearer and sufficient IMO.
- It's not clear what `message` contains in this context. Could be worth documenting or using a more meaningful name.

Otherwise looks good.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1173 @@
...
+        let iv = [0u8; 16];
+        let mut blocks = [[0u8; 16]; PIN_TOKEN_LENGTH / 16];
+        for (i, item) in blocks.iter_mut().take(PIN_TOKEN_LENGTH / 16).enumerate() {
+            item.copy_from_slice(&self.pin_uv_auth_token[i * 16..(i + 1) * 16]);
+        }
+        cbc_encrypt(&token_encryption_key, iv, &mut blocks);
+        let mut pin_token = vec![];
+        for item in blocks.iter().take(PIN_TOKEN_LENGTH / 16) {
+            pin_token.extend(item);
+        }
```
gendx, line 351: I think the same `iter.flatten().collect()` pattern applies here.

src/ctap/pin_protocol_v1.rs
gendx: I think quoting this sentence from the specs will be the clearest option then! I didn't realize it was that bad in the specs...

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1173 @@
...
+
+    #[test]
+    fn test_check_and_store_new_pin() {
+        let mut rng = ThreadRng256 {};
+        let mut persistent_store = PersistentStore::new(&mut rng);
+        let shared_secret = [0x88; 32];
+        let aes_enc_key = crypto::aes256::EncryptionKey::new(&shared_secret);
+        let aes_dec_key = crypto::aes256::DecryptionKey::new(&aes_enc_key);
+
+        // The PIN "1234" should be accepted.
```
gendx, line 1037: For these tests, it's not obvious that the pin is "1234" or "123" from the encrypted golden value. So I'd suggest presenting the tests more in terms of the decrypted blob (e.g. `[1, 2, 3, 4, 0, 0, 0, ...]`), which is then encrypted with `aes_enc_key` available in this function, and then passed to `check_and_store_new_pin`.

Also, `check_and_store_new_pin` returning only a boolean doesn't help to check that the rejection reason is indeed what is mentioned in each comment. It would be nice to check an error code and/or check what the storage contains afterwards. In particular have a more complete scenario (maybe a separate unit test) where we try to use the new pin, and confirm that the old pin doesn't work (or the other way around if the new pin was rejected as invalid).

I think it's quite important to have clear tests for this critical scenario (authentication).

That's also why I think having separate functions for the "encrypt pin" and for the "decrypt pin" logic would make things much clearer for testing.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done

src/ctap/pin_protocol_v1.rs
kaczmarczyck: I introduced `decrypt_pin` and a separate test. The tests for `test_check_and_store_new_pin` now also look into the storage. `encrypt_pin` is only extended for creating more flexible `new_pin_enc`.

src/ctap/pin_protocol_v1.rs
```diff
@@ -345,10 +352,7 @@ impl PinProtocolV1 {
             item.copy_from_slice(&self.pin_uv_auth_token[i * 16..(i + 1) * 16]);
         }
         cbc_encrypt(&token_encryption_key, iv, &mut blocks);
-        let mut pin_token = vec![];
-        for item in blocks.iter().take(PIN_TOKEN_LENGTH / 16) {
-            pin_token.extend(item);
-        }
+        let pin_token: Vec<u8> = blocks.iter().flatten().cloned().collect();
```
gendx, line 355: Nit: does `blocks.into_iter()` allow to remove the `.cloned()` later in the chain? Or is there no `into_iter` for arrays?

src/ctap/pin_protocol_v1.rs
```diff
@@ -644,21 +651,25 @@ mod test {
         persistent_store.set_pin_hash(&pin_hash);
     }
 
-    fn encrypt_standard_pin(shared_secret: &[u8; 32]) -> Vec<u8> {
+    // Fails on PINs bigger than 64 byte..
```
gendx, line 654: `byte..` => `bytes.`

src/ctap/pin_protocol_v1.rs
```diff
@@ -644,21 +651,25 @@ mod test {
...
-        blocks[0][2] = 0x33;
-        blocks[0][3] = 0x34;
+        let (b0, b1, b2, b3) = array_refs!(&padded_pin, 16, 16, 16, 16);
+        blocks[0][..].copy_from_slice(b0);
+        blocks[1][..].copy_from_slice(b1);
+        blocks[2][..].copy_from_slice(b2);
+        blocks[3][..].copy_from_slice(b3);
         let iv = [0u8; 16];
         cbc_encrypt(&aes_enc_key, iv, &mut blocks);
+        blocks.iter().flatten().cloned().collect::<Vec<u8>>()
```
gendx, line 668: Same question here.

src/ctap/pin_protocol_v1.rs
```diff
@@ -644,21 +651,25 @@ mod test {
...
+    }
 
-        let mut encrypted_pin = Vec::with_capacity(64);
-        for b in &blocks {
-            encrypted_pin.extend(b);
-        }
-        encrypted_pin
+    fn encrypt_standard_pin(shared_secret: &[u8; 32]) -> Vec<u8> {
```
gendx, line 671: Question: what is a "standard" PIN? The string "1234"? This function would be worth a 1-line documentation.

src/ctap/data_formats.rs
gendx: I think removing the `use` clauses entirely would be the most consistent approach (although the most verbose). We have explicit enums for CTAP2 error codes for example.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Again, `into_iter` is `iter`.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Done.

src/ctap/data_formats.rs
kaczmarczyck: I removed all inner use instructions.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1228 @@
...
+            #[cfg(feature = "with_ctap2_1")]
+            permissions_rp_id: None,
+        }
+    }
+
+    /// Decrypts the encrypted pin_hash and compares it to the stored pin_hash.
+    /// Resets or decreases the PIN retries, depending on success or failure.
+    fn verify_pin_hash_enc(
+        &mut self,
+        rng: &mut impl Rng256,
```
gendx, line 207: I just noticed it now, but I find it strange that a verification function takes a rng as input. Could you document that the key agreement key is updated to a random key, why, and when?

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1228 @@
...
+        let new_pin_enc = vec![
+            0xC0, 0xCF, 0xAE, 0x4C, 0x79, 0x56, 0x87, 0x99, 0xE5, 0x83, 0x4F, 0xE6, 0x4D, 0xFE,
+            0x53, 0x32, 0x36, 0x0D, 0xF9, 0x1E, 0x47, 0x66, 0x10, 0x5C, 0x63, 0x30, 0x1D, 0xCC,
+            0x00, 0x09, 0x91, 0xA4, 0x20, 0x6B, 0x78, 0x10, 0xFE, 0xC6, 0x2E, 0x7E, 0x75, 0x14,
+            0xEE, 0x01, 0x99, 0x6C, 0xD7, 0xE5, 0x2B, 0xA5, 0x7A, 0x5A, 0xE1, 0xEC, 0x69, 0x31,
+            0x18, 0x35, 0x06, 0x66, 0x97, 0x84, 0x68, 0xC2,
+        ];
+        assert_eq!(
+            decrypt_pin(&aes_dec_key, new_pin_enc),
+            Some(vec![0x31, 0x32, 0x33, 0x34]),
```
gendx, line 1082: Nit: `vec![b'1', b'2', b'3', b'4']` would be more readable. Or even `b"1234".into_vec()` (if that doesn't include any trailing zero). Same for the other tests.

gendx (commented):
Apart from a few nits that had unfortunately slipped through, and rebasing/merging with master, this looks good to me.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1228 @@
...
+        return false;
+    }
+    verify_hmac_256_first_128bits::<Sha256>(
+        hmac_key,
+        hmac_contents,
+        array_ref![pin_auth, 0, PIN_AUTH_LENGTH],
+    )
+}
+
+/// Decrypts the HMAC secret salt(s) that were encrypted with the shared secret.
```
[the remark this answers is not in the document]
kaczmarczyck: This function does encrypt the HMAC-secret output. To compute those, it has to decrypt the salt(s) first. I restructured the comment to mention the most important parts of it first, instead of a chronological description.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: There is no reason in the specification for why the key has to be reset, so I don't want to speculate. It's most likely to prevent attacks on the private platform key-agreement key by cycling it. Everything else is added as a comment.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Changed both tests in this function.

gendx: I still see the "manual vec! form" in the following test function.

kaczmarczyck: That is because I need a `0u8` there, and that was the most consistent and visible solution to me.

src/ctap/pin_protocol_v1.rs
```diff
@@ -0,0 +1,1229 @@
...
+            (
+                vec![0x31, 0x32, 0x33],
+                Err(Ctap2StatusCode::CTAP2_ERR_PIN_POLICY_VIOLATION),
+            ),
+            // Reject PIN "12'\0'4" (a zero byte at index 2).
+            (
+                vec![0x31, 0x32, 0x00, 0x34],
+                Err(Ctap2StatusCode::CTAP2_ERR_PIN_POLICY_VIOLATION),
+            ),
+            // Reject PINs not ending in 0u8 padding.
```
gendx, line 1129: This comment is not 100% clear. The accepted PIN "1234" also doesn't explicitly include 0u8 padding.

src/ctap/pin_protocol_v1.rs
gendx: On the other hand, the comment above the test mentions "12'\0'4", so why not using `b"12\04"` there?

kaczmarczyck: I wasn't sure if it's too easy to miss, but done.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Changed to a better explanation.

gendx: I don't see any change uploaded (yet). Please mark as resolved when the code is up-to-date.

gendx (approved):

gendx (approved):

Issue #130, opened 2020-07-23
The deploy.py script isn't compatible with Tockloader 1.5


jasonloveday:
Hi there,

I am fairly new to OpenSK, so please be gentle.

## Expected Behavior
Running ./deploy.py with --programmer=none should create hex files for manual programming

## Actual Behavior
Compiling appears to work, but intelHex stage produces error message as follows:

info: Generating Tock TAB file for application/example ctap2
Traceback (most recent call last):
  File "./deploy.py", line 901, in <module>
    main(main_parser.parse_args())
  File "./deploy.py", line 734, in main
    OpenSKInstaller(args).run()
  File "./deploy.py", line 684, in run
    self.create_hex_file(dest_file)
  File "./deploy.py", line 584, in create_hex_file
    app_tab.extract_app(board_props.arch).get_binary(),
TypeError: get_binary() missing 1 required positional argument: 'address'

## Steps to Reproduce the Problem

1. ./deploy.py --board=nrf52840_dongle --opensk --programmer=none

## Specifications

- Version: 
- Platform: Ubuntu 20.04 Desktop

gendx:
We have something very similar in CI (https://github.com/google/OpenSK/blob/master/reproduce_board.sh#L27). This looks like using an non-supported version of the [Tockloader library](https://pypi.org/project/tockloader/#history).

Can you run `./reset.sh && ./setup.sh` before re-running this command?

gendx:
I checked again, and it looks like the latest version of Tockloader (1.5.0) that was just released this week broke the `deploy.py` script on our side.

jasonloveday:
Excellent, thanks!

PR #131, opened 2020-07-23
Pin Tockloader to version 1.4, as a temporary workaround until we migrate to 1.5

gendx:
This is a temporary workaround for #130. The new Tockloader 1.5 release changed many APIs so this is a simple workaround until we investigate how to migrate.

- [x] Tests pass


gendx:
Updated the reference binaries.

PR #132, opened 2020-07-24
Add reference elf2tab and binary hashes to the uploaded reproduced.tar file

gendx:
This adds the `reproducible/elf2tab.txt` and `reproducible/binaries.sha256sum` files to the `reproduced.tar` archive that is uploaded in the "check reproducible" GitHub action, to make it easier to update the reference binaries in pull requests.

- [x] Tests pass


gendx:
Blocked on #131.

PR #137, opened 2020-08-11
Add fuzz target for CBOR

mingxguo27:
Fuzzing for the CBOR library. 

This PR contains a small fuzz target to be used with cargo-fuzz (subcommand for using LibFuzzer in Rust).


libraries/cbor/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,20 @@
...
+extern crate alloc;
+
+use alloc::vec::Vec;
+
+fuzz_target!(|data: &[u8]| {
+    let encoded = cbor::read(data);
+    match encoded{
+    	Ok(value) => {
+    		let mut decoded = Vec::new();
+    		let _ = cbor::write(value, &mut decoded);
```
kaczmarczyck, line 14: The `bool` output of `write` indicates success. It might fail if the encoded data is too deeply nested. I don't think this can happen if the structure was generated by `read` without errors. So I think you can also assert that `write` returns `true`.

libraries/cbor/fuzz/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies.cbor]
+path = ".."
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "fuzz_target_1"
+path = "fuzz_targets/fuzz_target_1.rs"
```
kaczmarczyck, line 24: Can you rename this file to something more meaningful?

libraries/cbor/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,20 @@
...
+#[macro_use] 
+extern crate libfuzzer_sys;
+extern crate cbor;
+extern crate alloc;
+
+use alloc::vec::Vec;
+
+fuzz_target!(|data: &[u8]| {
+    let encoded = cbor::read(data);
+    match encoded{
```
kaczmarczyck, line 11: There is a space missing here compared to our usual style. Usually an automated check would run `cargo fmt` to ensure correct code style. It seems that our github workflow does not cover your fuzzing code. Can you please add this path and fix the style?

libraries/cbor/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,20 @@
...
+    	Ok(value) => {
+    		let mut decoded = Vec::new();
+    		let _ = cbor::write(value, &mut decoded);
+    		assert_eq!(decoded, data);
+    	}
+    	Err(_) => {}
```
kaczmarczyck, line 17: Since you ignore the `Err` pattern, you don't need a match, but this should work too:

`if let Ok(value) = encoded { [..]`

libraries/cbor/fuzz/fuzz_targets/fuzz_target_1.rs
mingxguo27: Sorry for triggering the error, just wanted to make sure if I added the path correctly and it can be detected.

libraries/cbor/fuzz/fuzz_targets/fuzz_target_cbor.rs
```diff
@@ -0,0 +1,18 @@
...
+use alloc::vec::Vec;
+
+fuzz_target!(|data: &[u8]| {
+    let encoded = cbor::read(data);
+    if let Ok(value) = encoded {
+	let mut decoded = Vec::new();
+	let result = cbor::write(value, &mut decoded);
+	assert!(result);
+	assert_eq!(decoded, data);
+    };
```
ia0, line 16: If you want, this whole function could be written:

```rust
if let Ok(value) = cbor::read(data) {
    let mut result = Vec::new();
    assert!(cbor::write(value, &mut result));
    assert_eq!(result, data);
}
```

mingxguo27: Thank you for the suggestion!! ^^

ia0: No problem :-)

Also, in Rust there is no need for semi-colon `if` statements (it's only for `let` statements I think). But apparently, there is no lint about it.

reproducible/reference_binaries_ubuntu-18.04.sha256sum
```diff
@@ -1,9 +1,9 @@
 3feb5d29a3d669107b460a00391440be4ebc5e50461f9ef3248714f4f99c070e	third_party/tock/target/thumbv7em-none-eabi/release/nrf52840dk.bin
-3c4dde09e8e6082a15bbf9d9eedd12de63f70d85c240292aea246f2d83c54f07	target/nrf52840dk_merged.hex
```
jmichelp, line 2: I'm not sure I understand why adding some fuzzing testing would have some impact on the hash of the binary of the firmware.
Am I missing something?

kaczmarczyck: Could it be that we merged with somethign that changed the hashes?

PR #138, opened 2020-08-12
Bump required version for nrfutil to 6.1+

### LINKED ISSUE #135: nrfutil package version 6.1.x is not supported by OpenSK
## Expected Behavior

DFU flashing should work with `nrfutil` version 6.x

## Actual Behavior

DFU flashing works with `nrfutil` version 6.0.1 which was the latest when we released it.

Since then, Nordic released `nrfutil` package version 6.1.0 which changed the expected command line parameter from `usb-serial` to `usb_serial`

N/A

## Specifications

- Version: N/A
- Platform: all

jmichelp:
Fixes #135

Now requiring nrfutil package 6.1+ so that we don't have to handle the difference between v6.0 which expects `usb-serial` parameter and v6.1 which expects `usb_serial` instead.

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -607,8 +607,9 @@ class OpenSKInstaller:
       assert_python_library("intelhex")
       assert_python_library("nordicsemi.lister")
       nrfutil_version = __import__("nordicsemi.version").version.NRFUTIL_VERSION
-      if not nrfutil_version.startswith("6."):
-        fatal(("You need to install nrfutil python3 package v6.0 or above. "
+      nrfutil_major, nrfutil_minor, _ = nrfutil_version.split(".", 3)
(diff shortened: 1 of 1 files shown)


ia0:
It actually looks like the nrfutil version was a red-herring. The problem is https://github.com/pallets/click/issues/1123 and the correct code is the current one. If people have issues, they should reinstall click not nrfutil. I'll test this to confirm when I get access to my personal computer.

jmichelp:
Indeed, forcing an upgrade on click (`pip3 install --upgrade click`) installed Click 7.1.2 (previously 7.1.1) and now `usb-serial` is accepted instead of `usb_serial`

ia0:
Cool! Thanks for checking! So I'll simply do that on my other computer.

PR #146, opened 2020-09-14
fuzzing for OpenSK

mingxguo27:
I commited the fuzz targets that cover most code I could come up with:
1. fuzz_target_process_message: treats inputs as raw messages and tests message assembling, disassembling and processing at HID level (which then processes at ctap1 commands or ctap2 commands level etc., depending on the message type). 
2. fuzz_target_process_ctap{1,2}: same as above, but treats inputs as ctap2 command parameters encoded in cbor or ctap1 raw messages. This increases in a meaningful scale the achieved coverage (400 -> 1400) in the case of ctap2.
3. fuzz_target_split_assemble: this only tests message splitting and assembling messages, but includes correctness check.


fuzz/make_corpus.py
```diff
@@ -0,0 +1,23 @@
+import json
+import sys
+import os.path
+
+"""
```
kaczmarczyck, line 5: For reference, Google has a style guide for some popular languages, including [Python](https://google.github.io/styleguide/pyguide.html#382-modules). The link shows an example documentation for files.

fuzz/make_corpus.py
```diff
@@ -0,0 +1,23 @@
...
+Creates a directory containing seed inputs from a json file.
+Usage:
+- pass the resulting corpus directory path as the first argument
+- pass the json file path to make the corpus from as the second argument
+"""
+corpus_dir = sys.argv[1]
```
kaczmarczyck, line 11: This files looks like you outsiders might want to execute it? Then it needs more robust parameter handling, like using `argparse`.

fuzz/make_corpus.py
```diff
@@ -0,0 +1,23 @@
...
+corpus_file_json = sys.argv[2]
```
kaczmarczyck, line 12: Also, all these things should go into functions or guarded by the typical `if __name__ == '__main__':`.

fuzz/make_corpus.py
```diff
@@ -0,0 +1,23 @@
...
+if not os.path.exists(corpus_dir):
+    os.makedirs(corpus_dir)
```
kaczmarczyck, line 14: Can this line fail?

src/ctap/hid/mod.rs
```diff
@@ -50,7 +51,7 @@ pub enum ProcessedPacket<'a> {
 }
 
 // An assembled CTAPHID command.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Arbitrary)]
```
kaczmarczyck, line 54: This include should be conditional on being used for fuzzing, or it might unnecessarily increase binary size.

src/ctap/hid/mod.rs
```diff
@@ -163,7 +164,6 @@ impl CtapHid {
             Ok(Some(message)) => {
                 #[cfg(feature = "debug_ctap")]
                 writeln!(&mut Console::new(), "Received message: {:02x?}", message).unwrap();
-
```
kaczmarczyck, line 166: Seems unintended?

src/ctap/mod.rs
```diff
@@ -14,7 +14,7 @@
 
 pub mod command;
 #[cfg(feature = "with_ctap1")]
-mod ctap1;
+pub mod ctap1;
```
kaczmarczyck, line 17: If you have the `derive` conditional on a feature, this `pub` can also be conditional.

kaczmarczyck (changes requested):

mingxguo27:
Closing for splitting in 2 smaller pull requests.

Issue #118, opened 2020-06-09
Migrate syntax for extern crates and macros to Rust 2018 edition


gendx:
The 2018 edition of Rust simplified the syntax around usage of extern crates and exporting of macros. We should migrate to that.

More specifically:
- `extern crate` statements are [not needed anymore](https://doc.rust-lang.org/stable/edition-guide/rust-2018/module-system/path-clarity.html#no-more-extern-crate).
- `#[macro_use]` attributes are [replaced by importing each macro individually](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html). In this regard, macros can be imported like other symbols, such as functions.

I didn't check for other specifics of the 2018 edition, but we could track here any other syntax changes related to it.

PR #149, opened 2020-09-16
Fuzzing MVP

mingxguo27:
MVP of OpenSK HID fuzzing with dependencies and linking errors fixed.

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -17,7 +17,7 @@ cbor = { path = "../libraries/cbor" }
 
 [dependencies.ctap2]
 path = ".."
-features = ['std', 'ram_storage', 'fuzzing']
+features = ['std', 'ram_storage']
 
 # Prevent this from interfering with workspaces
 [workspace]
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,4 +1,3 @@
-
 [package]
 name = "ctap2-fuzz"
 version = "0.0.0"
@@ -14,7 +13,7 @@ libfuzzer-sys = { version = "0.3"}
 arrayref = "0.3.6"
 libtock_drivers = { path = "../third_party/libtock-drivers" }
 crypto = { path = "../libraries/crypto", features = ['std'] }
-cbor = { path = "../libraries/cbor"}
+cbor = { path = "../libraries/cbor" }
 
 [dependencies.ctap2]
 path = ".."
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -1,0 +1,31 @@
+
+[package]
+name = "ctap2-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = { version = "0.3"}
+arrayref = "0.3.6"
+libtock_drivers = { path = "../third_party/libtock-drivers" }
+crypto = { path = "../libraries/crypto", features = ['std'] }
+cbor = { path = "../libraries/cbor"}
+
+[dependencies.ctap2]
+path = ".."
+features = ['std', 'ram_storage', 'fuzzing']
+
+# Prevent this from interfering with workspaces
+[workspace]
+members = ["."]
+
+[[bin]]
+name = "fuzz_target_split_assemble"
+path = "fuzz_targets/fuzz_target_split_assemble.rs"
+test = false
+doc = false
--- a/fuzz/fuzz_targets/fuzz_target_split_assemble.rs
+++ b/fuzz/fuzz_targets/fuzz_target_split_assemble.rs
@@ -46,7 +46,7 @@ fuzz_target!(|data: &[u8]| {
             message.cmd &= !PACKET_TYPE_MASK;
             assert_eq!(
                 assembler.parse_packet(last_packet, DUMMY_TIMESTAMP),
-                Ok(Some(message.clone()))
+                Ok(Some(message))
             );
         }
     }
--- a/fuzz/fuzz_targets/fuzz_target_split_assemble.rs
+++ b/fuzz/fuzz_targets/fuzz_target_split_assemble.rs
@@ -1,0 +1,66 @@
+#![no_main]
+
+extern crate ctap2;
+extern crate libtock_drivers;
+#[macro_use]
+extern crate arrayref;
+
+use libfuzzer_sys::fuzz_target;
+use ctap2::ctap::hid::receive::MessageAssembler;
+use ctap2::ctap::hid::send::HidPacketIterator;
+use ctap2::ctap::hid::{Message, HidPacket};
+use libtock_drivers::timer::Timestamp;
+
+const DUMMY_TIMESTAMP: Timestamp<isize> = Timestamp::from_ms(0);
+const PACKET_TYPE_MASK: u8 = 0x80;
+
+// Converts a byte slice into Message
+fn raw_to_message(data: &[u8], len: usize) -> Message{
+    if len <= 4 {
+        let mut cid = [0;4];
+        cid[..len].copy_from_slice(data);
+        Message{
+            cid,
+            cmd: 0,
+            payload: vec![],
+        }
+    }
+    else if len == 5{
+        Message{
+            cid: array_ref!(data,0,4).clone(),
+            cmd: data[4],
+            payload: vec![],
+        }
+    }
+    else{
+        Message {
+            cid: array_ref!(data,0,4).clone(),
+            cmd: data[4],
+            payload: data[5..].to_vec(),
+        }
+    }
+}
+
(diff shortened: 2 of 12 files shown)


fuzz/fuzz_targets/fuzz_target_split_assemble.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/* Fuzzing HID packets splitting and assembling functions*/
+fuzz_target!(|data: &[u8]| {
+    let Message{cid, mut cmd, payload} = raw_to_message(data, data.len());
+    if let Some(hid_packet_iterator) = HidPacketIterator::new(Message{cid,cmd,payload:payload.clone()}){
+        let packets: Vec<HidPacket> = hid_packet_iterator.collect();
+        let mut assembler = MessageAssembler::new();
+        for (i, packet) in packets.iter().enumerate(){
+            if i != packets.len() - 1 {
```
kaczmarczyck, line 51: Copied from the last PR: Consider `split_last` here?

src/lib.rs
```diff
@@ -16,8 +16,15 @@
 
 #[macro_use]
 extern crate alloc;
+#[cfg(not(feature = "fuzzing"))]
 extern crate lang_items;
 extern crate libtock_core;
 extern crate libtock_drivers;
+#[macro_use]
+extern crate cbor;
+#[macro_use]
```
kaczmarczyck, line 25: When you run 

> ./deploy.py --opensk --board=nrf52840dk

you get

> error: unused `#[macro_use]` import

(You also see those with `./run_desktop_tests.sh`.)

kaczmarczyck (approved):
Let's wait for one more reviewer to do a pass.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+
```
ia0, line 1: Was the newline automatically generated? I think it's safe to remove.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+libfuzzer-sys = { version = "0.3"}
+arrayref = "0.3.6"
+libtock_drivers = { path = "../third_party/libtock-drivers" }
+crypto = { path = "../libraries/crypto", features = ['std'] }
+cbor = { path = "../libraries/cbor"}
```
ia0, line 17: Style: Missing space before closing brace.

fuzz/fuzz_targets/fuzz_target_split_assemble.rs
```diff
@@ -0,0 +1,65 @@
...
+use ctap2::ctap::hid::send::HidPacketIterator;
+use ctap2::ctap::hid::{HidPacket, Message};
+use libfuzzer_sys::fuzz_target;
+use libtock_drivers::timer::Timestamp;
+
+const DUMMY_TIMESTAMP: Timestamp<isize> = Timestamp::from_ms(0);
+const PACKET_TYPE_MASK: u8 = 0x80;
+
+// Converts a byte slice into Message
+fn raw_to_message(data: &[u8], len: usize) -> Message {
```
ia0, line 18: The function seems to have redundant information. In Rust, a reference to a slice is a "fat pointer", i.e. it contains both a pointer to the slice and the length of the slice. So here, the `len` argument is redundant with `data`. You can replace all usages of `len` with `data.len()` in this function and remove the argument.

fuzz/fuzz_targets/fuzz_target_split_assemble.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+
+/* Fuzzing HID packets splitting and assembling functions*/
+fuzz_target!(|data: &[u8]| {
+    let Message {
+        cid,
+        mut cmd,
+        payload,
+    } = raw_to_message(data, data.len());
```
ia0, line 42: This is probably just style, but it removes a bit of the clutter. What about `let mut message = raw_to_message(data);` and then `message.clone()` and `message.cmd &= !PACKET_TYPE_MASK`?

third_party/lang-items/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "MIT/Apache-2.0"
 edition = "2018"
 
 [dependencies]
-libtock_core = { path = "../../third_party/libtock-rs/core", default-features = false, features = ["alloc_init", "custom_panic_handler", "custom_alloc_error_handler"] }
+libtock_core = { path = "../../third_party/libtock-rs/core", default-features = false, features = ["custom_panic_handler", "custom_alloc_error_handler"] }
```
ia0, line 12: I guess that's the solution to the problem we faced.

fuzz/fuzz_targets/fuzz_target_split_assemble.rs
mingxguo27: Ahh I can't believe it :D I'm too dumb. I had to break it down into parts because I got an error saying Message didn't derive some trait, probably I tried .copy() instead of .clone() first.

third_party/lang-items/Cargo.toml
mingxguo27: Yes :')

ia0 (dismissed):

kaczmarczyck (dismissed):

kaczmarczyck (dismissed):

kaczmarczyck:
@jmichelp The binary seems to not be reproducible anymore. Is this related to this pull request? I'd propose to squash and merge on this one, and investigate later?

jmichelp:
Should be fixed by #153 
Please rebase

kaczmarczyck (dismissed):
Let's give @gendx a chance to take a look next week. Otherwise LGTM.

src/ctap/storage.rs
```diff
@@ -12,16 +12,19 @@
...
 
-use crate::crypto::rng256::Rng256;
 #[cfg(feature = "with_ctap2_1")]
 use crate::ctap::data_formats::{extract_array, extract_text_string};
 use crate::ctap::data_formats::{CredentialProtectionPolicy, PublicKeyCredentialSource};
 use crate::ctap::pin_protocol_v1::PIN_AUTH_LENGTH;
 use crate::ctap::status_code::Ctap2StatusCode;
 use crate::ctap::{key_material, USE_BATCH_ATTESTATION};
+#[cfg(feature = "fuzzing")]
+use crate::embedded_flash::{self, StoreConfig, StoreEntry, StoreError};
```
gendx, line 22: Looks good to me, although I still don't fully understand why the import path changes. But we can discuss splitting away the `embedded_flash` into its own crate as a separate pull request.

src/lib.rs
```diff
@@ -16,8 +16,15 @@
 
 #[macro_use]
 extern crate alloc;
+#[cfg_attr(feature = "fuzzing", macro_use)]
+extern crate arrayref;
+#[cfg_attr(feature = "fuzzing", macro_use)]
+extern crate cbor;
+#[cfg(not(feature = "fuzzing"))]
```
gendx, line 23: Looks good, although we won't need these anymore when migrating to Rust 2018 edition (#118).

third_party/lang-items/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "MIT/Apache-2.0"
 edition = "2018"
 
 [dependencies]
-libtock_core = { path = "../../third_party/libtock-rs/core", default-features = false, features = ["alloc_init", "custom_panic_handler", "custom_alloc_error_handler"] }
```
gendx, line 12: If we remove this feature, how is the allocator initialized in the real OpenSK application? Have you tried running OpenSK with this pull request?

gendx (changes requested):

gendx:
I'm concerned that removing the `alloc_init` feature from libtock-rs core breaks OpenSK, as no allocator is initialized.

gendx (commented):
Now that #156 is merged, the changes in `src/` and `third_party/` shouldn't be needed anymore, and the `fuzzing` feature can be removed. The `std` and `ram_storage` features should be sufficient.

**Note:** `cargo fuzz` requires `cargo install cargo-fuzz`. This should be added to either `setup.sh` or a `fuzz-setup.sh` script.

third_party/lang-items/Cargo.toml
gendx: Resolved by #156.

src/lib.rs
gendx: Resolved by #156.

src/ctap/storage.rs
gendx: Resolved by #156.

kaczmarczyck (dismissed):

gendx (commented):
It would be great to add a GitHub workflow that does `cargo fuzz build` to make sure this doesn't break in the future.

gendx (approved):

gendx:
No questions left from me. I'll leave it to @kaczmarczyck to merge this pull request.

PR #150, opened 2020-09-16
Update to a newer kernel version

### LINKED ISSUE #148: Update Tock kernel to a more recent version
This involves:
- [x] Updating the submodule to a more recent commit
- [x] Update the patches
  - [x] persistent storage
  - [x] USB
  - [x] USB debugging
  - [x] additional boards 
- [x] Update the in-tree boards
- [x] Run unit tests
- [ ] Test dev kit against CTAP2 Test tool

jmichelp:
Fixes #148 

- [x] Tests pass


jmichelp:
There's an issue with the reproducibility. I guess the AAGUID is changing. Will investigate that. In the meantime, this shouldn't prevent the code review to happen

ia0 (dismissed):
I only looked at the persistent storage patch.

ia0 (approved):

patches/tock/05-mpu-fix.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/kernel/src/sched.rs b/kernel/src/sched.rs
+index ed3ae8260..f109960d3 100644
+--- a/kernel/src/sched.rs
++++ b/kernel/src/sched.rs
+@@ -632,6 +632,7 @@ impl Kernel {
+                     let context_switch_reason = process.switch_to();
+                     scheduler_timer.disarm();
+                     chip.mpu().disable_app_mpu();
++                    chip.mpu().clear_mpu();
```
gendx, line 9: What's the purpose of this patch? Is there a bug tracked upstream?

jmichelp: Allistair changed the behavior of the MPU (https://github.com/tock/tock/commit/128782dad5ad410d3e48c16aadba60d9ed6d842b) and this makes our firmware panic when we try to touch the storage.
This one line patch reverts the change made upstream. We need to investigate why we panic and how to properly fix it without diverging from upstream

gendx: Can you file an upstream bug before we merge the pull request here?

jmichelp: I'm not sure it's an upstream bug. Maybe we were relying on the MPU being lenient in the kernel which is not the case anymore. I will open a bug on our side to investigate whether it's a problem upstream or on our side (e.g. maybe we need to add MPU configurations for the storage)

jmichelp: Created #155

gendx (dismissed):

kaczmarczyck:
Using the new kernel version for various tests didn't surface any problems for me. LGTM from the application side.

gendx (approved):

PR #158, opened 2020-09-22
Return errors and Vecs from CTAP storage

ia0:
This prepares for the new persistent store library API. All persistent storage methods may fail and return owned objects.

- [x] Tests pass


src/ctap/storage.rs
```diff
@@ -315,171 +326,170 @@ impl PersistentStore {
...
-    pub fn master_keys(&self) -> MasterKeys {
-        // We have as invariant that there is always exactly one MasterKeys entry in the store.
+    pub fn master_keys(&self) -> Result<MasterKeys, Ctap2StatusCode> {
         let (_, entry) = self.store.find_one(&Key::MasterKeys).unwrap();
-        let data = entry.data;
-        // And this entry is well formed: the encryption key followed by the hmac key.
-        let encryption = array_ref!(data, 0, 32);
-        let hmac = array_ref!(data, 32, 32);
-        MasterKeys { encryption, hmac }
+        if entry.data.len() != 64 {
```
kaczmarczyck, line 382: If we're doing this check here, why return a `struct` with a `Vec`? Could `MasterKeys` contain the arrays? Probably not a big difference, but it seems more natural to me to have the length check and the conversion in one place?

src/ctap/storage.rs
```diff
@@ -315,171 +326,170 @@ impl PersistentStore {
...
+            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        }
+        Ok(MasterKeys(entry.data.to_vec()))
     }
 
-    pub fn pin_hash(&self) -> Option<&[u8; PIN_AUTH_LENGTH]> {
-        self.store
-            .find_one(&Key::PinHash)
-            .map(|(_, entry)| array_ref!(entry.data, 0, PIN_AUTH_LENGTH))
+    pub fn pin_hash(&self) -> Result<Option<Vec<u8>>, Ctap2StatusCode> {
```
kaczmarczyck, line 388: Same here, why move from reference to arrays entirely, when `pin_hash` has a known fixed length?

src/ctap/storage.rs
```diff
@@ -565,7 +575,7 @@ impl PersistentStore {
         Ok(())
     }
 
-    pub fn aaguid(&self) -> Result<&[u8; AAGUID_LENGTH], Ctap2StatusCode> {
+    pub fn aaguid(&self) -> Result<Vec<u8>, Ctap2StatusCode> {
```
kaczmarczyck, line 578: As above again, what do you think about returning an array?

src/ctap/storage.rs
ia0: Good point, I was thinking about returning arrays but was worried about the stack usage. But actually those arrays are very small. I'll do the change, it's a strict improvement.

src/ctap/storage.rs
```diff
@@ -105,7 +105,7 @@ enum Key {
     MinPinLengthRpIds,
 }
 
-pub struct MasterKeys(Vec<u8>);
+pub struct MasterKeys([u8; 64]);
```
kaczmarczyck, line 108: If `MasterKeys` is already it's own struct, I'm wondering why the two arrays need to be stored in a single array and have getter functions, compared to how it was before with 2 public arrays you could just use?

kaczmarczyck (dismissed):
Just one comment left which is a matter of opinion I think. Otherwise LGTM.

src/ctap/storage.rs
ia0: Oh good point, I'll change that back.

kaczmarczyck (approved):

kaczmarczyck (approved):

PR #162, opened 2020-09-23
Setup a Clippy workflow and fix remaining warnings.

### LINKED ISSUE #122: Add a CI workflow to run Clippy
We should run [Clippy](https://github.com/rust-lang/rust-clippy) in one of the GitHub workflows.

### LINKED ISSUE #142: Clippy fails on `libraries/crypto/ec/int256.rs`
As part of #122 I was trying to check if clippy could be run now that we moved to a more recent libtock-rs.

We still have a couple of warnings from libtock-rs that are easy to fix if they're not already upstream.

The first blocking errors are encountered on the int256 arithmetic.

The error [suspicious_arithmetic_impl](https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_arithmetic_impl) is reported on:
* https://github.com/google/OpenSK/blob/57bf6235394692a95c5bfcae0970fe62800f305d/libraries/crypto/src/ec/int256.rs#L562
* https://github.com/google/OpenSK/blob/57bf6235394692a95c5bfcae0970fe62800f305d/libraries/crypto/src/ec/int256.rs#L592
* https://github.com/google/OpenSK/blob/57bf6235394692a95c5bfcae0970fe62800f305d/libraries/crypto/src/ec/int256.rs#L612

The error [suspicious_op_assign_impl](https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_op_assign_impl) is reported on
* https://github.com/google/OpenSK/blob/57bf6235394692a95c5bfcae0970fe62800f305d/libraries/crypto/src/ec/int256.rs#L576
* https://github.com/google/OpenSK/blob/57bf6235394692a95c5bfcae0970fe62800f305d/libraries/crypto/src/ec/int256.rs#L627

gendx:
Fixes #122 and #142.

Some lints are explicitly allowed:
- suspicious arithmetic in elliptic curves,
- `enum_variant_names` in the `Key` enum of storage, which has a variant ending by `Key`,
- `let_and_return`, where the let value is used under some features,
- `new_without_default` in the console driver and for the `CtapHid` struct - I don't think a `Default` makes sense there.

One remaining issue is that the workflow doesn't seem to deny the warnings, e.g. [this run](https://github.com/google/OpenSK/pull/162/checks?check_run_id=1160542267). This seems related to https://github.com/rust-lang/rust-clippy/issues/5749.

This allows to successfully run `cargo clippy --all-targets --features std`. However, "vanilla" `cargo clippy` doesn't work due to lang items being defined twice.

```rust
error[E0152]: found duplicate lang item `oom`
   --> third_party/lang-items/src/allocator.rs:96:1
    |
96  | / unsafe fn alloc_error_handler(_layout: Layout) -> ! {
97  | |     util::signal_oom();
98  | |     util::signal_panic();
99  | |
...   |
109 | |     util::cycle_leds()
110 | | }
    | |_^
    |
    = note: the lang item is first defined in crate `std` (which `libtock_core` depends on)

error[E0152]: found duplicate lang item `panic_impl`
  --> third_party/lang-items/src/panic_handler.rs:9:1
   |
9  | / fn panic_handler(_info: &PanicInfo) -> ! {
10 | |     util::signal_panic();
11 | |
12 | |     #[cfg(feature = "panic_console")]
...  |
25 | |     util::flash_all_leds();
26 | | }
   | |_^
   |
   = note: the lang item is first defined in crate `std` (which `libtock_core` depends on)

error: aborting due to 2 previous errors
```

- [x] Tests pass


jmichelp (approved):

gendx:
I fixed all the remaining warnings (or allowed some local exceptions). The only remaining issue seems to be that the workflow doesn't actually fail on a warning - but I think we can still start to experiment with it and have a separate issue to track that bug.

src/ctap/hid/mod.rs
```diff
@@ -137,6 +137,7 @@ impl CtapHid {
     const TIMEOUT_DURATION: Duration<isize> = Duration::from_ms(100);
     const WINK_TIMEOUT_DURATION: Duration<isize> = Duration::from_ms(5000);
 
+    #[allow(clippy::new_without_default)]
```
kaczmarczyck, line 140: Why do you sometimes `#allow`, and sometimes implement `Default`?

src/ctap/storage.rs
```diff
@@ -264,12 +265,10 @@ impl PersistentStore {
...
-        if check_cred_protect
-            && result.as_ref().map_or(false, |cred| {
-                cred.cred_protect_policy
-                    == Some(CredentialProtectionPolicy::UserVerificationRequired)
-            })
-        {
+        let user_verification_required = result.as_ref().map_or(false, |cred| {
+            cred.cred_protect_policy == Some(CredentialProtectionPolicy::UserVerificationRequired)
+        });
+        if check_cred_protect && user_verification_required {
```
kaczmarczyck, line 271: Might not use short-circuiting this way? (I don't particularly mind.)

gendx: With the 1-statement if with `&&`, Clippy complains about the complexity. If I nest two ifs, it complains that it can be collapsed into a `&&`. Here we don't really care about the short-circuiting so I think naming the result of this `map_or` as `user_verification_required` and then using it is the clearest.

src/ctap/hid/mod.rs
gendx: Let's just allow this lint everywhere, as we don't use it.

kaczmarczyck (approved):

PR #165, opened 2020-09-25
Fuzzing helper library and more fuzz targets

mingxguo27:
Created a fuzzing helper library / API to increase code reusability.

Added fuzz targets for specific input types. For example, we can treat inputs as ctap2 command parameters encoded in cbor or ctap1 raw messages. This increases the achieved coverage (from 400 to 1400 in ctap2 case) compared to treating inputs as raw messages.


fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
+// Copyright 2019 Google LLC
```
gendx, line 1: This should be 2020.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                // Only checks for assembling crashes, not for semantics.
+                let _ = assembler_reply.parse_packet(&pkt_reply, DUMMY_TIMESTAMP);
+            }
+        }
+    }
+}
+
+// Splits and reassembles the given data as HID packets.
+pub fn split_assemble(data: &[u8]) {
```
gendx, line 162: Maybe `split_assemble_hid_packets` would be clearer.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+        Message {
+            cid,
+            cmd: 0,
+            payload: vec![],
+        }
+    } else {
+        Message {
+            cid: array_ref!(data, 0, 4).clone(),
+            cmd: data[4],
+            payload: data[5..].to_vec(),
```
gendx, line 63: FYI, the `arrayref` crate also has a macro to split into multiple slices. One can either use

```rust
let (cid, cmd, payload) = array_refs![data, 4, 1, data.len() - 5]
```

or maybe clearer:

```rust
let (header, payload) = data.split_at(5);
let (cid, cmd) = array_refs![header, 4, 1];
```

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+
+const COMMAND_INIT: u8 = 0x06;
+const CHANNEL_BROADCAST: ChannelID = [0xFF, 0xFF, 0xFF, 0xFF];
+const PACKET_TYPE_MASK: u8 = 0x80;
+
+const CLOCK_FREQUENCY_HZ: usize = 32768;
+const DUMMY_TIMESTAMP: Timestamp<isize> = Timestamp::from_ms(0);
+const DUMMY_CLOCK_VALUE: ClockValue = ClockValue::new(0, CLOCK_FREQUENCY_HZ);
+
+#[derive(Clone, PartialEq)]
```
gendx, line 41: This can (and should) also derive `Copy`, as the enum is a "trivial" type that doesn't store any data.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+                result_cid.copy_from_slice(&result.payload[8..12]);
+            }
+        }
+    }
+    result_cid
+}
+
+// Checks whether the given data can be interpreted as the given type.
+fn is_type(data: &[u8], input_type: InputType) -> bool {
+    if input_type.clone() == InputType::Ctap1 {
```
gendx, line 97: This clone is then unnecessary.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+                AuthenticatorClientPinParameters::try_from(decoded_cbor).is_ok()
+            }
+            _ => true,
+        },
+    }
+}
+
+// Interprets the raw data as of the given input type and
+// invokes message splitting, assembling and packet processing at CTAP HID level.
+pub fn process_input(data: &[u8], input_type: InputType) {
```
gendx, line 119: It would be clearer to have a more specific function name than `process_input`. In a way, all functions process some input and return some output.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+    if !is_type(data, input_type.clone()) {
+        return ();
```
kaczmarczyck, line 121: Seems inefficient to throw away so many inputs. How slow is this fuzzer, and do you need to hand it some working examples?

Also, it seems splitting the fuzzing into three targets like that may benefit the corpus generation for the other tool, but not really necessary for OpenSK's sake. Is that correct? If so, it should at least be documented.

mingxguo27: Yes, in this case providing an initial corpus makes a big difference (from coverage stuck at 300+ for at least millions of executions to starting at 800+ from the beginning). The fuzzer always generates between 3000-4000 inputs per sec. When a corpus is provided, it's not that inefficient. The inputs get accepted really often.

Yes I split considering the use case for the test tool. Should I also include the one fuzz target for all CTAP2 cases?

kaczmarczyck: This new fuzz target would have more coverage, but produces a less specific corpus? Sounds like something you would want to run to test OpenSK specifically. Doesn't sound like a lot of work, so maybe just add it and add documentation as of what each target's purpose is.

fuzz/fuzz_helper/src/lib.rs
mingxguo27: Hi Guillaume.
I've tried both and there are problems I don't know how to fix:

For the first option, I cannot use it with dynamic values.
For the second option, I got type mismatch no matter how I cast. And to use array_refs![header,4,1] rust expects header to be an array of length 5 so I have to do some conversion, since split_at returns slices. 

Any ideas?

kaczmarczyck (approved):

fuzz/fuzz_helper/src/lib.rs
gendx: Yes, it's true that unfortunately `data.split_at(5);` doesn't return a sized `&[T; 5]` but a generic `&[T]`. Then the current code should be fine!

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+// Interprets the raw data as any ctap command (including the command byte) and
+// invokes message splitting, packet processing at CTAP HID level and response assembling
+// using an initialized and allocated channel.
+pub fn process_ctap_any_type(data: &[u8]) {
+    // Initialize ctap state and hid and get the allocated cid.
+    let mut rng = ThreadRng256 {};
+    let user_immediately_present = |_| Ok(());
```
gendx, line 146: Note: rather than defining this lambda function in each function, you can simply define it once (with the correct `ErrorType`) and then use it everywhere:

```rust
fn user_immediately_present() -> Result<(), ErrorType> {
    Ok(())
}

File: fuzz/fuzz_helper/src/lib.rs [hunk 9]
```diff
@@ -0,0 +1,206 @@
...
+    command.extend(data);
+    process_message(&command, &mut ctap_state, &mut ctap_hid);
+}
+
+// Interprets the raw data as of the given input type and
+// invokes message splitting, packet processing at CTAP HID level and response assembling
+// using an initialized and allocated channel.
+pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
+    if !is_type(data, input_type) {
+        return ();
```
gendx, line 161: This is equivalent to `return;`

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+        InputType::Ctap1 => {
+            command.extend(&[0x03]);
+        }
+    }
+    command.extend(data);
+    process_message(&command, &mut ctap_state, &mut ctap_hid);
+}
+
+// Splits and reassembles the given data as HID packets.
```
gendx, line 189: Nit: a more complete description would be: "Splits the given data as HID packets and reassembles it, verifying that the original input message is reconstructed."

gendx (commented):
Just left a few nits on the code, otherwise it looks good :)

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -186,7 +188,7 @@ pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
     process_message(&command, &mut ctap_state, &mut ctap_hid);
 }
 
-// Splits and reassembles the given data as HID packets.
+// Splits the given data as HID packets and reassembles it, verifying that the original input message is reconstructed.
```
kaczmarczyck, line 191: This line is 120 characters long. I'm curious why our tests pass. Maybe a workflow is missing a directory here?

I'll approve to unblock, and leave it to @gendx to merge.

kaczmarczyck (approved):

fuzz/fuzz_helper/src/lib.rs
gendx: Let's resolve it here, and investigate in a separate bug.

gendx (dismissed):

gendx (approved):

fuzz/fuzz_helper/src/lib.rs
gendx: FYI, this is intended as we don't enable the (unstable) `wrap_comments` option of Rustfmt:
- https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#wrap_comments
- https://github.com/rust-lang/rustfmt/issues/3347

PR #145, opened 2020-08-19
Elf2tab internally install

coelner:
Fixes #141 

- [ ] Tests pass
- [ ] Appropriate changes to README are included in PR

PR #167, opened 2020-09-29
Install elf2tab in its own directory.

gendx:
Since #145, `elf2tab` is installed in the OpenSK root directory.
- This is not clean as the `.crates.toml` and `.crates2.json` files appear in the root.
- The `reset.sh` script removes this local `elf2tab` installation, which means that a full compilation of `elf2tab` will happen again on the next `./setup.sh` run. This is not necessary, as the setup script will check the version and re-install it if a version upgrade is needed.

This pull request installs `elf2tab` in its own sub-directory instead, which is excluded from the `reset.sh` script.

- [x] Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,13 @@
-bin/
 target/
 Cargo.lock
 
+# Local installation of elf2tab.
+/elf2tab/
+
 # Prevent people from commiting sensitive files.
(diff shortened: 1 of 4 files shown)


ia0 (approved):
Thanks!

Issue #169, opened 2020-09-29
Make sure that the Clippy workflow denies warnings and causes the workflow to fail


gendx:
## Expected Behavior

When Clippy warns against a lint with `-D warnings` in a GitHub workflow, the lint is reported and the workflow fails.

## Actual Behavior

On the pull request page, the Clippy workflow appears as green. One has to manually check the workflow's output log to notice the lint.

This seems related to https://github.com/rust-lang/rust-clippy/issues/5749.

## Steps to Reproduce the Problem

1. Add code that triggers a Clippy warning.
1. Create a pull request with that code.
1. Look at the GitHub workflow checks.

kaczmarczyck:
The script should fail now, and with it the workflow that's running it.

PR #170, opened 2020-09-29
[Do not merge] Test pull-request for issue #169.

gendx:
This is a test for the behavior in #169. This pull request can be removed and #169 marked as fixed once the GitHub checks fail here.

- [ ] Tests pass


kaczmarczyck:
@gendx Is this still uesful to be kept open as a canary?

gendx:
> @gendx Is this still uesful to be kept open as a canary?

Just rebased this pull request to see if the Clippy workflow now fails.

jmichelp:
Anyway, due to recent changes on Github, the workflow for Clippy must be adjusted for it to work properly. It now must be split into 2 workflows: an untrusted one which analyses the PR and produces a structured file with the results and a trusted one which takes this file and creates the annotations.

gendx:
Looks like the current workflow passed instead of denying the lint, so leaving this PR open until #169 is fixed.

kaczmarczyck:
This PR does not work with the current CI, since crypto was excluded from clippy. I verified that CI fails when I add a let-and-return in the opensk library.

PR #173, opened 2020-10-02
Fix formatting.

gendx:
I ran `cargo fmt` in the patched Tock repository, and the following line was reformatted:

```rust
self.kernel.storage_locations().iter().any(|storage_location| {
```

- [x] Tests pass


ia0 (dismissed):
I'm surprised the current patch was missing the space for common empty lines and seemed to use more context lines. Maybe @jmichelp uses a custom version of patch.

ia0 (approved):

PR #178, opened 2020-10-09
Add initial corpus files

mingxguo27:
This is a JSON file containing the initial seed inputs I used for fuzzing CTAP2 commands and helped a lot in the fuzzing process (coverage 400 -> 1500). The python script parses it into a corpus.


fuzz/ctap2_commands_parameters_corpus.json
```diff
@@ -0,0 +1,152 @@
...
+        "cbor": "map {\"a\": 1, \"b\": [2, 3]}",
+        "description": "cbor value"
+    },
+    {
+        "hex": "bf6346756ef563416d7421ff",
+        "cbor": "indefinite length map {\"Fun\": true, \"Amt\": -2}",
+        "description": "cbor value"
+    },
+    {
+        "hex": "",
```
gendx, line 93: Why is the hex empty when the CBOR is not?

mingxguo27: Ups my bad. Fixed.

fuzz/make_corpus.py
```diff
@@ -0,0 +1,59 @@
+import json
+import argparse
```
kaczmarczyck, line 2: Please reorder. (Also, we probably need a Python linter if we start using Python.)

kaczmarczyck (changes requested):

fuzz/make_corpus.py
mingxguo27: There is, but it didn't detect this somehow :(

kaczmarczyck (approved):

PR #180, opened 2020-10-09
DO NOT MERGE: Check reproducibility

ia0:
This is just to make sure that our build is not hermetic. As a consequence binaries are not reproducible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+Make sure that binaries are not reproducible because the build is not hermetic.
+
 # <img alt="OpenSK logo" src="docs/img/OpenSK.svg" width="200px">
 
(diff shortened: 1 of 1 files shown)


gendx:
> This is just to make sure that our build is not hermetic. As a consequence binaries are not reproducible.

Any more context on that? To me, it seems that over time reproducibility breaks because dependencies get updated and we don't pin a specific version via a `Cargo.lock` file. There is also the OS discrepancies that we're aware of (see https://github.com/google/OpenSK/pull/94#issue-403232755 and https://github.com/rust-lang/cargo/issues/8140).

But other than that binaries should be reproducible, and if there is any other unknown reason for non-reproducibility we should investigate.

ia0:
The context is that the workflow for reproducibility is thus completely useless, because it tracks noise instead of signal. As such it should be disabled until this is fixed.

Issue #151, opened 2020-09-16
Modify the reproducible workflow behavior


jmichelp:
## Expected Behavior

We should have 2 different workflows:
1. On each PR, build twice the firmware for each operating system, removing the `crypto_data` directory and copying the `reference_crypto_data` between the two builds and ensure that the hashes remain the same. This should block PR and tell us that we have something wrong.
2. On merge, instead of checking the hashes against a reference, this should generate the new hashes and text outputs and automatically push a new commit with the this as a reference.

## Actual Behavior

Currently the two behaviors are mixed up and don't allow us to stop when a build is actually not reproducible.

## Steps to Reproduce the Problem

N/A

## Specifications

- Version: N/A
- Platform: N/A

kaczmarczyck:
Workflow is dropped.

PR #181, opened 2020-10-09
Disable reproducible workflow

ia0:
As illustrated by #180, our build is not hermetic. As a consequence this workflow fails for unrelated reasons. Disabling until it brings any value.

### DIFF
--- a/.github/workflows/reproducible.yml
+++ b/.github/workflows/reproducible.yml
@@ -38,9 +38,9 @@ jobs:
           name: reproduced-${{ matrix.os }}
           path: reproducible/reproduced.tar
 
-      - name: Comparing binary sizes
-        if: always()
-        run: git diff --no-index reproducible/reference_elf2tab_${{ matrix.os }}.txt reproducible/elf2tab.txt
-      - name: Comparing cryptographic hashes
-        if: always()
-        run: git diff --no-index reproducible/reference_binaries_${{ matrix.os }}.sha256sum reproducible/binaries.sha256sum
+      # - name: Comparing binary sizes
+      #   if: always()
+      #   run: git diff --no-index reproducible/reference_elf2tab_${{ matrix.os }}.txt reproducible/elf2tab.txt
+      # - name: Comparing cryptographic hashes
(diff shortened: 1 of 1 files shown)


gendx:
Quoting the rationale in https://github.com/google/OpenSK/pull/180#issuecomment-706099968:

> The context is that the workflow for reproducibility is thus completely useless, because it tracks noise instead of signal. As such it should be disabled until this is fixed.

I'd argue that the current workflow tracks signal+noise. Therefore "completely useless" is too strong a statement IMO - as the actual signal affects it.

One can argue that the core reproducibility is actually quite stable, i.e. it's really rare that a pull request makes the build totally unreproducible. By that I mean removing some flag like `--remap-path-prefix`: https://github.com/google/OpenSK/blob/f328971ae876c9ac7388bf715e435f0bf033e83d/deploy.py#L371
So we mostly observe noise, which is indeed a problem.

But I think we should improve the process to reduce the noise (e.g. #151) rather than outright removing any reproducibility goal.

ia0:
> So we mostly observe noise, which is indeed a problem.

Yes, that's what I mean by completely useless. A workflow should have a low rate of false positives otherwise it hinders engineering velocity much more than it has benefits. Here, there are no true positives and only false positives on almost every commit including commits that do not affect the build.

> But I think we should improve the process to reduce the noise

Yes obviously. This PR is a temporary measure until #182 is fixed. Then #151 is a possible improvement on the workflow system.

jmichelp:
I would suggest that either we decide that compiling/generating the build artefacts with this workflow are useful and then we continue running all the commands but we ensure this can't fail (e.g. `git --diff ... || true`) or we disable completely the workflow.

ia0:
The `git diff || true` seems strictly better because we get more information which could be used for debugging. Although this information can be recovered from the artifacts which are still produced, I still did the change.

Once debugging is not needed because #182 is fixed, we can either re-enable failing on diffs, or continue with the improvements suggested in #151 instead.

jaya98777 (approved):

PR #185, opened 2020-10-19
Add NFC driver interface

MirnaMuhammad98:
Adding a driver interface to support the syscalls interface.


third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    pub fn emulate(enabled: bool) -> bool {
+        let result_code =
+            syscalls::command(DRIVER_NUMBER, command_nr::EMULATE, enabled as usize, 0);
+        if result_code.is_err() {
+            return false;
+        }
+
+        true
```
kaczmarczyck, line 44: You can just return `!result_code.is_err()`, or even better `result_code.is_ok()`. And you probably don't need the variable, if you don't want it for documentation purposes, then `syscalls::command(...).is_ok()` as a 1-line function works.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,144 @@
...
+impl NfcTag {
+    pub fn enable_emulation() {
+        NfcTag::emulate(true);
+    }
+
+    pub fn disable_emulation() {
+        NfcTag::emulate(false);
+    }
+
+    pub fn emulate(enabled: bool) -> bool {
```
kaczmarczyck, line 37: I don't understand your public interface. Do you want people to call `emulate` with the correct bool? Or the `enable` and `disable` function? Do you care about the return value, and why is it only returned in `emulate`?

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,144 @@
...
+            syscalls::command(DRIVER_NUMBER, command_nr::EMULATE, enabled as usize, 0);
+        if result_code.is_err() {
+            return false;
+        }
+
+        true
+    }
+
+    /// Subscribe to the tag being SELECTED callback.
+    pub fn selected() -> bool {
```
kaczmarczyck, line 48: You could consider not returning `bool` everywhere, but just passing on the error code with a `Result<(), ErrorType>`. The you'd have access to the `?` operator below, like
`syscalls::subscribe(...)?;
instead of `is_err`.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Configure the tag type command.
+    pub fn configure(tag_type: u8) -> bool {
+        let result_code =
+            syscalls::command(DRIVER_NUMBER, command_nr::CONFIGURE, tag_type as usize, 0);
+        if result_code.is_err() {
+            return false;
+        }
+
+        true
```
kaczmarczyck, line 72: See above for `is_ok()` or switching to returning error codes. Also, you don't need a newline here in my opinion.

Same goes for functions below.

kaczmarczyck (changes requested):

deploy.py
```diff
@@ -921,6 +921,13 @@ def main(args):
       const="console_test",
       help=("Compiles and installs the console_test example that tests the "
             "console driver with messages of various lengths."))
+  apps_group.add_argument(
+      "--nfct_test",
```
jmichelp, line 925: It seems that the `examples/nfct_test.rs` file was added and then deleted as part of this PR.
Therefore running `deploy.py --nfct_test` will result in an error.

third_party/libtock-drivers/src/nfc.rs
jmichelp: Actually, libtock-rs has its own return type which should be used: `TockResult`

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -63,34 +65,21 @@ impl NfcTag {
...
-
-        true
+        syscalls::command(DRIVER_NUMBER, command_nr::FRAMEDELAYMAX, delay as usize, 0).is_ok()
     }
 
     /// 1. Share with the driver a buffer.
     /// 2. Subscribe to having a successful receive callback.
     /// 3. Issue the request for reception.
-    pub fn receive(buf: &mut [u8]) -> bool {
+    pub fn receive(buf: &mut [u8; 256]) -> TransOrRecvStatus {
```
kaczmarczyck, line 79: You use your own `enum` instead of reusing the libtock result because you want to convert them to something more readable? Why does i.e. `set_framedelaymax` still use `bool` as a return type?

kaczmarczyck, line 79: If this status wouldn't contain the success, it would instead return
`Result<(), TransOrRecvError>`
or
`Result<(), WhateverLibtockError>`
and you'd have access to `Result` magic like the `?` operator later.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -23,25 +24,26 @@ mod allow_nr {
     pub const RECEIVE: usize = 2;
 }
 
+pub enum TransOrRecvStatus {
+    Error,
+    InvalidBuffer,
+    OOM,
+    Success,
+}
```
kaczmarczyck, line 32: As suggested below, you could either reuse an existing error code `enum` or remove `Success` and use it inside of `Result`s.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -23,25 +24,26 @@ mod allow_nr {
     pub const RECEIVE: usize = 2;
 }
 
+pub enum TransOrRecvStatus {
+    Error,
```
kaczmarczyck, line 28: `Error` is too unspecific. Maybe make it clear that it is supposed to include all errors that are neither `InvalidBuffer` nur `OOM`?

third_party/libtock-drivers/src/nfc.rs
kaczmarczyck: Ah yeah, only saw your comment now. Please consider my suggestions below with `TockResult`.

third_party/libtock-drivers/src/nfc.rs
MirnaMuhammad98: > `set_framedelaymax` still use `bool` as a return type?

There's no variations in the kind of errors I would get. It either succeeds and the registers gets set correctly in the driver, or somehow the kernel fails to process the command.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -76,61 +76,53 @@ impl NfcTag {
...
-            return TransOrRecvStatus::InvalidBuffer;
-        }
-
+    pub fn transmit(buf: &mut [u8], amount: usize) -> Result<usize, TockError> {
+        let result = syscalls::allow(DRIVER_NUMBER, allow_nr::TRANSMIT, buf)?;
+        // set callback with 1 argument, to receive ReturnCode
         let done = Cell::new(false);
-        let mut alarm = || done.set(true);
-        let subscription = syscalls::subscribe::<callback::Identity0Consumer, _>(
+        let result_code = Cell::new(None);
```
kaczmarczyck, line 112: Your `Cell` contains an option, so you don't need the extra `bool` as an indicator? So if you don't want to keep the `done` pattern, you could replace the `dont.get()` check below? Same for `receive`.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -76,61 +76,53 @@ impl NfcTag {
...
+        // set callback with 2 arguments, to receive ReturnCode and RX Amount
         let done = Cell::new(false);
-        let mut alarm = || done.set(true);
-        let subscription = syscalls::subscribe::<callback::Identity0Consumer, _>(
+        let result_code = Cell::new(None);
+        let recv_amount = Cell::new(None);
+        let mut callback = |result, amount| {
+            result_code.set(Some(result));
+            recv_amount.set(Some(amount));
+            done.set(true)
```
kaczmarczyck, line 88: Did the underlying syscall interface change? Because the `callback` behaves differently now?

third_party/libtock-drivers/src/nfc.rs
kaczmarczyck: But you are using `.is_ok()` instead of `?`. Is the error here not a `TockError`, or are you saying it is not as interesting as the other places where we have `TockError`s?

third_party/libtock-drivers/src/nfc.rs
MirnaMuhammad98: No, just discovered how to send a callback with arguments 😅

third_party/libtock-drivers/src/nfc.rs
MirnaMuhammad98: I'm not receiving any information of value from the functions that contain `bool` as return type, so they either succeed or not.

For `receive()` and `transmit()`, I expect something from the kernel, so I want to return that, so I encapsulated the thing I want with the `Result<something, TockError>`. And also to be able to use `?` 😅

kaczmarczyck: It's a question of what we want. I'll let @jmichelp and @gendx do a pass of reviews.

kaczmarczyck:
The Rust code looks good like that! Please add the compile flag as discussed.

third_party/libtock-drivers/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 pub mod buttons;
 pub mod console;
 pub mod led;
+pub mod nfc;
```
kaczmarczyck, line 6: Please make this conditional:
#[cfg(feature = "with_nfc")]

kaczmarczyck (dismissed):

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn disable_emulation() -> bool {
+        NfcTag::emulate(false)
+    }
+
+    fn emulate(enabled: bool) -> bool {
+        syscalls::command(DRIVER_NUMBER, command_nr::EMULATE, enabled as usize, 0).is_ok()
+    }
+
+    /// Subscribe to the tag being SELECTED callback.
+    pub fn selected() -> bool {
```
jmichelp, line 50: This function will be blocking until the tag gets selected. I guess from the CTAP2 point of view, that's going to be a problematic behavior because we won't be able to receive packets from USB if we call this function.
@kaczmarczyck and @gendx  WDYT?
I have the impression that this function (as the documentation says) should only take a callback function as a parameter, issue the syscall and returns whether the subscription succeeded or not. The `yield_for` belongs to CTAP2 logic to be able to multiplex the transport channels (NFC, USB and maybe BLE at some point).
Otherwise we have to add the notion of timeouts like in USB but it means we need to poll sequentially USB and NFC which can lead to timing problems.
Last alternative would be to do the multiplexing directly in a capsule and then there's 1 single point to receive command (the capsule would probably need to tag the received packet/frame with the channel it came from for parsing/routing it properly)

kaczmarczyck: I see. They shouldn't both wait with timeout, true! Since USB was here first, I guess it has the privilege to stay blocking for now. Might make sense to revise that sometime.

MirnaMuhammad98: So should I change that to receive a callback and subscribe to it without yielding?

kaczmarczyck (approved):
LGTM from my side. Please merge from master and check in with @jmichelp to resolve his comments.

third_party/libtock-drivers/src/nfc.rs
jmichelp: At the moment we will move on with NFC XOR USB which removes the issue I pointed.
Multiplexing transport protocol may involve a significant amount of code changes so we'll deal with it at a later stage.

Following this decision, all NFC related code should be gated by a feature and `deploy.py` script will be responsible for ensuring that either USB or NFC are compiled but not both.

jmichelp (approved):

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,128 @@
...
+        let is_selected = Cell::new(false);
+        let mut is_selected_alarm = || is_selected.set(true);
+        let subscription = syscalls::subscribe::<callback::Identity0Consumer, _>(
```
gendx, line 59: I'm wondering about the sequence of events here. We subscribe to a callback but without triggering anything with a command, which is unusual with Tock.

How does the driver get notified that it should trigger the callback? Is it triggered just by the subscription (which although unusual is fine)?

I just want to make sure that the application doesn't miss any callback that the kernel would trigger before any subscription has happened.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /// 1. Share with the driver a buffer.
+    /// 2. Subscribe to having a successful receive callback.
+    /// 3. Issue the request for reception.
+    pub fn receive(buf: &mut [u8; 256]) -> TockResult<RecvOp> {
+        let result = syscalls::allow(DRIVER_NUMBER, allow_nr::RECEIVE, buf)?;
+        // set callback with 2 arguments, to receive ReturnCode and RX Amount
+        let result_code = Cell::new(None);
+        let recv_amount = Cell::new(None);
```
gendx, line 89: Here given that both values are set together, wouldn't it make sense to bundle them in a single `Cell<Option<RecvOp>>`?

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,128 @@
...
+        };
+        let subscription = syscalls::subscribe::<callback::Identity2Consumer, _>(
+            DRIVER_NUMBER,
+            subscribe_nr::RECEIVE,
+            &mut callback,
+        )?;
+        syscalls::command(DRIVER_NUMBER, command_nr::RECEIVE, 0, 0)?;
+        util::yieldk_for(|| recv_amount.get().is_some());
+        mem::drop(subscription);
+        mem::drop(result);
```
gendx, line 102: I don't think a manual drop is necessary here. Rust will drop them at the end of the function, and unless there is a bug in the driver, there should be a single callback per receive command, so at this point the received `Cell` should not change value anymore.

third_party/libtock-drivers/src/nfc.rs
MirnaMuhammad98: So this is actually an issue that I didn't consider. I can't think of a solution other than adding a command to unmask the `Selected` interrupt.

third_party/libtock-drivers/src/nfc.rs
MirnaMuhammad98: Done, will be modified in the following commit.

third_party/libtock-drivers/src/nfc.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Set the maximum frame delay value to support transmission with the reader.
+    pub fn set_framedelaymax(delay: u32) -> bool {
+        syscalls::command(DRIVER_NUMBER, command_nr::FRAMEDELAYMAX, delay as usize, 0).is_ok()
+    }
+
+    /// 1. Share with the driver a buffer.
+    /// 2. Subscribe to having a successful receive callback.
+    /// 3. Issue the request for reception.
+    pub fn receive(buf: &mut [u8; 256]) -> TockResult<RecvOp> {
+        let _result = syscalls::allow(DRIVER_NUMBER, allow_nr::RECEIVE, buf)?;
```
gendx, line 93: Ah, now I understand why there was a manual `drop`. The other drivers generally use the result variable by manually checking for `result.is_err()`, but here you directly used `?`.

I think the `?` pattern is good, but indeed this makes the remaining of the returned value look unused from the Rust compiler's point of view.

In the end, both using `let _foo = ...` or `let foo = ...; [...]; drop(foo);` work (https://users.rust-lang.org/t/unused-variables-that-need-to-not-be-prematurely-dropped/17192).

Since this is a new pattern for Tock drivers, either `let _result` or `drop(result)` styles should work - we should just have a consistent style across functions. For now, `receive` uses `let _result` whereas `transmit` uses `drop(result)` - could you change it so that both use the same style (of your choosing)?

third_party/libtock-drivers/src/nfc.rs
jmichelp: Isn't an `OptionalCell<RecvOp>` the same as a `Cell<Option<RecvOp>>`?

jmichelp (dismissed):
OK for me modulo my nit about the choice between `Cell<Option<...>>` and `OptionalCell<...>` on Guillaume's comment

jmichelp (approved):

kaczmarczyck (approved):

gendx (approved):

PR #186, opened 2020-10-19
Avoid erasing uicr

xiongyihui:
Do not use `nrf52_components::startup::NrfStartupComponent::new` to change the reset pin, which will erase UICR.
As the start address of the bootloader is stored at UICR, erasing UICR will make the bootloader invalid.

- [x] Tests pass

### DIFF
--- a/boards/nordic/nrf52840_mdk_dfu/src/main.rs
+++ b/boards/nordic/nrf52840_mdk_dfu/src/main.rs
@@ -26,7 +26,7 @@ const LED1_B_PIN: Pin = Pin::P0_24;
 
 // The nRF52840 Dongle button
 const BUTTON_PIN: Pin = Pin::P0_18;
-const BUTTON_RST_PIN: Pin = Pin::P0_02;
+// const BUTTON_RST_PIN: Pin = Pin::P0_02;
 
 const UART_RTS: Option<Pin> = Some(Pin::P0_21);
 const UART_TXD: Pin = Pin::P0_20;
@@ -209,12 +209,12 @@ pub unsafe fn reset_handler() {
     let chip = static_init!(nrf52840::chip::Chip, nrf52840::chip::new());
     CHIP = Some(chip);
 
-    nrf52_components::startup::NrfStartupComponent::new(
-        false,
-        BUTTON_RST_PIN,
-        nrf52840::uicr::Regulator0Output::V3_0,
-    )
-    .finalize(());
+    // nrf52_components::startup::NrfStartupComponent::new(
+    //     false,
+    //     BUTTON_RST_PIN,
+    //     nrf52840::uicr::Regulator0Output::V3_0,
+    // )
+    // .finalize(());
 
     // Create capabilities that the board needs to call certain protected kernel
     // functions.


jmichelp:
The startup component is also responsible for starting all the clocks.
As (from Tock point of view) we can't predict what the bootloader does, we can have a situation where the clocks are not started and many things will not work properly without HF clock

xiongyihui:
`NrfStartupComponent` does not set the clocks. It is `NrfClockComponent` to set the clocks.
`NrfStartupComponent` only changes the settings in UICR.

```
impl NrfStartupComponent {
    pub fn new(nfc_as_gpios: bool, button_rst_pin: Pin, reg_vout: Regulator0Output) -> Self {
        Self {
            nfc_as_gpios,
            button_rst_pin,
            reg_vout,
        }
    }
}
```

jmichelp:
Indeed. I overlooked it.

If the configuration in the code is the same than the one that is programmed, no erase should happen and the component should be a no-op. Is it just the RST pin that is problematic here and triggers an erase? I think I took it from the schematics so I'm a bit surprised that UICR would be re-programmed.

Is the data you mention regarding the bootloader programmed in one of the registers for hardware (memory region 0x014-0x080) and customer firmware (region 0x080 - 0x100) inside UICR? If that's the case, I'm considering fixing the issue upstream by reading this region and flashing it back after the erase operation in order to avoid data loss.

xiongyihui:
Changing the RST pin, the NFC setting or REGOUT0 will trigger an erase. If NFC is enabled in the previous app, the UICR will be erased after flashing the OpenSK app.
[The bootloader address is stored at `0x10001014`](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.sdk5.v11.0.0%2Fbledfu_memory_relocating.html). Reading the address and flashing it back is a good option.

jmichelp:
After doing some tests on a new dongle it seems that there's also another address located at `0x10001018` so I guess we should save both and restore them.
I have prepared a fix to push upstream in Tock directly. It also fixes a few other things. I just need to update it and split it to individual fixes before creating a PR for review.

As a side note the fact that we don't use signatures when programming over DFU and the fact that we don't kill the bootloader after flashing creates a security issue: one could flash a small code instead of Tock + OpenSK application and leak all the internal storage. But on the other hand we don't disable SWD port neither.

alishir:
Hi all,

May I ask when this pull request will be merge?

axmmisaka:
I can confirm with nRF52840 MDK USB Dongle, after merging this branch (I do not know what will happen if it is not merged) AND writing back UICR entries that you two just mentioned, the bootloader (UF2 Bootloader) can be properly loaded by pressing reset button while booting. I deployed with command `./deploy.py --board=nrf52840_mdk_dfu --opensk --programmer=jlink`.
image
image

Thus, I second the suggestion @xiongyihui gave: read the bootloader address and write it back afterwards. A potential problem I suspect might be bootstraping: NordicDFU might already be dead after OpenSK is deployed, which prevents us from writing contents to specific addresses. Nevertheless, deploying with JLink will not have such problem.

Writing back bootloader address (with a pirated SEGGER JLink):
```bash
misaka@tokiwadai:~/makerdiary$ nrfjprog --version
nrfjprog version: 10.12.0 
JLinkARM.dll version: 6.88b
misaka@tokiwadai:~/makerdiary$ nrfjprog --memrd 0x10001014 --n 16
0x10001014: 000E0000 000FE000 FFFFFFFF FFFFFFFF   |................|
misaka@tokiwadai:~/makerdiary$ nrfjprog --memwr 0x10001014 --val 0x000E0000
Parsing parameters.
Writing.
misaka@tokiwadai:~/makerdiary$ nrfjprog --memwr 0x10001018 --val 0x000FE000
Parsing parameters.
Writing.
misaka@tokiwadai:~/makerdiary$ nrfjprog --memrd 0x10001014 --n 16
0x10001014: 000E0000 000FE000 FFFFFFFF FFFFFFFF   |................|
```

axmmisaka:
What is indeed VERY WEIRD is: after writing these two UICR entries back, flashing BLE Connectivity -> OpenSK again with UF2 bootloader does not seem to destroy UICR/registers. 
Nevertheless, the behaviour is, if I recall correctly, not that, when OpenSK is flashed with a freshly installed UF2 bootloader - UICR will be erased. Maybe it has something to do with what is quoted below?

I used  `./deploy.py --board=nrf52840_mdk_dfu --opensk --programmer=none` to build the `.hex` file, then I use `python3 uf2conv.py OpenSK/target/nrf52840_mdk_dfu_merged.hex --convert --family 0xADA52840 --output nrf52840_mdk_dfu_merged.uf2` where `uf2conv.py` is [this](https://github.com/microsoft/uf2/blob/master/utils/uf2conv.py).

> 
> 
> Changing the RST pin, the NFC setting or REGOUT0 will trigger an erase. If NFC is enabled in the previous app, the UICR will be erased after flashing the OpenSK app.

jmichelp:
Closing this PR as it would need to be rebased and I included the changes as part of the kernel version update in PR #374 

I also updated our patch for UICR which now saves and restored the DFU bootloader settings when UICR is erased. I'll create a PR to upstream it.

PR #190, opened 2020-10-27
Add model for new store

ia0:


libraries/persistent_store/src/store.rs
```diff
@@ -60,4 +62,55 @@ impl From<StorageError> for StoreError {
...
+    pub used: usize,
+
+    /// How much of the metric can be used at most.
+    pub total: usize,
+}
+
+impl StoreRatio {
+    /// How much of the metric is remaining.
+    pub fn remaining(self) -> usize {
+        self.total - self.used
```
kaczmarczyck, line 86: What happens when `used` > `total`? It supposedly doesn't happen, but would be interesting to catch?

libraries/persistent_store/src/model.rs
```diff
@@ -0,0 +1,164 @@
...
+
+impl StoreModel {
+    /// Creates an empty model for a given storage configuration.
+    pub fn new(format: Format) -> StoreModel {
+        let map = HashMap::new();
+        StoreModel { map, format }
+    }
+
+    /// Returns the modeled content.
+    pub fn map(&self) -> &HashMap<usize, Box<[u8]>> {
```
kaczmarczyck, line 61: This name might be confusing, since it conflicts with the commonly used `map` for i.e. `Result`. You have lines like:

> self.map.values().map(|x| self.entry_size(x)).sum();

And it's not super hard to understand, but it wouldn't hurt to rename the variable and getter?

libraries/persistent_store/src/model.rs
```diff
@@ -0,0 +1,164 @@
...
+            return Err(StoreError::InvalidArgument);
+        }
+        // Fail if there is not enough capacity.
+        let capacity = match updates.len() {
+            0 => 0,
+            1 => match &updates[0] {
+                StoreUpdate::Insert { value, .. } => self.entry_size(value),
+                StoreUpdate::Remove { .. } => 0,
+            },
+            _ => 1 + updates.iter().map(|x| self.update_size(x)).sum::<usize>(),
```
kaczmarczyck, line 108: This refers to a special property of the store, maybe document / link / mention something?

kaczmarczyck (dismissed):

libraries/persistent_store/src/store.rs
ia0: Oops, I've made the fields public such that they can be accessed by the model, but the user should not be able to build such object. I fixed the visibility of the fields and added accessors such that the user cannot shot themselves in the foot. I also documented the invariant that `used <= total`.

Those kind of bugs are already caught by tests and fuzzing (both run in debug mode), which also catch some logical bugs too by regularly checking some internal invariants that are not always caught on Rust-level operations like here.

On a related note to arithmetic overflow, I'm looking at a possible intermediate PR to clarify the different integer types used in the crate (what is a natural number, what is a sequence of bits, etc) because you had a related question before (why use `u32` somewhere and `usize` otherwhere). In particular it would improve fuzzing because `usize` is variable-sized which is an issue because fuzzing is usually done with `usize = u64` while we are interested in `usize = u32`. This would catch additional bugs and seems cleaner because the behavior would depend less on the target architecture.

libraries/persistent_store/src/model.rs
ia0: Right, I didn't see possible confusion here. I renamed to `content` which also fits the current documentation.

libraries/persistent_store/src/model.rs
ia0: Done.

libraries/persistent_store/src/store.rs
kaczmarczyck: Sounds good concerning arithmetic overflow. I was curious about your choice of `-` instead of i.e. a `checked_sub`.

kaczmarczyck (approved):

libraries/persistent_store/src/store.rs
ia0: Yes, it's because the subtraction never fails (so `checked_sub` would introduce dead-code for the `None` branch), but turns out the question was useful to see that there was a visibility problem :-)

PR #192, opened 2020-10-29
Clarify different integer types

ia0:
This PR does the following things:
- Give incompatible representations for integers with different semantics (this hopefully further clarifies comments from previous PRs: [`usize` vs `u32`](https://github.com/google/OpenSK/pull/166#discussion_r496904845) and [`Position`](https://github.com/google/OpenSK/pull/172#discussion_r500997209)):
  - `usize` is used for natural numbers for the public API.
  - `Nat` is used internally for natural numbers (essentially a stable `usize`).
  - `Word` is used for sequences of bits representing words in flash.
  - `Position` is used for word positions in the virtual storage.
- Only use fixed size integers to preserve overflow behavior between targets (this improves fuzzing coverage).
- Use little-endian representation instead of native representation for `Word` (this improves fuzzing realism, although a no-op in current setup).

Alternatives:
- Run tests and fuzzing on 32-bits architecture (or some compatibility mode). This approach would have better readability than the current solution (less conversions at public API). However it would require additional setup and might not be viable long-term by restricting machines on which fuzzing is possible.
- Accept the behavior difference for tests and fuzzing. This approach would also have better readability. However checking for arithmetic overflow (and other `usize` related concerns like memory size) is more important.

### DIFF
--- a/libraries/persistent_store/src/format.rs
+++ b/libraries/persistent_store/src/format.rs
@@ -15,68 +15,104 @@
 // TODO(ia0): Remove when the module is used.
 #![allow(dead_code)]
 
-use crate::bitfield::*;
-use crate::{Storage, StorageIndex, StoreError, StoreResult};
+#[macro_use]
+mod bitfield;
+
+use self::bitfield::*;
+use crate::{usize_to_nat, Nat, Storage, StorageIndex, StoreError, StoreResult};
 use alloc::vec::Vec;
 use core::cmp::min;
+use core::convert::TryFrom;
 
+/// Internal representation of a word in flash.
+///
+/// Currently, the store only supports storages where a word is 32 bits.
 type WORD = u32;
+
+/// Abstract representation of a word in flash.
+///
+/// This type is kept abstract to avoid possible confusion with `Nat` if they happen to have the
+/// same representation. This is because they have different semantics, `Nat` represents natural
+/// numbers while `Word` represents sequences of bits (and thus has no arithmetic).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Word(WORD);
+
+/// Byte slice representation of a word in flash.
+///
+/// The slice is in little-endian representation.
+pub type WordSlice = [u8; core::mem::size_of::<WORD>()];
+
+impl Word {
+    /// Converts a byte slice into a word.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `slice.len() != WORD_SIZE`.
+    pub fn from_slice(slice: &[u8]) -> Word {
+        Word(WORD::from_le_bytes(<WordSlice>::try_from(slice).unwrap()))
+    }
+
+    /// Converts a word into a byte slice.
+    pub fn as_slice(self) -> WordSlice {
+        self.0.to_le_bytes()
+    }
+}
 
 /// Size of a word in bytes.
 ///
 /// Currently, the store only supports storages where a word is 4 bytes.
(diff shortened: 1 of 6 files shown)


libraries/persistent_store/src/format.rs
```diff
@@ -15,68 +15,104 @@
...
+/// Currently, the store only supports storages where a word is 32 bits.
 type WORD = u32;
 
+/// Abstract representation of a word in flash.
+///
+/// This type is kept abstract to avoid possible confusion with `Nat` if they happen to have the
+/// same representation. This is because they have different semantics, `Nat` represents natural
+/// numbers while `Word` represents sequences of bits (and thus has no arithmetic).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Word(WORD);
```
kaczmarczyck, line 38: Why is there a `WORD` and a `Word`? You don't need that for i.e. `Nat`. What if you just had a

> pub type Word = u32;

?

kaczmarczyck (approved):

libraries/persistent_store/src/format.rs
ia0: It's because I want to have `Word` (and `Position`) abstract to avoid confusing them by mistake with `u32` because many of the operations that are possible on `u32` should be forbidden on those types (for modules that only see the public interface).

I gave a name to the internal representation of `Word` (i.e. I did not write `pub struct Word(u32)`) because it's used a few times in this module and writing `WORD` documents better the intent than writing `u32`.

The reason I did not do it for `Nat` is because it's more convenient. Many of the operations on `u32` are also applicable to `Nat` so I don't need to reimplement them (like I did for `Position`). And the risk to not have it abstract is low, because there is no other "`u32`" in the crate with which `Nat` could be confused.

kaczmarczyck: Makes sense!

kaczmarczyck:
Since it's related to your discussion, I added @gendx as a reviewer. I don't mind you merging anytime.

ia0:
Ok, then I'll merge and send the next PR, which is finally the actual code of the library :D

libraries/persistent_store/src/lib.rs
```diff
@@ -362,3 +360,25 @@ pub use self::buffer::{BufferCorruptFunction, BufferOptions, BufferStorage};
...
+///
+/// In Rust natural numbers are represented as `usize`. However, internally we represent them as
+/// `u32`. This is done to preserve semantics across different targets. This is useful when tests
+/// run with `usize = u64` while the actual target has `usize = u32`.
+///
+/// To avoid too many conversions between `usize` and `Nat` which are necessary when interfacing
+/// with Rust, `usize` is used instead of `Nat` in code meant only for tests.
+///
+/// Currently, the store only supports targets with `usize = u32`.
+type Nat = u32;
```
gendx, line 374: Nit: I'd suggest gating this declaration by `cfg(target_pointer_width = "32")` to trigger an error if someone tries it with `usize != 64`.

ia0: I'm not sure I understand. The whole point of introducing this type is to be able to also run the code when `usize = u64` and still preserve the overflow conditions of `usize = u32`.

gendx: I was referring to the following piece of documentation, which is currently not enforced by code.

> Currently, the store only supports targets with `usize = u32`.

ia0: Ok, I think I got the idea. I introduced the correct gating with a comment to make it explicit.

PR #193, opened 2020-11-02
Add new store (without tests yet)

ia0:
Tests need the driver which will be the next PR.

### DIFF
--- a/libraries/persistent_store/src/store.rs
+++ b/libraries/persistent_store/src/store.rs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::format::*;
+use crate::format::{
+    is_erased, CompactInfo, Format, Header, InitInfo, InternalEntry, Padding, ParsedWord, Position,
+    Word, WordState,
+};
 #[cfg(feature = "std")]
 pub use crate::model::{StoreModel, StoreOperation};
 #[cfg(feature = "std")]
--- a/libraries/persistent_store/src/store.rs
+++ b/libraries/persistent_store/src/store.rs
@@ -12,8 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{Nat, StorageError};
+use crate::format::*;
+#[cfg(feature = "std")]
+pub use crate::model::{StoreModel, StoreOperation};
+#[cfg(feature = "std")]
+pub use crate::BufferStorage;
+use crate::{usize_to_nat, Nat, Storage, StorageError, StorageIndex};
 use alloc::vec::Vec;
+use core::cmp::{max, min, Ordering};
+#[cfg(feature = "std")]
+use std::collections::HashSet;
 
 /// Errors returned by store operations.
 #[derive(Debug, PartialEq, Eq)]
@@ -101,6 +109,38 @@ impl StoreRatio {
     }
 }
 
+/// Safe pointer to an entry.
+///
+/// A store handle stays valid at least until the next mutable operation. Store operations taking a
+/// handle as argument always verify that the handle is still valid.
+#[derive(Clone, Debug)]
+pub struct StoreHandle {
+    /// The key of the entry.
+    key: Nat,
+
+    /// The position of the entry.
+    pos: Position,
+
+    /// The length in bytes of the value.
+    len: Nat,
+}
+
+impl StoreHandle {
+    /// Returns the key of the entry.
+    pub fn get_key(&self) -> usize {
+        self.key as usize
+    }
+
+    /// Returns the value of the entry.
+    ///
+    /// # Errors
+    ///
+    /// Returns `InvalidArgument` if the entry has been deleted or compacted.
+    pub fn get_value<S: Storage>(&self, store: &Store<S>) -> StoreResult<Vec<u8>> {
+        store.get_value(self)
+    }
+}
+
 /// Represents an update to the store as part of a transaction.
 #[derive(Clone, Debug)]
 pub enum StoreUpdate {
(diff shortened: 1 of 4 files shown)


libraries/persistent_store/src/store.rs
```diff
@@ -128,3 +168,1089 @@ impl StoreUpdate {
...
+    fn wipe_span(&mut self, pos: Position, length: Nat) -> StoreResult<()> {
+        let length = (length * self.format.word_size()) as usize;
+        self.write_slice(pos, &vec![0x00; length])
+    }
+
+    /// Returns an extremum page.
+    ///
+    /// With `Greater` returns the most recent page (or the tail). With `Less` returns the oldest
+    /// page (or the head).
+    fn get_extremum_page_head(&self, ordering: Ordering) -> StoreResult<Position> {
```
kaczmarczyck, line 839: I'm wondering if you ever tested the storage implementation's speed vs. the old version. You are looping here, and this function gets called by `head()`, and `tail()`. So quite often, I think?

libraries/persistent_store/src/store.rs
```diff
@@ -128,3 +168,1089 @@ impl StoreUpdate {
...
+    }
+
+    /// Returns the number of words that can be written without compaction.
+    ///
+    /// This can be temporarily negative during compaction.
+    fn immediate_capacity(&self) -> StoreResult<isize> {
+        let tail = self.tail()?;
+        let end = self.head()? + self.format.virt_size();
+        Ok(if tail > end {
+            -((tail - end) as isize)
```
kaczmarczyck, line 862: This seems to be the only reason to use `isize`. No callers use the negative value, so you could also return 0 in this case. Would that make sense from what you expect from a "capacity"?

libraries/persistent_store/src/store.rs
```diff
@@ -128,3 +168,1089 @@ impl StoreUpdate {
...
+    User(Header),
+
+    /// Internal entry.
+    Internal(InternalEntry),
+
+    /// Partial user entry with non-erased footer.
+    ///
+    /// The fact that the footer is not erased and does not checksum, means that the header is
+    /// valid. In particular, the length is valid. We cannot wipe the entry because wiping the
+    /// footer may validate the checksum. So we mark the entry as deleted, which also wipes it.
```
kaczmarczyck, line 1171: Double-checking this comment: "We cannot wipe the entry" does not mean that data may not be wiped, but talks about meta-bytes?

libraries/persistent_store/src/store.rs
```diff
@@ -128,3 +168,1089 @@ impl StoreUpdate {
...
+        let mut pos = self.get_extremum_page_head(Ordering::Greater)?;
+        let end = pos.next_page(&self.format);
+        while pos < end {
+            let entry_pos = pos;
+            match self.parse_entry(&mut pos)? {
+                ParsedEntry::Internal(InternalEntry::Erase { .. }) => {
+                    return self.compact_erase(entry_pos)
+                }
+                ParsedEntry::Padding | ParsedEntry::User(_) => (),
+                _ => break,
```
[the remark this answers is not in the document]
ia0: Yes this is intended, the store format invariant implies that there is at most one internal entry (marker, erase, clear) in the whole storage (the remove internal entries only appear after a marker).

The order in which we recover stuff is important. We first recover compaction (because otherwise the virtual storage is not well defined), then entry-level operations (now that the virtual storage is defined). And for compaction recovery we do it in opposite order as compaction would occur (to detect and continue from the latest change).

libraries/persistent_store/src/store.rs
ia0: I did not do a proper comparison, I just have a latency example application for each but they don't really do the same (since it depends on the actual implementation). However, I would say both behave similarly (because all operations are linear and the slowest things are writes, erases, and syscalls, which are done a similar number of times).

The loop here is short, it's in the number of pages. But yes, the first optimization we should do if we want to improve latency and lose a bit of memory would be to store `head` and `tail` in memory. That was planned but I prefer to do it later and only if needed since it's an optimization.

libraries/persistent_store/src/store.rs
ia0: Good point, done.

The only problem I would see with returning 0 is that one cannot call `prepare` to finish a compaction cycle on a full store (very rare and probably meaningless circumstance).

libraries/persistent_store/src/store.rs
ia0: > "We cannot wipe the entry" does not mean that data may not be wiped

Yes

> but talks about meta-bytes?

Not sure what you mean. The explanation is that "we cannot wipe the entry" means "wiping the entry before marking the entry as deleted could introduce corruption, so we first have to mark the entry as deleted, then let the usual deletion mechanism wipe the entry."

libraries/persistent_store/src/store.rs
```diff
@@ -128,3 +168,1089 @@ impl StoreUpdate {
...
+        Ok(StoreRatio { used, total })
+    }
+
+    /// Applies a sequence of updates as a single transaction.
+    ///
+    /// # Errors
+    ///
+    /// Returns `InvalidArgument` in the following circumstances:
+    /// - There are too many updates.
+    /// - The updates overlap, i.e. their keys are not disjoint.
```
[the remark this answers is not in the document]
kaczmarczyck: I was wondering if optimizing out the duplicate keys could be a convenience feature. But if you think this should never happen, it make sense to require it.

libraries/persistent_store/src/store.rs
kaczmarczyck: The store wasn't really stress tested for lookup speed in v1 either, as far as I recall, so I guess that's fair.

libraries/persistent_store/src/store.rs
kaczmarczyck: "meta-bytes" was the marker, so my assumption was correct.

kaczmarczyck (approved):
I think my review isn't at the level of "I'm sure this should work."
I don't see any obvious mistakes, but I'd need more time to have a full mental model of all that is going on. So take my approval with that disclaimer, and I can take a deeper look later if you want.

ia0:
> I think my review isn't at the level of "I'm sure this should work."

I think that's normal for code reviews. This is not an audit. This can be done later once all the code (including fuzzing) is available for review. What matters is that you understand the overall design and it makes sense to you. The rest can be improved later.

> I'd need more time to have a full mental model of all that is going on

Yes, and also the code cannot really be read linearly (which is why I added the documentation in `lib.rs` to try to bootstrap, but this is very light). The best is to have a first glance over all the code, and only then try to dive in. Also, I don't expect humans to be able to ensure the code is correct in reasonable time. Trying to prove this formally would take multiple months (just writing down the design, invariant, and theorems on paper is already long and I tried to skip as much details as possible), which is why I went with fuzzing instead when reaching the limits of my intuition.

> I can take a deeper look later if you want

That would be welcome, but there is no urgency. In particular it would be best once fuzzing is submitted and you have more time. Because then it's very easy to change the code, run fuzzing, and see what breaks to understand how things fit together. There are many subtle parts that would be too long to document and would thus hurt actual readability. It's much simpler to understand them by trying out.

PR #199, opened 2020-11-04
NFC example application

MirnaMuhammad98:
An example application that uses the NFC driver interface added in PR #185 

It writes in the console the following:
* The bytes of the received packet
* A short description on what was the reply/transmission
* Time elapsed for the transmission, and the amount of bytes transmitted


examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+use libtock_drivers::console::Console;
+use libtock_drivers::nfc::NfcTag;
+use libtock_drivers::nfc::RecvOp;
+use libtock_drivers::result::FlexUnwrap;
+use libtock_drivers::result::TockError;
+use libtock_drivers::timer;
+use libtock_drivers::timer::Timer;
+use libtock_drivers::timer::Timestamp;
+
+#[allow(dead_code)]
```
kaczmarczyck, line 18: Why are these functions potentially dead code?

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Helper function to write on console the received packet.
+fn print_rx_buffer(buf: &mut [u8], amount: usize) {
```
kaczmarczyck, line 20: nit: I just noticed here, that `amount` is an usual name for a buffer size. It's `size` or `length` most of the time.

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+    if amount < 1 || amount > buf.len() {
```
kaczmarczyck, line 21: Since `amont` is `usize`, it's probably more readable if you check `== 0` instead of `< 1`.

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+        return;
+    }
+    let mut console = Console::new();
+    write!(console, "RX:").unwrap();
+    for byte in buf.iter().take(amount - 1) {
```
kaczmarczyck, line 26: For what you are trying to do here, there is [`split_last`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_last). To use that, the slice length would need to match the `amount`. So maybe it makes sense to pass in the sub-slice of the correct length? Then this function wouldn't look so C-style.

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+    .unwrap();
+    console.flush();
+}
+
+#[derive(PartialEq, Eq)]
+/// enum for reserving the NFC tag state.
+enum State {
+    Enabled,
+    Disabled,
+}
```
kaczmarczyck, line 61: Are you expecting more variants, or why not just use a simple `bool is_enabled`?

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+    writeln!(
+        console,
+        "Clock frequency: {} Hz",
+        timer.clock_frequency().hz()
+    )
+    .unwrap();
+
+    let mut state = State::Disabled;
+    // Variable to count the change in the tag's state
+    let mut state_change_cntr = 0;
```
kaczmarczyck, line 81: Google style usually spells words out, like `counter`.

run_desktop_tests.sh
```diff
@@ -31,7 +31,7 @@ cargo fmt --all -- --check
 cd ../..
 
 echo "Running Clippy lints..."
-cargo clippy --all-targets --features std -- -A clippy::new_without_default -D warnings
+cargo clippy --all-targets --features std --features with_nfc -- -A clippy::new_without_default -D warnings
```
gendx, line 34: I don't know what using `--features` twice does, but maybe you meant `--features std,with_nfc`?

run_desktop_tests.sh
```diff
@@ -53,7 +53,7 @@ cargo check --release --target=thumbv7em-none-eabi --features debug_ctap,with_ct
 cargo check --release --target=thumbv7em-none-eabi --features debug_ctap,with_ctap1,panic_console,debug_allocations,verbose
 
 echo "Checking that examples build properly..."
-cargo check --release --target=thumbv7em-none-eabi --examples
+cargo check --release --target=thumbv7em-none-eabi --examples --features with_nfc
```
gendx, line 56: Ideally, we should be able to check (1) all examples except for nfct_test without providing the feature (2) the nfct_test with providing the feature.

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
+#![no_std]
+
+extern crate alloc;
+extern crate lang_items;
+extern crate libtock_drivers;
+
+use core::fmt::Write;
+use libtock_core::result::CommandError;
+use libtock_drivers::console::Console;
+use libtock_drivers::nfc::NfcTag;
```
gendx, line 10: I'm not entirely sure what to do here. Can we gate the whole file by `cfg(feature = "with_nfc")` (and provide a dummy empty example when the feature is off)?

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+            }
+            State::Disabled => {
+                NfcTag::enable_emulation();
+                // Configure Type 4 tag
+                if NfcTag::configure(4) {
+                    state = State::Enabled;
+                }
+            }
+        }
+        state_change_cntr += 1;
```
kaczmarczyck, line 164: It's a little bit hard to catch that you `break` inside the loop. I started wondering why you don't just use a `for` loop. I think a general reason is the nesting depth. Could the function be more readable if you refactor out the inner part? It could return a `bool` where it currently breaks.

examples/nfct_test.rs
```diff
@@ -0,0 +1,170 @@
...
+                                    let mut reply = [rx_buf[0], 0x12, 0x90, 0x00,];
+                                    bench_transmit(&mut console, &timer, "TX: File Size", &mut reply);
+                                }
+                                0x12 => {
+                                    let mut reply = [rx_buf[0], 0xd1, 0x01, 0x0e, 0x55, 0x77, 0x77, 0x77, 0x2e, 0x6f, 0x70, 0x65, 0x6e, 0x73, 0x6b, 0x2e, 0x64, 0x65, 0x76, 0x90, 0x00,];
+                                    bench_transmit(&mut console, &timer, "TX: NDEF", &mut reply);
+                                }
+                                0x0f => {
+                                    let mut reply = [rx_buf[0], 0x00, 0x0f, 0x20, 0x00, 0x7f, 0x00, 0x7f, 0x04, 0x06, 0xe1, 0x04, 0x00, 0x7f, 0x00, 0x00, 0x90, 0x00,];
+                                    bench_transmit(&mut console, &timer, "TX: CC", &mut reply);
```
kaczmarczyck, line 133: GitHub doesn't correctly display this line on my screen. I assume it's too long :)

kaczmarczyck (changes requested):
There are errors in the code, and I added more general comments on code style and quality. I'll take another look when those are fixed!

examples/nfct_test.rs
MirnaMuhammad98: I wanted to use the match pattern with what I thought a more readable tag state.

examples/nfct_test.rs
MirnaMuhammad98: At the time I used the feature flag that hid all the nfc code, so the functions were left unused. And I'll reuse it again now as Guillaume suggested it would be better for cargo checks.

examples/nfct_test.rs
kaczmarczyck: Not wrong per se, but for me, the additional code for the `enum` is not worth it. Maybe others disagree, but it has other readability advantages. The nesting is a bit more shallow. `Enabled` vs `Disabled` are direct opposites, so you can reuse the `bool` idea, that people are used to.

examples/nfct_test.rs
MirnaMuhammad98: `amount` reflects the amount of data received and not the whole size of the buffer.

examples/nfct_test.rs
```diff
@@ -5,166 +5,194 @@ extern crate lang_items;
...
+
+    fn receive_packet(console: &mut Console, mut buf: &mut [u8; 256]) {
+        match NfcTag::receive(&mut buf) {
+            Ok(RecvOp {
+                recv_amount: amount,
+                ..
+            }) => {
+                if amount > 0 && amount <= buf.len() {
+                    print_rx_buffer(&mut buf[..amount]);
+                }
```
kaczmarczyck, line 76: FYI: If you want to handle the cases with different amounts too, you could move the `if` before `=>` (maybe even in 2 separate arms). Depends on if you want to silently ignore the 0 check.

If you do, I'd probably remove the check here, pass an empty slice into the function, and move the `console` declaration in `print_rx_buffer` inside the `if let`. That way, you have a more general function (because it can also handle empty slices). And a little bit easier code here.

examples/nfct_test.rs
```diff
@@ -5,166 +5,194 @@ extern crate lang_items;
...
+            "Clock frequency: {} Hz",
+            timer.clock_frequency().hz()
+        )
+        .unwrap();
+
+        let mut is_enabled = false;
+        let mut state_change_counter = 0;
+        loop {
+            match is_enabled {
+                true => {
```
kaczmarczyck: Please use `if` and `else` here.

examples/nfct_test.rs
```diff
@@ -5,166 +5,194 @@ extern crate lang_items;
...
                 }
             }
-        }
-        state_change_cntr += 1;
-        if state_change_cntr > 100 && state == State::Disabled {
-            break;
+            state_change_counter += 1;
+            if state_change_counter > 100 && is_enabled == false {
+                break;
+            }
```
kaczmarczyck, line 183: Checking into your control flow again, please let me know if this is correct. You are doing the following (shortened):

counter = 0;
> loop {
>   while !configure(4) {
>     enable_emulation();
>     counter += 1;
>   }
>   while !transmit_reply() && counter != 100 {
>     receive_packet();
>     counter += 1;
>   }
> }

If that's roughly the case, you could consider this alternative way to implement this flow. It moves the conditions for the loops into it's head. It reads more naturally than `while(True) break` constructions. Also, it is more obvious that you are alternating between 2 loops, starting with the disabled branch, without the outer `bool` to control that.

run_desktop_tests.sh
```diff
@@ -31,7 +31,7 @@ cargo fmt --all -- --check
 cd ../..
 
 echo "Running Clippy lints..."
-cargo clippy --all-targets --features std --features with_nfc -- -A clippy::new_without_default -D warnings
+cargo clippy --all-targets --features std -- -A clippy::new_without_default -D warnings
```
kaczmarczyck: Very nice, that you got to work around the import problem! Can you also add an additional check with NFC enabled here and below?

kaczmarczyck (commented):
Looks much more readable already! Left a few more comments.

examples/nfct_test.rs
MirnaMuhammad98: I don't understand where `if let` should go as I'm matching on some errors after the `Ok()`. I'm ignoring the 0 as I won't be printing anything inside of the function, so I don't want to make an unnecessary function call.

examples/nfct_test.rs
MirnaMuhammad98: I made that change in 0eaa2b5

examples/nfct_test.rs
kaczmarczyck: Ignoring is perfectly fine. What I'm saying is: Remove the `amount > 0` check here, and then change `print_rx_buffer` to gracefully handle the empty slice:

        if let Some((last, bytes)) = buf.split_last() {
          let mut console = Console::new();
          write!(console, "RX:").unwrap();
            for byte in bytes {
                write!(console, " {:02x?}", byte).unwrap();
            }
            writeln!(console, " {:02x?}", last).unwrap();
            console.flush();
        }

examples/nfct_test.rs
MirnaMuhammad98: The first `while` is somewhat correct to what should happen, I did something similar in the upcoming commit. The second loop won't work as the `transmit` check can't be evaluated before receiving first.

kaczmarczyck (approved):
Much nicer to look at and easier to understand!

kaczmarczyck (approved):

jmichelp (approved):

PR #200, opened 2020-11-06
Adds a UP check for device identification

kaczmarczyck:
The spec introduced a feature where empty PIN auth parameters are used to trigger user presence checks. This is already checked in the test tool, and now also reflected here.

- [x] Tests pass

#106

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -344,6 +344,33 @@ where
         }
     }
 
+    fn pin_uv_auth_precheck(
+        &mut self,
+        pin_uv_auth_param: &Option<Vec<u8>>,
+        pin_uv_auth_protocol: Option<u64>,
+        cid: ChannelID,
+    ) -> Result<(), Ctap2StatusCode> {
+        if let Some(auth_param) = &pin_uv_auth_param {
+            // This case was added in FIDO 2.1.
+            if auth_param.is_empty() {
+                (self.check_user_presence)(cid)?;
+                if self.persistent_store.pin_hash()?.is_none() {
+                    return Err(Ctap2StatusCode::CTAP2_ERR_PIN_NOT_SET);
+                } else {
+                    return Err(Ctap2StatusCode::CTAP2_ERR_PIN_INVALID);
+                }
+            }
+
+            match pin_uv_auth_protocol {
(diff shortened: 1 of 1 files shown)


src/ctap/mod.rs
```diff
@@ -342,6 +342,7 @@ where
         if let Some(auth_param) = &pin_uv_auth_param {
             // This case was added in FIDO 2.1.
             if auth_param.is_empty() {
+                let _ = (self.check_user_presence)(cid);
```
jmichelp, line 345: We do a UP check but we don't care about the result. Seems odd.

kaczmarczyck: The reason is that we return a specific error code right after.

> If authenticator supports either pinUvAuthToken or clientPin features and the platform sends a zero length pinUvAuthParam, wait for user touch and then return either CTAP2_ERR_PIN_NOT_SET if PIN is not set or CTAP2_ERR_PIN_INVALID if PIN has been set.

The reason for this behaviour:

> This is done for the case where multiple authenticators are attached to the platform and the platform wants to enforce pinUvAuthToken semantics, but the user has to select which authenticator to get the pinUvAuthToken from.

jmichelp: But if the user doesn't validate UP (i.e. timeout) the authenticator hasn't really been selected but yet will return one of the 2 error codes.
I would expect that the authenticator should return that UP timed out in such case. Maybe something to clarify on the spec.

kaczmarczyck: Discussion opened for clarification on the [spec](https://github.com/fido-alliance/fido-2-specs/issues/1092) (link for those with permissions). I'll keep this PR open until resolved.

kaczmarczyck: https://github.com/fido-alliance/fido-2-specs/pull/1096
Actually, the user touch shouldn't be ignored. Spec PR on the way. Changed in the next commit.

jmichelp:
Having a second look, the 2 snippets shown in the diff look identical. Maybe worth a code refactoring to avoid code duplication?

kaczmarczyck:
Done, moved to a function and made it a bit nicer.

jmichelp (approved):

PR #205, opened 2020-11-13
Add APDUs Parser

MirnaMuhammad98:
The target of this PR is to have a separate module for parsing APDUs.

Supported APDUs:
- [X] Short
- [ ] Extended Length

---
- [X] Tests pass

### DIFF
--- a/src/ctap/apdu.rs
+++ b/src/ctap/apdu.rs
@@ -3,6 +3,8 @@ use core::convert::TryFrom;
 
 type ByteArray = &'static [u8];
 
+const APDU_HEADER_LEN: usize = 4;
+
 #[cfg_attr(test, derive(Clone, Debug))]
 #[allow(non_camel_case_types)]
 #[derive(PartialEq)]
@@ -39,11 +41,11 @@ impl From<ApduStatusCode> for ByteArray {
     }
 }
 
-#[allow(non_camel_case_types, dead_code)]
-pub enum ApduIns {
-    SELECT = 0xA4,
-    READ_BINARY = 0xB0,
-    GET_RESPONSE = 0xC0,
+#[allow(dead_code)]
+pub enum ApduInstructions {
+    Select = 0xA4,
+    ReadBinary = 0xB0,
+    GetResponse = 0xC0,
 }
 
 #[cfg_attr(test, derive(Clone, Debug))]
@@ -68,6 +70,33 @@ impl From<&[u8]> for ApduHeader {
 }
 
 #[cfg_attr(test, derive(Clone, Debug))]
+#[derive(PartialEq)]
+/// The APDU cases
+pub enum Case {
+    Le,
+    LcData,
+    LcDataLe,
+    // TODO: More cases to add as extended length APDUs
+    // Le could be 2 or 3 Bytes
+}
+
+#[cfg_attr(test, derive(Clone, Debug))]
+#[allow(dead_code)]
+#[derive(PartialEq)]
+pub enum ApduType {
+    Instruction,
+    Short(Case),
+    Extended(Case),
+    Unknown,
+}
+
+impl Default for ApduType {
+    fn default() -> ApduType {
+        ApduType::Unknown
+    }
+}
+
+#[cfg_attr(test, derive(Clone, Debug))]
 #[allow(dead_code)]
 #[derive(Default, PartialEq)]
 pub struct APDU {
@@ -75,10 +104,8 @@ pub struct APDU {
     lc: u16,
     data: Vec<u8>,
     le: u32,
-    case_type: u8,
+    case_type: ApduType,
 }
-
-const APDU_HEADER_LEN: u8 = 4;
 
 impl TryFrom<&[u8]> for APDU {
     type Error = ApduStatusCode;
@@ -90,24 +117,23 @@ impl TryFrom<&[u8]> for APDU {
         //        +-----+-----+----+----+
         // header | CLA | INS | P1 | P2 |
         //        +-----+-----+----+----+
-        let (header, payload) = frame.split_at(APDU_HEADER_LEN as usize);
+        let (header, payload) = frame.split_at(APDU_HEADER_LEN);
 
         let mut apdu = APDU {
             header: header.into(),
             lc: 0x00,
             data: Vec::new(),
             le: 0x00,
-            case_type: 0x00,
+            case_type: ApduType::default(),
         };
 
         // case 1
         if payload.is_empty() {
-            apdu.case_type = 0x01;
+            apdu.case_type = ApduType::Instruction;
         } else {
             let byte_0 = payload[0];
-            // case 2S (Le)
             if payload.len() == 1 {
-                apdu.case_type = 0x02;
+                apdu.case_type = ApduType::Short(Case::Le);
                 apdu.le = if byte_0 == 0x00 {
                     // Ne = 256
                     0x100
@@ -115,15 +141,13 @@ impl TryFrom<&[u8]> for APDU {
                     byte_0.into()
                 }
             }
-            // case 3S (Lc + data)
             if payload.len() == (1 + byte_0) as usize && byte_0 != 0 {
-                apdu.case_type = 0x03;
+                apdu.case_type = ApduType::Short(Case::LcData);
                 apdu.lc = byte_0.into();
                 apdu.data = payload[1..].to_vec();
             }
-            // case 4S (Lc + data + Le)
             if payload.len() == (1 + byte_0 + 1) as usize && byte_0 != 0 {
-                apdu.case_type = 0x04;
+                apdu.case_type = ApduType::Short(Case::LcDataLe);
                 apdu.lc = byte_0.into();
                 apdu.data = payload[1..(payload.len() - 1)].to_vec();
                 apdu.le = (*payload.last().unwrap()).into();
@@ -133,7 +157,7 @@ impl TryFrom<&[u8]> for APDU {
             }
         }
         // TODO: Add extended length cases
-        if apdu.case_type == 0x00 {
+        if apdu.case_type == ApduType::default() {
             return Err(ApduStatusCode::SW_COND_USE_NOT_SATISFIED);
         }
         Ok(apdu)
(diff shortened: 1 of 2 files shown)


src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+#[allow(dead_code)]
+#[derive(Default, PartialEq)]
+pub struct ApduHeader {
+    cla: u8,
+    ins: u8,
+    p1: u8,
+    p2: u8,
+}
+
+impl From<&[u8]> for ApduHeader {
```
jmichelp, line 59: Considering the header has a fixed size, shouldn't we enforce it instead of using a slice?
Otherwise I guess the app will panic if the slice is less than 4 bytes long.

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+
+#[cfg_attr(test, derive(Clone, Debug))]
+#[allow(dead_code)]
+#[derive(Default, PartialEq)]
+pub struct APDU {
+    header: ApduHeader,
+    lc: u16,
+    data: Vec<u8>,
+    le: u32,
+    case_type: u8,
```
jmichelp, line 78: This is not part of an APDU. Can you describe what's the purpose of `case_type`?
Also by the name of it, it somehow screams to be an `enum` ;-)

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+        Ok(apdu)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    fn pass_frame(frame: &[u8]) -> Result<APDU, ApduStatusCode> {
```
jmichelp, line 147: What's the purpose of this one-liner instead of inlining?

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+}
+
+#[cfg_attr(test, derive(Clone, Debug))]
+#[allow(dead_code)]
+#[derive(Default, PartialEq)]
+pub struct APDU {
+    header: ApduHeader,
+    lc: u16,
```
jmichelp, line 75: Is `lc` actually needed once parsed? It's supposed to be equivalent to `data.len()` at the end otherwise the APDU is ill-formed, isn't it?

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+        };
+        assert_eq!(expected, response.unwrap());
+    }
+
+    #[test]
+    fn test_case_type_4_short() {
+        let frame: [u8; 13] = [
+            0x00, 0xa4, 0x04, 0x00, 0x07, 0xd2, 0x76, 0x00, 0x00, 0x85, 0x01, 0x01, 0xff,
+        ];
+        let payload = [0xd2, 0x76, 0x00, 0x00, 0x85, 0x01, 0x01];
```
jmichelp, line 237: I would rather compute `payload` as a slice of `frame` here. Otherwise it's not obvious that `payload` is actually already contained in `frame`. Same thing for the tests below.
Alternatively you may leave `payload` as is and construct `frame` using `Vec`:
```rust
let payload = [0xd2, 0x76, 0x00, 0x00, 0x85, 0x01, 0x01];
let frame = {
  let mut frame = vec![0x00, 0xa4, 0x04, 0x00, 0x07];
  frame.extend(&payload);
  let le = [0x01, 0xff];
  frame.extend(&le);
  frame
};
```

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+            ApduStatusCode::SW_INS_INVALID => b"\x6d\x00",
+            ApduStatusCode::SW_CLA_INVALID => b"\x6e\x00",
+            ApduStatusCode::SW_INTERNAL_EXCEPTION => b"\x6f\x00",
+        }
+    }
+}
+
+#[allow(non_camel_case_types, dead_code)]
+pub enum ApduIns {
+    SELECT = 0xA4,
```
jmichelp, line 44: nit: although it's common to write these in capital letters, I wonder if it wouldn't be more readable in Rust to convert them into camel case: `Select`, `ReadBinary`, `GetResponse`.

I defer to other reviewer's opinion on that.

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+        // case 1
+        if payload.is_empty() {
+            apdu.case_type = 0x01;
+        } else {
+            let byte_0 = payload[0];
+            // case 2S (Le)
+            if payload.len() == 1 {
+                apdu.case_type = 0x02;
+                apdu.le = if byte_0 == 0x00 {
+                    // Ne = 256
```
jmichelp, line 112: `Ne` or `Le`?

src/ctap/apdu.rs
MirnaMuhammad98: The purpose is to just look at that field in the future to know which type of the APDU it was, as there are multiple cases for the short APDUs, and as well as for the extended ones.

src/ctap/apdu.rs
MirnaMuhammad98: I was trying to incorporate the same notations as this [link](https://en.wikipedia.org/wiki/Smart_card_application_protocol_data_unit).

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+            ApduStatusCode::SW_SUCCESS => b"\x90\x00",
+            ApduStatusCode::SW_GET_RESPONSE => b"\x61\x00",
+            ApduStatusCode::SW_WRONG_DATA => b"\x6A\x80",
+            ApduStatusCode::SW_WRONG_LENGTH => b"\x67\x00",
+            ApduStatusCode::SW_COND_USE_NOT_SATISFIED => b"\x69\x85",
+            ApduStatusCode::SW_FILE_NOT_FOUND => b"\x6a\x82",
+            ApduStatusCode::SW_INCORRECT_P1P2 => b"\x6a\x86",
+            ApduStatusCode::SW_INS_INVALID => b"\x6d\x00",
+            ApduStatusCode::SW_CLA_INVALID => b"\x6e\x00",
+            ApduStatusCode::SW_INTERNAL_EXCEPTION => b"\x6f\x00",
```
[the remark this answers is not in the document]
MirnaMuhammad98: I have the assumption that the replies would be `ByteArray` it would be easier to copy over slices from one another if most things are already kept as byte arrays.

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+                apdu.case_type = 0x04;
+                apdu.lc = byte_0.into();
+                apdu.data = payload[1..(payload.len() - 1)].to_vec();
+                apdu.le = (*payload.last().unwrap()).into();
+                if apdu.le == 0x00 {
+                    apdu.le = 0x100;
+                }
+            }
+        }
+        // TODO: Add extended length cases
```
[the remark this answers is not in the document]
MirnaMuhammad98: Which NFC issue are you referring to?

src/ctap/apdu.rs
```diff
@@ -0,0 +1,303 @@
...
+    header: ApduHeader,
+    lc: u16,
+    data: Vec<u8>,
+    le: u32,
+    case_type: u8,
+}
+
+const APDU_HEADER_LEN: u8 = 4;
+
+impl TryFrom<&[u8]> for APDU {
```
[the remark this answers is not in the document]
MirnaMuhammad98: Nope, the received APDUs will vary in length.

src/ctap/apdu.rs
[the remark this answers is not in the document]
kaczmarczyck: Not super hard to convert, so this `enum` might actually merge with the status codes in CTAP1 later and move to `data_formats`. So okay to keep it like that until we know what to do with it exactly.

src/ctap/apdu.rs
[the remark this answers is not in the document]
kaczmarczyck: Hmm, the open NFC issue is a bit to broad for this piece of code. Also, is extended length compatible with our drivers?

src/ctap/apdu.rs
jmichelp: OK then it's fine

src/ctap/apdu.rs
jmichelp: Ok but that sounds more like something internal to parsing logic.
Unless if it needs to be known for answering, this shouldn't be an attribute of the parsed APDU.
If the APDUs have distinct properties we want to keep track of then we should either have:
1.  `case_type` being an enum
1. or we have different structs combined into an enum.

For now, if you can simply move `case_type` for `u8` to something like `enum ApduType`, that would help us later when we'll continue with this code.

src/ctap/apdu.rs
MirnaMuhammad98: So I shouldn't keep it even if it is to keep the structure of what the actual APDU looked like?

src/ctap/apdu.rs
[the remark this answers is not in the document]
MirnaMuhammad98: The existing [parser](https://github.com/google/OpenSK/blob/f204005ecaa3410fb03e0d04cd9c98b6f0950de7/src/ctap/ctap1.rs#L114) works only with extended length APDUs. So I think if this module would be used in the future either with NFC or USB, then it will support both APDU types 🤔

src/ctap/apdu.rs
jmichelp: Actually the code currently doesn't take `Lc` into account. So it's fine to keep it.
We'll add sanity checks later.

src/ctap/apdu.rs
MirnaMuhammad98: Enforcing the header length results in this
```
   --> src/ctap/apdu.rs:122:28
    |
122 |             header: header.into(),
    |                            ^^^^ the trait `std::convert::From<&[u8]>` is not implemented for `ctap::apdu::ApduHeader`
```

MirnaMuhammad98: @kaczmarczyck this is also something that you commented on.

src/ctap/apdu.rs
MirnaMuhammad98: because for example:
```
        let frame: [u8; 4] = [0x00, 0x12, 0x00, 0x80];
        let response = APDU::try_from(&frame);
```
This results in:
```
error[E0277]: the trait bound `ctap::apdu::APDU: std::convert::From<&[{integer}; 4]>` is not satisfied
the trait `std::convert::From<&[{integer}; 4]>` is not implemented for `ctap::apdu::APDU`
```

src/ctap/apdu.rs
kaczmarczyck: Yes, you'd have to call it with an array. `array_ref!` could help.

src/ctap/apdu.rs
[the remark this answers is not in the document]
kaczmarczyck: I think leaving it open is fine.

PR #206, opened 2020-11-13
Add histogram for fuzzing

ia0:
cc @mingxguo27 (same issue as before, can't add to reviewer)

As before, full code for context is here: https://github.com/ia0/OpenSK/tree/v2/libraries/persistent_store/fuzz.

### DIFF
--- a/libraries/persistent_store/fuzz/src/histogram.rs
+++ b/libraries/persistent_store/fuzz/src/histogram.rs
@@ -1,0 +1,100 @@
+// Copyright 2019-2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::num_bits;
+use std::collections::HashMap;
+
+/// Histogram with logarithmic buckets.
+///
+/// This is used to compute coverage statistics of the fuzzing runs of a corpus. This is not used
+/// during actual fuzzing, only when replaying a corpus to compute statistics.
+#[derive(Default)]
+pub struct Histogram {
+    /// Maps each bucket to its count.
+    ///
+    /// Buckets are numbers sharing the same highest bit. The first buckets are: only 0, only 1, 2
+    /// to 3, 4 to 7, 8 to 15. Buckets are identified by their lower-bound.
+    buckets: HashMap<usize, usize>,
+}
+
+impl Histogram {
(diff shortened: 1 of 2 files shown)


libraries/persistent_store/fuzz/src/histogram.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Increases the count of the bucket of an item.
+    ///
+    /// The bucket of `item` is the highest power of two, lower or equal to `item`. If `item` is
+    /// zero, then its bucket is also zero.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the item is too big, i.e. it uses its most significant bit.
+    pub fn add(&mut self, item: usize) {
+        assert!(item <= usize::max_value() / 2);
```
kaczmarczyck, line 41: Why is this a constraint? Why not allow one extra bucket and never panic?
`bucket_lim` is problematic with that extra bucket, but by returning `bucket_lim() - 1` instead, this could work.

ia0: It's because it never happens in practice and simplifies code (in `stats.rs`). The highest value we store is `50000`, so far below this limit (even on 32-bits machines). To understand the simplification we should first look at the alternative: use `bucket_max() -> Option<usize>` instead of `bucket_lim() -> usize`. (Note that the `bucket_lim - 1` would force signedness when there's no bucket and essentially end up to the same as current code where max bucket is `2^62`.) With `bucket_max`, one has to deal with `None` which adds one branch each time we use this function. This branch does not exist with `bucket_lim` because the code is polymorphic for when the statistics are empty or not.

ia0: The intuition behind this is actually quite common. It's the same as using `[min, lim)` instead of `[min, max]` because then one can use the same representation for both empty (`min = lim`) and non-empty (`min < lim`) intervals.

kaczmarczyck: Yeah, the extra `Option` might not be worth it. The intention for using `max + 1` is clear, I find that more natural (compare our discussion on the random `range(min,max)`).

kaczmarczyck (approved):

libraries/persistent_store/fuzz/src/histogram.rs
ia0: Just to be sure there's no confusion (since you said `lim - 1`previously), it's not really `max + 1` but `max << 1` or equivalently `pow2(log2(max) + 1)` in case you were in log2-space.

Yeah, for `range(min, max)` we use `max` instead of `lim` because ranges there cannot be empty, while here they can. So it's strictly better to use `max` there, while here it would require `Option`.

ia0: Actually realized I can just return the bit-width of the bucket since it's what I use in the end (I recompute the bucket value before display anyway). So I added the [commit](https://github.com/google/OpenSK/pull/209/commits/bbb73c77a8d7f61470e93f4092ba28fd68ea6845) to #209.

kaczmarczyck: Got it, will take a look.

PR #208, opened 2020-11-15
nano33ble support

lucadulgheru:
Hello guys! I've recently managed to make the OpenSK project work on the nano33ble and I am interested in adding it to the OpenSK repo. I've got a question, though. I've added the necessary linker file + additions to the main.rs file of the board (in order to support USB HID and a button) to the boards/nano33ble directory, but I am not sure how to "transfer" it to the Tock patch that OpenSK uses. For example, I would like to replace the third_party/tock/boards/nano33ble directory that OpenSK currently uses.

Additional tools needed in order to install OpenSK and TockOS on the board: **bossac**. Details on how to get it up and running are in the board's README file.

If there is something I'm missing or any adjustments are needed, please let me know!

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -26,7 +26,6 @@ import os
 import shutil
 import subprocess
 import sys
-
 import colorama
 from six.moves import input
 import tockloader
@@ -35,8 +34,8 @@ from tockloader import tbfh
 from tockloader import tockloader as loader
 from tockloader.exceptions import TockLoaderException
 
-PROGRAMMERS = frozenset(("jlink", "openocd", "pyocd",
-                        "nordicdfu", "none", "bossac"))
+PROGRAMMERS = frozenset(
+    ("jlink", "openocd", "pyocd", "nordicdfu", "none", "bossac"))
 
 # This structure allows us to support out-of-tree boards as well as (in the
 # future) more achitectures.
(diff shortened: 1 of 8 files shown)


deploy.py
```diff
@@ -664,6 +686,41 @@ def run(self):
...
+      if self.args.tockos:
+        # The Arduino Nano requires a double tap on the reset button
+        # in order to enter bootloader mode
+        # We must wait for the user to confirm that
+        # the board is ready to load something on it
+        info("""Double tap the reset button on your Arduino Nano board
+            in order to begin the flashing process""")
+        info("Press [ENTER] when ready.")
+        _ = input()
+        self.checked_command(["make", "program"], cwd=props.path)
```
jmichelp, line 701: Considering that at that point the kernel and the application are already compiled, why not writing both at the same time?
That's what we do for OpenOCD and DFU flashing by first calling
```py
self.create_hex_file(dest_file)
```
and then use the produced IntelHEX file to flash only once.

deploy.py
```diff
@@ -664,6 +686,41 @@ def run(self):
...
+      # Install the app using bossac
+      if self.args.application:
+        info("""Double tap the reset button on your Arduino Nano board
+            in order to load the application onto the board""")
+        info("Press [ENTER] when ready.")
+        _ = input()
+
+        self.checked_command(
+            [
+                "bossac",
```
jmichelp, line 712: Please use long options when available as they are easier to understand/review and (normally) more stable overtime than the short ones.

deploy.py
```diff
@@ -664,6 +686,41 @@ def run(self):
...
+                "-i",
+                "-e",
+                "-o",
+                "0x30000",
```
jmichelp, line 716: I assume this should then be `props.padding_address` instead.

jmichelp (commented):
Thanks for your contribution!

Usually we try to keep the amount of board we have to maintain as low as possible and upstream everything we can. What's the difference between the nano33ble board files you submitted as part of this PR and the ones directly available upstream (https://github.com/tock/tock/tree/master/boards/nano33ble)?
If it's just that the kernel we pin is too old but the files are identical then we'll soon update the kernel so it's fine to remove the files and only keep the changes in `deploy.py`.

deploy.py
lucadulgheru: Hello and thank you for the review! Yeah, this is a very good point, it would make more sense to create the hex file and flash only once. Regarding the changes, nano33ble does not support USB at all, so I've added that. Moreover, the board itself does not come with a button, which is why I've added the button to a certain GPIO (can change that, obviously). 
Compared to the upstream version, the nano33ble has some sensors which are not available on the tock version that OpenSK uses, which is why I've removed them for this scenario. (OpenSK does not require them anyways.)

So, in order to conclude, the board files which I've submitted have a different linker file (found some address mismatching when I tried to flash OpenSK, so I had to change that.), support for USB (which upstream does not have) and support for button(which upstream does not have)

deploy.py
lucadulgheru: Yeah, I should've thought about that. My bad... :smiley:

deploy.py
lucadulgheru: Yes, that's correct.

lucadulgheru:
Hey guys! Following @jmichelp 's suggestions, I've been stuck on making the hex file work for the `nano33ble`. I've tried to flash only the kernel using the hexfile, but no success either. I've set the kernel address to `0x10000` as stated in the linker file, built the hexfile, flashed it onto the board using `bossac` but it doesn't seem to "do" anything, as in not even Tock is being detected. I've had many attempts at making this work using different addresses, flashing the file at different addreses (even though it's kind of stupid, but I've tried anyways) with no success. Would you guys mind pointing me to the right direction? Or better yet, is there any other way to achieve this? (Obtain both Tock and the app itself in one file so we avoid the need of flashing twice onto the board - perhaps "concatenating" the two with a certain padding in between)

jmichelp:
> I've set the kernel address to `0x10000` as stated in the linker file

Have you also updated the `kernel_address` parameter of the board declared in `deploy.py`?
This should produce a valid HEX file and you should be able to flash it with `bossac` without specifying an offset (i.e. without `-o 0x30000`)

lucadulgheru:
Yeah, I was talking about the` kernel_address` from `deploy.py` . I've flashed with `bossac` exactly as you mentioned, without specifying an offset. Even with that, I still cannot get at least the kernel to show up, let alone the app itself.

kaczmarczyck:
No activity since last year. @jmichelp @lucadulgheru please reopen if you still want to work on nano33ble support!

PR #210, opened 2020-11-19
Add store fuzzing

ia0:
+Reviewer @mingxguo27

### DIFF
--- a/libraries/persistent_store/fuzz/src/store.rs
+++ b/libraries/persistent_store/fuzz/src/store.rs
@@ -1,0 +1,533 @@
+// Copyright 2019-2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::stats::{StatKey, Stats};
+use crate::Entropy;
+use persistent_store::{
+    BufferOptions, BufferStorage, Store, StoreDriver, StoreDriverOff, StoreDriverOn,
+    StoreInterruption, StoreInvariant, StoreOperation, StoreUpdate,
+};
+use rand_core::{RngCore, SeedableRng};
+use rand_pcg::Pcg32;
+use std::collections::HashMap;
+use std::convert::TryInto;
+
+// NOTE: We should be able to improve coverage by only checking the last operation. Because
+// operations before the last could be checked with a shorter entropy.
+
+/// Checks the store against a sequence of manipulations.
+///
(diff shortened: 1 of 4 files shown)


libraries/persistent_store/fuzz/src/store.rs
```diff
@@ -0,0 +1,533 @@
...
+
+/// Compact stack of bits.
+// NOTE: This would probably go away once the delay map is simplified.
+#[derive(Default, Clone, Debug)]
+struct BitStack {
+    /// Bits stored in little-endian (for bytes and bits).
+    data: Vec<u8>,
+
+    /// Number of bits stored.
+    len: usize,
```
kaczmarczyck, line 443: `len` is actually `len() % 8`, right? So the comment should be something like:

> Number of extra bits stored that didn't fill bytes.

libraries/persistent_store/fuzz/src/store.rs
```diff
@@ -0,0 +1,533 @@
...
+    bits.push(false);
+    bits.push(true);
+    assert_eq!(bits.pop(), Some(true));
+    assert_eq!(bits.pop(), Some(false));
+    assert_eq!(bits.pop(), None);
+
+    for i in 0..27 {
+        assert_eq!(bits.len(), i);
+        bits.push(true);
+    }
```
kaczmarczyck, line 532: Maybe `pop` them all back? Since you never `pop` over a byte border.

libraries/persistent_store/fuzz/src/store.rs
```diff
@@ -0,0 +1,533 @@
...
+        });
+        StoreInterruption { delay, corrupt }
+    }
+}
+
+/// The initial fuzzing mode.
+enum Init {
+    /// Fuzzing starts from a clean storage.
+    ///
+    /// All invariant are checked.
```
kaczmarczyck, line 398: nit: "invariants"

libraries/persistent_store/fuzz/src/store.rs
```diff
@@ -0,0 +1,533 @@
...
+    Clean,
+
+    /// Fuzzing starts from a dirty storage.
+    ///
+    /// Only crashing is checked.
+    Dirty,
+
+    /// Fuzzing starts from a simulated old storage.
+    ///
+    /// All invariant are checked.
```
kaczmarczyck, line 408: same nit: "invariants"

libraries/persistent_store/fuzz/src/store.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    /// Increments a counter if statistics are enabled.
+    fn increment(&mut self, key: StatKey) {
+        if self.stats.is_some() {
+            *self.counters.get_mut(&key).unwrap() += 1;
+        }
+    }
+
+    /// Initializes all counters if statistics are enabled.
+    fn init_counters(&mut self) {
```
kaczmarczyck, line 243: This is called once after `new`. Why not just call it inside of `new` so users can't forget?

libraries/persistent_store/fuzz/src/store.rs
ia0: Good point. Done.

kaczmarczyck (approved):

ia0:
Thanks for merging into master to avoid re-approval! Merging.

Issue #196, opened 2020-11-03
OpenSK passes the conformance tool


jmichelp:
This issue will be used to track the different runs of the official conformance tool from FIDO Alliance against commits of OpenSK.

jmichelp:
As of commit f475aa5 all tests are green and have been submitted as part of the certification process.

PR #212, opened 2020-11-20
Change response in HID_INIT for sync

kaczmarczyck:
Resolves an old TODO in HID to correctly respond to syncs.

- [x] Tests pass

### DIFF
--- a/src/ctap/hid/mod.rs
+++ b/src/ctap/hid/mod.rs
@@ -227,44 +227,35 @@ impl CtapHid {
                     }
                     // CTAP specification (version 20190130) section 8.1.9.1.3
                     CtapHid::COMMAND_INIT => {
-                        if cid == CtapHid::CHANNEL_BROADCAST {
-                            if message.payload.len() != 8 {
-                                return CtapHid::error_message(cid, CtapHid::ERR_INVALID_LEN);
-                            }
+                        if message.payload.len() != 8 {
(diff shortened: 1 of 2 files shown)


kaczmarczyck:
Blocking #196

jmichelp:
With this PR, how do we keep track of allocated channels and how do we "expire" them?
Because I could imagine that someone can "hijack" a channel, right?
Nothing seems to prevent someone from allocating a chosen channel which might already be in use.

BTW I think this behavior is legacy from U2F before 1.1 specification.

kaczmarczyck:
This PR doesn't change channel management, or behaviour on a receiving a packet. The only thing that should be different is the output, that changes from empty in some cases to always return a payload.

jmichelp:
It somehow does change it.
By returning a payload, you're implicitly allocating the channel (from the platform point of view). If the sent `cid` is not considered as allocated (i.e. if `self.is_allocated(cid)` is `false`), then we will reject following packets coming on this channel.

kaczmarczyck:
The check for `self.has_valid_channel` is done in both cases, so we only send this response if it comes from a valid channel.
Are there any other concerns about channel hijacking? I don't think this change facilitates them, because sending `INIT` with a channel ID other than broadcast should always just cancel the current sequence, and wouldn't happen by accident.

jmichelp (approved):

PR #218, opened 2020-11-23
Allow HMAC secret with rk=false

kaczmarczyck:
Blocks #196. Allows the `rk=false` option together with the HMAC-secret extension.

- [x] Tests pass

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -86,6 +86,7 @@ const USE_SIGNATURE_COUNTER: bool = true;
 // - (optional) 32 byte for HMAC-secret,
 // - 32 byte HMAC-SHA256 over everything else.
 pub const CREDENTIAL_ID_BASE_SIZE: usize = 112;
+pub const CREDENTIAL_ID_MAX_SIZE: usize = CREDENTIAL_ID_BASE_SIZE + 32;
 // Set this bit when checking user presence.
 const UP_FLAG: u8 = 0x01;
 // Set this bit when checking user verification.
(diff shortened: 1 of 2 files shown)


src/ctap/mod.rs
```diff
@@ -228,11 +235,15 @@ where
         credential_id: Vec<u8>,
         rp_id_hash: &[u8],
     ) -> Result<Option<PublicKeyCredentialSource>, Ctap2StatusCode> {
-        if credential_id.len() != ENCRYPTED_CREDENTIAL_ID_SIZE {
+        let has_cred_random = if credential_id.len() == CREDENTIAL_ID_BASE_SIZE {
```
jmichelp, line 238: nit: It looks odd to me to see 3 branches and only 2 resulting in an assigned value while the other one will early stop the function.

jmichelp (approved):

src/ctap/mod.rs
kaczmarczyck: Good point, should be more readable with a `match`.

src/ctap/mod.rs
```diff
@@ -203,14 +206,19 @@ where
...
         blocks[2].copy_from_slice(&application[..16]);
         blocks[3].copy_from_slice(&application[16..]);
+        if let Some(cred_random) = cred_random {
+            blocks[4].copy_from_slice(&cred_random[..16]);
+            blocks[5].copy_from_slice(&cred_random[16..]);
+        }
         cbc_encrypt(&aes_enc_key, iv, &mut blocks);
 
-        let mut encrypted_id = Vec::with_capacity(ENCRYPTED_CREDENTIAL_ID_SIZE);
+        let mut encrypted_id = Vec::with_capacity(16 * (block_len + 3));
```
jmichelp, line 221: Something to keep in mind for a later improvement: maybe we should replace all occurence of `16` constant.
I'm thinking like having something like a `CredentialWrapper` which would provide `encrypt`, `decrypt`, and `block_size` methods for example. This would allow to change the way we wrap them in a more flexible way.

jmichelp: Otherwise if we have the constant already defined in some way in the crypto we could import this constant, replace all `16` by this constant and compute the `block_len` above with proper maths.

jmichelp (approved):

src/ctap/mod.rs
kaczmarczyck: True, this function has never been really nice. I will have to touch it again for 2.1 for a similar modification. I understand moving this change to a later PR is okay?

jmichelp:
Yes. That's why I approved. This can wait

PR #219, opened 2020-11-23
Adding metadata file used for certification.

jmichelp:


metadata/metadata.json
```diff
@@ -0,0 +1,47 @@
...
+   "keyProtection" : 1,
+   "attestationRootCertificates" : [
+   ],
+   "aaguid" : "664d9f67-84a2-412a-9ff7-b4f7d8ee6d05",
+   "publicKeyAlgAndEncoding" : 260,
+   "protocolFamily" : "fido2",
+   "upv" : [
+      {
+         "major" : 1,
+         "minor" : 0
```
kaczmarczyck, line 12: If this is what I think it is, we report different numbers in HID:
`https://github.com/google/OpenSK/blob/3867497c4bab9c2772ba3a59fc4f1675fd83ce0e/src/ctap/hid/mod.rs#L122`

We report serial number `v0.1` over HID to the test tool for some reason. Maybe that's somewhere in the driver?

metadata/metadata.json
```diff
@@ -0,0 +1,47 @@
...
+   "description" : "OpenSK authenticator",
+   "authenticatorVersion" : 1,
+   "isSecondFactorOnly" : false,
+   "userVerificationDetails" : [
+      [
+         {
+            "userVerification" : 1
+         },
+         {
+         	"userVerification" : 4
```
kaczmarczyck, line 37: nit: indentation differs

metadata/metadata.json
jmichelp: It's the version of the supported upv protocol.
1.2 means UAF, 1.0 means U2F, etc.

See this [link](https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-metadata-statement-v2.0-rd-20180702.html#dictionary-metadatastatement-members).

The shown [FIDO2 Example](https://fidoalliance.org/specs/fido-v2.0-rd-20180702/fido-metadata-statement-v2.0-rd-20180702.html#fido2-example) there also reports 1.0

metadata/metadata.json
jmichelp: Nice catch. Tab vs. spaces. That's the drawback of having edited this file with different editors :)

metadata/metadata.json
kaczmarczyck: Oh okay, then my comment is unrelated. We might still want to change the version reported in HID. Not in this PR, but in general.

kaczmarczyck (approved):

Issue #195, opened 2020-11-03
Add vendor commands to inject crypto materials


jmichelp:
Once the crypto materials are removed from the firmware (i.e. when issue #194 is fixed), we need a way to inject these values into the FOB.
The easiest way to do so is by adding vendor specific commands and provide a command line tool to run them.

For security reasons, theses commands must only be executed once and the firmware should make sure to never overwrite a value that has already been programmed.

ia0:
Some code pointers regarding storage:
- Those [lines](https://github.com/google/OpenSK/blob/c9fc621f597e987035634aadac5c66f3e0b3fdcc/src/ctap/storage.rs#L234-L247) should be removed. Thus, the `key_material` module can be removed too.
- The vendor commands should use the `set_attestation_private_key`, `set_attestation_certificate`, and `set_aaguid` methods of the storage. If it is important that those are set at most once, I can write a PR for just that part since it's purely storage related.

jmichelp:
AAGUID will not be set through these commands. I'd like it to still be part of the firmware at the moment.
The certification is done for a specific AAGUID and a root CA so this makes sense to have AAGUID in the firmware. Then if you inject crypto materials that are signed by the corresponding CA you run a certified key otherwise not.

jmichelp:
List of pull requests required to achieve the goal:
- [ ] CTAP1/U2F accesses data through the store
- [ ] Vendor command is add to configure the authenticator (attestation crypto materials, JTAG lockdown)
- [ ] Capsule and syscalls are added to enable JTAG lockdown
- [ ] Configure tool to inject crypto materials in the authenticator
- [ ] Cleanup in the build script (e.g. removing openssl code/dependancy)

ia0:
Hopefully the first point is already fixed by #112. If it's not the case, let me know. I hopefully should be able to fix it.

kaczmarczyck:
We missed usage in [CTAP1](https://github.com/google/OpenSK/blob/master/src/ctap/ctap1.rs).

ia0:
I see @jmichelp just sent #222 for that :-) I don't know how I missed that usage.

PR #222, opened 2020-11-25
Ensure CTAP1 uses the storage to access attestation material

jmichelp:
Part of the changes needed for #195 

This also prepares removing the crypto material from the firmware by:
* defining `attestation_private_key` as sensitive in the storage
* ensuring that the certificate and the private key are one-time programmable

- [x] Tests pass

### DIFF
--- a/src/ctap/ctap1.rs
+++ b/src/ctap/ctap1.rs
@@ -35,6 +35,8 @@ pub enum Ctap1StatusCode {
     SW_WRONG_LENGTH = 0x6700,
     SW_CLA_NOT_SUPPORTED = 0x6E00,
     SW_INS_NOT_SUPPORTED = 0x6D00,
+    SW_MEMERR = 0x6501,
+    SW_COMMAND_ABORTED = 0x6F00,
     SW_VENDOR_KEY_HANDLE_TOO_LONG = 0xF000,
 }
 
@@ -49,6 +51,8 @@ impl TryFrom<u16> for Ctap1StatusCode {
             0x6700 => Ok(Ctap1StatusCode::SW_WRONG_LENGTH),
             0x6E00 => Ok(Ctap1StatusCode::SW_CLA_NOT_SUPPORTED),
             0x6D00 => Ok(Ctap1StatusCode::SW_INS_NOT_SUPPORTED),
+            0x6501 => Ok(Ctap1StatusCode::SW_MEMERR),
+            0x6F00 => Ok(Ctap1StatusCode::SW_COMMAND_ABORTED),
             0xF000 => Ok(Ctap1StatusCode::SW_VENDOR_KEY_HANDLE_TOO_LONG),
             _ => Err(()),
         }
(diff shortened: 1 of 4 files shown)


src/ctap/storage.rs
```diff
@@ -525,29 +530,33 @@ impl PersistentStore {
...
     pub fn attestation_private_key(
         &self,
-    ) -> Result<Option<&[u8; ATTESTATION_PRIVATE_KEY_LENGTH]>, Ctap2StatusCode> {
+    ) -> Result<Option<&[u8; key_material::ATTESTATION_PRIVATE_KEY_LENGTH]>, Ctap2StatusCode> {
         let data = match self.store.find_one(&Key::AttestationPrivateKey) {
             None => return Ok(None),
             Some((_, entry)) => entry.data,
         };
-        if data.len() != ATTESTATION_PRIVATE_KEY_LENGTH {
+        if data.len() != key_material::ATTESTATION_PRIVATE_KEY_LENGTH {
```
ia0, line 538: Why not import those constants? The name seems unambiguous enough that `key_material` is not needed.

src/ctap/storage.rs
```diff
@@ -231,22 +229,29 @@ impl PersistentStore {
...
+        #[cfg(not(any(test, feature = "ram_storage")))]
+        self.load_attestation_from_firmware();
+
         if self.store.find_one(&Key::Aaguid).is_none() {
             self.set_aaguid(key_material::AAGUID).unwrap();
         }
     }
 
+    // TODO(jmichel): remove this function when vendor command is in place.
+    #[cfg(not(any(test, feature = "ram_storage")))]
```
ia0, line 242: Just wondering, but do we still need this `ram_storage` feature? Asking because it uses "testing" code. So we could have a custom module (different from the one we use in testing) if we want to reduce memory footprint.

ia0 (approved):

src/ctap/storage.rs
jmichelp: I could. Usually for constants I like knowing where they come from without looking at the top :D

src/ctap/storage.rs
jmichelp: I don't know. I was simply copying the conditions used above to ensure all the test were working as expected.

src/ctap/ctap1.rs
```diff
@@ -295,14 +294,22 @@ impl Ctap1Command {
...
-        let mut response =
-            Vec::with_capacity(105 + key_handle.len() + ATTESTATION_CERTIFICATE.len());
+        let certificate = match ctap_state.persistent_store.attestation_certificate() {
+            Ok(Some(value)) => value,
+            _ => return Err(Ctap1StatusCode::SW_INS_NOT_SUPPORTED),
+        };
+        let private_key = match ctap_state.persistent_store.attestation_private_key() {
+            Ok(Some(value)) => value,
+            _ => return Err(Ctap1StatusCode::SW_INS_NOT_SUPPORTED),
+        };
```
kaczmarczyck, line 304: The `Err` and `None` are technically 2 different errors. The first is a storage error and we might want to use something vendor specific for that, to later translate it to `CTAP2_ERR_VENDOR_INTERNAL_ERROR`.

So how about
> let certificate = ctap_state.persistent_store.attestation_certificate()
 .map_err(Ctap1StatusCode::SW_VENDOR_INTERNAL_ERROR)?
.ok_or(Ctap1StatusCode::SW_INS_NOT_SUPPORTED)?;

src/ctap/storage.rs
ia0: Ok there's 2 questions then :-)

1. Do we still need the `ram_storage` feature in general? Unrelated to this PR.
2. When do we want to load the baked-in values? If it's in prod keys, then we should just do `not(test)`

src/ctap/storage.rs
```diff
@@ -231,22 +229,29 @@ impl PersistentStore {
...
+    fn load_attestation_from_firmware(&mut self) {
```
kaczmarczyck, line 243: Are we loading an "attestation" or the data used for attestation? Maybe rename?

src/ctap/storage.rs
ia0: Ah ok, it's not too disambiguate but just to underline the definition origin.

ia0 (approved):

src/ctap/ctap1.rs
jmichelp: I don't think CTAP1 has vendor specific error codes. They are inherited from ISO7816.
I could use 0x6387 which means "Non-volatile memory is not available". WDYT?

src/ctap/storage.rs
jmichelp: Done

src/ctap/ctap1.rs
kaczmarczyck: There is a `SW_VENDOR_KEY_HANDLE_TOO_LONG` that is currently misused also for a storage error in line 292 (currently 291). I think changing yours and that one to `0x6F00` would make most sense?

jmichelp: Done for all 3.
I'm tempted to use `0x6387` instead of `SW_INS_NOT_SUPPORTED`.
First it allows us to distinguish between a key that is missing attestation crypto material and a storage failure. But also because we do support this instruction, it's just that the data is missing in the storage.

jmichelp: So, I found that `0x6501` is standard code for memory failure. So I mapped storage errors to `0x6501` and missing entries to `0x6f00` (command aborted)

kaczmarczyck (approved):

PR #223, opened 2020-11-26
GetNextAssertion

### LINKED ISSUE #196: OpenSK passes the conformance tool

kaczmarczyck:
Fixes #196.

This PR implements the GetNextAssertion. This includes saving the command state in GetAssertion, and discarding it on timeout or intersecting commands. The correct behaviour for timeouts landed in CTAP 2.1, and should be backwards compatible to 2.0.

Added @jmichelp for the CTAP changes and double-checking with the conformance tool.
Added @ia0 for reviewing the storage changes.

- [x] Tests pass

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -33,9 +33,9 @@ use self::command::{
 #[cfg(feature = "with_ctap2_1")]
 use self::data_formats::AuthenticatorTransport;
 use self::data_formats::{
-    CredentialProtectionPolicy, PackedAttestationStatement, PublicKeyCredentialDescriptor,
-    PublicKeyCredentialParameter, PublicKeyCredentialSource, PublicKeyCredentialType,
-    PublicKeyCredentialUserEntity, SignatureAlgorithm,
+    CredentialProtectionPolicy, GetAssertionHmacSecretInput, PackedAttestationStatement,
+    PublicKeyCredentialDescriptor, PublicKeyCredentialParameter, PublicKeyCredentialSource,
+    PublicKeyCredentialType, PublicKeyCredentialUserEntity, SignatureAlgorithm,
 };
 use self::hid::ChannelID;
 #[cfg(feature = "with_ctap2_1")]
@@ -133,6 +133,14 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
     }
 }
 
+#[derive(Clone)]
+struct AssertionInput {
+    client_data_hash: Vec<u8>,
+    auth_data: Vec<u8>,
+    uv: bool,
+    hmac_secret_input: Option<GetAssertionHmacSecretInput>,
+}
+
 // This struct currently holds all state, not only the persistent memory. The persistent members are
 // in the persistent store field.
 pub struct CtapState<'a, R: Rng256, CheckUserPresence: Fn(ChannelID) -> Result<(), Ctap2StatusCode>>
@@ -147,6 +155,8 @@ pub struct CtapState<'a, R: Rng256, CheckUserPresence: Fn(ChannelID) -> Result<(
     accepts_reset: bool,
     #[cfg(feature = "with_ctap1")]
     pub u2f_up_state: U2fUserPresenceState,
+    next_credentials: Vec<PublicKeyCredentialSource>,
+    next_assertion_input: Option<AssertionInput>,
 }
 
 impl<'a, R, CheckUserPresence> CtapState<'a, R, CheckUserPresence>
@@ -173,6 +183,8 @@ where
                 U2F_UP_PROMPT_TIMEOUT,
                 Duration::from_ms(TOUCH_TIMEOUT_MS),
             ),
+            next_credentials: vec![],
+            next_assertion_input: None,
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/ctap/mod.rs
```diff
@@ -167,18 +189,26 @@ where
...
 
-    pub fn check_disable_reset(&mut self, timestamp: Timestamp<isize>) {
-        if timestamp - Timestamp::<isize>::from_ms(0) > Duration::from_ms(RESET_TIMEOUT_MS) {
-            self.accepts_reset = false;
+    pub fn update_command_permission(&mut self, now: ClockValue) {
+        self.stateful_command_permission = self.stateful_command_permission.check_expiration(now);
+    }
+
+    fn check_command_permission(&mut self, now: ClockValue) -> Result<(), Ctap2StatusCode> {
+        self.stateful_command_permission = self.stateful_command_permission.check_expiration(now);
```
jmichelp, line 207: why not calling `self.update_command_permission(now);` here?

src/ctap/mod.rs
```diff
@@ -307,20 +343,35 @@ where
...
                     }
+                    Command::AuthenticatorGetNextAssertion => self.process_get_next_assertion(now),
                     Command::AuthenticatorGetInfo => self.process_get_info(),
                     Command::AuthenticatorClientPin(params) => self.process_client_pin(params),
-                    Command::AuthenticatorReset => self.process_reset(cid),
+                    Command::AuthenticatorReset => self.process_reset(cid, now),
                     #[cfg(feature = "with_ctap2_1")]
                     Command::AuthenticatorSelection => self.process_selection(cid),
-                    // TODO(kaczmarczyck) implement GetNextAssertion and FIDO 2.1 commands
-                    _ => self.process_unknown_command(),
```
jmichelp, line 323: Why removing this?
If the command is unsupported we must reply.

kaczmarczyck: We do return an error_code in `Command::deserialize`. We needed `process_unknown_command` before when we correctly parsed the inputs of some commands, but didn't actually implement them. From now on, I don't expect that we will have a command that we parse, but don't handle here.

src/ctap/mod.rs

ia0 (dismissed):
LGTM but I have a few questions regarding the semantics of `creation_order` that might require change:
- Should different credentials have different creation order? Currently all existing credentials have a creation order of 0.
- Should the order actually be creation order time? Or would any stable order for the same set of credentials be enough? In store v2, credentials have unique stable identifiers.
- Should different credentials across time have different creation order? Currently adding a credential, removing it, then adding a new one, would share the same creation order.
- (depends on previous questions but) Was a monotonic creation order counter considered?

kaczmarczyck:
To answer your questions:
- The `creation_order` is used for sorting, so non-uniqueness means we can't correctly sort. For existing credentials, we don't know their order of creation, we only know that they are earlier.
- The exact clock value is not important, but we need to know the order in which they are created. So a unique identifier isn't good enough.
- Removed credentials can never be discovered in `GetAssertion`, and we can still sort all found credentials.
- I did consider monotonic counter. It shouldn't be a big difference: We store one more integer, and it's a bit easier to query, since we don't have to compute a max over all credentials. On the other hand, we technically run out of integers a little bit faster, since there is a possibility that removing credentials through credential management (2.1 command) frees up some numbers, and frequent creation and deletion doesn't hurt that much.

We could even filter by `rp_id` for these integers, since we will only have to sort credentials of the same RP ID. Makes caching harder, but keeps numbers lower. Probably no big difference in practise though?

src/ctap/mod.rs
jmichelp: I see. Then we're good.

jmichelp (approved):

src/ctap/mod.rs
```diff
@@ -619,92 +756,74 @@ where
...
-
+        // This check comes before CTAP2_ERR_NO_CREDENTIALS in CTAP 2.0.
+        // For CTAP 2.1, it was moved to a later protocol step.
         if options.up {
             (self.check_user_presence)(cid)?;
         }
 
-        self.increment_global_signature_counter()?;
+        let (credential, remaining_credentials) = credentials
+            .split_last()
```
ia0, line 785: This seems costly. You convert back the remaining credentials to a `Vec` later. Why not use `pop` here?

ia0:
Ok I think I see. Let me try to recap and if it's good then it's good on my side (modulo the small performance issue).

- We just need to sort credentials with the same RP ID by creation order (updating a credential is like a creation, it just happens that it also deletes the old credential).
- We actually sort credentials by creation order regardless of RP ID.
- We just need to sort credentials at a given point in time (because we can't break a chain of GetNext without "resetting" it, so in particular we can't change the credential states within a loop of GetNext).
- The creation order is just an time-dependent injection of the actual creation time of credentials into `u64` preserving order.

kaczmarczyck:
How to update a credential is a good question, and the specification does not explicitly mention how this case is handled with respect to ordering. It might be more in the spirit of the specification to bump, and that's what's happening.

So: Yes to all points in your summary. Do you have an opinion about this one?

> We actually sort credentials by creation order regardless of RP ID.

We currently just `+ 1` the current maximum, I'll make it a more explicit `wrapping_add`. That would mean that ordering is still correct for all credentials created between to overflows. And we can still cache the current maximum if we ever feel like it.

ia0:
I don't have opinion about RP ID extra filtering. The 2 sides I see are:
- Our current solution (order all credentials) seems more robust to future changes (in case we can list all credentials in creation order).
- Extra filtering seems more robust regarding wraparounds (not sure it's an actual concern).

Good point about making the wrapping explicit. This could have failed in tests or fuzzing, while working correctly in prod.

Also I forgot to add to what I understand that we return in reverse creation order (most recent first). It doesn't seem to impact the choices we have though.

kaczmarczyck:
Yes to most recent first. Background is probably that the browser will show you a list of user names and you choose which one you'd like. And you find the most interesting ones easiest.

I also like the current version best, so we're good.

src/ctap/mod.rs
kaczmarczyck: Leftover from an old revision. Now using `pop`, and also moving the credential later to avoid a few `clone`s later.

jmichelp (approved):

PR #224, opened 2020-11-27
Adds user data to stored credentials

kaczmarczyck:
Blocks #196 . And hopefully fixes it this time.

Returning user data is mandatory, so we added it to the store. We deviate from `WebAuthn`'s `CredentialSource` now, so I renamed `OtherUi` to what we actually used it for, the `DisplayName` in user. This should still be backwards compatible with the store, just adding @ia0 for checking.

- [x] Tests pass


ia0 (commented):
Just checked `data_formats.rs` and `storage.rs`. Yes, this change is backward (new fields are `None` and `other_ui` is the same as `display_name`) and forward compatible (new fields are ignored and `display_name` is the same as `other_ui`).

jmichelp (approved):

PR #226, opened 2020-11-30
Add debug helper for fuzzing

ia0:
This permits to debug a single artifact or get statistics about a corpus.


kaczmarczyck (approved):
The file name actually confused me a bit at first, because there is a `store.rs` already somewhere. Since it's a `main`, maybe name the file after an action?

ia0:
Good point. I'll maybe split the fuzz targets in 3, but I'm not sure yet. I would use action names.

The reason why it's called "store" currently is such that to debug `cargo fuzz run X` you do `cd fuzz && cargo run --example=X -- artifact`. For now `X` is "store".

PR #227, opened 2020-11-30
Add extended APDU parser

krkhan:
This builds on #205 to add a parser for extended APDUs.


src/ctap/apdu.rs
```diff
@@ -302,7 +343,56 @@ mod test {
...
+            0xb0, 0x1b, 0x8e, 0xbc, 0xf0, 0x2d, 0xc9, 0x99, 0x8d, 0x6f, 0xe4, 0x8a, 0xb2, 0x70,
+            0x9a, 0x70, 0x3a, 0x27, 0x71, 0x88, 0x3c, 0x75, 0x30, 0x16, 0xfb, 0x02, 0x11, 0x4d,
+            0x30, 0x54, 0x6c, 0x4e, 0x8c, 0x76, 0xb2, 0xf0, 0xa8, 0x4e, 0xd6, 0x90, 0xe4, 0x40,
+            0x25, 0x6a, 0xdd, 0x64, 0x63, 0x3e, 0x83, 0x4f, 0x8b, 0x25, 0xcf, 0x88, 0x68, 0x80,
+            0x01, 0x07, 0xdb, 0xc8, 0x64, 0xf7, 0xca, 0x4f, 0xd1, 0xc7, 0x95, 0x7c, 0xe8, 0x45,
+            0xbc, 0xda, 0xd4, 0xef, 0x45, 0x63, 0x5a, 0x7a, 0x65, 0x3f, 0xaa, 0x22, 0x67, 0xe7,
+            0x8a, 0xf2, 0x5f, 0xe8, 0x59, 0x2e, 0x0b, 0xc6, 0x85, 0xc6, 0xf7, 0x0e, 0x9e, 0xdb,
+            0xb6, 0x2b, 0x00, 0x00,
+        ];
+        let payload = [
```
jmichelp, line 363: Should we define payload as a slice of frame to avoid duplicating things?

```rust
let payload = frame[7..frame.len() - 2];
```

Or simply use a slice of `frame` when creating the `expected` value should also be fine.

.gitmodules
```diff
@@ -1,6 +1,8 @@
 [submodule "third_party/libtock-rs"]
 	path = third_party/libtock-rs
 	url = https://github.com/tock/libtock-rs
+	ignore = dirty
```
kaczmarczyck, line 4: Why is this necessary now, and wasn't before?

src/ctap/apdu.rs
```diff
@@ -58,8 +59,8 @@ pub struct ApduHeader {
     p2: u8,
 }
 
-impl From<&[u8]> for ApduHeader {
-    fn from(header: &[u8]) -> Self {
+impl From<&[u8; 4]> for ApduHeader {
+    fn from(header: &[u8; 4]) -> Self {
```
kaczmarczyck, line 63: This `4` should be `APDU_HEADER_LEN` here and below.

src/ctap/apdu.rs
```diff
@@ -120,20 +124,23 @@ impl TryFrom<&[u8]> for APDU {
...
             data: Vec::new(),
             le: 0x00,
             case_type: ApduType::default(),
         };
 
-        // case 1
         if payload.is_empty() {
+            // Lc is zero-bytes in length
             apdu.case_type = ApduType::Instruction;
         } else {
```
kaczmarczyck, line 137: General question, since this is a pretty big if-else-construction. The structure is
- `case_type: ApduType::default()`
- lots of `if-else`
- check `case_type`, return error if `Unknown`
- return `apdu`

Would it be more readable if we early return in all known cases, and then return the error as a default in the end? That might mean to not have a `mut apdu`, but actually construct `APDU`s for returning in each case. WDYT?

.gitmodules
krkhan: Without this `git status` considers `third_party/tock` to be modified. Since that automatically gets modified during build/deployment I considered ignoring it.

E.g., without any manual modifications to `tock` folder.
```
On branch apdu-parser-kk
Your branch is up to date with 'origin/apdu-parser-kk'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   .gitmodules
        modified:   third_party/tock (modified content, untracked content)

no changes added to commit (use "git add" and/or "git commit -a")
```

src/ctap/apdu.rs
krkhan: Addressed in b9ffe7e4ce0664a0ebe4e5088efbbaa16a75e1a0

src/ctap/apdu.rs
```diff
@@ -73,11 +74,14 @@ impl From<&[u8]> for ApduHeader {
...
-    LcDataLe,
-    // TODO: More cases to add as extended length APDUs
-    // Le could be 2 or 3 Bytes
+    Le1,
+    Lc1Data,
+    Lc1DataLe1,
+    Lc3Data,
+    Lc3DataLe1,
+    Lc3DataLe2,
+    Lc3DataLe3,
```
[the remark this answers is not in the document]
krkhan: Good catch, removed in 2c49718fee700d349e0f5df9d1079823eca5259b . Would love to have a test-case for this case :) (i.e., from an actual packet, we can obviously create test-cases with hand-written packets)

src/ctap/apdu.rs
```diff
@@ -0,0 +1,426 @@
...
+        if frame.len() < APDU_HEADER_LEN as usize {
+            return Err(ApduStatusCode::SW_WRONG_DATA);
+        }
+        //        +-----+-----+----+----+
+        // header | CLA | INS | P1 | P2 |
+        //        +-----+-----+----+----+
+        let (header, payload) = frame.split_at(APDU_HEADER_LEN);
+
+        let mut apdu = APDU {
+            header: array_ref!(header, 0, 4).into(),
```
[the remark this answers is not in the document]
krkhan: Addressed in 0420ad8de6ebb9157202dd5333d1b49e012ed79a

src/ctap/apdu.rs
krkhan: That's a pretty great idea! Mutation is evil :) , addressed in 1d8c103d9b8180b93994e35ea9365c780e15803b

src/ctap/apdu.rs
```diff
@@ -142,21 +149,55 @@ impl TryFrom<&[u8]> for APDU {
                 }
             }
             if payload.len() == (1 + byte_0) as usize && byte_0 != 0 {
```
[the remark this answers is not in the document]
krkhan: Addressed in 524ebe3fce0877dc019fb504d6c98aab31d751d3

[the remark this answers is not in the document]
krkhan: Addressed in 524ebe3fce0877dc019fb504d6c98aab31d751d3

src/ctap/apdu.rs
```diff
@@ -142,21 +149,55 @@ impl TryFrom<&[u8]> for APDU {
...
-                apdu.case_type = ApduType::Short(Case::LcData);
+                // Lc is one-byte long and since the size specified by Lc covers the rest of the
+                // payload there's no Le at the end
+                apdu.case_type = ApduType::Short(Case::Lc1Data);
                 apdu.lc = byte_0.into();
                 apdu.data = payload[1..].to_vec();
             }
             if payload.len() == (1 + byte_0 + 1) as usize && byte_0 != 0 {
```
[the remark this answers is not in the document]
krkhan: Yup, done in 524ebe3fce0877dc019fb504d6c98aab31d751d3

src/ctap/apdu.rs
```diff
@@ -142,21 +149,55 @@ impl TryFrom<&[u8]> for APDU {
...
                 apdu.lc = byte_0.into();
                 apdu.data = payload[1..(payload.len() - 1)].to_vec();
                 apdu.le = (*payload.last().unwrap()).into();
                 if apdu.le == 0x00 {
                     apdu.le = 0x100;
                 }
             }
+            if payload.len() > 2 {
+                // Lc is possibly three-bytes long
+                let extended_apdu_lc: usize = BigEndian::read_u16(&payload[1..]) as usize;
```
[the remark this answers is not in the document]
krkhan: It's already checked at https://github.com/google/OpenSK/pull/227/files#diff-dedfa70623a95fd74474b0dc83c98c5468d0739e8bc3a8fe5dea90d199df6147R185

.gitmodules
kaczmarczyck: Ah got it. The change makes sense to me, I was just wondering if it's somehow related to this PR. Good to have fresh eyes on this.

src/ctap/apdu.rs
kaczmarczyck: Nice! I'd personally remove the else, since the if block before always returns.

src/ctap/apdu.rs
```diff
@@ -119,48 +123,103 @@ impl TryFrom<&[u8]> for APDU {
...
+                                &payload[payload.len() - extended_apdu_le_len..],
+                            ),
+                            _ => 0,
+                        },
+                        case_type: ApduType::Extended(match extended_apdu_le_len {
+                            0 => Case::Lc3Data,
+                            1 => Case::Lc3DataLe1,
+                            2 => Case::Lc3DataLe2,
+                            3 => Case::Le3,
+                            _ => Case::Unknown,
```
jmichelp, line 216: Not sure that building an APDU we haven't fully "understood" is a good idea.
Should we delete the Unknown case and simply return an error instead?

src/ctap/apdu.rs
```diff
@@ -119,48 +123,103 @@ impl TryFrom<&[u8]> for APDU {
...
+                    });
                 }
             }
         }
-        // TODO: Add extended length cases
-        if apdu.case_type == ApduType::default() {
-            return Err(ApduStatusCode::SW_COND_USE_NOT_SATISFIED);
-        }
-        Ok(apdu)
+        return Err(ApduStatusCode::SW_COND_USE_NOT_SATISFIED);
```
jmichelp, line 222: This error in the context of CTAP means that the authenticator is waiting for user presence.
I would put something else.

src/ctap/apdu.rs
kaczmarczyck: Also, `ApduType` doesn't seem to need `default` and `Unknown` anymore?

src/ctap/apdu.rs
kaczmarczyck, line 216: While I was thinking about `Unknown`, I was wondering why we return an unknown case, and not
`Err(ApduStatusCode::SW_COND_USE_NOT_SATISFIED);`
?

kaczmarczyck: Maybe I just don't see the use case? It makes the logic uses your APDUs a bit simpler if you remove the enum variant. If we don't need it?

krkhan: Addressed in 71ec2cf937260c075222960405f478108cbe0106

src/ctap/apdu.rs
krkhan: Addressed in 71ec2cf937260c075222960405f478108cbe0106

src/ctap/apdu.rs
krkhan: Done

src/ctap/apdu.rs
krkhan: Done

src/ctap/apdu.rs
```diff
@@ -142,21 +149,55 @@ impl TryFrom<&[u8]> for APDU {
...
+                    });
+                    apdu.data = payload[3..(payload.len() - extended_apdu_le_len)].to_vec();
+                    apdu.lc = extended_apdu_lc as u16;
+                    apdu.le = match extended_apdu_le_len {
+                        0 => 0,
+                        1 => (*payload.last().unwrap()).into(),
+                        2 => BigEndian::read_u16(&payload[payload.len() - extended_apdu_le_len..])
+                            as u32,
+                        3 => BigEndian::read_u32(&payload[payload.len() - extended_apdu_le_len..]),
+                        _ => 0,
```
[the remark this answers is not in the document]
krkhan: Done

src/ctap/apdu.rs
```diff
@@ -142,21 +149,55 @@ impl TryFrom<&[u8]> for APDU {
...
+                        _ => Case::Unknown,
+                    });
+                    apdu.data = payload[3..(payload.len() - extended_apdu_le_len)].to_vec();
+                    apdu.lc = extended_apdu_lc as u16;
+                    apdu.le = match extended_apdu_le_len {
+                        0 => 0,
+                        1 => (*payload.last().unwrap()).into(),
+                        2 => BigEndian::read_u16(&payload[payload.len() - extended_apdu_le_len..])
+                            as u32,
+                        3 => BigEndian::read_u32(&payload[payload.len() - extended_apdu_le_len..]),
```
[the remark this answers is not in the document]
krkhan: we'll need different indices for the `Le3` case anyway, so won't be any more concise i think

src/ctap/apdu.rs
kaczmarczyck: Sorry, I had two comments in one thread. What do you think about removing indentation levels after you return?

krkhan: done, addressed in 4bfce88

kaczmarczyck (approved):
LGTM from my side. Letting @jmichelp do the final approval!

src/ctap/apdu.rs
```diff
@@ -119,48 +115,118 @@ impl TryFrom<&[u8]> for APDU {
...
+                            if last_byte == 0x00 {
+                                0x100
+                            } else {
+                                last_byte
+                            }
+                        }
+                        2 => {
+                            let le_parsed = BigEndian::read_u16(&payload[payload.len() - 2..]);
+                            if le_parsed == 0x00 {
+                                0x100
```
jmichelp, line 198: If I'm correct, the value should be `0x10000`.
When `Le` is encoded on 2 or 3 bytes, a 0 is interpreted as 65 536.

jmichelp (approved):

PR #229, opened 2020-12-04
Signature counter increments

kaczmarczyck:
Changes the signature counter to be in line with [FIDO requirements](https://fidoalliance.org/wp-content/uploads/2019/01/Authenticator_Sample_VQ_Answers.pdf).

@ia0 This change affects storage v2, in case it's merged ealier.

- [x] Tests pass


src/ctap/mod.rs
```diff
@@ -1094,9 +1097,43 @@ mod test {
                 // The expected response is split to only assert the non-random parts.
                 assert_eq!(fmt, "packed");
                 let mut expected_auth_data = vec![
-                    0xA3, 0x79, 0xA6, 0xF6, 0xEE, 0xAF, 0xB9, 0xA5, 0x5E, 0x37, 0x8C, 0x11, 0x80,
-                    0x34, 0xE2, 0x75, 0x1E, 0x68, 0x2F, 0xAB, 0x9F, 0x2D, 0x30, 0xAB, 0x13, 0xD2,
-                    0x12, 0x55, 0x86, 0xCE, 0x19, 0x47, 0x41, 0x00, 0x00, 0x00, 0x00,
+                    0xA3,
```
jmichelp, line 1100: This formatting is not really convenient.

src/ctap/ctap1.rs
```diff
@@ -642,6 +643,16 @@ mod test {
         assert_eq!(response, Err(Ctap1StatusCode::SW_WRONG_DATA));
     }
 
+    fn check_signature_counter(response: &[u8; 4], signature_counter: u32) {
+        if USE_SIGNATURE_COUNTER {
+            let mut signature_counter_bytes = [0u8; 4];
+            BigEndian::write_u32(&mut signature_counter_bytes, signature_counter);
+            assert_eq!(response, &signature_counter_bytes);
```
ia0, line 650: What about the following?

```rust
assert_eq!(u32::from_be_bytes(*response), signature_counter);
```

src/ctap/mod.rs
ia0: Since it's a test we can just do `expected_auth_data.push(INITIAL_SIGNATURE_COUNTER)` which should hopefully make cargo-fmt compress the rest.

ia0 (dismissed):
Storage change looks good to me. I'll apply it to v2. And actually I realized I should check the length of the global counter and return an error if it's not 4...

src/ctap/ctap1.rs
kaczmarczyck: That's a bit nicer, true!

src/ctap/mod.rs
kaczmarczyck: I didn't notice `cargo fmt` changed that. Changed to push the last byte, and fixed.

ia0 (approved):
Still good for me.

PR #230, opened 2020-12-08
Use new APDU parser in CTAP1 code

krkhan:


src/ctap/apdu.rs
```diff
@@ -168,12 +168,17 @@ impl TryFrom<&[u8]> for APDU {
         }
         if payload.len() > 2 {
             // Lc is possibly three-bytes long
-            let extended_apdu_lc: usize = BigEndian::read_u16(&payload[1..]) as usize;
-            let extended_apdu_le_len: usize = if payload.len() > extended_apdu_lc {
-                payload.len() - extended_apdu_lc - 3
-            } else {
-                0
+            let extended_apdu_lc: usize = BigEndian::read_u16(&payload[1..3]) as usize;
```
kaczmarczyck, line 171: Do you need to make the type explicit, if you already assign `as usize`?

src/ctap/apdu.rs
```diff
@@ -168,12 +168,17 @@ impl TryFrom<&[u8]> for APDU {
...
+            if payload.len() < extended_apdu_lc + 3 {
+                return Err(ApduStatusCode::SW_WRONG_LENGTH);
+            }
+            let extended_apdu_le_len: usize = match payload.len() - extended_apdu_lc {
+                // There's some possible Le bytes at the end
+                2..=5 => payload.len() - extended_apdu_lc - 3,
```
kaczmarczyck, line 177: Is this intended? What if `payload.len() - extended_apdu_lc == 2`? Wouldn't you underflow?

src/ctap/ctap1.rs
```diff
@@ -118,11 +119,16 @@ impl TryFrom<&[u8]> for U2fCommand {
...
     fn try_from(message: &[u8]) -> Result<Self, Ctap1StatusCode> {
-        if message.len() < Ctap1Command::APDU_HEADER_LEN as usize {
-            return Err(Ctap1StatusCode::SW_WRONG_DATA);
-        }
-
-        let (apdu, payload) = message.split_at(Ctap1Command::APDU_HEADER_LEN as usize);
+        let apdu: APDU = match APDU::try_from(message) {
+            Ok(apdu) => apdu,
+            // Todo: Better conversion between ApduStatusCode and Ctap1StatusCode
+            // Maybe use TryFrom?
```
kaczmarczyck, line 125: Our `TODO` style would be
> // TODO(kaczmarczyck)

or 
> // TODO(#123456)

for a GitHub issue number. In this case, why not resolve it? By merging these two status codes? I don't think we need both.

src/ctap/ctap1.rs
```diff
@@ -131,30 +137,29 @@ impl TryFrom<&[u8]> for U2fCommand {
...
             return Err(Ctap1StatusCode::SW_CLA_NOT_SUPPORTED);
         }
 
-        let lc = (((apdu[4] as u32) << 16) | ((apdu[5] as u32) << 8) | (apdu[6] as u32)) as usize;
-
         // Since there is always request data, the expected length is either omitted or
         // encoded in 2 bytes.
-        if lc != payload.len() && lc + 2 != payload.len() {
+        // Todo: support extended APDUs now that the new parser can work with those
+        if apdu.lc as usize != apdu.data.len() && (apdu.lc as usize) + 2 != apdu.data.len() {
```
kaczmarczyck, line 147: Same comment for `TODO` style. Also, maybe set `let lc = apdu.lc as usize` here once and the some old code runs as before?

src/ctap/ctap1.rs
```diff
@@ -571,13 +574,7 @@ mod test {
...
             create_authenticate_message(&application, Ctap1Flags::CheckOnly, &key_handle);
 
-        message.push(0x00);
-        let response = Ctap1Command::process_command(&message, &mut ctap_state, START_CLOCK_VALUE);
-        assert_eq!(response, Err(Ctap1StatusCode::SW_WRONG_LENGTH));
-
-        // Two extra zeros are okay, they could encode the expected response length.
-        message.push(0x00);
-        message.push(0x00);
+        message.extend_from_slice(&[0x00, 0x00, 0x00, 0x00]);
```
kaczmarczyck, line 577: Before, we were checking if both types work: 1 or 3 zeros. Now, you are checking only once, and with 4 zeros. Was the unit test wrong before? And why not check both cases?

src/ctap/apdu.rs
krkhan: @kaczmarczyck the check a couple lines above would ensure that's not the case

src/ctap/ctap1.rs
krkhan: @kaczmarczyck before any extra bytes were bad. Now, there might be 1-3 additional bytes at the end, so the old test will now simply be parsed as Le=0. Appending four bytes to really overflow length.

src/ctap/apdu.rs
krkhan: yup it wasn't needed, removed

src/ctap/ctap1.rs
krkhan: @kaczmarczyck good idea! moved to a local `lc` variable to remove redundant casting

src/ctap/ctap1.rs
```diff
@@ -60,6 +60,24 @@ impl TryFrom<u16> for Ctap1StatusCode {
     }
 }
 
+impl TryFrom<ApduStatusCode> for Ctap1StatusCode {
+    type Error = ();
+
+    fn try_from(apdu_status_code: ApduStatusCode) -> Result<Ctap1StatusCode, ()> {
```
kaczmarczyck, line 66: Can you say why we convert between those, instead of removing one of the types and only using the other?

src/ctap/apdu.rs
kaczmarczyck: Ah, now the range is `0..=3` and it matches. The comparison seems to be a bit convoluted, still. What you basically want is:

`let extended_apdu_le_len = match payload.len() - extended_apdu_lc - 3`
`if not 0 <= extended_apdu_le_len <= 3 return WRONG_LENGTH`

But because of `usize`, you need an `if` and a `match`?

src/ctap/ctap1.rs
kaczmarczyck: Okay, can you add tests for all different possible lengths then, to at least match the test strength from before?

src/ctap/apdu.rs
krkhan: @kaczmarczyck yeah the reason I was doing the `if` dance was to prevent subtraction overflows. Now I case them as unsigned integers and evaluate the expression only one. Let me know if this one is more readable.

src/ctap/ctap1.rs
krkhan: @kaczmarczyck @jmichelp  not sure if this is the case, but wouldn't APDU be possibly be used by non CTAP-1 scenarios as well?

src/ctap/ctap1.rs
krkhan: @kaczmarczyck done

src/ctap/ctap1.rs
jmichelp: `Ctap1StatusCode` is a practically a subset of `ApduStatusCode` where a few values have a slightly different meaning in the context of Ctap1. E.g. SW_COND_USE_NOT_SATISFIED means in Ctap1 that we're waiting for user presence whereas in generic APDU context it doesn't have a specific meaning.
We can merge them and only keep the error codes in `apdu.rs`. I have a slight preference of having the representation being integers instead of a `ByteArray`.

If that makes sense at a higher level, we can even have `CtapStatusCode` being an enum containing both `ApduStatusCode` and `Ctap2StatusCode`. This way all the processing functions return `CtapStatusCode` errors.

kaczmarczyck: I certainly agree to the first paragraph. For the `CtapStatusCode` refactoring, how many functions allow interchanging these error codes? I thought they have pretty separate domains?

src/ctap/apdu.rs
kaczmarczyck: There is an
`if payload.len() < extended_apdu_lc + 3 {`
So why cast? Why not just

`let extended_apdu_le_len = payload.len() - extended_apdu_lc - 3;`
`if extended_apdu_le_len > 3 { return [...]`

krkhan: @kaczmarczyck `payload.len()` and `extended_apdu_lc` are `usize` so `payload.len() - extended_apdu_lc - 3` on first look feels like it might possible overflow. you are right though that in this case it's not possible for them to both be `0` because of the checks above. updated as per your suggested snippet

kaczmarczyck: If this makes you feel uneasy, you could also do something like
`let extended_apdu_le_len payload.len().checked_sub(extended_apdu_lc + 3).map_err(...)?;`
I'm happy either way.

src/ctap/ctap1.rs
krkhan: @jmichelp @kaczmarczyck not sure how much it helped but I moved `ApduStatusCode` to use integers instead of `ByteArray`s and did a best-effort match for places `Ctap1StatusCode` was in use. Not familiar with provenance of the codes mentioned in the enums so maybe the mapping wasn't perfect

(commit 29dbff7a40a1207e19fc6f2e0249eebd6b6d5db8 )

src/ctap/apdu.rs
krkhan: good idea, updated to use `checked_sub` and `unwrap_or` the result

src/ctap/ctap1.rs
```diff
@@ -292,22 +255,22 @@ impl Ctap1Command {
         let pk = sk.genpk();
         let key_handle = ctap_state
             .encrypt_key_handle(sk, &application)
-            .map_err(|_| Ctap1StatusCode::SW_COMMAND_ABORTED)?;
+            .map_err(|_| Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?;
         if key_handle.len() > 0xFF {
             // This is just being defensive with unreachable code.
-            return Err(Ctap1StatusCode::SW_VENDOR_KEY_HANDLE_TOO_LONG);
+            return Err(Ctap1StatusCode::SW_WRONG_LENGTH);
```
jmichelp, line 261: Please don't change the return codes as this may break interop.

src/ctap/ctap1.rs
```diff
@@ -292,22 +255,22 @@ impl Ctap1Command {
...
         }
 
         let certificate = ctap_state
             .persistent_store
             .attestation_certificate()
-            .map_err(|_| Ctap1StatusCode::SW_MEMERR)?
-            .ok_or(Ctap1StatusCode::SW_COMMAND_ABORTED)?;
+            .map_err(|_| Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?
```
jmichelp, line 267: We decided in a PR #222  to report separate errors for potential storage/hardware failure (SW_MEMERR) or the entry not being present (SW_COMMAND_ABORTED)

src/ctap/ctap1.rs
```diff
@@ -292,22 +255,22 @@ impl Ctap1Command {
...
+            .ok_or(Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?;
         let private_key = ctap_state
             .persistent_store
             .attestation_private_key()
-            .map_err(|_| Ctap1StatusCode::SW_MEMERR)?
-            .ok_or(Ctap1StatusCode::SW_COMMAND_ABORTED)?;
+            .map_err(|_| Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?
```
jmichelp, line 272: Same here

src/ctap/ctap1.rs
```diff
@@ -362,14 +325,14 @@ impl Ctap1Command {
...
-                return Err(Ctap1StatusCode::SW_CONDITIONS_NOT_SATISFIED);
+                return Err(Ctap1StatusCode::SW_COND_USE_NOT_SATISFIED);
             }
             ctap_state
                 .increment_global_signature_counter()
                 .map_err(|_| Ctap1StatusCode::SW_WRONG_DATA)?;
             let mut signature_data = ctap_state
                 .generate_auth_data(&application, Ctap1Command::USER_PRESENCE_INDICATOR_BYTE)
-                .map_err(|_| Ctap1StatusCode::SW_WRONG_DATA)?;
+                .map_err(|_| Ctap1StatusCode::SW_COND_USE_NOT_SATISFIED)?;
```
jmichelp, line 335: This error in CTAP1 means that we're waiting for user presence. Is there a reason you changed the returned error?

src/ctap/hid/mod.rs
```diff
@@ -417,7 +417,7 @@ impl CtapHid {
     #[cfg(feature = "with_ctap1")]
     fn ctap1_success_message(cid: ChannelID, payload: &[u8]) -> HidPacketIterator {
         let mut response = payload.to_vec();
-        let code: u16 = ctap1::Ctap1StatusCode::SW_NO_ERROR.into();
+        let code: u16 = ctap1::Ctap1StatusCode::SW_INTERNAL_EXCEPTION.into();
```
jmichelp, line 420: Here we need to either change the name or the error code.
A function called `ctap1_success_message` shouldn't expect an `INTERNAL_EXCEPTION` error code.

src/ctap/apdu.rs
```diff
@@ -168,7 +174,10 @@ impl TryFrom<&[u8]> for APDU {
                 return Err(ApduStatusCode::SW_WRONG_LENGTH);
             }
 
-            let extended_apdu_le_len: usize = payload.len() - extended_apdu_lc - 3;
+            let extended_apdu_le_len: usize = payload
+                .len()
+                .checked_sub(extended_apdu_lc + 3)
+                .unwrap_or(0xff);
```
kaczmarczyck, line 180: You are implicitly trying to reuse the error code handling in the `if` below? Should work, but explicit is usually better than implicit. That's why I proposed `ok_or(ERROR_CODE)?`.

src/ctap/ctap1.rs
```diff
@@ -282,7 +227,7 @@ impl Ctap1Command {
             U2fCommand::Version => Ok(Vec::<u8>::from(super::U2F_VERSION_STRING)),
 
             // TODO: should we return an error instead such as SW_INS_NOT_SUPPORTED?
-            U2fCommand::VendorSpecific { .. } => Err(Ctap1StatusCode::SW_NO_ERROR),
+            U2fCommand::VendorSpecific { .. } => Err(Ctap1StatusCode::SW_INTERNAL_EXCEPTION),
```
kaczmarczyck, line 230: This case is for vendor specific commands. The functionally equivalent replacement would be `SW_SUCCESS`, right? As the comment suggests, I'd also support `SW_INS_INVALID` without any expected interoperability issues. Can you explain your rational behind `SW_INTERNAL_EXCEPTION`?

src/ctap/ctap1.rs
kaczmarczyck: In this case, the if should be unreachable and we're just being defensive here. I think this should be `SW_INTERNAL_EXCEPTION` with our new policy.
The risk of interoperability issues is low in this particular case, but I see the point of interop breaking changes. So I'm just leaving my opinion here, if we want to get rid of `SW_VENDOR_KEY_HANDLE_TOO_LONG`.

src/ctap/ctap1.rs
kaczmarczyck: Thanks for the changes! I left comments on the best-effort. I think generally, we probably want to return the same code, if we don't have a strong reason to change it. There is one custom error code, that should be non problematic to change. All others are usually implemented as the specification demands.

src/ctap/apdu.rs
krkhan: done

src/ctap/ctap1.rs
krkhan: updated

src/ctap/ctap1.rs
krkhan: updated to `SW_INTERNAL_EXCEPTION` as per team chat

src/ctap/hid/mod.rs
krkhan: the only excuse i can come up with is that i wasn't thinking straight. in my defense the line number tripped me off

src/ctap/ctap1.rs
krkhan: fixed

krkhan (commented):
addressed comments about the error codes

jmichelp (approved):

src/ctap/apdu.rs
```diff
@@ -177,10 +178,7 @@ impl TryFrom<&[u8]> for APDU {
             let extended_apdu_le_len: usize = payload
                 .len()
                 .checked_sub(extended_apdu_lc + 3)
-                .unwrap_or(0xff);
-            if extended_apdu_le_len > 3 {
-                return Err(ApduStatusCode::SW_WRONG_LENGTH);
-            }
```
kaczmarczyck: Don't you still want the length check? (And if yes, and this goes undetected, maybe add a unit test?)

src/ctap/apdu.rs
kaczmarczyck: Opened a new comment for a new possible issue.

src/ctap/apdu.rs
krkhan: fixed with 35bdfe90ed52a5fac37d84b1ea6702b7d8c78001

kaczmarczyck: Ah, this was handled in the case below: `extended_apdu_le_len <= 3`
My oversight!

kaczmarczyck (approved):

PR #231, opened 2020-12-08
implicitly generate HMAC-secret

kaczmarczyck:
This PR implements the new version of the HMAC-secret extension. It does so by reverting the extension of key handles in #218, now implicitly creating `cred_random` values deterministically from the credential and 2 new device secrets.

Part of #106 , but backwards compatible to the 2.0 version.

@jmichelp for checking the idea
@ia0 for storage
optional FYI @gendx for crypto (i.e. `generate_cred_random`)

- [x] Tests pass

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -86,10 +86,8 @@ pub const INITIAL_SIGNATURE_COUNTER: u32 = 1;
 // - 16 byte initialization vector for AES-256,
 // - 32 byte ECDSA private key for the credential,
 // - 32 byte relying party ID hashed with SHA256,
-// - (optional) 32 byte for HMAC-secret,
 // - 32 byte HMAC-SHA256 over everything else.
-pub const CREDENTIAL_ID_BASE_SIZE: usize = 112;
-pub const CREDENTIAL_ID_MAX_SIZE: usize = CREDENTIAL_ID_BASE_SIZE + 32;
+pub const CREDENTIAL_ID_SIZE: usize = 112;
 // Set this bit when checking user presence.
 const UP_FLAG: u8 = 0x01;
 // Set this bit when checking user verification.
@@ -142,6 +140,7 @@ struct AssertionInput {
     client_data_hash: Vec<u8>,
     auth_data: Vec<u8>,
     hmac_secret_input: Option<GetAssertionHmacSecretInput>,
+    has_uv: bool,
 }
 
 struct AssertionState {
@@ -231,7 +230,6 @@ where
         &mut self,
         private_key: crypto::ecdsa::SecKey,
         application: &[u8; 32],
-        cred_random: Option<&[u8; 32]>,
     ) -> Result<Vec<u8>, Ctap2StatusCode> {
         let master_keys = self.persistent_store.master_keys()?;
         let aes_enc_key = crypto::aes256::EncryptionKey::new(&master_keys.encryption);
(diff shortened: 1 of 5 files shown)


src/ctap/storage.rs
```diff
@@ -987,7 +1011,7 @@ mod test {
         let mut rng = ThreadRng256 {};
         let mut persistent_store = PersistentStore::new(&mut rng);
 
-        // Master keys stay the same between resets.
+        // Master keys stay the same between calls.
```
ia0, line 1014: I guess the English was wrong. We should probably keep the intent by suffixing "within the same CTAP reset cycle" or something like that, no?

ia0 (approved):
LGTM for storage.

src/ctap/storage.rs
kaczmarczyck: I'm not an authority for the English language, but I wasn't sure about its meaning, so I changed it. :)
You other formulation is good, so I changed it to that.

ia0 (approved):

jmichelp (commented):
We still pass the conformance tool (only tested HMAC-Extension) with this PR.

src/ctap/mod.rs
```diff
@@ -622,10 +598,25 @@ where
...
+    // The computation is deterministic, and private_key expected to be unique.
+    fn generate_cred_random(
+        &mut self,
+        private_key: &crypto::ecdsa::SecKey,
+        has_uv: bool,
+    ) -> Result<[u8; 32], Ctap2StatusCode> {
+        let mut private_key_bytes = [0u8; 32];
+        private_key.to_bytes(&mut private_key_bytes);
+        let salt = crypto::sha256::Sha256::hash(&private_key_bytes);
+        // TODO(kaczmarczyck) KDF? hash salt together with rp_id_hash?
```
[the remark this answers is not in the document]
kaczmarczyck: Discussed offline, and SHA256 removed.

jmichelp (approved):
LGTM
I'll let @gendx approve too and this can be merged

src/ctap/mod.rs
```diff
@@ -240,16 +238,12 @@ where
         let mut iv = [0; 16];
         iv.copy_from_slice(&self.rng.gen_uniform_u8x32()[..16]);
 
-        let block_len = if cred_random.is_some() { 6 } else { 4 };
+        let block_len = 4;
```
gendx, line 241: Maybe this `block_len` is not needed anymore now that all the sizes are known at compile time.

src/ctap/mod.rs
```diff
@@ -240,16 +238,12 @@ where
...
         blocks[1].copy_from_slice(&sk_bytes[16..]);
         blocks[2].copy_from_slice(&application[..16]);
         blocks[3].copy_from_slice(&application[16..]);
-        if let Some(cred_random) = cred_random {
-            blocks[4].copy_from_slice(&cred_random[..16]);
-            blocks[5].copy_from_slice(&cred_random[16..]);
-        }
         cbc_encrypt(&aes_enc_key, iv, &mut blocks);
 
         let mut encrypted_id = Vec::with_capacity(16 * (block_len + 3));
```
gendx, line 249: And this could be a fixed-size array instead of a Vec, now that `block_len == 4`.

src/ctap/mod.rs
```diff
@@ -622,10 +598,23 @@ where
...
+    // The computation is deterministic, and private_key expected to be unique.
+    fn generate_cred_random(
+        &mut self,
+        private_key: &crypto::ecdsa::SecKey,
+        has_uv: bool,
+    ) -> Result<[u8; 32], Ctap2StatusCode> {
+        let mut private_key_bytes = [0u8; 32];
+        private_key.to_bytes(&mut private_key_bytes);
+        let key = self.persistent_store.cred_random_secret(has_uv)?;
+        Ok(hmac_256::<Sha256>(&key, &private_key_bytes[..]))
```
gendx, line 611: - This should be fine given the assumptions (i.e. `private_key_bytes` was chosen uniformly at random).
- Some nit on formatting: the `[..]` part is not really needed here. The only difference I see is that it creates a `&[u8]` rather than a `&[u8; 32]`, but the conversion to `&[u8]` is implicit IIRC.

src/ctap/mod.rs
kaczmarczyck: Changed!

src/ctap/mod.rs
kaczmarczyck: The assumption is correct. Removed the `[..]`.

src/ctap/mod.rs
kaczmarczyck: Both `block_len` removed.

jmichelp (approved):

PR #234, opened 2020-12-10
Remove ram_storage feature

ia0:
We don't need to build a production key without persistent storage. Tests and fuzzing continue to use the std feature to use the RAM implementation (that does sanity checks).


src/ctap/storage.rs
```diff
@@ -31,9 +31,9 @@ use cbor::cbor_array_vec;
 use core::convert::TryInto;
 use crypto::rng256::Rng256;
 
-#[cfg(any(test, feature = "ram_storage"))]
+#[cfg(feature = "std")]
 type Storage = persistent_store::BufferStorage;
-#[cfg(not(any(test, feature = "ram_storage")))]
+#[cfg(not(feature = "std"))]
```
kaczmarczyck, line 36: Why `std` instead of `test`? Because fuzzing is `std`, but not `test`?

ia0: Yes exactly. We can't control the `test` condition while we can control features condition like `feature = "std"`. Since tests and fuzzing need `feature = "std"` and production needs `no_std`, this makes a clean distinction between:
 - Production environment: `not(feature = "std")`.
- Testing environment (including fuzzing): `feature = "std"`.

The `test` condition is only needed if something is specific to Rust unit tests, instead of being specific to testing/non-production.

kaczmarczyck: Sounds good. I tested the other fuzzing calls, they still work.

kaczmarczyck (approved):

PR #235, opened 2020-12-10
Add latency example

ia0:


kaczmarczyck (approved):
Latencies and code look okay.

Your code tries to get the storage into specific states to test those. How did you test if the states you are trying to reach are actually what is in the device?

ia0:
> How did you test if the states you are trying to reach are actually what is in the device?

I used to have asserts. I removed them because they still assumed some stuff and the actual latency numbers you get while running are a better proof. For example, if the code is wrong, the compaction would be much faster. But I can add them back if you think it helps.

kaczmarczyck:
I think I'd like to see them, if it's not super implementation dependent and breaks on each change to the persistent store? (Wouldn't happen too often anyway?)

ia0:
Yeah, they use the stable API of the store, so it shouldn't break because of that. However, they rely on the relation between virtual capacity and user capacity. But this should not change unless we add new features to the store that require reserved internal capacity.

kaczmarczyck (dismissed):

ia0:
I had to add a commit to make the latency example compile when running `cargo test --features=std` which not only runs unit tests but also tries to build all examples. Since the latency example was meant to measure latency of the flash storage, it didn't support compiling with "std" feature. I made it compile with the "std" feature such that we don't break the build of this example by accident. This also remove some logic from the CTAP storage to the embedded_flash module.

kaczmarczyck (approved):

PR #238, opened 2020-12-10
Make store operations constant wrt flash operations

ia0:
This is not clear whether we need this, but it's easier to decide with code.

This makes all store operations (except initial boot) constant-time with respect to flash operations to the price of 2 bytes per alive entry. In particular:
- The initial boot is still linear because we have to check for possible interrupted operations or corruption. Although it might possible be improved if needed.
- The operations are not linear with respect to RAM operations. They are still linear, but the time of RAM operations is order-of-magnitudes faster than flash operations, so it doesn't matter to get the benefit.
- The memory cost is static, i.e. if there are `N` non-deleted entries in the store, then the store uses `2 * N` bytes of RAM. With our current setup, this would be at most 350 bytes.

This was fuzzed for one week for now (still running). We might want to fuzz longer and also wait for a review of fuzzing (I want to clean it up a bit which may improve fuzzing efficience as a side-effect).


libraries/persistent_store/src/store.rs
```diff
@@ -863,6 +923,30 @@ impl<S: Storage> Store<S> {
...
+        let entries = match &mut self.entries {
+            None => return Ok(()),
+            Some(x) => x,
+        };
+        let head = self.head?;
+        let offset = u16::try_from(pos - head).ok()?;
+        let i = entries.iter().position(|x| *x == offset)?;
+        entries.swap_remove(i);
+        Ok(())
+    }
```
kaczmarczyck, line 948: `push_entry` and `remove_entry` could use some unit testing?

kaczmarczyck (approved):
The code looks good. Let's discuss the change internally before merging.

libraries/persistent_store/src/store.rs
ia0: Done. I checked their behavior on the `entries` field.

kaczmarczyck (approved):

libraries/persistent_store/src/format.rs
```diff
@@ -1077,4 +1077,12 @@ mod tests {
             0xff800000
         );
     }
+
+    #[test]
+    fn position_offsets_fit_in_a_halfword() {
+        // The store stores the entry positions as their offset from the head. Those offsets are
+        // represented as u16. The bound below is a large over-approximation of the maximal offset
+        // but it already fits.
+        assert_eq!((MAX_PAGE_INDEX + 1) * MAX_VIRT_PAGE_SIZE, 0xff80);
```
jmichelp: According to the name of the test, shouldn't you check that the computed value is less or equal than `u16::MAX` then?

.github/workflows/persistent_store_test.yml
```diff
@@ -13,6 +13,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          override: true
```
jmichelp, line 19: This means that the persistent storage is going to be tested/built against a toolchain that is not the one use to build and test OpenSK (i.e. the one referred in the `rust-toolchain` file at the root of the repository.

ia0: But this is already not the case no? I had to change this otherwise it was using stable. If there's a way to use `rust-toolchain` I'd like to do that. I think the issue is that `rust-toolchain` is a symlink to a gitmodule. So I see 2 solutions:
1. Clone the submodules (waste of resources but then I use the same toolchain).
2. Keep as is since any difference should be a compiler bug.

I'm more in favor of keeping as is since this is just to run tests, so there would anyway be a difference between x86 and arm.

libraries/persistent_store/src/format.rs
ia0: Yes, but this is a consequence of the test. The 2 alternatives are:
1. `assert_eq!(BOUND, 0xff80)` which implies `BOUND <= u16::MAX`
2. `assert!(BOUND <= u16::MAX)`

I think the first is stable enough that it should not increase maintenance while documenting the actual maximum value (which fits in `u16` which is the goal of the test).

Actually there's a 3rd option if you prefer:
```rust
const MAX_POS: Nat = (MAX_PAGE_INDEX + 1) * MAX_VIRT_PAGE_SIZE;
assert_eq!(MAX_POS, 0xff80);  // documentation
assert!(MAX_POS <= u16::MAX as Nat);  // redundant but what actually matters
```

jmichelp: I think the third option makes more sense provided you swap the 2 assertions:
```rust
const MAX_POS: Nat = (MAX_PAGE_INDEX + 1) * MAX_VIRT_PAGE_SIZE;
assert!(MAX_POS <= u16::MAX as Nat);  // what actually matters
assert_eq!(MAX_POS, 0xff80);  // redundant but documentation
```

.github/workflows/persistent_store_test.yml
jmichelp: Ah I see, it's a workaround because the symlink is broken until you get the submodules. Makes sense then.
I'm simply worried that this may creates some discrepancies between the workflows by not using the same toolchain (e.g. 1 workflow being broken but not the other)

ia0: Yeah, we should do something. One solution that looks OK would be:
- Don't use a symlink and have duplicated data.
- The `setup.sh` script should check that file at the root is equal (content) to the reference file (currently `libtock-rs`)
- We could also emit a warning that the `libtock-rs` and `tock` toolchains are different, but we probably can't fix it on our side.

This would fix all github worflows that don't update the submodules.

libraries/persistent_store/src/format.rs
ia0: Sounds good. Done.

.github/workflows/persistent_store_test.yml
jmichelp: Rather than `setup.sh` this should be part of the checks from the CI to ensure that when we bump the kernel version we also update this file.

`libtock-rs` and `tock` don't use the same toolchain IIRC. Or at least that used to be case if they fixed it.

ia0: Right that's even better to do it in a workflow. So should I wait until we have this check and update my workflow, or should I create a bug to remember to update the workflow once we have this check?

jmichelp (approved):

.github/workflows/persistent_store_test.yml
jmichelp: Create a bug and merge this PR.
I have to think about revamping the GitHub actions anyway now that we have multiple branches.

ia0: Sounds good, created https://github.com/google/OpenSK/issues/261.

PR #244, opened 2020-12-18
updates command bytes

kaczmarczyck:
Continuing on #106. Some command codes were incorrect, this reflects the state of review draft. Some command codes are conceptual only (i.e. `AUTHENTICATOR_VENDOR_FIRST_UNUSED`), they are commented out to eventually be able to drop `#[allow(dead_code)]`.

- [x] Tests pass

### DIFF
--- a/src/ctap/command.rs
+++ b/src/ctap/command.rs
@@ -65,12 +65,13 @@ impl Command {
     const AUTHENTICATOR_GET_NEXT_ASSERTION: u8 = 0x08;
     // TODO(kaczmarczyck) use or remove those constants
     const AUTHENTICATOR_BIO_ENROLLMENT: u8 = 0x09;
-    const AUTHENTICATOR_CREDENTIAL_MANAGEMENT: u8 = 0xA0;
-    const AUTHENTICATOR_SELECTION: u8 = 0xB0;
-    const AUTHENTICATOR_CONFIG: u8 = 0xC0;
+    const AUTHENTICATOR_CREDENTIAL_MANAGEMENT: u8 = 0x0A;
+    const AUTHENTICATOR_SELECTION: u8 = 0x0B;
+    const AUTHENTICATOR_LARGE_BLOBS: u8 = 0x0C;
+    const AUTHENTICATOR_CONFIG: u8 = 0x0D;
+    const _AUTHENTICATOR_VENDOR_FIRST: u8 = 0x40;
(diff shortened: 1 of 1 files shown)


src/ctap/command.rs
```diff
@@ -65,12 +65,13 @@ impl Command {
...
-    const AUTHENTICATOR_CONFIG: u8 = 0xC0;
+    const AUTHENTICATOR_CREDENTIAL_MANAGEMENT: u8 = 0x0A;
+    const AUTHENTICATOR_SELECTION: u8 = 0x0B;
+    const AUTHENTICATOR_LARGE_BLOBS: u8 = 0x0C;
+    const AUTHENTICATOR_CONFIG: u8 = 0x0D;
     const AUTHENTICATOR_VENDOR_CONFIGURE: u8 = 0x40;
-    const AUTHENTICATOR_VENDOR_FIRST_UNUSED: u8 = 0x41;
-    const AUTHENTICATOR_VENDOR_LAST: u8 = 0xBF;
+    // const AUTHENTICATOR_VENDOR_FIRST_UNUSED: u8 = 0x41;
+    // const AUTHENTICATOR_VENDOR_LAST: u8 = 0xBF;
```
ia0, line 74: What about the following 2 alternatives?
1. Name them with underscore, e.g. `_AUTHENTICATOR_VENDOR_LAST`.
2. Simply have a comment that explains the range:

```rust
// Those constants are vendor commands. They should take value between 0x40 and 0xBF inclusive.
const AUTHENTICATOR_VENDOR_CONFIGURE: u8 = 0x40;
```

ia0 (approved):

src/ctap/command.rs
kaczmarczyck: No strong preference. All of these allow removing the `dead_code` warning eventually. Maybe the underscore is a bit more readable. I'd vote against the comment since names like `AUTHENTICATOR_VENDOR_LAST` are actually taken from the specification. So it reads easier when you have the implementation and the spec side by side.

Changing to `_AUTHENTICATOR_VENDOR_LAST` works for me?

ia0: Ah ok I didn't know those names were official. In that case yes, `_AUTHENTICATOR_VENDOR_LAST` seems the best (since it avoids having commented code which always ends up rotting because the compiler doesn't check it).

kaczmarczyck: Done

ia0 (approved):

PR #246, opened 2020-12-21
Authenticator config: SetMinPinLength

kaczmarczyck:
Continues implementation of #106 .

This implements the subcommand `SetMinPinLength` and the `minPinLength` extension. Their functionality started in the `ClientPin` command, but was moved to `AuthenticatorConfig`.

To complete `SetMinPinLength`, the `ClientPin` command needs to implement forced PIN changes. I decided to move that to the next PR, since this one is already pretty big.

- [x] Tests pass

### DIFF
--- a/src/ctap/config_command.rs
+++ b/src/ctap/config_command.rs
@@ -24,7 +24,6 @@ use alloc::vec;
 /// Processes the subcommand setMinPINLength for AuthenticatorConfig.
 fn process_set_min_pin_length(
     persistent_store: &mut PersistentStore,
-    pin_protocol_v1: &mut PinProtocolV1,
     params: SetMinPinLengthParams,
 ) -> Result<ResponseData, Ctap2StatusCode> {
     let SetMinPinLengthParams {
@@ -44,8 +43,10 @@ fn process_set_min_pin_length(
     if let Some(old_length) = persistent_store.pin_code_point_length()? {
         force_change_pin |= new_min_pin_length > old_length;
     }
-    pin_protocol_v1.force_pin_change |= force_change_pin;
-    // TODO(kaczmarczyck) actually force a PIN change
+    if force_change_pin {
+        // TODO(kaczmarczyck) actually force a PIN change in PinProtocolV1
+        persistent_store.force_pin_change()?;
+    }
     persistent_store.set_min_pin_length(new_min_pin_length)?;
     if let Some(min_pin_length_rp_ids) = min_pin_length_rp_ids {
         persistent_store.set_min_pin_length_rp_ids(min_pin_length_rp_ids)?;
@@ -86,7 +87,7 @@ pub fn process_config(
     match sub_command {
         ConfigSubCommand::SetMinPinLength => {
             if let Some(ConfigSubCommandParams::SetMinPinLength(params)) = sub_command_params {
-                process_set_min_pin_length(persistent_store, pin_protocol_v1, params)
+                process_set_min_pin_length(persistent_store, params)
             } else {
                 Err(Ctap2StatusCode::CTAP2_ERR_MISSING_PARAMETER)
             }
--- a/src/ctap/config_command.rs
+++ b/src/ctap/config_command.rs
@@ -1,0 +1,269 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::check_pin_uv_auth_protocol;
+use super::command::AuthenticatorConfigParameters;
+use super::data_formats::{ConfigSubCommand, ConfigSubCommandParams, SetMinPinLengthParams};
+use super::pin_protocol_v1::PinProtocolV1;
+use super::response::ResponseData;
+use super::status_code::Ctap2StatusCode;
+use super::storage::PersistentStore;
+use alloc::vec;
+
+fn process_set_min_pin_length(
+    persistent_store: &mut PersistentStore,
+    pin_protocol_v1: &mut PinProtocolV1,
+    params: SetMinPinLengthParams,
+) -> Result<ResponseData, Ctap2StatusCode> {
+    let SetMinPinLengthParams {
+        new_min_pin_length,
+        min_pin_length_rp_ids,
+        force_change_pin,
+    } = params;
+    let store_min_pin_length = persistent_store.min_pin_length()?;
+    let new_min_pin_length = new_min_pin_length.unwrap_or(store_min_pin_length);
+    if new_min_pin_length < store_min_pin_length {
+        return Err(Ctap2StatusCode::CTAP2_ERR_PIN_POLICY_VIOLATION);
+    }
+    let mut force_change_pin = force_change_pin.unwrap_or(false);
+    if force_change_pin && persistent_store.pin_hash()?.is_none() {
+        return Err(Ctap2StatusCode::CTAP2_ERR_PIN_NOT_SET);
+    }
+    if let Some(old_length) = persistent_store.pin_code_point_length()? {
+        force_change_pin |= new_min_pin_length > old_length;
+    }
+    pin_protocol_v1.force_pin_change |= force_change_pin;
+    // TODO(kaczmarczyck) actually force a PIN change
+    persistent_store.set_min_pin_length(new_min_pin_length)?;
+    if let Some(min_pin_length_rp_ids) = min_pin_length_rp_ids {
+        persistent_store.set_min_pin_length_rp_ids(min_pin_length_rp_ids)?;
+    }
+    Ok(ResponseData::AuthenticatorConfig)
+}
+
+pub fn process_config(
+    persistent_store: &mut PersistentStore,
+    pin_protocol_v1: &mut PinProtocolV1,
+    params: AuthenticatorConfigParameters,
+) -> Result<ResponseData, Ctap2StatusCode> {
+    let AuthenticatorConfigParameters {
+        sub_command,
+        sub_command_params,
+        pin_uv_auth_param,
+        pin_uv_auth_protocol,
+    } = params;
+
+    if persistent_store.pin_hash()?.is_some() {
+        // TODO(kaczmarczyck) The error code is specified inconsistently with other commands.
+        check_pin_uv_auth_protocol(pin_uv_auth_protocol)
+            .map_err(|_| Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED)?;
+        let auth_param = pin_uv_auth_param.ok_or(Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED)?;
+        let mut config_data = vec![0xFF; 32];
+        config_data.extend(&[0x0D, sub_command as u8]);
+        if let Some(sub_command_params) = sub_command_params.clone() {
+            if !cbor::write(sub_command_params.into(), &mut config_data) {
+                return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+            }
+        }
+        if !pin_protocol_v1.verify_pin_auth_token(&config_data, &auth_param) {
+            return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID);
+        }
+    }
+
+    match sub_command {
+        ConfigSubCommand::SetMinPinLength => {
+            if let Some(ConfigSubCommandParams::SetMinPinLength(params)) = sub_command_params {
+                process_set_min_pin_length(persistent_store, pin_protocol_v1, params)
+            } else {
+                Err(Ctap2StatusCode::CTAP2_ERR_MISSING_PARAMETER)
+            }
+        }
+        _ => Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER),
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::super::command::AuthenticatorConfigParameters;
+    use super::*;
+    use crypto::rng256::ThreadRng256;
+
+    fn create_min_pin_config_params(
+        min_pin_length: u8,
+        min_pin_length_rp_ids: Option<Vec<String>>,
+    ) -> AuthenticatorConfigParameters {
+        let set_min_pin_length_params = SetMinPinLengthParams {
+            new_min_pin_length: Some(min_pin_length),
+            min_pin_length_rp_ids,
+            force_change_pin: None,
+        };
+        AuthenticatorConfigParameters {
+            sub_command: ConfigSubCommand::SetMinPinLength,
+            sub_command_params: Some(ConfigSubCommandParams::SetMinPinLength(
+                set_min_pin_length_params,
+            )),
+            pin_uv_auth_param: None,
+            pin_uv_auth_protocol: Some(1),
+        }
+    }
+
+    #[test]
+    fn test_process_set_min_pin_length() {
+        let mut rng = ThreadRng256 {};
+        let mut persistent_store = PersistentStore::new(&mut rng);
+        let key_agreement_key = crypto::ecdh::SecKey::gensk(&mut rng);
+        let pin_uv_auth_token = [0x55; 32];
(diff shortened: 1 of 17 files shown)


src/ctap/storage.rs
```diff
@@ -313,14 +312,33 @@ impl PersistentStore {
         Ok(Some(*array_ref![pin_hash, 0, PIN_AUTH_LENGTH]))
     }
 
-    /// Sets the PIN hash.
+    /// Returns the length of the currently set PIN if defined.
+    ///
+    /// This function is used starting with CTAP 2.1.
+    #[allow(dead_code)]
```
ia0: Why not prefix with underscore? Either it's used with some feature in which case we can protect with that feature, or it's not yet used and we can prefix with underscore, no?

src/ctap/storage.rs
```diff
@@ -313,14 +312,33 @@ impl PersistentStore {
...
+    pub fn set_pin(
         &mut self,
         pin_hash: &[u8; PIN_AUTH_LENGTH],
+        pin_code_point_length: u8,
     ) -> Result<(), Ctap2StatusCode> {
-        Ok(self.store.insert(key::PIN_HASH, pin_hash)?)
+        self.store.insert(key::PIN_HASH, pin_hash)?;
+        Ok(self
+            .store
+            .insert(key::PIN_CODE_POINT_LENGTH, &[pin_code_point_length])?)
```
ia0: Should those be in the same transaction?

Also, if we always set them together, why not have them in the same entry? First byte is `pin_code_point_length` followed by `pin_hash`.

src/ctap/storage.rs
```diff
@@ -914,16 +932,35 @@ mod test {
...
         assert_eq!(persistent_store.pin_hash().unwrap(), Some(pin_hash_1));
-        persistent_store.set_pin_hash(&pin_hash_2).unwrap();
+        assert_eq!(
+            persistent_store.pin_code_point_length().unwrap(),
+            Some(pin_length_1)
+        );
+        assert_eq!(
+            persistent_store.pin_code_point_length().unwrap(),
+            Some(pin_length_1)
+        );
```
ia0, line 1007: I don't know why I duplicated all asserts. I guess it was to test that reading the value doesn't modify it. I think we should rely on the persistent store tests for those persistent store properties and only test CTAP storage properties here. So I would just deduplicate each line.

ia0 (commented):
Just reviewed the storage. The main question is whether we need to store the pin hash and pin length separately. If we always write them together then it's strictly better (in terms of flash lifetime) to write them in the same entry. This would be similar to the master keys for example.

src/ctap/storage.rs
kaczmarczyck: Having the same entry solves the transaction problem!

src/ctap/storage.rs
kaczmarczyck: Don't know why I did that. Fixed.

src/ctap/storage.rs
kaczmarczyck: Removed.

src/ctap/storage.rs
```diff
@@ -303,24 +302,41 @@ impl PersistentStore {
...
+    pub fn pin_code_point_length(&self) -> Result<Option<u8>, Ctap2StatusCode> {
+        let pin_properties = match self.store.find(key::PIN_PROPERTIES)? {
+            None => return Ok(None),
+            Some(pin_properties) => pin_properties,
+        };
+        if pin_properties.len() != PIN_AUTH_LENGTH + 1 {
+            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        }
+        Ok(Some(pin_properties[0]))
+    }
```
ia0, line 378: We can have a private `fn pin_properties(&self) -> Result<PinProperties, Ctap2StatusCode>` to factorize both functions.

```rust
pub struct PinProperties {
    hash: [u8; PIN_AUTH_LENGTH],
    code_point_length: u8,
}
```

Then `set_pin` could also take a `PinProperties` if it's convenient at call-site.

ia0 (commented):
Thinking about it, this change breaks compatibility if a pin was set. Since we're not in v1 that's probably not an issue. Otherwise we might want to update the `pin_properties` function such  that if the value is PIN_AUTH_LENGTH only, then we use a default value for the code length.

src/ctap/storage.rs
kaczmarczyck: `PinProperties` done. Since the getters are version dependent and separate, I decided to also leave the setter as is and not expose `PinProperties`.

ia0 (dismissed):
LGTM for storage.

kaczmarczyck:
I just had a thought about backwards compatibility. I think I like it better if we fail for CTAP 2.1 authenticators if we don't read the length byte. We will write to the storage with forward compatible lengths, so it's only a problem if you have pre-1.0 storages and update to CTAP 2.1.

And if we ever make CTAP 2.1 the default and deprecate 2.0, then this makes it easier to remove legacy.

src/ctap/storage.rs
```diff
@@ -301,26 +310,53 @@ impl PersistentStore {
...
-        if pin_hash.len() != PIN_AUTH_LENGTH {
-            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        const PROPERTIES_LENGTH: usize = PIN_AUTH_LENGTH + 1;
+        match pin_properties.len() {
+            #[cfg(not(feature = "with_ctap2_1"))]
+            PIN_AUTH_LENGTH => Ok(Some(PinProperties {
+                hash: *array_ref![pin_properties, 0, PIN_AUTH_LENGTH],
+                // Return a dummy value for backwards compatibility.
+                #[cfg(feature = "with_ctap2_1")]
+                code_point_length: 0,
```
ia0: Those 3 lines can be removed now.

ia0 (dismissed):

src/ctap/storage.rs
kaczmarczyck: Thanks, removed.

ia0 (dismissed):

kaczmarczyck:
Thanks for the review so far! If I have to do changes on the storage side, I'll come back to you. Otherwise, all future changes are to be reviewed by @jmichelp .

jmichelp:
Considering the amount of complexity that CTAP2.1 adds compared to CTAP2.0 and that now we have separate branches, I'm wondering if we shouldn't simply get rid of the feature `with_ctap2_1` in the `develop` branch.

CTAP2.0 shouldn't change a lot in the upcoming months which means that people interested in CTAP2.0 can use a pre-built firmware or checkout the corresponding tagged commit.
Worst case scenario is that we end up with `stable-2.0`, `stable-2.1` branches and corresponding staging bugfix branches but we have a better readability in the code.

WDYT?

kaczmarczyck:
Sure! I'll submit a PR to do that. Additionally, I can move this commands code to a separate file like I did for credential management in the other PR?

kaczmarczyck:
This is now on top of #248 to make it easier to review without the 2.1 flag. No changes to storage, so no action required for @ia0 I think. Also, the command is moved to a separate file, that is likely going to grow with more subcommands.

The failing tests are because the upstream workflows are not up to date with those in #248. To make that cleaner, I'll send a PR for the workflows only first.

src/ctap/config_command.rs
```diff
@@ -0,0 +1,271 @@
+// Copyright 2020 Google LLC
```
jmichelp: You need to add 2021 now :)

src/ctap/config_command.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+    let mut force_change_pin = force_change_pin.unwrap_or(false);
+    if force_change_pin && persistent_store.pin_hash()?.is_none() {
+        return Err(Ctap2StatusCode::CTAP2_ERR_PIN_NOT_SET);
+    }
+    if let Some(old_length) = persistent_store.pin_code_point_length()? {
+        force_change_pin |= new_min_pin_length > old_length;
+    }
+    pin_protocol_v1.force_pin_change |= force_change_pin;
+    // TODO(kaczmarczyck) actually force a PIN change
```
jmichelp: Maybe that's part of the TODO but shouldn't this `force_change_pin` be part of the persistent storage in case the key is power cycled in the middle?

src/ctap/config_command.rs
```diff
@@ -0,0 +1,271 @@
...
+        pin_uv_auth_param,
+        pin_uv_auth_protocol,
+    } = params;
+
+    if persistent_store.pin_hash()?.is_some() {
+        // TODO(kaczmarczyck) The error code is specified inconsistently with other commands.
+        check_pin_uv_auth_protocol(pin_uv_auth_protocol)
+            .map_err(|_| Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED)?;
+        let auth_param = pin_uv_auth_param.ok_or(Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED)?;
+        let mut config_data = vec![0xFF; 32];
```
jmichelp, line 76: where does this 32 comes from? Fixed length by the spec? Alignment? Other? If it's related to a specific field length in the spec, I suggest defining the const and use it

src/ctap/config_command.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    #[test]
+    fn test_process_config_vendor_prototype() {
+        let mut rng = ThreadRng256 {};
+        let mut persistent_store = PersistentStore::new(&mut rng);
+        let key_agreement_key = crypto::ecdh::SecKey::gensk(&mut rng);
+        let pin_uv_auth_token = [0x55; 32];
+        let mut pin_protocol_v1 = PinProtocolV1::new_test(key_agreement_key, pin_uv_auth_token);
+
+        let config_params = AuthenticatorConfigParameters {
```
jmichelp, line 259: Now that the specification implements this, we could potentially merge our vendor specific command for configuration into this one to inject private key, certificate as well as removing JTAG access.

src/ctap/mod.rs
```diff
@@ -117,6 +122,17 @@ pub const ES256_CRED_PARAM: PublicKeyCredentialParameter = PublicKeyCredentialPa
 // - Some(CredentialProtectionPolicy::UserVerificationRequired)
 const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
 
+// Checks the PIN protocol parameter against all supported versions.
+pub fn check_pin_uv_auth_protocol(
+    pin_uv_auth_protocol: Option<u64>,
+) -> Result<(), Ctap2StatusCode> {
+    match pin_uv_auth_protocol {
+        Some(PIN_PROTOCOL_VERSION) => Ok(()),
+        Some(_) => Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID),
```
jmichelp: Considering the error code is the same, can't you simply write:
```rust
Some(PIN_PROTOCOL_VERSION) => Ok(()),
_ => Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID),
```
?

src/ctap/mod.rs
```diff
@@ -427,22 +442,29 @@ where
...
-            }
-            (extensions.hmac_secret, cred_protect)
-        } else {
-            (false, DEFAULT_CRED_PROTECT)
-        };
+        let rp_id = rp.rp_id;
+        let (use_hmac_extension, cred_protect_policy, min_pin_length) =
+            if let Some(extensions) = extensions {
+                let mut cred_protect = extensions.cred_protect;
+                if cred_protect.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
```
jmichelp, line 448: This code block seems equivalent to
```rust
let cred_protect = Some(cmp::max(
  extensions.cred_protect.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional),
  DEFAULT_CRED_PROTECT.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
));
```

src/ctap/mod.rs
```diff
@@ -1028,6 +1059,49 @@ mod test {
...
+                    0x12, 0x55, 0x86, 0xCE, 0x19, 0x47, flags, 0x00, 0x00, 0x00,
+                ];
+                expected_auth_data.push(INITIAL_SIGNATURE_COUNTER as u8);
+                expected_auth_data.extend(expected_aaguid);
+                expected_auth_data.extend(&[0x00, expected_credential_id_size]);
+                assert_eq!(
+                    auth_data[0..expected_auth_data.len()],
+                    expected_auth_data[..]
+                );
+                /*assert_eq!(
```
jmichelp: Why is this code commented out?

src/ctap/pin_protocol_v1.rs
```diff
@@ -170,6 +172,7 @@ pub struct PinProtocolV1 {
     consecutive_pin_mismatches: u8,
     permissions: u8,
     permissions_rp_id: Option<String>,
+    pub force_pin_change: bool,
 }
 
 impl PinProtocolV1 {
```
jmichelp: Do you plan to define a trait for PinProtocol at some point?
By brushing the 2.1 spec, I see they now define also PinProtocolV2.

src/ctap/config_command.rs
kaczmarczyck: It's taken from the [specification](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#authenticatorConfig), step 4.2. It's only used in one place, and they are nameless constants in the specification. WDYT?

src/ctap/config_command.rs
kaczmarczyck: The vendor specific command range is also [still mentioned](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#commands). Moving it to a subcommand here means one difference: If a PIN is set on the device, you'd have to get a token first. Do we want that?

Outside of the question if we want it, the deploy script would be much more difficult, and shouldn't be part of this PR.

src/ctap/mod.rs
kaczmarczyck: There is a slight difference. Practically, `Some(CredentialProtectionPolicy::UserVerificationOptional)` and `None` should be equivalent, but I need to support both for backwards compatibility. And I rather return `None` in case the platform doesn't speak `credProtect`. Because having `Some` policy means we return the extension output. I don't think the current version is elegant, but I didn't find a better alternative so far.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: Yes, that is the plan indeed. That was one of the reasons to factor out the PinProtocolV1 implementation, to make it easy to add a PinProtocolV2.

src/ctap/config_command.rs
kaczmarczyck: Done, for this and all other files in CTAP.

src/ctap/config_command.rs
kaczmarczyck: I didn't put much thought into the stub, but I can see it's misleading. It's now a storage stub instead.

src/ctap/mod.rs
kaczmarczyck: That's a leftover from before the `#cfg` for CTAP 2.1 was dropped. Fixed.

src/ctap/mod.rs
kaczmarczyck: Forgot to remove, deleted now. This is the random part of the returned data.

src/ctap/config_command.rs
jmichelp: Ok then let's leave it as-is and maybe just put a comment that this directly comes from the specification without any explanation/justification.

src/ctap/mod.rs
jmichelp: It's just the way Github takes code snippet that is confusing here I think.
The equivalent code block I referred to is:
```rust
let mut cred_protect = extensions.cred_protect;
if cred_protect.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
    < DEFAULT_CRED_PROTECT
        .unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
{
    cred_protect = DEFAULT_CRED_PROTECT;
}
```

which should be a perfect equivalent of:
```rust
let cred_protect = Some(cmp::max(
  extensions.cred_protect.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional),
  DEFAULT_CRED_PROTECT.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
));
```

kaczmarczyck: This is what I thought you mean, and my comment applies. I want this piece of code to always return `None` instead of `Some(CredentialProtectionPolicy::UserVerificationOptional)`.

src/ctap/config_command.rs
kaczmarczyck: Comment added.

jmichelp (approved):

PR #247, opened 2020-12-22
Add erase_storage application example

ia0:
Permits to update the firmware to an incompatible storage version or to deliberately fully erase the storage (which is different from a CTAP reset, which keeps lifetime tracking and attestation private key and certificate):

```
./deploy.py --board=nrf52840_dongle_dfu --erase_storage --programmer=nordicdfu
./deploy.py --board=nrf52840_dongle_dfu --opensk --programmer=nordicdfu
```

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -947,7 +947,16 @@ if __name__ == "__main__":
       dest="application",
       action="store_const",
       const="store_latency",
-      help=("Compiles and installs the store_latency example."))
+      help=("Compiles and installs the store_latency example which print "
+            "latency statistics of the persistent store library."))
+  apps_group.add_argument(
+      "--erase_storage",
+      dest="application",
+      action="store_const",
+      const="erase_storage",
+      help=("Compiles and installs the erase_storage example which erases "
+            "the storage. During operation the dongle red light is on. Once "
+            "the operation is completed the dongle green light is on."))
   apps_group.add_argument(
       "--panic_test",
       dest="application",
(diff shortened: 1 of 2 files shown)


deploy.py
```diff
@@ -947,7 +947,16 @@ def main(args):
...
+      help=("Compiles and installs the store_latency example which print "
+            "latency statistics of the persistent store library."))
+  apps_group.add_argument(
+      "--erase_storage",
+      dest="application",
+      action="store_const",
+      const="erase_storage",
+      help=("Compiles and installs the erase_storage example which erases "
+            "the storage. During operation the dongle red light is on. Once "
+            "the operation is completed the dongle green light is on."))
```
kaczmarczyck, line 959: Does this generalize to the dev kit? I'd guess it's just different LEDs instead of different colors? Might make sense to change the comment, so dev kit users know they are supported.

ia0: If we plan to keep the `--clear-storage` flag, then dev-kit users should continue using it (as well as dongle users with jlink), it's much simpler to use since it requires only one firmware flash.

The `erase_storage` binary is only for dongle users that only have DFU to flash firmware. This binary is a work-around for them not being able to use `--clear-storage`.

kaczmarczyck: Okay, gotcha!

kaczmarczyck (dismissed):

kaczmarczyck (approved):

PR #248, opened 2021-01-06
Remove CTAP 2.1 flag

kaczmarczyck:
In our new develop branch, we drop support for CTAP 2.0 separately from CTAP 2.1.

This is related to #106 and makes #128 much less important since the amount of cfg statements is significantly reduced. (I leave it to @jmichelp if we want to close #128.)

- [x] Tests pass
- [x] Appropriate changes to README are included in PR

PR #249, opened 2021-01-07
Updates workflows to remove the CTAP 2.1 flag

kaczmarczyck:
Necessary to submit before #248 to not fail the old tests. Just removes the workflows for 2.1, so #248 should be submitted right after this one.

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -947,7 +947,16 @@ if __name__ == "__main__":
       dest="application",
       action="store_const",
       const="store_latency",
-      help=("Compiles and installs the store_latency example."))
+      help=("Compiles and installs the store_latency example which print "
+            "latency statistics of the persistent store library."))
+  apps_group.add_argument(
+      "--erase_storage",
+      dest="application",
+      action="store_const",
+      const="erase_storage",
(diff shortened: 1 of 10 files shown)


deploy.py
```diff
@@ -947,7 +947,16 @@ def main(args):
       dest="application",
       action="store_const",
       const="store_latency",
-      help=("Compiles and installs the store_latency example."))
+      help=("Compiles and installs the store_latency example which print "
```
jmichelp, line 950: nit: prints

examples/erase_storage.rs
```diff
@@ -0,0 +1,53 @@
+// Copyright 2020 Google LLC
```
jmichelp, line 1: Why is this file part of this PR?

kaczmarczyck:
Aaah sorry old habit. This PR goes into stable, not develop. My bad. Closing and opening a new PR.

deploy.py
kaczmarczyck: Will be fixed in a different PR :)

PR #250, opened 2021-01-08
removes all occurences of CTAP 2.1 flags from workflows

kaczmarczyck:
Necessary to submit before #248 to not fail the old tests. Just removes the workflows for 2.1, so #248 should be submitted right after this one.

- [x] Tests pass


kaczmarczyck:
This is the new version of #249 .

jmichelp (approved):

PR #252, opened 2021-01-08
Remove CTAP 2.1 flag

### LINKED ISSUE #128: #cfg in macros
As noted in #127 , we should support `#cfg` inside macro calls, i.e. `cbor_map`.

kaczmarczyck:
In our new develop branch, we drop support for CTAP 2.0 separately from CTAP 2.1.

This is related to #106 and makes #128 much less important since the amount of cfg statements is significantly reduced. (I leave it to @jmichelp if we want to close #128.)

- [x] Tests pass

This is a copy of #248 and requests against the correct branch. It is based on #250 and is implemented on top of its changes.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -1016,14 +1016,11 @@ mod test {
         let info_reponse = ctap_state.process_command(&[0x04], DUMMY_CHANNEL_ID, DUMMY_CLOCK_VALUE);
 
         let mut expected_response = vec![0x00, 0xAA, 0x01];
-        // The difference here is a longer array of supported versions.
-        let mut version_count = 0;
-        // CTAP 2.0 and 2.1 are always supported
-        version_count += 2;
+        // The version array differs with CTAP1, always including 2.0 and 2.1.
+        #[cfg(not(feature = "with_ctap1"))]
+        let version_count = 2;
         #[cfg(feature = "with_ctap1")]
-        {
-            version_count += 1;
-        }
+        let version_count = 3;
         expected_response.push(0x80 + version_count);
         #[cfg(feature = "with_ctap1")]
         expected_response.extend(&[0x66, 0x55, 0x32, 0x46, 0x5F, 0x56, 0x32]);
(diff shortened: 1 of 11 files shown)


src/ctap/mod.rs
```diff
@@ -1036,42 +1015,31 @@ mod test {
...
-        #[cfg(feature = "with_ctap2_1")]
         let mut expected_response = vec![0x00, 0xAA, 0x01];
-        #[cfg(not(feature = "with_ctap2_1"))]
-        let mut expected_response = vec![0x00, 0xA6, 0x01];
         // The difference here is a longer array of supported versions.
         let mut version_count = 0;
-        // CTAP 2 is always supported
-        version_count += 1;
+        // CTAP 2.0 and 2.1 are always supported
+        version_count += 2;
```
jmichelp: why not `let mut version_count = 2;` then?

src/ctap/storage.rs
```diff
@@ -374,11 +366,10 @@ impl PersistentStore {
                 _deserialize_min_pin_length_rp_ids(&value)
             });
         debug_assert!(rp_ids.is_some());
-        Ok(rp_ids.unwrap_or(vec![]))
+        Ok(rp_ids.unwrap_or_default())
     }
 
     /// Sets the list of RP IDs that are used to check if reading the minimum PIN length is allowed.
-    #[cfg(feature = "with_ctap2_1")]
     pub fn _set_min_pin_length_rp_ids(
```
jmichelp, line 373: why are these (this function and the next 2 function show in this diff) unused? Is it WAI?

kaczmarczyck: They are used in #246. Basically, implementing new commands is pretty hard to split into reasonably sized PRs.

src/ctap/mod.rs
kaczmarczyck: Because if we compile without CTAP1, we don't mutate a `mut` variable. This style was the most readable when we still had multiple version flags. I changed it to not use `mut`.

jmichelp (approved):

Issue #90, opened 2020-03-26
OpenSSH key generation fails if OpenSK has a pin set.


Confined-:
## Expected Behavior
Key generation proceeds after entering the correct pin at the prompt.

## Actual Behavior
After entering the set pin, openssh errors out with `Key enrollment failed: requested feature not supported` 

`brittle@archdesktop .ssh]$ ssh-keygen -t ecdsa-sk -f ~/.ssh/id_ecdsa_sk`
`Generating public/private ecdsa-sk key pair.`
`You may need to touch your authenticator to authorize key generation.`
`Enter PIN for authenticator:` 
`Key enrollment failed: requested feature not supported`

If OpenSK does not have a pin set, the key pair is generated as expected.

## Steps to Reproduce the Problem

1. Set a pin on OpenSK. I used [https://demo.yubico.com/playground](url) with passwordless login on Chromium to set the pin. 
2. Try to create a fido backed key pair with OpenSSH. 
`ssh-keygen -t ecdsa-sk -f ~/.ssh/id_ecdsa_sk`

## Specifications

- Version: 284b1426ecaf71d7a9b2e1b445ed55953efbd370
- Platform: Arch Linux

Confined-:
A more verbose log.

> [brittle@archdesktop ~]$ ssh-keygen -t ecdsa-sk -f ~/.ssh/id_ecdsa_sk -v
Generating public/private ecdsa-sk key pair.
You may need to touch your authenticator to authorize key generation.
debug1: start_helper: starting /usr/lib/ssh/ssh-sk-helper 
debug1: sshsk_enroll: provider "internal", device "(null)", application "ssh:", userid "(null)", flags 0x01, challenge len 0
debug1: sshsk_enroll: using random challenge
debug1: ssh_sk_enroll: using device /dev/hidraw0
debug1: ssh_sk_enroll: fido_dev_make_cred: FIDO_ERR_PIN_REQUIRED
debug1: sshsk_enroll: provider "internal" returned failure -3
debug1: ssh-sk-helper: Enrollment failed: incorrect passphrase supplied to decrypt private key
debug1: ssh-sk-helper: reply len 8
debug1: client_converse: helper returned error -43
Enter PIN for authenticator: 
debug1: start_helper: starting /usr/lib/ssh/ssh-sk-helper 
debug1: sshsk_enroll: provider "internal", device "(null)", application "ssh:", userid "(null)", flags 0x01, challenge len 0 with-pin
[...]

PR #253, opened 2021-01-08
COSE conversion for ECDH and ECDSA

kaczmarczyck:
For the bigger picture, please see #107 (reviewed by @ia0). We need COSE conversion for both ECDH and ECDSA keys. This PR unifies these conversions, and makes the crypto library purer again.

- [x] Tests pass

### DIFF
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -85,7 +85,7 @@ impl PubKey {
         self.p.to_bytes_uncompressed(bytes);
     }
 
-    /// Creates a new PubKey from their coordinates.
+    /// Creates a new PubKey from its coordinates on the elliptic curve.
     pub fn from_coordinates(x: &[u8; NBYTES], y: &[u8; NBYTES]) -> Option<PubKey> {
         PointP256::new_checked_vartime(Int256::from_bin(x), Int256::from_bin(y))
             .map(|p| PubKey { p })
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -62,8 +62,10 @@ impl SecKey {
         // - https://www.secg.org/sec1-v2.pdf
     }
 
-    // DH key agreement method defined in the FIDO2 specification, Section 5.5.4. "Getting
-    // sharedSecret from Authenticator"
+    /// Creates a shared key using the Diffie Hellman key agreement.
+    ///
+    /// The key agreement is defined in the FIDO2 specification,
+    /// Section 6.5.5.4. "Obtaining the Shared Secret"
     pub fn exchange_x_sha256(&self, other: &PubKey) -> [u8; 32] {
         let p = self.exchange_raw(other);
         let mut x: [u8; 32] = [Default::default(); 32];
@@ -83,11 +85,13 @@ impl PubKey {
         self.p.to_bytes_uncompressed(bytes);
     }
 
+    /// Creates a new PubKey from their coordinates.
     pub fn from_coordinates(x: &[u8; NBYTES], y: &[u8; NBYTES]) -> Option<PubKey> {
         PointP256::new_checked_vartime(Int256::from_bin(x), Int256::from_bin(y))
             .map(|p| PubKey { p })
     }
 
+    /// Writes the coordinates into the passed in arrays.
     pub fn to_coordinates(&self, x: &mut [u8; NBYTES], y: &mut [u8; NBYTES]) {
         self.p.getx().to_int().to_bin(x);
         self.p.gety().to_int().to_bin(y);
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -738,7 +738,8 @@ impl TryFrom<CoseKey> for ecdh::PubKey {
 
         // Since algorithm can be used for different COSE key types, we check
         // whether the current type is correct for ECDH. For an OpenSSH bugfix,
-        // the algorithm ES256_ALGORITHM is allowed here too. See #90.
+        // the algorithm ES256_ALGORITHM is allowed here too.
+        // https://github.com/google/OpenSK/issues/90
         if algorithm != CoseKey::ECDH_ALGORITHM && algorithm != ES256_ALGORITHM {
             return Err(Ctap2StatusCode::CTAP2_ERR_UNSUPPORTED_ALGORITHM);
         }
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -23,16 +23,8 @@ use crypto::{ecdh, ecdsa};
 #[cfg(test)]
 use enum_iterator::IntoEnumIterator;
 
-// This is the algorithm specifier that is supposed to be used in a COSE key
-// map in ECDH. CTAP requests -25 which represents ECDH-ES + HKDF-256 here:
-// https://www.iana.org/assignments/cose/cose.xhtml#algorithms
-const ECDH_ALGORITHM: i64 = -25;
-// This is the identifier used for ECDSA and ECDH in OpenSSH.
+// Used as the identifier for ECDSA in assertion signatures and COSE.
 const ES256_ALGORITHM: i64 = -7;
-// The COSE key parameter behind map key 1.
-const EC2_KEY_TYPE: i64 = 2;
-// The COSE key parameter behind map key -1.
-const P_256_CURVE: i64 = 1;
 
 // https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialrpentity
 #[cfg_attr(any(test, feature = "debug_ctap"), derive(Debug, PartialEq))]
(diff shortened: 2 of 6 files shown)


libraries/crypto/src/ecdh.rs
```diff
@@ -83,11 +85,13 @@ impl PubKey {
         self.p.to_bytes_uncompressed(bytes);
     }
 
+    /// Creates a new PubKey from their coordinates.
```
gendx: I'd suggest "from their coordinates" => "from its coordinates on the elliptic curve"

src/ctap/data_formats.rs
```diff
@@ -631,26 +631,39 @@ const ES256_ALGORITHM: i64 = -7;
...
+        } else {
+            Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)
+        }
+    }
+}
+
+fn cose_key_from_bytes(x_bytes: [u8; ecdh::NBYTES], y_bytes: [u8; ecdh::NBYTES]) -> CoseKey {
+    let x_byte_cbor: cbor::Value = cbor_bytes_lit!(&x_bytes);
+    let y_byte_cbor: cbor::Value = cbor_bytes_lit!(&y_bytes);
+    // TODO(kaczmarczyck) do not write optional parameters, spec is unclear
```
gendx: Note: IIRC `cbor_map_options` won't write anything for values equal to `None`. Although in this case I don't see any `None`, so you can probably use `cbor_map!` directly.

src/ctap/data_formats.rs
```diff
@@ -631,26 +631,39 @@ const ES256_ALGORITHM: i64 = -7;
 const EC2_KEY_TYPE: i64 = 2;
 const P_256_CURVE: i64 = 1;
 
+impl TryFrom<cbor::Value> for CoseKey {
```
gendx: I don't think that having the `CoseKey` type publicly expose the underlying `BTreeMap` is a robust design, because anyone can construct a `CoseKey` from a `BTreeMap` which contains something else. Likewise, publicly exposing this `TryFrom` conversion allows anyone to construct a `CoseKey` from an arbitrary `BTreeMap` without any validation of the input.

All we want to expose publicly is the `impl From<ecdh::PubKey> for CoseKey`, as well as a read-only API to convert the `CoseKey` into a `cbor::Value` (without giving access to the underlying `BTreeMap`), i.e. an API that does the `cbor::Value::Map(CoseKey::from(pk).0)` conversion done elsewhere.

For now, I only see this `TryFrom` being used in `cose_key_from_bytes`, so It'd be better to rather have this code inlined there.

src/ctap/data_formats.rs
```diff
@@ -631,26 +631,39 @@ const ES256_ALGORITHM: i64 = -7;
...
+    type Error = Ctap2StatusCode;
+
+    fn try_from(cbor_value: cbor::Value) -> Result<Self, Ctap2StatusCode> {
+        if let cbor::Value::Map(cose_map) = cbor_value {
+            Ok(CoseKey(cose_map))
```
gendx: Here there is no validation that the `cose_map` is indeed structured as a COSE map. But this `TryFrom` is publicly available outside of this module, which is problematic.

src/ctap/data_formats.rs
```diff
@@ -631,26 +631,39 @@ const ES256_ALGORITHM: i64 = -7;
...
+        } else {
+            Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)
+        }
+    }
+}
+
+fn cose_key_from_bytes(x_bytes: [u8; ecdh::NBYTES], y_bytes: [u8; ecdh::NBYTES]) -> CoseKey {
```
gendx: Might be worth explaining that this function is used to convert both `ecdh::PubKey` and `ecdsa::PubKey` into a `CoseKey`.

src/ctap/mod.rs
```diff
@@ -534,11 +534,9 @@ where
         }
         auth_data.extend(vec![0x00, credential_id.len() as u8]);
         auth_data.extend(&credential_id);
-        let cose_key = match pk.to_cose_key() {
-            Some(cose_key) => cose_key,
-            None => return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR),
-        };
-        auth_data.extend(cose_key);
+        if !cbor::write(cbor::Value::Map(CoseKey::from(pk).0), &mut auth_data) {
```
gendx: Here it should rather be `cbor::write(cbor::Value::from(CoseKey::from(pk)), &mut auth_data)` or `cbor::write(CoseKey::from(pk).into(), &mut auth_data)`, without exposing `CoseKey`'s data publicly.

gendx (changes requested):

libraries/crypto/src/ecdh.rs
kaczmarczyck: Done

src/ctap/data_formats.rs
kaczmarczyck: The function is removed.

src/ctap/data_formats.rs
kaczmarczyck: I changed the whole `CoseKey` data type. This PR was initially only about the crypto library changes. But this seems to be a good time to tackle one of my TODOs to implement a proper `CoseKey` data format since it's usage won't change much anymore.

src/ctap/data_formats.rs
kaczmarczyck: Done in a new conversion.

src/ctap/data_formats.rs
kaczmarczyck: Done for the new conversion function.

src/ctap/mod.rs

src/ctap/data_formats.rs
```diff
@@ -17,12 +17,23 @@ use alloc::collections::BTreeMap;
...
+// This is the algorithm specifier that is supposed to be used in a COSE key
+// map in ECDH. CTAP requests -25 which represents ECDH-ES + HKDF-256 here:
+// https://www.iana.org/assignments/cose/cose.xhtml#algorithms
+const ECDH_ALGORITHM: i64 = -25;
+// This is the identifier used for ECDSA and ECDH in OpenSSH.
+const ES256_ALGORITHM: i64 = -7;
+// The COSE key parameter behind map key 1.
+const EC2_KEY_TYPE: i64 = 2;
+// The COSE key parameter behind map key -1.
+const P_256_CURVE: i64 = 1;
```
gendx: All of these constants are specific to the COSE specification. It would be worth to either put them in the `CoseKey` namespace (i.e. `impl CoseKey { const FOO: i64 = bar }`, or prefix them with `COSE_`.

src/ctap/data_formats.rs
```diff
@@ -676,22 +670,76 @@ impl TryFrom<CoseKey> for ecdh::PubKey {
...
-        ecdh::PubKey::from_coordinates(x_array_ref, y_array_ref)
+    fn try_from(cose_key: CoseKey) -> Result<Self, Ctap2StatusCode> {
+        let CoseKey {
+            x_bytes,
+            y_bytes,
+            algorithm,
+        } = cose_key;
+
+        // Since algorithm can be used for different COSE key types, we check
+        // whether the current type is correct for ECDH. For an OpenSSH bugfix,
```
gendx, line 740: Can you document which OpenSSH bug you're referring to?

src/ctap/data_formats.rs
```diff
@@ -1322,7 +1370,84 @@ mod test {
...
+                -1 => P_256_CURVE,
+                -2 => [0u8; 32],
+                -3 => [0u8; 32],
+            };
+            let cose_key = CoseKey::try_from(cbor_value.clone()).unwrap();
+            let created_cbor_value = cbor::Value::from(cose_key);
+            assert_eq!(created_cbor_value, cbor_value);
+        }
+
+        let cbor_value = cbor_map! {
```
gendx, line 1394: Nit: It's usually better to have many small tests than one large test, so that:
- if the first assertion fails, it doesn't block the rest of the test cases from running,
- it's clear at first glance what failed and what succeeded, allowing easier debugging and bug fixing if some test failed.

gendx (changes requested):

src/ctap/data_formats.rs
kaczmarczyck: Sounds good, split!

src/ctap/data_formats.rs
kaczmarczyck: I mentioned #90 in the comment.

src/ctap/data_formats.rs
kaczmarczyck: I left `ES256_ALGORITHM` on top since it's used across in `SignatureAlgorithm` and `CoseKey`. All other constants are moved into `impl CoseKey`.

src/ctap/data_formats.rs
gendx: It might be preferable to expand it to the full URL for clarity: `https://github.com/google/OpenSK/issues/90`

gendx (dismissed):

gendx (approved):

PR #254, opened 2021-01-11
Storage management

kaczmarczyck:
Fixes #<issue_number_goes_here>

Implements storage changes necessary for credential management. The first commit is part of #107. The second commit implements changes suggested in [this comment](https://github.com/google/OpenSK/pull/107#discussion_r553940602).

`delete_credential` and `update_credential` will be used when the command is implemented (see #107 for reference). The plan is to send a new PR for a rebased #107 after #253 and this PR are submitted.

- [x] Tests pass

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -761,11 +761,23 @@ where
                 vec![],
             )
         } else {
-            let mut stored_credentials =
-                self.persistent_store.filter_credentials(&rp_id, !has_uv)?;
-            stored_credentials.sort_unstable_by_key(|c| c.1);
-            let mut stored_credentials: Vec<usize> =
-                stored_credentials.into_iter().map(|c| c.0).collect();
+            let mut iter_result = Ok(());
+            let iter = self.persistent_store.iter_credentials(&mut iter_result)?;
+            let mut stored_credentials: Vec<(usize, u64)> = iter
+                .filter_map(|(key, credential)| {
+                    if credential.rp_id == rp_id && (has_uv || credential.is_discoverable()) {
+                        Some((key, credential.creation_order))
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+            iter_result?;
+            stored_credentials.sort_unstable_by_key(|&(_key, order)| order);
+            let mut stored_credentials: Vec<usize> = stored_credentials
+                .into_iter()
+                .map(|(key, _order)| key)
+                .collect();
             let credential = stored_credentials
                 .pop()
                 .map(|key| self.persistent_store.get_credential(key))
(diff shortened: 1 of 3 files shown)


src/ctap/storage.rs
```diff
@@ -126,22 +168,17 @@ impl PersistentStore {
...
-        if let Some(cred) = &result {
-            let user_verification_required = cred.cred_protect_policy
-                == Some(CredentialProtectionPolicy::UserVerificationRequired);
-            if check_cred_protect && user_verification_required {
-                return Ok(None);
-            }
+        let credential = match self.find_credential_item(credential_id) {
+            Err(Ctap2StatusCode::CTAP2_ERR_NO_CREDENTIALS) => return Ok(None),
+            Err(e) => return Err(e),
+            Ok(credential) => credential.1,
```
ia0: I think it's more readable to have:

```rust
Ok((_, credential)) => credential,
// or
Ok((_key, credential)) => credential,
```

src/ctap/storage.rs
```diff
@@ -196,32 +233,60 @@ impl PersistentStore {
...
+        credential.user_name = user.user_name;
+        credential.user_display_name = user.user_display_name;
+        credential.user_icon = user.user_icon;
+        let value = serialize_credential(credential)?;
+        Ok(self.store.insert(key, &value)?)
+    }
+
     /// Returns the list of matching credentials.
     ///
     /// Does not return credentials that are not discoverable if `check_cred_protect` is set.
```
ia0: Let's update the documentation to say that the function returns pairs of key and creation order. But this is a bit weird since it looks quite custom logic. Shouldn't we inline this logic in the only place where `filter_credentials` is used?

src/ctap/mod.rs
```diff
@@ -749,36 +755,31 @@ where
...
+                self.get_any_credential_from_allow_list(allow_list, &rp_id, &rp_id_hash, has_uv)?,
+                vec![],
+            )
         } else {
-            self.persistent_store.filter_credential(&rp_id, !has_uv)?
+            let mut stored_credentials =
+                self.persistent_store.filter_credentials(&rp_id, !has_uv)?;
+            stored_credentials.sort_unstable_by_key(|c| c.1);
+            let mut stored_credentials: Vec<usize> =
+                stored_credentials.into_iter().map(|c| c.0).collect();
```
ia0: Instead of `|c| c.x` shouldn't we use `|(key, creation_order)| key or creation_order` for better readability?

src/ctap/mod.rs
```diff
@@ -848,13 +850,19 @@ where
...
                 max_credential_id_length: Some(CREDENTIAL_ID_SIZE as u64),
                 transports: Some(vec![AuthenticatorTransport::Usb]),
                 algorithms: Some(vec![ES256_CRED_PARAM]),
                 default_cred_protect: DEFAULT_CRED_PROTECT,
                 min_pin_length: self.persistent_store.min_pin_length()?,
                 firmware_version: None,
+                max_cred_blob_length: None,
+                // TODO(kaczmarczyck) update when extension is implemented
+                max_rp_ids_for_set_min_pin_length: None,
+                remaining_discoverable_credentials: Some(
```
ia0, line 875: Just making sure this is alright with information leakage here. On one side we only expose some credentials according to some permissions and here we expose the count (including those that are not "visible") according to the same permissions?

ia0 (dismissed):

src/ctap/mod.rs
kaczmarczyck: The general idea is to protect the relationship to a specific relying party. It should be impossible to somehow call `has_credential(rp_id)`. Your attack on `GetInfo` only let's you check if a credential was created, not for whom. If you are in a position to guess that relying party, this extra bit probably doesn't help much.

ia0: I see, all good then.

src/ctap/storage.rs
kaczmarczyck: Done for all cases.

src/ctap/storage.rs
kaczmarczyck: `filter_credentials` is moved to the call side.

src/ctap/mod.rs
kaczmarczyck: Done.

ia0 (approved):

Issue #184, opened 2020-10-16
Persistent store


ia0:
- Do not panic if store doesn't initialize, just fail all storage operations. Done in [opt_store](https://github.com/google/OpenSK/compare/develop...ia0:opt_store) branch but increases binary size from 124.1k to 125.7k.
- [Blocked by Tock 2] Use read-only slices now that https://github.com/tock/tock/issues/1274 is fixed.\
- Look at fuzzing [coverage](https://rust-fuzz.github.io/book/cargo-fuzz/coverage.html), some functions (like `remove_handle`) or error conditions are not covered.
- Add a fuzzing mode 2 full cycles before end of lifetime.
- Update the fuzzing moder for dirty storage to start from one of the other mode, then fork and modify the storage, then continue. There's probably a way to use the fuzzer to mutate the storage.

ia0:
The code moved out of the repo.

PR #256, opened 2021-01-13
Command logic for credential management

kaczmarczyck:
Together with #253 and #254, this PR implements the credential management command. Its logic is equivalent to #107, with the difference of only storing keys for credentials as state between calls to `EnumerateCredentialsGetNextCredential`.

- [x] Tests pass

### DIFF
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -22,11 +22,7 @@ use super::pin_protocol_v1::{PinPermission, PinProtocolV1};
 use super::response::{AuthenticatorCredentialManagementResponse, ResponseData};
 use super::status_code::Ctap2StatusCode;
 use super::storage::PersistentStore;
-use super::timed_permission::TimedPermission;
-use super::{
-    check_command_permission, check_pin_uv_auth_protocol, StatefulCommand,
-    STATEFUL_COMMAND_TIMEOUT_DURATION,
-};
+use super::{check_pin_uv_auth_protocol, StatefulCommand, StatefulPermission};
 use alloc::collections::BTreeSet;
 use alloc::string::String;
 use alloc::vec;
@@ -125,8 +121,7 @@ fn process_get_creds_metadata(
 /// Processes the subcommand enumerateRPsBegin for CredentialManagement.
 fn process_enumerate_rps_begin(
     persistent_store: &PersistentStore,
-    stateful_command_permission: &mut TimedPermission,
-    stateful_command_type: &mut Option<StatefulCommand>,
+    stateful_command_permission: &mut StatefulPermission,
     now: ClockValue,
 ) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
     let rp_set = get_stored_rp_ids(persistent_store)?;
@@ -134,9 +129,7 @@ fn process_enumerate_rps_begin(
 
     // TODO(kaczmarczyck) should we return CTAP2_ERR_NO_CREDENTIALS if empty?
     if total_rps > 1 {
-        *stateful_command_permission =
-            TimedPermission::granted(now, STATEFUL_COMMAND_TIMEOUT_DURATION);
-        *stateful_command_type = Some(StatefulCommand::EnumerateRps(1));
+        stateful_command_permission.set_command(now, StatefulCommand::EnumerateRps(1));
     }
     // TODO https://github.com/rust-lang/rust/issues/62924 replace with pop_first()
     enumerate_rps_response(rp_set.into_iter().next(), Some(total_rps as u64))
@@ -145,19 +138,16 @@ fn process_enumerate_rps_begin(
 /// Processes the subcommand enumerateRPsGetNextRP for CredentialManagement.
 fn process_enumerate_rps_get_next_rp(
     persistent_store: &PersistentStore,
-    stateful_command_permission: &mut TimedPermission,
-    stateful_command_type: &mut Option<StatefulCommand>,
-    now: ClockValue,
+    stateful_command_permission: &mut StatefulPermission,
 ) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
-    check_command_permission(stateful_command_permission, now)?;
-    if let Some(StatefulCommand::EnumerateRps(rp_id_index)) = stateful_command_type {
+    if let StatefulCommand::EnumerateRps(rp_id_index) = stateful_command_permission.get_command()? {
         let rp_set = get_stored_rp_ids(persistent_store)?;
         // A BTreeSet is already sorted.
         let rp_id = rp_set
             .into_iter()
             .nth(*rp_id_index)
             .ok_or(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED)?;
-        *stateful_command_type = Some(StatefulCommand::EnumerateRps(*rp_id_index + 1));
+        *rp_id_index += 1;
         enumerate_rps_response(Some(rp_id), None)
     } else {
         Err(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED)
@@ -167,8 +157,7 @@ fn process_enumerate_rps_get_next_rp(
 /// Processes the subcommand enumerateCredentialsBegin for CredentialManagement.
 fn process_enumerate_credentials_begin(
     persistent_store: &PersistentStore,
-    stateful_command_permission: &mut TimedPermission,
-    stateful_command_type: &mut Option<StatefulCommand>,
+    stateful_command_permission: &mut StatefulPermission,
     sub_command_params: CredentialManagementSubCommandParameters,
     now: ClockValue,
 ) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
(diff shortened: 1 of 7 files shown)


kaczmarczyck:
For storage review ( @ia0 ) this should be very quick, only using all changes made in #254.

ia0 (dismissed):
I reviewed `Ctrl-F persistent_storage` in all files.

I added a task that maybe `iter_credentials` usability could be improved. I wonder if I can remove the need to have `iter_result` (with compromise of each item being a `Result`).

kaczmarczyck:
I think that would be easier indeed!

kaczmarczyck:
The command is tested against Chrome. To test, change the [default permission in pin_protocol_v1.rs](https://github.com/google/OpenSK/blob/d5761018ab0178a2d3db3b7647b2865fa5b30fd5/src/ctap/pin_protocol_v1.rs#L341) to
```rust
self.permissions = 0x03 | PinPermission::CredentialManagement as u8;
```

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,884 @@
...
+    let credential_id = PublicKeyCredentialDescriptor {
+        key_type,
+        key_id: credential_id,
+        transports: None, // You can set USB as a hint here.
+    };
+    let public_key = CoseKey::from(private_key.genpk());
+    Ok(AuthenticatorCredentialManagementResponse {
+        existing_resident_credentials_count: None,
+        max_possible_remaining_resident_credentials_count: None,
+        rp: None,
```
jmichelp: I see this pattern a lot where we create an object but most of the field are left to `None`.
Couldn't it improve readability by doing `impl Default for ...`?
IIRC this would allow us to only put the values we care about and finish constructing the object by putting `..` at end, right?

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,884 @@
...
+    iter_result?;
+    let mut rp_ids = Vec::from_iter(rp_set);
+    let total_rps = rp_ids.len();
+
+    // TODO(kaczmarczyck) behaviour with empty list?
+    let rp_id = rp_ids.pop();
+    if total_rps > 1 {
+        *stateful_command_permission =
+            TimedPermission::granted(now, STATEFUL_COMMAND_TIMEOUT_DURATION);
+        *stateful_command_type = Some(StatefulCommand::EnumerateRps(rp_ids));
```
jmichelp: It's a bit hard for me to be sure here...
When we receive an enumeration, we'll generate a `Vec<string>` (which can consume quite a lot of RAM). But even if the permission expires, the object will not be destroyed until the next time we receive a command that will overwrite this, correct?

kaczmarczyck: Yes, the main loop logic is very lightweight. The timing of destruction shouldn't change worst case memory consumption IMO. If you are worried about the `Vec<String>`, we could instead keep a storage key of `usize`, iterate the credential keys, keep a set of relying parties that were already used and respond with the next unseen RP ID. @ia0 this should work? Something like the following and an initial counter for how many RP IDs exist.

```rust
self.first_unseen_key = key::CREDENTIAL.start;

fn next_rp() {
  let mut seen_rp_ids = BTreeSet::new();
  for key in key::CREDENTIAL.start..self.first_unseen_key {
    seen_rps.add(rp_id_of_credential_at_key);
  }
  for key in self.first_unseen_key..key::CREDENTIAL.end {
    if seen_rps.contains(rp_id_of_credential_at_key) {
      break;
    }
    self.first_unseen_key += 1;
  }
  // Use RP ID at self.first_unseen_key, increment again, ...
}
```

src/ctap/credential_management.rs
kaczmarczyck: Magic! I think it's
`..Default::default()` and I can actually `derive` the `Default`.

src/ctap/credential_management.rs
ia0: It's not the best to iterate on keys, because looking up a key is currently linear (would be constant with #238). There would be solutions to resume `iter_credentials`, but there might be some lifetime issues. I think the best in this case would simply be to recompute the set of RPids at each command (initial command and each follow-up) since we have a new time budget for each command and we know it's enough because we anyway need to do it once at some point. So the state could simply be the index in the ordered set of the next RP id. What do you think?

kaczmarczyck: That works too, as a way to reduce the state we need to keep to `usize`. Thanks for the input. @jmichelp would that help with your concern?

jmichelp: It should be fine provided we can't get dirty-read cases: e.g. we have an iterator, then we do some modification on the list (e.g. delete an entry) then we continue iterating.
But I don't think it's very different from this point of view from the current code as we would keep in `stateful_command_type` a list that may point to deleted entries.

kaczmarczyck: Done as proposed by @ia0 . I added a test to make sure the sorting is the same each time.

kaczmarczyck:
The 2 new commits don't show up in the conversation, since I merged after. The first is addressing the memory state issue. The other reworks the command state more to always reset timeout and state when either check fails. @jmichelp PTAL

src/ctap/mod.rs
```diff
@@ -165,19 +166,54 @@ pub struct AssertionState {
...
+    pub fn check_command_permission(&mut self, now: ClockValue) -> Result<(), Ctap2StatusCode> {
+        if self.permission.is_granted(now) {
+            Ok(())
+        } else {
+            self.clear();
+            Err(Ctap2StatusCode::CTAP2_ERR_NOT_ALLOWED)
+        }
+    }
+
+    pub fn get_command(&mut self) -> Result<&mut StatefulCommand, Ctap2StatusCode> {
```
jmichelp: Returning a mutable value is not the typical idiom you expect for a getter.
Why not having `get_command` return a immutable and use `set_command` to modify the value? If the issue is the timestamp that shouldn't be updated, you may also add `update_command` that calls `set_command` with the stored timestamp.

src/ctap/credential_management.rs
```diff
@@ -0,0 +1,924 @@
...
+        ..Default::default()
+    })
+}
+
+/// Processes the subcommand getCredsMetadata for CredentialManagement.
+fn process_get_creds_metadata(
+    persistent_store: &PersistentStore,
+) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
+    Ok(AuthenticatorCredentialManagementResponse {
+        existing_resident_credentials_count: Some(persistent_store.count_credentials()? as u64),
```
ia0, line 113: Just as FYI: I've added an item to #184 to improve `count_credentials`. We currently do unnecessary work and I'm not sure how costly it is. In particular given that it may look cheap and it's called by `remaining_credentials` used below.

src/ctap/mod.rs
```diff
@@ -345,27 +382,23 @@ where
                         Duration::from_ms(TOUCH_TIMEOUT_MS),
                     );
                 }
-                match (&command, &self.stateful_command_type) {
-                    (
-                        Command::AuthenticatorGetNextAssertion,
-                        Some(StatefulCommand::GetAssertion(_)),
-                    ) => (),
-                    (
+                match (&command, self.stateful_command_permission.get_command()) {
```
jmichelp, line 444: Should such logic actually be part of `StatefulPermission`?
e.g. in `check_command_permission`

ia0 (dismissed):
I just checked occurrences of `persistent_store`.

src/ctap/mod.rs
kaczmarczyck: I'd need different `check_command` commands, one for the `command` here, the other for the `subcommand` in credential management. Since these `match`es were specific to the call site, I thought I was best to keep them there.
Another problem is that merging with `check_command_permission` means we start returning a `Result` here, and this caller function does not return an error, but converts to the byte later. It might mean that we have to refactor this `process_command` function.
WDYT?

jmichelp: Ok.
I thought that the logic was "generic" in the way that if you start iterating through the credentials, we know which set of commands are allowed without dropping the state. So you could have checked both the command and the timestamp at the same time. Any invalid command of timestamp would result in invalidating the state.
And the rest of the code would only have to care about the requested command, not the internal state command.

src/ctap/mod.rs
kaczmarczyck: Instead of one `update_command` function, I now have individual update functions to handle the correct states. This makes the API even more strict, as commands can't update their state outside more than necessary, e.g. to a different command. Updating now always restarts the whole command permissions.

src/ctap/credential_management.rs
kaczmarczyck: Thanks!

src/ctap/mod.rs
kaczmarczyck: I'll resolve since you're okay, just a comment on idea to conserve my thoughts. I think doing one check for everything would be nice, but there are a few problems in that idea:

- The timeout check has to be made periodically even when no commands arrive, so we don't miss when our timestamps overflow.
- The subcommands are parsed later than the command byte, and we need both for the decision. If CTAP2 wouldn't have so many nested commands, parsing would be easier and cleaner.
- We'd need to refactor `process_command` a bit.

I'd rather keep these matches where the commands are parsed, so our struct doesn't need any knowledge of other commands or subcommands.

src/ctap/mod.rs
```diff
@@ -149,47 +149,63 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
...
 }
 
+/// Stores the current CTAP command state and when it times out.
+///
+/// Some commands are executed in a series of calls to the authenticator.
+/// Interleaving calls to other commands interrupt the current command and
+/// remove all state and permissions. Power cycling allows the Reset command,
+/// and to prevent misuse or accidents, we disallow Reset after receiving
+/// different commands. Therefore, Reset behaves just like all other stateful
+/// commands and is included here. Please not that the allowed time for Reset
```
jmichelp: nit: note

jmichelp (approved):

PR #263, opened 2021-01-20
Large blobs

kaczmarczyck:
This PR implements the `LargeBlobs` command. It is based on #256 right now, and will be rebased on top of develop after #256 is merged.
All work is in this [commit](https://github.com/google/OpenSK/commit/8d1d5ac250fd33c48502cc990ff10efebb5c5182). It is pushed here as a preview and to discuss how to best access the storage. The goal is to end up with an implementation of this command that is functionally equivalent, but more efficient in how it `get`s and `commit`s the data.

- [x] Tests pass


src/ctap/storage.rs
```diff
@@ -458,6 +462,70 @@ impl PersistentStore {
...
+            let shard_length = cmp::min(SHARD_SIZE - shard_offset, byte_count);
+
+            let shard_key = key::LARGE_BLOB_SHARDS.start + shard;
+            if !key::LARGE_BLOB_SHARDS.contains(&shard_key) {
+                // This request should have been caught at application level.
+                return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+            }
+            let shard_entry = self.store.find(shard_key)?.unwrap_or_default();
+            if shard_entry.len() < shard_offset + shard_length {
+                output.extend(&shard_entry[..]);
```
ia0: This looks wrong. Probably `&shard_entry[shard_offset..]` with an additional check that `shard_offset <= shard_entry.len()`.

src/ctap/storage.rs
```diff
@@ -458,6 +462,70 @@ impl PersistentStore {
...
+                return Ok(output);
+            }
+            output.extend(&shard_entry[shard_offset..shard_offset + shard_length]);
+            offset += shard_length;
```
ia0: This also looks wrong. I would expect something like `offset += shard_length - shard_offset`.

src/ctap/storage.rs
```diff
@@ -458,6 +462,70 @@ impl PersistentStore {
...
+            byte_count -= shard_length;
```
ia0: Same here.

src/ctap/storage.rs
```diff
@@ -458,6 +462,70 @@ impl PersistentStore {
...
+            self.store.insert(
+                shard_key,
+                &large_blob_array[large_blob_index..large_blob_index + shard_length],
+            )?;
+            large_blob_index += shard_length;
+            shard_key += 1;
+        }
+        // The length is not stored, so overwrite old entries explicitly.
+        for key in shard_key..key::LARGE_BLOB_SHARDS.end {
+            // Assuming the store optimizes out unnecessary writes.
```
ia0: I guess you mean unnecessary `remove`. Yes, removing an absent entry is a no-op in terms of flash (and thus lifetime).

src/ctap/storage.rs
```diff
@@ -458,6 +462,70 @@ impl PersistentStore {
...
+            self.store.remove(key)?;
```
ia0: Those `insert` and `remove` should be part of the same transaction and encoded as `StoreUpdate`s. I'm planning to improve the `StoreUpdate` to avoid passing a `Vec<u8>` and instead be also able to pass a `&[u8]` which would be sufficient here.

src/ctap/storage.rs
kaczmarczyck: The specification asks to return all remaining bytes, if more are requested than exist. This is what this line is about. Maybe the confusion here and below is from `shard_length` not being the length of the shard in storage, but the length of what we want to read from this shard. Maybe another variable name works better for you?
(Also, if it was wrong, it should have been caught by unit tests.)

src/ctap/storage.rs
kaczmarczyck: I added / changed a unit test to double check, and it should be correct. Maybe this `shard_length - shard_offset` is implicitly covered by

```rust
let shard_length = cmp::min(SHARD_SIZE - shard_offset, byte_count);
```

?

src/ctap/storage.rs
kaczmarczyck: Since this whole method seems to be poorly written and confusing, I'll think about if a rewrite make sense.

src/ctap/storage.rs
kaczmarczyck: Yes, exactly that. Comment changed.

src/ctap/storage.rs
ia0: The issue I have in mind is:
- one shard of 10 bytes
- asking to read from offset 1 (i.e. skipping first byte)
- asking to read 20 bytes

My understanding is that we don't skip the first byte. So I guess when you say "return all remaining bytes" it's not just the ones after the offset, but the whole content. Then there would still be a problem with the following case:

- 2 shards A and B
- offset is A.len + 1
- length is A.len + B.len + 1

My understanding is that we return only B and not AB.

Regarding unit tests, how could they catch all bugs? Unit tests are just basic checks, they can't guarantee anything about the code (as long as there are a few ifs or a loop).

src/ctap/storage.rs
ia0: Oh right, yeah, `shard_length` is actually alright. I confused it with `SHARD_SIZE`.

src/ctap/storage.rs
ia0: Or maybe at least describe the expected behavior in the documentation, including corner-cases like when the reading span is not included in the content span. Because it seems CTAP specification is not the intuitive version where you just crop the reading span to the content span.

src/ctap/storage.rs
kaczmarczyck: Oh true, I thought I had this covered. Fixed.

kaczmarczyck: Also, unit test added for coverage.

src/ctap/storage.rs
kaczmarczyck: Documentation changed. You're supposed to crop, I think. I actually like the current loop implementation, since each loop iteration is like having called the function with the next set of `byte_count` and `offset`, just that more bytes are already read.

kaczmarczyck:
I'll rebase as announced now, then work on the transactions.

src/ctap/storage.rs
```diff
@@ -59,6 +60,9 @@ const DEFAULT_MIN_PIN_LENGTH_RP_IDS: Vec<String> = Vec::new();
 // This constant is an attempt to limit storage requirements. If you don't set it to 0,
 // the stored strings can still be unbounded, but that is true for all RP IDs.
 pub const MAX_RP_IDS_LENGTH: usize = 8;
+const SHARD_SIZE: usize = 1023;
```
ia0: Oh, actually I'm realizing just now. But to be more generic, we should use `self.store.max_value_length()`. The value of 1023 is correct only for flash pages of size 2048 or bigger (Nordic is 4096).

src/ctap/storage.rs
```diff
@@ -59,6 +60,9 @@ const DEFAULT_MIN_PIN_LENGTH_RP_IDS: Vec<String> = Vec::new();
...
+pub const MAX_LARGE_BLOB_ARRAY_SIZE: usize =
+    SHARD_SIZE * (key::LARGE_BLOB_SHARDS.end - key::LARGE_BLOB_SHARDS.start);
```
ia0: This would probably become a function since the shard size would be dependent of the board.

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
     }
 
+    /// Reads the byte vector stored as the serialized large blobs array.
+    ///
+    /// If too few bytes exist at that offset, return the maximum number
+    /// available. This includes cases of offset being beyond the stored array.
+    pub fn get_large_blob_array(
+        &self,
+        mut byte_count: usize,
+        mut offset: usize,
```
ia0: Just style, in case it wasn't deliberate: Usually the order is `offset` then `length`. This is quite unusual and could lead to bugs since those have the same type.

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
+    ) -> Result<Vec<u8>, Ctap2StatusCode> {
+        if self.store.find(key::LARGE_BLOB_SHARDS.start)?.is_none() {
+            return Ok(vec![
+                0x80, 0x76, 0xBE, 0x8B, 0x52, 0x8D, 0x00, 0x75, 0xF7, 0xAA, 0xE9, 0x8D, 0x6F, 0xA5,
+                0x7A, 0x6D, 0x3C,
+            ]);
```
ia0: The function should document that if the blob is absent, a default value is returned (maybe with some description of this default value).

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
+                // This request should have been caught at application level.
+                return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+            }
+            let shard_entry = self.store.find(shard_key)?.unwrap_or_default();
+            if shard_entry.len() < shard_offset + shard_length {
+                // If fewer bytes exist than requested, return them all.
+                output.extend(&shard_entry[shard_offset..]);
+                return Ok(output);
+            }
+            output.extend(&shard_entry[shard_offset..shard_offset + shard_length]);
```
ia0: Those 5 lines still look buggy when the shard is smaller than `shard_offset`. What about the following more natural alternative?

```rust
if shard_entry.len() < shard_offset {
    break;
}
let shard_length = cmp::min(shard_entry.len() - shard_offset, byte_count);
output.extend(&shard_entry[shard_offset..][..shard_length]);
if shard_entry.len() < SHARD_SIZE {
    break;
}
```

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
+        &mut self,
+        large_blob_array: &[u8],
+    ) -> Result<(), Ctap2StatusCode> {
+        if large_blob_array.len() > MAX_LARGE_BLOB_ARRAY_SIZE {
+            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        }
+        const MIN_SHARD_KEY: usize = key::LARGE_BLOB_SHARDS.start;
+        const SHARD_COUNT: usize = key::LARGE_BLOB_SHARDS.end - MIN_SHARD_KEY;
+        let mut transactions = Vec::with_capacity(SHARD_COUNT);
+        for shard_key in MIN_SHARD_KEY..key::LARGE_BLOB_SHARDS.end {
```
ia0: This can simply be `for shard_key in key::LARGE_BLOB_SHARDS`.

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
+    pub fn commit_large_blob_array(
+        &mut self,
+        large_blob_array: &[u8],
+    ) -> Result<(), Ctap2StatusCode> {
+        if large_blob_array.len() > MAX_LARGE_BLOB_ARRAY_SIZE {
+            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        }
+        const MIN_SHARD_KEY: usize = key::LARGE_BLOB_SHARDS.start;
+        const SHARD_COUNT: usize = key::LARGE_BLOB_SHARDS.end - MIN_SHARD_KEY;
+        let mut transactions = Vec::with_capacity(SHARD_COUNT);
```
ia0: This should probably be called `updates`, not `transactions`.

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
+            if large_blob_array.len() > large_blob_index {
+                let shard_length = cmp::min(SHARD_SIZE, large_blob_array.len() - large_blob_index);
+                transactions.push(StoreUpdate::Insert {
+                    key: shard_key,
+                    value: &large_blob_array[large_blob_index..large_blob_index + shard_length],
+                });
+            } else {
+                transactions.push(StoreUpdate::Remove { key: shard_key });
+            }
+        }
```
ia0: To avoid arithmetic, this loop could be more simply written as:

```rust
let mut shards = large_blob_array.chunks(SHARD_SIZE);
let updates = key::LARGE_BLOB_SHARDS.map(|key| match shards.next() {
    None => StoreUpdate::Remove { key },
    Some(value) => StoreUpdate::Insert { key, value },
}).collect();
```

src/ctap/storage.rs
```diff
@@ -467,6 +471,70 @@ impl PersistentStore {
...
+        Ok(self.store.transaction(&transactions)?)
```
ia0: Just as a note (maybe we want to add as a comment in code), but we might want to optimize this function (although it's probably not useful since this should be called very rarely):
1. Only write a shard if it is different from before. This optimization could actually be done in the persistent storage using a flag (not sure about the design yet). This would permit use-cases that are append-only or modify bytes only (not insert or delete) to be much more efficient.
2. Do not add useless remove to the transaction. Each remove is 1 word of lifetime. This is not very useful though compared to the other costs. But it would come for free if the previous step is implemented here (and not in the persistent store).

kaczmarczyck: In fact, a `largeBlob` always ends with a 16 byte hash (so far, the storage did not care). If we had the old blob length available, we could check easily. I'm not sure if we need this kind of protection though. In any case, I can add more `find`s to not adding unused `remove`s.

src/ctap/storage.rs
kaczmarczyck: `chunks` is super useful here! I don't think I can collect as nice with the improvement mentioned below to skip some `remove`s. Maybe with a filter_map, but the break seems more natural to me. Also FYI, your code give me a:

> error[E0277]: the size for values of type `[persistent_store::store::StoreUpdate<_>]` cannot be known at compilation time

src/ctap/storage.rs
kaczmarczyck: Okay for the private `SHARD_SIZE`. I decided to assign a fixed value to the total allowed length. Else, even parsing CBOR would require a store call for a constant, and (worse) OpenSK would behave differently on different hardware by default. Tuning is okay, and tests ensure `MAX_LARGE_BLOB_ARRAY_SIZE` is within bounds if people play around with it.

src/ctap/storage.rs
kaczmarczyck: Done for `SHARD_SIZE`.

src/ctap/storage.rs
kaczmarczyck: Good to know, changed.

src/ctap/storage.rs
kaczmarczyck: Documented, and changed the implementation a bit to work for small offsets in get.

src/ctap/storage.rs
kaczmarczyck: Thanks, good catch. Tests added.

src/ctap/storage.rs
kaczmarczyck: Very true.

src/ctap/storage.rs
kaczmarczyck: Not sure why, but I'm happy either way, so changed.

src/ctap/storage.rs
kaczmarczyck: If we still want (1), I'd rather implement it one level higher even. But I don't think we have to, I don't expect a platform to write without reading first.
For me (2) is more universally useful, as it helps with store lifetime after a single commit. Therefore (2) is implemented.

src/ctap/storage.rs
ia0: Right, we probably need to specify the collect return type: `.collect::<Vec<StoreUpdate<&[u8]>>>()`.

src/ctap/storage.rs
ia0: There's probably a misunderstanding on (1). The platform has no power for this. This is an optimization at the storage level. I don't mean to skip writing the whole blob, just to skip writing a shard. This can have huge benefits in the case mentioned (localized byte updates, append and remove from tail).

I implemented a generic fragmentation module in #270 because this might be useful at other places too and it will keep the CTAP storage simpler. Also, we can more easily fuzz it (not sure it's useful but I still added to my list in #184).

kaczmarczyck: Using the new fragment, so should be fixed.

src/ctap/storage.rs
kaczmarczyck: Obsolete.

src/ctap/storage.rs
```diff
@@ -28,9 +28,11 @@ use alloc::vec;
 use alloc::vec::Vec;
 use arrayref::array_ref;
 use cbor::cbor_array_vec;
+use core::cmp;
 use core::convert::TryInto;
 use crypto::rng256::Rng256;
 use persistent_store::StoreUpdate;
+use persistent_store::fragment::{read_range, write};
```
ia0: Since the names are generic, I had in mind to use them with the module name:

```rust
use persistent_store::fragment;
// ...
fragment::read_range()
fragment::write()
```

As you want.

src/ctap/storage.rs
```diff
@@ -467,6 +470,44 @@ impl PersistentStore {
...
+    /// large blob entry possible).
+    pub fn get_large_blob_array(
+        &self,
+        offset: usize,
+        byte_count: usize,
+    ) -> Result<Vec<u8>, Ctap2StatusCode> {
+        let byte_range = offset..offset + byte_count;
+        let output = read_range(&self.store, &key::LARGE_BLOB_SHARDS, byte_range)?;
+        Ok(output.unwrap_or_else(|| {
+            let empty_large_blob = vec![
```
ia0: This could be a constant:

```rust
const EMPTY_LARGE_BLOB: [u8; 17] = [...];
```

src/ctap/storage.rs
```diff
@@ -563,6 +604,15 @@ impl PersistentStore {
...
         Ok(self.store.insert(key::FORCE_PIN_CHANGE, &[])?)
     }
+
+    /// The size used for shards of large blobs.
+    ///
+    /// This value is constant during the lifetime of the device.
+    #[cfg(test)]
+    fn shard_size(&self) -> usize {
+        self.store.max_value_length()
+    }
```
ia0: Do we still need this?

ia0 (dismissed):
Only looked at storage.

libraries/persistent_store/src/fragment.rs
```diff
@@ -24,6 +24,7 @@ use alloc::vec::Vec;
 use core::ops::Range;
 
 /// Represents a sequence of keys.
+#[allow(clippy::len_without_is_empty)]
```
ia0, line 27: Interesting lint :-) Thanks for the fix.

src/ctap/storage.rs
kaczmarczyck: Seems better, changed.

src/ctap/storage.rs
kaczmarczyck: Done

src/ctap/storage.rs
kaczmarczyck: The second commit reduces the test complexity to be independent of shard sizes. Doing that, we can remove this function. Probably makes sense after shard tests are in the store now, but if you think I reduced testing too much, let me know.

ia0 (approved):

libraries/persistent_store/src/fragment.rs
kaczmarczyck: I was wondering if I should add the function itself, but `len` is O(1), so it should be okay.

ia0: Yes, and also we would never call this function (because it's useless for keys to be empty), so it's silly. The lint is just over-triggering.

PR #265, opened 2021-01-20
credBlob extension

kaczmarczyck:
This PR implements the credBlob extension.

At the same time, it facilitates the Extension data structures to have a default instead of being optional.

- [x] Tests pass


jmichelp (approved):
Seems that my suggestion about deriving from Default was useful ;-)

kaczmarczyck:
Still enough concepts in Rust that I don't know about, so keep them coming :)

PR #267, opened 2021-01-21
Force PIN change

kaczmarczyck:
This PR implements the logic for forcing PIN changes, that were started in #246.

Also, it changes the test for GetInfo to be more readable and more meaningful.

- [x] Tests pass


src/ctap/storage.rs
```diff
@@ -384,7 +384,9 @@ impl PersistentStore {
         let mut pin_properties = [0; 1 + PIN_AUTH_LENGTH];
         pin_properties[0] = pin_code_point_length;
         pin_properties[1..].clone_from_slice(pin_hash);
-        Ok(self.store.insert(key::PIN_PROPERTIES, &pin_properties)?)
+        self.store.insert(key::PIN_PROPERTIES, &pin_properties)?;
+        // If this second transaction fails, you are forced to retry.
+        Ok(self.store.remove(key::FORCE_PIN_CHANGE)?)
```
ia0: Just  checking the semantics we want in case of power-loss. Is it ok to only insert and not remove? Is there a reason to not use a transaction? There are at least 2 reasons not to:
1. It costs 2 words of lifetime.
2. Those 2 words of lifetime are also temporary required capacity, so if the store is almost full, the transaction would fail, while both updates in a sequence would pass. But this scenario should not be considered since we should probably never hit capacity because of how we dimensionned the CTAP storage.

Not sure there are many reasons to use a transaction, but we could still add a comment to say that it's ok to do in 2 steps (because the relation between both entries is only in one direction and not an equivalence).

src/ctap/storage.rs
```diff
@@ -541,9 +543,18 @@ impl PersistentStore {
         Ok(())
     }
 
+    /// Returns whether the PIN needs to be changed before its next usage.
+    pub fn has_force_pin_change(&self) -> Result<bool, Ctap2StatusCode> {
+        match self.store.find(key::FORCE_PIN_CHANGE)? {
+            None => Ok(false),
+            Some(value) if value.len() == 1 && value[0] == 1 => Ok(true),
```
ia0: If the value is constant, why not have a value of length zero? Then it's just about the presence of the entry. Is it because it might change in the future?

ia0 (commented):
I only reviewed the storage part.

src/ctap/storage.rs
kaczmarczyck: I assume that this is a very rare event, that the second transaction fails after the first succeeded.
From a specification perspective, both are okay. We force the user to change their PIN until `remove(key::FORCE_PIN_CHANGE)`. Which means they'll just need do the PIN change again.

From a user experience perspective, the current implementation has the downside that the user expects the old PIN to be active, since they received an error. And the new PIN actually works. Upside is that they are protected by the stronger PIN earlier. The confusion with this implementation could lock users out that never try the new PIN. Maybe this is reason to change it to be a transaction? WDYT?

src/ctap/storage.rs
kaczmarczyck: Zero length is okay, too. This was just supposed to be an explicit `true`. I can change it.

I was thinking if I should piggy-back it onto some other key like PIN related keys, since they need to be changed anyway and we save a delete. But went with the explicit and clean solution instead. It's a rare enough event that we don't have to care I think.

src/ctap/storage.rs
ia0: Yes, I think that's actually a good reason to use a transaction. I didn't think of it. So the alternative would simply be:

```rust
Ok(self.store.transaction(&[
    StoreUpdate::Insert { key: key::PIN_PROPERTIES, value: &pin_properties },
    StoreUpdate::Remove { key: key::FORCE_PIN_CHANGE }
])?)
```

src/ctap/storage.rs
ia0: The advantage of zero-length is:
1. It only costs 1 word of flash instead of 2.
2. There are less invariants.

src/ctap/pin_protocol_v1.rs
```diff
@@ -328,6 +328,9 @@ impl PinProtocolV1 {
         let token_encryption_key = crypto::aes256::EncryptionKey::new(&shared_secret);
         let pin_decryption_key = crypto::aes256::DecryptionKey::new(&token_encryption_key);
         self.verify_pin_hash_enc(rng, persistent_store, &pin_decryption_key, pin_hash_enc)?;
+        if persistent_store.has_force_pin_change()? {
```
jmichelp, line 332: Why not moving this test right at the beginning, after the one that checks for pin retries?
It seems from this logic that if invalid PIN is provided, you will return `CTAP2_ERR_PIN_INVALID` from `verify_pin_hash_enc()` 1 line above. If the PIN is valid you'll return `CTAP2_ERR_PIN_INVALID` because of this check. So maybe we can save some time.

src/ctap/storage.rs
kaczmarczyck: FYI your snippet needed a minor change to work, or gets a

> the trait bound `&[u8; 17]: std::borrow::Borrow<[u8]>` is not satisfied

The array reference has to be a slice instead.

src/ctap/storage.rs
kaczmarczyck: Good to know for size on flash! Changed.

src/ctap/storage.rs
ia0: Oh right yes, I forgot that with generics it doesn't do the conversion from array reference to slice reference.

ia0 (approved):
Storage change LGTM.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: This is a very good question, and I was wondering why the [specification](https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#getPinToken) asks for this behaviour in the order of the protocol. It might be related to allowing to decrement / reset the PIN counter. I added a TODO to investigate, but this is the purpose I can think of.

jmichelp (approved):

PR #270, opened 2021-01-23
Add helper functions for fragmented entries

ia0:


libraries/persistent_store/src/fragment.rs
```diff
@@ -0,0 +1,179 @@
...
+    let mut handles: Vec<Option<StoreHandle>> = vec![None; keys_len as usize];
+    for handle in store.iter()? {
+        let handle = handle?;
+        let pos = match keys.pos(handle.get_key()) {
+            Some(pos) => pos,
+            None => continue,
+        };
+        if pos >= keys_len {
+            return Err(StoreError::InvalidArgument);
+        }
```
kaczmarczyck, line 157: Why is this extra check necessary? I'd expect `keys.pos()` to be `None` for invalid arguments?

libraries/persistent_store/src/fragment.rs
```diff
@@ -0,0 +1,179 @@
...
+        return Ok(None);
+    }
+    let mut result = Vec::with_capacity(range_len);
+    let mut offset = 0;
+    for handle in handles {
+        let start = range.start.saturating_sub(offset);
+        let length = handle.get_length(store)?;
+        let end = core::cmp::min(range.end.saturating_sub(offset), length);
+        offset += length;
+        if start < end && end <= length {
```
kaczmarczyck: How could `end > length` happen here?

libraries/persistent_store/src/fragment.rs
```diff
@@ -0,0 +1,179 @@
...
+            result.extend(&handle.get_value(store)?[start..end]);
+        }
```
kaczmarczyck, line 101: Do we want to `break` if `offset >= range.end`?

kaczmarczyck (dismissed):
Looks good, comments are just minor optimizations.

libraries/persistent_store/src/fragment.rs
ia0: Good catch, that's a leftover. I first wrote the condition for the `[start..end]` to not panic and forgot to simplify.

libraries/persistent_store/src/fragment.rs
ia0: I don't think it's useful, but we can. It's just CPU and negligible to the rest. At first I wasn't even sure it's useful to have `read_range` compare to just calling `read` and taking the range after. I'm still not sure actually, but since you had it in your PR, I did it here too.

libraries/persistent_store/src/fragment.rs
ia0: The problem is that the `Keys` trait is safe (not marked `unsafe`) so anybody can implement it, and they can do mistakes, like returning `Some(out_of_bound_pos)` from `Keys::pos()`. Since I do `handles[pos]` below, I need to check this invariant to avoid panics. The alternative would be to mark the `Keys` trait as `unsafe` such that it's the responsibility of the implementor if this code panics. Since the check is cheap, I prefer to have a safe API.

kaczmarczyck: Understood.

libraries/persistent_store/src/fragment.rs
kaczmarczyck: I don't have a strong preference, so resolved.

kaczmarczyck (approved):

PR #274, opened 2021-01-27
largeBlobKey extension

kaczmarczyck:
Implements the last extension missing for CTAP 2.1. Also cleans up the `GetInfo` output for all options.

One design choice made here is how we want to store the per-credential `largeBlobKey`. The specification mandates:

- only discoverable (aka resident) credentials need a `largeBlobKey`
- only return a `largeBlobKey` when created during `MakeCredential`

Therefore, we have to store per-credential information. I decided to go with the 32 byte array itself, as the cryptographically most straight-forward implementation.
We could sometimes save some memory by only storing a bool indicating existence, and later derive the `largeBlobKey` using a master secret similar to `credRandom`. This approach only pays off with the second created credentials that uses `largeBlobKey`, because of the cost for master secrets. For users (including those never using `largeBlobKey` at all), I expect this to be more expensive on average.

- [x] Tests pass

### DIFF
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -81,6 +81,7 @@ fn enumerate_credentials_response(
         user_name,
         user_icon,
         cred_blob: _,
+        large_blob_key,
     } = credential;
     let user = PublicKeyCredentialUserEntity {
         user_id: user_handle,
@@ -100,8 +101,7 @@ fn enumerate_credentials_response(
         public_key: Some(public_key),
         total_credentials,
         cred_protect: cred_protect_policy,
-        // TODO(kaczmarczyck) add when largeBlobKey extension is implemented
-        large_blob_key: None,
+        large_blob_key,
         ..Default::default()
     })
 }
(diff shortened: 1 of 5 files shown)


ia0 (commented):
How come the default cred protect became force pin change?

Non-logical changes LGTM, but I'll let @jmichelp approve if the logic needs to be reviewed too.

kaczmarczyck:
It changed somewhere between review draft 1 and 2. I can go search for the specific PR and dig out the reasoning. But I'd assume the default cred protect isn't really necessary since you can get individual levels per credentials from credential management.

ia0:
> It changed somewhere between review draft 1 and 2. I can go search for the specific PR and dig out the reasoning. But I'd assume the default cred protect isn't really necessary since you can get individual levels per credentials from credential management.

Sounds good, no need to search for references. I was just wondering why and the answer is the draft changed.

jmichelp (approved):
Sorry, this took me a while because I had to review it side by side with the specification due to non-obvious changes :)

PR #278, opened 2021-02-02
Remove derive_debug feature

### LINKED ISSUE #251: Do not feature-gate generated dead-code
Generated code (like derives) don't generate warnings when unused. They also don't end up in the final binary when unused (library size and compilation time shouldn't matter and are negligible). So there seems to be no reason to feature-gate them, which currently decreases readability and increases maintenance.

Current example snippet:
```rust
#[cfg_attr(feature = "derive_debug", derive(Debug))]
```

Proposed example snippet:
```rust
#[derive(Debug)]
```

kaczmarczyck:
Fixes (with some side effects we'll discuss on the issue) #251

Removes most conditional trait imports. `IntoEnumIterator` is an exception as it is only in `dev-dependencies`. Binary sizes seem to be unaffected, after checking with our workflows:

```
./reproduce_hashes.sh
git diff --no-index reproducible/reference_elf2tab_ubuntu-18.04.txt reproducible/elf2tab.txt || true
```

- [x] Tests pass

### DIFF
--- a/.github/workflows/crypto_test.yml
+++ b/.github/workflows/crypto_test.yml
@@ -33,10 +33,10 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --manifest-path libraries/crypto/Cargo.toml --release --features std,derive_debug
+          args: --manifest-path libraries/crypto/Cargo.toml --release --features std
 
       - name: Unit testing of crypto library (debug mode)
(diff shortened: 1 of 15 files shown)


jmichelp (approved):

ia0 (approved):

kaczmarczyck:
I changed the PR description to not resolve #251 right away, in case we still want the linter.

PR #280, opened 2021-02-04
Allow RP ID in token permissions for some subcommands of credential management

kaczmarczyck:
Related to #106 . Adds missing functionality related to RP ID permissions checking in credential management.

- [x] Tests pass

### DIFF
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -106,6 +106,22 @@ fn enumerate_credentials_response(
     })
 }
 
+/// Check if the token permissions have the correct associated RP ID.
+///
+/// Either no RP ID is associated, or the RP ID matches the stored credential.
+fn check_rp_id_permissions(
+    persistent_store: &mut PersistentStore,
+    pin_protocol_v1: &mut PinProtocolV1,
+    credential_id: &[u8],
+) -> Result<(), Ctap2StatusCode> {
+    // Pre-check a sufficient condition before calling the store.
+    if pin_protocol_v1.has_no_rp_id_permission().is_ok() {
+        return Ok(());
+    }
+    let (_, credential) = persistent_store.find_credential_item(credential_id)?;
+    pin_protocol_v1.has_no_or_rp_id_permission(&credential.rp_id)
+}
+
 /// Processes the subcommand getCredsMetadata for CredentialManagement.
 fn process_get_creds_metadata(
     persistent_store: &PersistentStore,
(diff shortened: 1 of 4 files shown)


src/ctap/credential_management.rs
```diff
@@ -155,12 +171,14 @@ fn process_enumerate_rps_get_next_rp(
 fn process_enumerate_credentials_begin(
     persistent_store: &PersistentStore,
     stateful_command_permission: &mut StatefulPermission,
+    pin_protocol_v1: &mut PinProtocolV1,
```
jmichelp, line 176: Should we define a PinProtocol trait instead, anticipating that in a short term we will implement PinProtocolV2?

src/ctap/pin_protocol_v1.rs
```diff
@@ -510,22 +511,47 @@ impl PinProtocolV1 {
...
-        } else {
-            self.permissions_rp_id = Some(String::from(rp_id));
+            _ => Ok(()),
+        }
+    }
+
+    /// Check if the passed RP ID is associated with the token permission.
+    ///
+    /// If no RP ID is associated, associate the passed RP ID as a side effect.
+    pub fn require_rp_id_permission(&mut self, rp_id: &str) -> Result<(), Ctap2StatusCode> {
```
jmichelp: Not very happy about the function names here. `has_` functions makes me think they will return `bool` whereas `require_` makes me think about assertions and I expect then to have `Result` as a return value. But I'm not expecting a `require_` function to alter the object.
I know it's not easy but could we try finding better names?

jmichelp (dismissed):

src/ctap/credential_management.rs
kaczmarczyck: I plan to, but I have to design that first. Should go into a separate PR.

src/ctap/pin_protocol_v1.rs
kaczmarczyck: I called it `require_` because it is like `has_`, but also sets the RP ID if not present. How about...
- `set_`, but it can't actually overwrite
- `check_`, but it also sets
- `check_and_set_` is a bit long?

If we settle with `check_and_set_`, then `has_` would become `check_`?
To defend the `has_` choice, it is kind of a `bool` with `?` support.

ia0: I usually call those functions `ensure_` in my personal projects. It ensures the property holds as post-condition regardless of pre-condition. So it may mutate stuff to make it hold. But if it already holds it does nothing, just checks it holds.

jmichelp: It seems there's even [a crate](https://docs.rs/ensure/0.3.1/ensure/) that agrees with Julien.
I'm fine leaving tha `has_*` functions as is and change `require_` for `ensure_`.
Is that also good for you Fabian?

kaczmarczyck: Very good!

jmichelp (approved):

PR #281, opened 2021-02-04
Enterprise attestation

kaczmarczyck:
Adds enterprise attestation for #106 .

There is not just a single attestation to use for enterprise, as it is up to the enterprise to choose. This PR has useful defaults and implements all logic around the actual attestation cryptography to make it easy to extend the implementation. With the constants set as is, this PR does not actually change OpenSK's behaviour. (Except for fixing the formerly unused command parameter type in MakeCredential.)

- [x] Tests pass

### DIFF
--- a/src/ctap/command.rs
+++ b/src/ctap/command.rs
@@ -161,7 +161,7 @@ pub struct AuthenticatorMakeCredentialParameters {
     pub options: MakeCredentialOptions,
     pub pin_uv_auth_param: Option<Vec<u8>>,
     pub pin_uv_auth_protocol: Option<u64>,
-    pub enterprise_attestation: Option<bool>,
+    pub enterprise_attestation: Option<u64>,
 }
 
 impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
@@ -219,7 +219,7 @@ impl TryFrom<cbor::Value> for AuthenticatorMakeCredentialParameters {
 
         let pin_uv_auth_param = pin_uv_auth_param.map(extract_byte_string).transpose()?;
         let pin_uv_auth_protocol = pin_uv_auth_protocol.map(extract_unsigned).transpose()?;
-        let enterprise_attestation = enterprise_attestation.map(extract_bool).transpose()?;
+        let enterprise_attestation = enterprise_attestation.map(extract_unsigned).transpose()?;
 
         Ok(AuthenticatorMakeCredentialParameters {
             client_data_hash,
@@ -601,7 +601,7 @@ mod test {
             0x05 => cbor_array![],
             0x08 => vec![0x12, 0x34],
             0x09 => 1,
-            0x0A => true,
+            0x0A => 2,
         };
         let returned_make_credential_parameters =
             AuthenticatorMakeCredentialParameters::try_from(cbor_value).unwrap();
@@ -635,7 +635,7 @@ mod test {
             options,
             pin_uv_auth_param: Some(vec![0x12, 0x34]),
             pin_uv_auth_protocol: Some(1),
-            enterprise_attestation: Some(true),
+            enterprise_attestation: Some(2),
         };
 
         assert_eq!(
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -939,6 +939,24 @@ impl From<SetMinPinLengthParams> for cbor::Value {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub enum EnterpriseAttestationMode {
+    VendorFacilitated = 0x01,
+    PlatformManaged = 0x02,
+}
+
+impl TryFrom<u64> for EnterpriseAttestationMode {
+    type Error = Ctap2StatusCode;
+
+    fn try_from(value: u64) -> Result<Self, Ctap2StatusCode> {
+        match value {
+            1 => Ok(EnterpriseAttestationMode::VendorFacilitated),
+            2 => Ok(EnterpriseAttestationMode::PlatformManaged),
+            _ => Err(Ctap2StatusCode::CTAP2_ERR_INVALID_OPTION),
+        }
+    }
+}
+
 #[derive(Clone, Copy, Debug, PartialEq)]
 #[cfg_attr(test, derive(IntoEnumIterator))]
 pub enum CredentialManagementSubCommand {
@@ -1793,6 +1811,22 @@ mod test {
             0x03 => true,
         };
         assert_eq!(cbor::Value::from(config_sub_command_params), cbor_params);
+    }
+
+    #[test]
+    fn test_from_enterprise_attestation_mode() {
+        assert_eq!(
+            EnterpriseAttestationMode::try_from(1),
+            Ok(EnterpriseAttestationMode::VendorFacilitated),
+        );
+        assert_eq!(
+            EnterpriseAttestationMode::try_from(2),
+            Ok(EnterpriseAttestationMode::PlatformManaged),
+        );
+        assert_eq!(
+            EnterpriseAttestationMode::try_from(3),
+            Err(Ctap2StatusCode::CTAP2_ERR_INVALID_OPTION),
+        );
     }
 
     #[test]
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -1816,6 +1816,10 @@ mod test {
     #[test]
     fn test_from_enterprise_attestation_mode() {
         assert_eq!(
+            EnterpriseAttestationMode::try_from(0),
+            Err(Ctap2StatusCode::CTAP2_ERR_INVALID_OPTION),
+        );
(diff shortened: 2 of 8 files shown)


src/ctap/command.rs
```diff
@@ -601,7 +601,7 @@ mod test {
             0x05 => cbor_array![],
             0x08 => vec![0x12, 0x34],
             0x09 => 1,
-            0x0A => true,
+            0x0A => 2,
```
ia0, line 604: I'm assuming this is within the same draft, so there's no compatibility concerns.

src/ctap/data_formats.rs
```diff
@@ -939,6 +939,24 @@ impl From<SetMinPinLengthParams> for cbor::Value {
...
+    PlatformManaged = 0x02,
+}
+
+impl TryFrom<u64> for EnterpriseAttestationMode {
+    type Error = Ctap2StatusCode;
+
+    fn try_from(value: u64) -> Result<Self, Ctap2StatusCode> {
+        match value {
+            1 => Ok(EnterpriseAttestationMode::VendorFacilitated),
+            2 => Ok(EnterpriseAttestationMode::PlatformManaged),
```
ia0, line 954: Why not `0x02`? I think in another CL you explicitly used the `0x` syntax everywhere.

src/ctap/data_formats.rs
```diff
@@ -1795,6 +1813,22 @@ mod test {
...
+        assert_eq!(
+            EnterpriseAttestationMode::try_from(1),
+            Ok(EnterpriseAttestationMode::VendorFacilitated),
+        );
+        assert_eq!(
+            EnterpriseAttestationMode::try_from(2),
+            Ok(EnterpriseAttestationMode::PlatformManaged),
+        );
+        assert_eq!(
+            EnterpriseAttestationMode::try_from(3),
```
ia0, line 1831: Should we test 0 too?

src/ctap/mod.rs
```diff
@@ -86,6 +87,18 @@ const USE_BATCH_ATTESTATION: bool = false;
...
+// - EnterpriseAttestationMode::VendorFacilitated,
+// - EnterpriseAttestationMode::PlatformManaged.
+// For VendorFacilitated, choose an appriopriate ENTERPRISE_RP_ID_LIST.
+// To enable the feature, send the subcommand enableEnterpriseAttestation in
+// AuthenticatorConfig. An enterprise might want to customize the type of
+// attestation that is used. OpenSK defaults to batch attestation. Configuring
+// individual certificates then makes authenticators identifiable. Do NOT set
+// USE_BATCH_ATTESTATION to true at the same time in this case!
+pub const ENTERPRISE_ATTESTATION_MODE: Option<EnterpriseAttestationMode> = None;
+const ENTERPRISE_RP_ID_LIST: Vec<String> = Vec::new();
```
ia0: I don't think this can be instantiated with non-empty values, because I don't think non-empty string can be constant-created. What about using `&'static [&'static str]` instead?

ia0 (dismissed):
LGTM for storage (added some random non-CTAP/Rust-only comments outside storage though)

src/ctap/command.rs
kaczmarczyck: This actually didn't change within the specification, it's just that I assigned the wrong type to the stub.

src/ctap/data_formats.rs
kaczmarczyck: I used the hex notation in all commands and responses to reflect the specification. In this case, it reads e.g.
> If the enterpriseAttestation parameter’s value is not 1 or 2, then end the operation by returning CTAP2_ERR_INVALID_OPTION.

I had conflicting feelings about this, but decided to stick with the reasoning.

ia0: I see, makes sense.

src/ctap/data_formats.rs
kaczmarczyck: Done

src/ctap/mod.rs
kaczmarczyck: Changed both stubs, here and in `storage.rs`.

src/ctap/storage.rs
```diff
@@ -1203,7 +1209,8 @@ mod test {
             persistent_store.set_min_pin_length_rp_ids(rp_ids.clone()),
             Ok(())
         );
-        for rp_id in DEFAULT_MIN_PIN_LENGTH_RP_IDS {
+        for rp_id in DEFAULT_MIN_PIN_LENGTH_RP_IDS.iter() {
+            let rp_id = rp_id.to_string().to_string();
```
ia0: That's weird. You're sure you need both `to_string()`? What about `for &rp_id in ...`? Same for the loop above.

ia0 (dismissed):

src/ctap/storage.rs
kaczmarczyck: Ah, I forgot to fix the test to match the implementation.

ia0 (dismissed):
Approve for storage.

src/ctap/mod.rs
```diff
@@ -86,6 +87,18 @@ const USE_BATCH_ATTESTATION: bool = false;
...
+// disabled by default. You can choose between
+// - EnterpriseAttestationMode::VendorFacilitated,
+// - EnterpriseAttestationMode::PlatformManaged.
+// For VendorFacilitated, choose an appriopriate ENTERPRISE_RP_ID_LIST.
+// To enable the feature, send the subcommand enableEnterpriseAttestation in
+// AuthenticatorConfig. An enterprise might want to customize the type of
+// attestation that is used. OpenSK defaults to batch attestation. Configuring
+// individual certificates then makes authenticators identifiable. Do NOT set
+// USE_BATCH_ATTESTATION to true at the same time in this case!
+pub const ENTERPRISE_ATTESTATION_MODE: Option<EnterpriseAttestationMode> = None;
```
jmichelp, line 105: Can be done in a future PR but considering how many configuration constants we have, this could make sense now to move all of them into a `configuration.rs` file.

src/ctap/mod.rs
```diff
@@ -86,6 +87,18 @@ const USE_BATCH_ATTESTATION: bool = false;
...
+// This flag allows usage of enterprise attestation. For privacy reasons, it is
+// disabled by default. You can choose between
+// - EnterpriseAttestationMode::VendorFacilitated,
+// - EnterpriseAttestationMode::PlatformManaged.
+// For VendorFacilitated, choose an appriopriate ENTERPRISE_RP_ID_LIST.
+// To enable the feature, send the subcommand enableEnterpriseAttestation in
+// AuthenticatorConfig. An enterprise might want to customize the type of
+// attestation that is used. OpenSK defaults to batch attestation. Configuring
+// individual certificates then makes authenticators identifiable. Do NOT set
+// USE_BATCH_ATTESTATION to true at the same time in this case!
```
jmichelp: If this is not an allowed configuration, do we have a way to assert it at compilation time and prevent a mis-configured firmware from being compiled?

src/ctap/mod.rs
kaczmarczyck: I put it on our agenda. We could also allow configuration through a command. I'll do a new PR with whatever we decide is the best solution.

src/ctap/mod.rs
kaczmarczyck: Test added, with a lengthy description to explain what it is about and what it is not about.

src/ctap/mod.rs
```diff
@@ -2720,4 +2764,19 @@ mod test {
             ))
         );
     }
+
+    #[test]
```
jmichelp: I would have put the `assert!` directly in the code, not as a test.
Otherwise nothing prevents me from configuring a wrong firmware and compiling it. I will never see the error unless I run `cargo test`.

kaczmarczyck: Assertion moved to code. Also I updated and merged the documentation. If we move to a configuration command, this property will probably be asserted there.

jmichelp (approved):

src/ctap/mod.rs
ia0: Note that the current test still doesn't prevent from compiling. The key will just crash when the assert is executed. If we want static asserts we need a crate (it's still not supported by the language).

kaczmarczyck: True, but I decided to not introduce a new crate before we agreed on a long term plan for all constants. Also, if you make this invalid change after getting a pretty big warning in the documentation directly above that constant, this should be okay for now? Users' privacy should be sufficiently protected IMO.

ia0: Yeah no problem, I was just mentioning it because the initial message from @jmichelp says:

> nothing prevents me from configuring a wrong firmware and compiling it

And this is still the case :D But yeah I hope if someone attempts to build a non-vanilla firmware, they would read the documentation surrounding their change.

PR #283, opened 2021-02-05
Add the option alwaysUv

kaczmarczyck:
More progress on #106, implementing the `alwaysUv` option.

This command completes the implementation of `authenticatorConfig`. (There is the subcommand `Vendor Prototype Command`, but we don't need it - yet, at least.)

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,11 +126,8 @@ a few things you can personalize:
 1.  Increase the `MAX_CRED_BLOB_LENGTH` in `ctap/mod.rs`, if you expect blobs
     bigger than the default value.
 1.  If a certification (additional to FIDO's) requires that all requests are
-    protected with user verification, set `CAN_DISABLE_ALWAYS_UV` in
-    `ctap/config_command.rs` to `false`. In that case, consider deploying
-    authenticators after calling `toggleAlwaysUv` to activate the feature.
-    Alternatively, you could change `ctap/storage.rs` to set `alwaysUv` in its
-    initialization.
+    protected with user verification, set `ENFORCE_ALWAYS_UV` in
+    `ctap/config_mod.rs` to `true`.
 
 ### 3D printed enclosure
 
--- a/README.md
+++ b/README.md
@@ -125,6 +125,12 @@ a few things you can personalize:
     length.
 1.  Increase the `MAX_CRED_BLOB_LENGTH` in `ctap/mod.rs`, if you expect blobs
     bigger than the default value.
+1.  If a certification (additional to FIDO's) requires that all requests are
+    protected with user verification, set `CAN_DISABLE_ALWAYS_UV` in
+    `ctap/config_command.rs` to `false`. In that case, consider deploying
+    authenticators after calling `toggleAlwaysUv` to activate the feature.
+    Alternatively, you could change `ctap/storage.rs` to set `alwaysUv` in its
+    initialization.
 
 ### 3D printed enclosure
 
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -341,6 +341,14 @@ where
         Ok(())
     }
 
+    // Returns whether CTAP1 commands are currently supported.
+    // If alwaysUv is enabled and the authenticator does not support internal UV,
+    // CTAP1 needs to be disabled.
+    #[cfg(feature = "with_ctap1")]
+    pub fn allows_ctap1(&self) -> Result<bool, Ctap2StatusCode> {
+        Ok(!self.persistent_store.has_always_uv()?)
+    }
+
     // Encrypts the private key and relying party ID hash into a credential ID. Other
     // information, such as a user name, are not stored, because encrypted credential IDs
     // are used for credentials stored server-side. Also, we want the key handle to be
@@ -642,7 +650,11 @@ where
                 UP_FLAG | UV_FLAG | AT_FLAG | ed_flag
             }
             None => {
-                if self.persistent_store.pin_hash()?.is_some() {
+                if self.persistent_store.has_always_uv()? {
+                    return Err(Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED);
+                }
+                // Corresponds to makeCredUvNotRqd set to true.
+                if options.rk && self.persistent_store.pin_hash()?.is_some() {
                     return Err(Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED);
                 }
                 if options.uv {
@@ -927,8 +939,10 @@ where
                 UV_FLAG
             }
             None => {
+                if self.persistent_store.has_always_uv()? {
+                    return Err(Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED);
+                }
                 if options.uv {
-                    // The specification (inconsistently) wants CTAP2_ERR_UNSUPPORTED_OPTION.
                     return Err(Ctap2StatusCode::CTAP2_ERR_INVALID_OPTION);
                 }
                 0x00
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -125,6 +125,12 @@ a few things you can personalize:
     length.
 1.  Increase the `MAX_CRED_BLOB_LENGTH` in `ctap/mod.rs`, if you expect blobs
     bigger than the default value.
+1.  If a certification (additional to FIDO's) requires that all requests are
+    protected with user verification, set `CAN_DISABLE_ALWAYS_UV` in
```
ia0: Negative flags are always a bit confusing because you get double negations. Is this constant name used in the specification? Otherwise what about something like ENABLE_ALWAYS_UV?

src/ctap/mod.rs
```diff
@@ -1017,6 +1031,23 @@ where
...
+        {
+            if !has_always_uv {
+                versions.insert(0, String::from(U2F_VERSION_STRING))
+            }
+        }
+        #[cfg(not(feature = "with_ctap1"))]
+        let versions = vec![
+            String::from(FIDO2_VERSION_STRING),
+            String::from(FIDO2_1_VERSION_STRING),
+        ];
```
ia0: I guess this is convoluted because of the "mut" that should only be present with ctap1. What about disabling the lint for that line instead and only define the versions variable once? The lint may actually be disable only when not ctap1.

ia0 (dismissed):
LGTM for storage.

README.md
kaczmarczyck: I want to express a slight difference between ENABLE and CAN_DISABLE. Enabling `alwaysUv` is always possible, but disabling needs a device reset. The default is `true`, that you can disable `alwaysUv` again. In that sense, it is kind of positive? Or would any of those be better?
- `IS_ALWAYS_UV_PERMANENT`
- `LOCK_ALWAYS_UV`

(And then revert the bool value.)

Quoting the specification:
> Authenticators SHOULD support users disabling the Always Require User Verification feature unless required not to by specific external certifications such as [CMVP].

src/ctap/mod.rs
kaczmarczyck: Done!

README.md
ia0: I see. We have some boolean feature A (always UV) and want to control whether that feature is active with a boolean constant B. When B forbids A, the default value of A is true. We can't rename A because it's in the spec, so we have to live with the fact that the default value for A (when it is absent) is true. That's quite confusing since usually absent features have a value of false. But that's the spec...

Does the spec also specifies that the default value for A should be false in the storage when B forbids A? This looks even more confusing. I would expect that when B forbids A then the value in the storage is true, without requiring to manually toggle A from false to true with a command.

kaczmarczyck: Let me restate what this does with A and B.

There is a command that toggle A on and off. Our constant B makes this toggle one-way: If B is `false` (not the default with my current choice), then A can only be toggled on, not off.

So if you want the default to be `false`, then we could call it any of the above, or with your word choice, maybe:

- `FORBID_DISABLE_ALWAYS_UV`
- `ONLY_ENABLE_ALWAYS_UV`

ia0: My concern was about the default for A, not B. In the current setup B makes an operation irreversible. I think it would be less confusing to have A automatically on when B is on, thus there's nothing to toggle. Does the specification really states that B means "can't toggle A off" and not "A should always be on"?

kaczmarczyck: At some point, it says:

> Set the alwaysUv option ID to its default.

Which means custom defaults should be possible. So we could do the following:

`ENFORCE_ALWAYS_UV = false` is the new default constant. This default behaves exactly like the current implementation. If it's `true`, it behaves like the current implementation after the first call to `toggle_always_uv`. It means we don't need an extra configuration steps that you probably always want to do after changing the constant. WDYT?

ia0: Yes the following changes sound more intuitive and simpler for the user:
- Renaming B to `ENFORCE_ALWAYS_UV`
- Setting A to true during storage init if B is true
- Rejecting toggle commands if B is true
- Removing the last 2 sentences in the readme

In particular:
- Setting `ENFORCE_ALWAYS_UV = true` means that you can't toggle A and it's always true.
- Setting `ENFORCE_ALWAYS_UV = false` means that you can toggle A and it's initially false.

src/ctap/mod.rs
```diff
@@ -129,6 +129,9 @@ pub const ES256_CRED_PARAM: PublicKeyCredentialParameter = PublicKeyCredentialPa
 const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
 // Maximum size stored with the credBlob extension. Must be at least 32.
 const MAX_CRED_BLOB_LENGTH: usize = 32;
+// Enforce the alwaysUv option. With this constant set to true, commands require
+// a PIN to be set up. The command toggleAlwaysUv will fail to disable alwaysUv.
```
ia0: "The command toggleAlwaysUv will always fail."?

src/ctap/storage.rs
```diff
@@ -622,6 +625,9 @@ impl PersistentStore {
 
     /// Enables alwaysUv, when disabled, and vice versa.
     pub fn toggle_always_uv(&mut self) -> Result<(), Ctap2StatusCode> {
+        if ENFORCE_ALWAYS_UV {
+            return Ok(());
```
ia0: Would it make sense to directly return `Ctap2StatusCode::CTAP2_ERR_OPERATION_DENIED` here? Alternatively return `INTERNAL_ERROR`.

src/ctap/storage.rs
```diff
@@ -1331,11 +1337,15 @@ mod test {
         let mut rng = ThreadRng256 {};
         let mut persistent_store = PersistentStore::new(&mut rng);
 
-        assert!(!persistent_store.has_always_uv().unwrap());
-        assert_eq!(persistent_store.toggle_always_uv(), Ok(()));
-        assert!(persistent_store.has_always_uv().unwrap());
-        assert_eq!(persistent_store.toggle_always_uv(), Ok(()));
-        assert!(!persistent_store.has_always_uv().unwrap());
+        if ENFORCE_ALWAYS_UV {
+            assert!(persistent_store.has_always_uv().unwrap());
```
ia0, line 1377: We can also test `toggle_always_uv`.

README.md
kaczmarczyck: Implemented, except
> Setting A to true during storage init if B is true

I instead just treat A as if it was set to true during init. The only difference is if you flash a new firmware without `ENFORCE_ALWAYS_UV`, you disable `alwaysUv`. But we save one flash write. Preferences?

ia0 (dismissed):
We currently don't test half of the change, but hopefully there's a plan to either move those configurations to compile time and run the tests many times, or to parametrize the configurations and run the tests with different values for the relevant parameters.

README.md
ia0: I actually prefer the current version, so all good.

src/ctap/storage.rs
kaczmarczyck: Done

src/ctap/storage.rs
kaczmarczyck: I wasn't sure if I want that logic to live inside `storage.rs`, but it's more strict, so can't hurt. :) Changed.

src/ctap/mod.rs
kaczmarczyck: I rephrased, and made it more general.

ia0 (approved):

jmichelp (approved):

PR #284, opened 2021-02-10
Fix `config.py` tool according to the new API of fido2 python package

### LINKED ISSUE #269: OpenSK does not work properly on Microsoft Edge
## Expected Behavior
I can register the OpenSK key on my Google account

## Actual Behavior
When I try to register the OpenSK key on my Google account, it gets stuck on the confirmation page
Here are the logs:
```
two-step-verification?rapt=***:1 Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
two-step-verification?rapt=***:1 Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
two-step-verification?rapt=***:1 Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
two-step-verification?rapt=***:1 Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
```

1. Flash the OpenSK firmware to nRF52840-Dongle
2. Open https://myaccount.google.com/signinoptions/two-step-verification on Microsoft Edge
3. Click "Add security key", and click "Next"
4. Tap the button on the OpenSK key

## Specifications

- Version: latest stable
- Platform: Windows

jmichelp:
Fixes #269

Since fido2 v0.8.1 which we developed our configuration tool on, the device descriptor
moved from python `dict` to `NamedTuple`. This PR updates to this new API and ensure 
that we're using the correct version of fido2 module.

- [x] Tests pass

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -46,4 +46,4 @@ mkdir -p elf2tab
 cargo install elf2tab --version 0.6.0 --root elf2tab/
 
 # Install python dependencies to factory configure OpenSK (crypto, JTAG lockdown)
-pip3 install --user --upgrade colorama tqdm cryptography fido2
+pip3 install --user --upgrade colorama tqdm cryptography "fido2>=0.9.1"
--- a/tools/configure.py
+++ b/tools/configure.py
@@ -64,8 +64,7 @@ def info(msg):
 def get_opensk_devices(batch_mode):
   devices = []
   for dev in hid.CtapHidDevice.list_devices():
-    if (dev.descriptor.vid,
-        dev.descriptor.pid) == OPENSK_VID_PID:
+    if (dev.descriptor.vid, dev.descriptor.pid) == OPENSK_VID_PID:
       if dev.capabilities & hid.CAPABILITY.CBOR:
         if batch_mode:
           devices.append(ctap2.CTAP2(dev))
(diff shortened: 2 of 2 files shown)


jmichelp:
Considering we're not using virtualenv here, I'm unsure whether we should use `fido2>=0.9.1` or `fido2==0.9.1` as a version description.
The latter may prevent people from updating the package or even cause an error (I'm not sure how pip behaves if it has to downgrade a package)

setup.sh
```diff
@@ -46,4 +46,4 @@ mkdir -p elf2tab
 cargo install elf2tab --version 0.6.0 --root elf2tab/
 
 # Install python dependencies to factory configure OpenSK (crypto, JTAG lockdown)
-pip3 install --user --upgrade colorama tqdm cryptography fido2
+pip3 install --user --upgrade colorama tqdm cryptography "fido2>=0.9.1"
```
kaczmarczyck, line 49: If this list is growing, it might make sense to go for
```shell
pip3 install -r requirements.txt
```

and have the packets in a `requirements.txt` doc? We also have i.e. `tools/configure.py`, and the requirements could then indicate more easily which requirement is from where when things get more complicated.

kaczmarczyck (dismissed):

setup.sh
jmichelp: That's something we can do in the `develop` branch for the next release. For `stable` I would like to keep it as a small bugfix.

kaczmarczyck: Sounds good!

kaczmarczyck (approved):

PR #287, opened 2021-02-18
Improved documentation for customization

kaczmarczyck:
I moved all customizable constants to `customization.rs`. The documentation is now in one place, with parts of the README and old comments moved, updated and added.

This change is the non-controversial part of a bigger effort to improve unit tests for different configurations. Plan is to inject these constants into `CtapState`.

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -352,8 +352,8 @@ class OpenSKInstaller:
 
   def build_opensk(self):
     info("Building OpenSK application")
-    self._build_app_or_example(is_example=False)
     self._check_invariants()
+    self._build_app_or_example(is_example=False)
 
   def _build_app_or_example(self, is_example):
     assert self.args.application
--- a/deploy.py
+++ b/deploy.py
@@ -353,6 +353,7 @@ class OpenSKInstaller:
   def build_opensk(self):
     info("Building OpenSK application")
     self._build_app_or_example(is_example=False)
+    self._check_invariants()
 
   def _build_app_or_example(self, is_example):
     assert self.args.application
@@ -389,6 +390,11 @@ class OpenSKInstaller:
     app_path = os.path.join(app_path, self.args.application)
     # Create a TAB file
     self.create_tab_file({props.arch: app_path})
+
+  def _check_invariants(self):
+    print("Testing invariants in customization.rs...")
+    self.checked_command_output(
+        ["cargo", "test", "--features=std", "--lib", "customization"])
 
   def generate_crypto_materials(self, force_regenerate):
     has_error = subprocess.call([
@@ -616,15 +622,9 @@ class OpenSKInstaller:
     if self.args.programmer == "none":
       assert_python_library("intelhex")
 
-  def check_invariants(self):
-    print("Testing invariants in customization.rs...")
-    self.checked_command_output(
-        ["cargo", "test", "--features=std", "--lib", "customization"])
-
   def run(self):
     self.check_prerequisites()
     self.update_rustc_if_needed()
-    self.check_invariants()
 
     if not (self.args.tockos or self.args.application or
             self.args.clear_storage):
--- a/deploy.py
+++ b/deploy.py
@@ -618,8 +618,8 @@ class OpenSKInstaller:
 
   def check_invariants(self):
     print("Testing invariants in customization.rs...")
-    self.checked_command_output(["cargo", "test", "--features=std", "--lib",
-                                 "customization"])
+    self.checked_command_output(
+        ["cargo", "test", "--features=std", "--lib", "customization"])
 
   def run(self):
     self.check_prerequisites()
--- a/deploy.py
+++ b/deploy.py
@@ -616,9 +616,15 @@ class OpenSKInstaller:
     if self.args.programmer == "none":
       assert_python_library("intelhex")
 
+  def check_invariants(self):
+    print("Testing invariants in customization.rs...")
+    self.checked_command_output(["cargo", "test", "--features=std", "--lib",
+                                 "customization"])
+
   def run(self):
     self.check_prerequisites()
     self.update_rustc_if_needed()
+    self.check_invariants()
 
     if not (self.args.tockos or self.args.application or
             self.args.clear_storage):
--- a/src/ctap/customization.rs
+++ b/src/ctap/customization.rs
@@ -12,11 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! This file contains all customizable constants.
+//!
+//! If you adapt them, make sure to run the tests before flashing the firmware.
+//! Our deploy script enforces the invariants.
+
 use crate::ctap::data_formats::{CredentialProtectionPolicy, EnterpriseAttestationMode};
 
-/// ###########################################################################
-/// Constants for adjusting privacy and protection levels.
-/// ###########################################################################
+// ###########################################################################
+// Constants for adjusting privacy and protection levels.
+// ###########################################################################
 
 /// Changes the default level for the credProtect extension.
 ///
@@ -25,16 +30,18 @@ use crate::ctap::data_formats::{CredentialProtectionPolicy, EnterpriseAttestatio
 /// - CredentialProtectionPolicy::UserVerificationRequired
 ///
 /// UserVerificationOptionalWithCredentialIdList
-/// Resident credentials that are not in the allowList or excludeList can not be
-/// discovered without user verification.
+/// Resident credentials are discoverable with
+/// - an allowList,
+/// - an excludeList,
+/// - user verification.
 ///
 /// UserVerificationRequired
-/// No resident credentials can be discovered without user verification.
+/// Resident credentials are discoverable with user verification only.
 ///
 /// This can improve privacy, but can make usage less comfortable.
 pub const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
 
-/// Sets the intial minimum PIN length in code points.
+/// Sets the initial minimum PIN length in code points.
 ///
 /// # Invariant
 ///
@@ -42,8 +49,6 @@ pub const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
 /// - The minimum PIN length must be at most 63.
 /// - DEFAULT_MIN_PIN_LENGTH_RP_IDS must be non-empty if MAX_RP_IDS_LENGTH is 0.
 ///
-/// Only the RP IDs listed in DEFAULT_MIN_PIN_LENGTH_RP_IDS are allowed to read
-/// the minimum PIN length with the minPinLength extension.
 /// Requiring longer PINs can help establish trust between users and relying
 /// parties. It makes user verification harder to break, but less convenient.
 /// NIST recommends at least 6-digit PINs in section 5.1.9.1:
@@ -51,6 +56,15 @@ pub const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
 ///
 /// Reset reverts the minimum PIN length to this DEFAULT_MIN_PIN_LENGTH.
 pub const DEFAULT_MIN_PIN_LENGTH: u8 = 4;
+
+/// Lists relying parties that can read the minimum PIN length.
+///
+/// # Invariant
+///
+/// - DEFAULT_MIN_PIN_LENGTH_RP_IDS must be non-empty if MAX_RP_IDS_LENGTH is 0
+///
+/// Only the RP IDs listed in DEFAULT_MIN_PIN_LENGTH_RP_IDS are allowed to read
+/// the minimum PIN length with the minPinLength extension.
 pub const DEFAULT_MIN_PIN_LENGTH_RP_IDS: &[&str] = &[];
 
 /// Enforces the alwaysUv option.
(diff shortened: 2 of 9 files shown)


src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::ctap::data_formats::{CredentialProtectionPolicy, EnterpriseAttestationMode};
+
+/// ###########################################################################
+/// Constants for adjusting privacy and protection levels.
+/// ###########################################################################
```
ia0: This doesn't make sense. `///` is used for documentation not for comments. Documentation is either attached to the following item using `///` or the containing item using `//!`. It doesn't make sense to have a `///` block followed by an empty line. Is the intent here to describe the file content? In which case it would be:

```rust
//! Constants for adjusting privacy and protection levels.
```

If you just want to cut the file into different parts (and don't want to use modules), you just want comments:

```rust
// ###########################################################################
// Constants for adjusting privacy and protection levels.
// ###########################################################################
```

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+/// You can change this value to one of the following for more privacy:
+/// - CredentialProtectionPolicy::UserVerificationOptionalWithCredentialIdList
+/// - CredentialProtectionPolicy::UserVerificationRequired
+///
+/// UserVerificationOptionalWithCredentialIdList
+/// Resident credentials that are not in the allowList or excludeList can not be
+/// discovered without user verification.
+///
+/// UserVerificationRequired
+/// No resident credentials can be discovered without user verification.
```
ia0: Shouldn't those be part of the `CredentialProtectionPolicy` documentation?

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+///
+/// This can improve privacy, but can make usage less comfortable.
+pub const DEFAULT_CRED_PROTECT: Option<CredentialProtectionPolicy> = None;
+
+/// Sets the intial minimum PIN length in code points.
```
ia0: typo: initial

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+/// Only the RP IDs listed in DEFAULT_MIN_PIN_LENGTH_RP_IDS are allowed to read
+/// the minimum PIN length with the minPinLength extension.
+/// Requiring longer PINs can help establish trust between users and relying
+/// parties. It makes user verification harder to break, but less convenient.
+/// NIST recommends at least 6-digit PINs in section 5.1.9.1:
+/// https://pages.nist.gov/800-63-3/sp800-63b.html
+///
+/// Reset reverts the minimum PIN length to this DEFAULT_MIN_PIN_LENGTH.
+pub const DEFAULT_MIN_PIN_LENGTH: u8 = 4;
+pub const DEFAULT_MIN_PIN_LENGTH_RP_IDS: &[&str] = &[];
```
ia0, line 68: This item is not documented. Not sure if intended.

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+    assert!(MAX_LARGE_BLOB_ARRAY_SIZE >= 1024);
+    // Storage tests check if MAX_LARGE_BLOB_ARRAY_SIZE fits the shards.
+    if MAX_RP_IDS_LENGTH == 0 {
+        assert!(!DEFAULT_MIN_PIN_LENGTH_RP_IDS.is_empty());
+    }
+    // Storage keys tests check if MAX_SUPPORTED_RESIDENT_KEYS fits CREDENTIALS.
+}
+
+#[cfg(test)]
+mod test {
```
jmichelp, line 242: I'm pretty sure that nobody's going to run `cargo test` after changing the customization.
And if I remember our previous discussion, compile-time asserts are not supported yet without a crate.

This leaves us with one solution: can we run only this unit tests from `customization.rs` before we compile OpenSK app in `deploy.py`?

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+/// - EnterpriseAttestationMode::VendorFacilitated
+/// - EnterpriseAttestationMode::PlatformManaged
+///
+/// VendorFacilitated
+/// Enterprise attestation is restricted to ENTERPRISE_RP_ID_LIST. Add your
+/// enterprises domain, e.g. "example.com", to the list below.
+///
+/// PlatformManaged
+/// All relying parties can request an enterprise attestation. The authenticator
+/// trusts the platform to filter requests.
```
ia0, line 97: Same, those should probably just be documented in `EnterpriseAttestationMode`.

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+/// individual certificates then makes authenticators identifiable.
+///
+/// OpenSK prevents activating batch and enterprise attestation together. The
+/// current implementation uses the same key material at the moment, and these
+/// two modes have conflicting privacy guarantees.
+/// If you implement your own enterprise attestation mechanism, and you want
+/// batch attestation at the same time, proceed carefully and remove the
+/// assertion.
+pub const ENTERPRISE_ATTESTATION_MODE: Option<EnterpriseAttestationMode> = None;
+pub const ENTERPRISE_RP_ID_LIST: &[&str] = &[];
```
ia0, line 120: This item is not documented. Not sure if intended.

src/ctap/customization.rs
```diff
@@ -0,0 +1,247 @@
...
+///
+/// https://www.w3.org/TR/webauthn/#signature-counter
+pub const USE_SIGNATURE_COUNTER: bool = true;
+
+/// ###########################################################################
+/// Constants for performance optimization or adapting to different hardware.
+///
+/// Those constants may be modified before compilation to tune the behavior of
+/// the key.
+/// ###########################################################################
```
ia0: Same here. This should probably be a comment.

src/ctap/customization.rs
ia0: Those seem to be check when the board boots (and crash it). But yeah, would be good to have something at compile time. We could run something like `cargo test --features=std --lib customization` or `cargo test --features=std --lib static_test` where all tests that should run statically contain `static_test` in their name.

kaczmarczyck: This PR does not change the current behaviour. It still panics when the configuration is wrong, just tests more. It additionally has a test to check the invariants. Adding this test to deploy sounds good. Question is if we still need to check invariants in `CtapState::new()` then?

ia0: If broken invariants would have bad consequences (like corrupting data, leaking data, etc.) then we need them, if the device would otherwise crash without other bad consequences then we probably don't need them, but that's just optimization. Those tests look fast.

jmichelp: Preventing compilation of a bad firmware seems to me a better behavior than a device crashing at boot (and potentially people starting opening issues). I'm fine with merging this PR until we agree :)

src/ctap/customization.rs
kaczmarczyck: I split the documentation. Initial thought was that these two constants should be explained together, but I can split for consistency and relevance of documentation.

src/ctap/customization.rs
kaczmarczyck: Split like above.

src/ctap/customization.rs
kaczmarczyck: Done

src/ctap/customization.rs
kaczmarczyck: It is just a comment, thanks, fixed. I also added a short top level documentation.

src/ctap/customization.rs
kaczmarczyck: IMO Usually all tests should pass before you deploy. The difference here is that we expect non-developers to change these values, so we want the extra protection. The test in main wasn't really the end-game solution anyway, so I went ahead and added
```rust
cargo test --features=std --lib customization
```

to `deploy.py`.

src/ctap/customization.rs
kaczmarczyck: Fixed

src/ctap/customization.rs
kaczmarczyck: I think the intent is different:

Here, documentation is about the implications of setting a _default_ level. In `data_formats.rs`, it is about the data structure, that is used in the default level and the per-credential level. The documentation is similar, of course. I added it there.
But, I really want this information to also be in `customization.rs`, as users shouldn't have to go look for the information they need to make an educated choice.

src/ctap/customization.rs
kaczmarczyck: As above, documentation added to `data_formats.rs`, and kept here.

deploy.py
```diff
@@ -616,9 +616,15 @@ def check_prerequisites(self):
...
 
+  def check_invariants(self):
+    print("Testing invariants in customization.rs...")
+    self.checked_command_output(
+        ["cargo", "test", "--features=std", "--lib", "customization"])
+
   def run(self):
     self.check_prerequisites()
     self.update_rustc_if_needed()
+    self.check_invariants()
```
jmichelp: This should only be run in `self.build_opensk()` method.

src/ctap/customization.rs
ia0: It's not only consistency and relevance, it's the actual semantics of documentation. Previously only one item was documented. I think this comes from the fact that you expect users to read the file (in which case documentation looks a lot like comments if you don't know Rust) rather than read the generated documentation, which makes sense if they are expected to modify the file (and they don't really need to know Rust for that). So it would also makes sense for me if all documentation was switched to comments such that the file is meant to be read linearly and then multiple items can be described at once. I'm fine with both.

src/ctap/data_formats.rs
```diff
@@ -504,6 +504,22 @@ impl TryFrom<cbor::Value> for SignatureAlgorithm {
...
+/// The credential is always discoverable, as if it had no protection level.
+///
+/// UserVerificationOptionalWithCredentialIdList
+/// The credential is discoverable with
+/// - an allowList,
+/// - an excludeList,
+/// - user verification.
+///
+/// UserVerificationRequired
+/// The credentials is discoverable with user verification only.
```
ia0: Those documentation should be directly on the enum values.

```rust
/// Documentation of Foo
enum Foo {
    /// Documentation of Bar
    Bar,
    /// Documentation of Baz
    Baz,
}
```

src/ctap/data_formats.rs
```diff
@@ -939,6 +955,15 @@ impl From<SetMinPinLengthParams> for cbor::Value {
...
 
+/// The level of enterprise attestation allowed in MakeCredential.
+///
+/// VendorFacilitated
+/// Enterprise attestation is restricted to ENTERPRISE_RP_ID_LIST. Add your
+/// enterprises domain, e.g. "example.com", to the list below.
+///
+/// PlatformManaged
+/// All relying parties can request an enterprise attestation. The authenticator
+/// trusts the platform to filter requests.
```
ia0: Same here.

src/ctap/customization.rs
```diff
@@ -0,0 +1,268 @@
...
+        assert!(MAX_CRED_BLOB_LENGTH >= 32);
+        if let Some(count) = MAX_CREDENTIAL_COUNT_IN_LIST {
+            assert!(count >= 1);
+        }
+        assert!(MAX_LARGE_BLOB_ARRAY_SIZE >= 1024);
+        // Storage tests check if MAX_LARGE_BLOB_ARRAY_SIZE fits the shards.
+        if MAX_RP_IDS_LENGTH == 0 {
+            assert!(!DEFAULT_MIN_PIN_LENGTH_RP_IDS.is_empty());
+        }
+        // Storage keys tests check if MAX_SUPPORTED_RESIDENT_KEYS fits CREDENTIALS.
```
jmichelp: There's no test after this comment. Is it a left-over or a TODO?

src/ctap/customization.rs
ia0: Same as in another comment, it looks like you see this file as a linear read for the user. In which case it's not really source code but more instructions to the user to update some constants. In which case maybe you just want to use comments instead of documentation and then freely organize the file (ignoring that it's actually a Rust file and that documentation is code and has a semantics).

Documentation (and comments) is maintenance burden so it's preferable to avoid writing too much and also avoid duplication. If you still prefer to go for documentation instead of comments, I would consider the fact that from IDEs (like emacs, etc) one can easily jump to the definition of an item and thus its documentation. But it is also possible to read the documentation in a browser and click on hyperlinks:

```
cargo doc --document-private-items --features=std
google-chrome target/doc/ctap2/ctap/customization/index.html
```

To create a hyperlink inside documentation you just need to write `See [CredentialProtectionPolicy] for a detailed description of possible values.` but it's also possible in this case to click directly on the type of the constant.

ia0 (dismissed):

src/ctap/data_formats.rs
kaczmarczyck: Done

src/ctap/data_formats.rs
kaczmarczyck: Done

deploy.py

src/ctap/customization.rs
kaczmarczyck: The tests were in the same order as the Invariants in the file. These comments address missing tests, that are currently located elsewhere. I figured they would need more refactoring, and decided it is outside of the scope of this PR to include them here.
I moved the comments to the top of the function, with more explanation what they are about.

deploy.py
```diff
@@ -353,6 +353,7 @@ def build_example(self):
   def build_opensk(self):
     info("Building OpenSK application")
     self._build_app_or_example(is_example=False)
+    self._check_invariants()
```
jmichelp, line 355: Almost... Here you compile the firmware then run the tests. Just swap the 2 lines and we're good.

ia0 (dismissed):

deploy.py
kaczmarczyck: Sure, done, thx!

jmichelp (approved):

PR #290, opened 2021-03-03
HKDF for CTAP2.1

kaczmarczyck:
This implements the HKDF function as used in ClientPin. Only usage is

> HKDF-SHA-256(salt = 32 zero bytes, IKM = Z, L = 32, info = "CTAP2 HMAC key") ||
> HKDF-SHA-256(salt = 32 zero bytes, IKM = Z, L = 32, info = "CTAP2 AES key")

Therefore, I prefer to use an interface that returns a `[u8; 32]` directly, instead of a `Vec<u8>`.

Necessary for #106.

- [x] Tests pass


libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,207 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::hmac::hmac_256;
+use super::{Hash256, HashBlockSize64Bytes};
+
+const HASH_SIZE: usize = 32;
+
+/// Computes the HKDF with empty salt and 256 bit (one block) output.
+pub fn hkdf_256<H>(ikm: &[u8], info: &[u8]) -> [u8; 32]
```
gendx: Can you remind in the documentation what "ikm" stands for? Same for "info".

libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,207 @@
...
+    hmac_256::<H>(&prk, t.as_slice())
+}
+
+/// Computes the HKDF.
+///
+/// RFC: https://tools.ietf.org/html/rfc5869
+/// Argument order is taken from:
+/// https://fidoalliance.org/specs/fido-v2.1-rd-20201208/fido-client-to-authenticator-protocol-v2.1-rd-20201208.html#pinProto2
+#[cfg(test)]
+pub fn hkdf<H>(salt: &[u8], ikm: &[u8], l: u8, info: &[u8]) -> Vec<u8>
```
gendx, line 56: Same here for "l".

libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,207 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::hmac::hmac_256;
+use super::{Hash256, HashBlockSize64Bytes};
+
+const HASH_SIZE: usize = 32;
+
+/// Computes the HKDF with empty salt and 256 bit (one block) output.
```
gendx, line 20: I'd rename this function "hkdf_empty_salt_256bit_output" or at least "hkdf_empty_salt_256".

libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,207 @@
...
+pub fn hkdf_256<H>(ikm: &[u8], info: &[u8]) -> [u8; 32]
+where
+    H: Hash256 + HashBlockSize64Bytes,
+{
+    // Salt is a zero block here.
+    let prk = hmac_256::<H>(&[0; HASH_SIZE], ikm);
+    // l is implicitly 32, so we iterate exactly once.
+    let mut t = info.to_vec();
+    t.push(1);
```
gendx, line 37: Note: in the future we may want to make the HMAC API more flexible (init/update/finish API) to allow for multiple updates and remove the need to allocate a Vec here.

libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,207 @@
...
+        .unwrap();
+        assert_eq!(
+            hkdf::<Sha256>(salt.as_slice(), ikm.as_slice(), l, info.as_slice()),
+            okm
+        );
+    }
+
+    #[test]
+    fn test_hkdf_256_sha256_vectors() {
+        let test_okms = [
```
gendx, line 156: Can you document where the vectors of this function come from?

gendx (commented):
I didn't look into the algorithm details nor the test vectors, but overall the implementation looks good and the unit tests give good confidence that it matches the reference.

My only nit would be to better document the parameters taken from the RFC, explaining here what "ikm", "prk" and "l" mean. Re-using the reference names is fine with good documentation, or you may choose better names where it makes sense (e.g. "l" => "output_length"?).

libraries/crypto/src/hkdf.rs
kaczmarczyck: Renamed to the latter.

libraries/crypto/src/hkdf.rs
kaczmarczyck: Arguments section added, it reflects the RFC mostly.

libraries/crypto/src/hkdf.rs
kaczmarczyck: Done

libraries/crypto/src/hkdf.rs
kaczmarczyck: Noted. We have to reconsider the crypto API when we move away from software crypto eventually.

libraries/crypto/src/hkdf.rs
kaczmarczyck: Just some other implementation, documented.

libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,226 @@
...
+/// Computes the HKDF with empty salt and 256 bit (one block) output.
+///
+/// # Arguments
+///
+/// * `ikm` - Input keying material
+/// * `info` - Optional context and application specific information
+///
+/// This implementation is equivalent to the below hkdf, with `salt` set to the
+/// default block of zeros and the output length l as 32.
+pub fn hkdf_empty_salt_256<H>(ikm: &[u8], info: &[u8]) -> [u8; 32]
```
jmichelp: Shouldn't the return value be `[u8; HASH_SIZE]` ?

libraries/crypto/src/hkdf.rs
```diff
@@ -0,0 +1,226 @@
...
+        .unwrap();
+        assert_eq!(
+            hkdf::<Sha256>(salt.as_slice(), ikm.as_slice(), l, info.as_slice()),
+            okm
+        );
+    }
+
+    #[test]
+    fn test_hkdf_empty_salt_256_sha256_vectors() {
+        // Test vectors generated by i.e. pycryptodome using:
```
gendx: Nit: no need for "i.e." here.

libraries/crypto/src/hkdf.rs
kaczmarczyck: Done

libraries/crypto/src/hkdf.rs
kaczmarczyck: Removed.

gendx (approved):

PR #291, opened 2021-03-03
Renaming PIN protocol related variables to prepare v2

kaczmarczyck:
This is the first PR in a series to implement PIN protocol v2. It does not add or change any functionality, just renames variables and a file. `pin_uv_auth_protocol` now matches the specification name. Also, I removed the dummy `process_client_pin` function and now call `client_pin.process.command` directly to match the other command calls.

Plan is to have a new `pin_protocol.rs` later that implements the trait. The now called `client_pin.rs` file keeps its role to have the generic command implementation, hence `pin_protocol_v1.rs` did not fit anymore.

- [x] Tests pass

### DIFF
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -166,7 +166,7 @@ pub enum PinPermission {
     AuthenticatorConfiguration = 0x20,
 }
 
-pub struct PinProtocolV1 {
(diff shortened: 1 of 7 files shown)


jmichelp (approved):
Seems indeed to be a no-op.

PR #292, opened 2021-03-08
adds the PIN protocol trait

kaczmarczyck:
Continues #291 in implementing ClientPin for CTAP2.1.

The main work is the new `PinProtocol` struct and the `SharedSecret` trait that allow a generic ClientPin implementation. The ClientPin struct itself only moves to the new `PinProtocol` implementation for V1, without yet allowing the new V2. `client_pin.rs` and its tests are only adapted to deal with the new interface. This change does not change OpenSK's behaviour (with the small exception of one error code for wrong length arrays that was proactively updated to 2.1). Therefore, ClientPin tests do not reflect the upcoming changes yet, but were not refactored to make it obvious this change is refactoring only.

Plan for the next PRs:
1. use V2
2. add timeout for tokens
3. fully update the command algorithms

- [x] Tests pass

### DIFF
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -31,14 +31,21 @@ use subtle::ConstantTimeEq;
 
 /// The prefix length of the PIN hash that is stored and compared.
 ///
-/// The code assumes that this value is a multiple of the AES block length and
-/// fits an u8.
+/// The code assumes that this value is a multiple of the AES block length, fits
+/// an u8 and is at most as long as a SHA256. The value is fixed for all PIN
+/// protocols.
 pub const PIN_AUTH_LENGTH: usize = 16;
 
 /// The length of the pinUvAuthToken used throughout PIN protocols.
+///
+/// The code assumes that this value is a multiple of the AES block length. It
+/// is fixed since CTAP2.1.
 pub const PIN_TOKEN_LENGTH: usize = 32;
 
 /// The length of the encrypted PINs when received by SetPin or ChangePin.
+///
+/// The code assumes that this value is a multiple of the AES block length. It
+/// is fixed since CTAP2.1.
 const PIN_PADDED_LENGTH: usize = 64;
 
 /// Computes and encrypts the HMAC-secret outputs.
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -55,7 +55,7 @@ const PIN_PADDED_LENGTH: usize = 64;
 /// for those salts.
 fn encrypt_hmac_secret_output(
     rng: &mut impl Rng256,
-    shared_secret: &mut dyn SharedSecret,
+    shared_secret: &dyn SharedSecret,
     salt_enc: &[u8],
     cred_random: &[u8; 32],
 ) -> Result<Vec<u8>, Ctap2StatusCode> {
@@ -73,7 +73,7 @@ fn encrypt_hmac_secret_output(
 
 /// Decrypts the new_pin_enc and outputs the found PIN.
 fn decrypt_pin(
-    shared_secret: &mut dyn SharedSecret,
+    shared_secret: &dyn SharedSecret,
     new_pin_enc: Vec<u8>,
 ) -> Result<Vec<u8>, Ctap2StatusCode> {
     let decrypted_pin = shared_secret.decrypt(&new_pin_enc)?;
@@ -93,7 +93,7 @@ fn decrypt_pin(
 /// truncated for persistent storage.
 fn check_and_store_new_pin(
     persistent_store: &mut PersistentStore,
-    shared_secret: &mut dyn SharedSecret,
+    shared_secret: &dyn SharedSecret,
     new_pin_enc: Vec<u8>,
 ) -> Result<(), Ctap2StatusCode> {
     let pin = decrypt_pin(shared_secret, new_pin_enc)?;
@@ -148,7 +148,7 @@ impl ClientPin {
         &mut self,
         rng: &mut impl Rng256,
         persistent_store: &mut PersistentStore,
-        shared_secret: &mut dyn SharedSecret,
+        shared_secret: &dyn SharedSecret,
         pin_hash_enc: Vec<u8>,
     ) -> Result<(), Ctap2StatusCode> {
         match persistent_store.pin_hash()? {
@@ -211,10 +211,10 @@ impl ClientPin {
         if persistent_store.pin_hash()?.is_some() {
             return Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID);
         }
-        let mut shared_secret = self.pin_protocol_v1.decapsulate(key_agreement, 1)?;
+        let shared_secret = self.pin_protocol_v1.decapsulate(key_agreement, 1)?;
         shared_secret.verify(&new_pin_enc, &pin_auth)?;
 
-        check_and_store_new_pin(persistent_store, shared_secret.as_mut(), new_pin_enc)?;
+        check_and_store_new_pin(persistent_store, shared_secret.as_ref(), new_pin_enc)?;
         persistent_store.reset_pin_retries()?;
         Ok(())
     }
@@ -231,13 +231,13 @@ impl ClientPin {
         if persistent_store.pin_retries()? == 0 {
             return Err(Ctap2StatusCode::CTAP2_ERR_PIN_BLOCKED);
         }
-        let mut shared_secret = self.pin_protocol_v1.decapsulate(key_agreement, 1)?;
+        let shared_secret = self.pin_protocol_v1.decapsulate(key_agreement, 1)?;
         let mut auth_param_data = new_pin_enc.clone();
         auth_param_data.extend(&pin_hash_enc);
         shared_secret.verify(&auth_param_data, &pin_auth)?;
-        self.verify_pin_hash_enc(rng, persistent_store, shared_secret.as_mut(), pin_hash_enc)?;
+        self.verify_pin_hash_enc(rng, persistent_store, shared_secret.as_ref(), pin_hash_enc)?;
 
-        check_and_store_new_pin(persistent_store, shared_secret.as_mut(), new_pin_enc)?;
+        check_and_store_new_pin(persistent_store, shared_secret.as_ref(), new_pin_enc)?;
         self.pin_protocol_v1.reset_pin_uv_auth_token(rng);
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


src/ctap/client_pin.rs
```diff
@@ -14,141 +14,90 @@
...
-        return false;
-    }
-    verify_hmac_256_first_128bits::<Sha256>(
-        hmac_key,
-        hmac_contents,
-        array_ref![pin_auth, 0, PIN_AUTH_LENGTH],
-    )
-}
+/// The length of the pinUvAuthToken used throughout PIN protocols.
+pub const PIN_TOKEN_LENGTH: usize = 32;
```
jmichelp, line 43: I guess the same constraint apply here: this constant and the following one should be multiples of AES block size.

src/ctap/client_pin.rs
```diff
@@ -167,34 +116,32 @@ pub enum PinPermission {
...
+    ///
     /// Decrypts the encrypted pin_hash and compares it to the stored pin_hash.
     /// Resets or decreases the PIN retries, depending on success or failure.
     /// Also, in case of failure, the key agreement key is randomly reset.
     fn verify_pin_hash_enc(
         &mut self,
         rng: &mut impl Rng256,
         persistent_store: &mut PersistentStore,
-        aes_dec_key: &crypto::aes256::DecryptionKey,
+        shared_secret: &mut dyn SharedSecret,
```
jmichelp: why do we need a `mut SharedSecret`? Encryption and decryption shouldn't alter the original data, right?
This comment also applies to the methods above of course.

src/ctap/pin_protocol.rs
```diff
@@ -0,0 +1,460 @@
...
+    ) -> Result<Vec<u8>, Ctap2StatusCode>;
+
+    /// Returns the decrypted ciphertext.
+    fn decrypt(&mut self, ciphertext: &[u8]) -> Result<Vec<u8>, Ctap2StatusCode>;
+
+    /// Verifies that the signature is a valid MAC for the given message.
+    fn verify(&mut self, message: &[u8], signature: &[u8]) -> Result<(), Ctap2StatusCode>;
+}
+
+fn encrypt(
```
jmichelp: `aes256_cbc_encrypt` would be more descriptive here.

src/ctap/pin_protocol.rs
```diff
@@ -0,0 +1,460 @@
...
+        blocks.push(*array_ref!(block, 0, 16));
+    }
+    let aes_enc_key = crypto::aes256::EncryptionKey::new(key);
+    cbc_encrypt(&aes_enc_key, iv, &mut blocks);
+    let mut ciphertext = if has_iv { iv.to_vec() } else { vec![] };
+    ciphertext.extend(blocks.iter().flatten());
+    Ok(ciphertext)
+}
+
+fn decrypt(key: &[u8; 32], ciphertext: &[u8], has_iv: bool) -> Result<Vec<u8>, Ctap2StatusCode> {
```
jmichelp: Same here with `aes256_cbc_decrypt`.

src/ctap/pin_protocol.rs
```diff
@@ -0,0 +1,460 @@
...
+        1 => verify_v1(token, message, signature),
+        2 => verify_v2(token, message, signature),
+        _ => Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER),
+    }
+}
+
+pub trait SharedSecret {
+    /// Returns the encrypted plaintext.
+    fn encrypt(
+        &mut self,
```
jmichelp: As above, I don't see why the object as to be mutable in this trait.

src/ctap/pin_protocol.rs
```diff
@@ -0,0 +1,460 @@
...
+            handshake,
+            aes_key: None,
+            hmac_key: None,
+        }
+    }
+
+    fn get_aes_key(&mut self) -> &[u8; 32] {
+        let handshake = &self.handshake;
+        self.aes_key
+            .get_or_insert_with(|| hkdf_empty_salt_256::<Sha256>(handshake, b"CTAP2 AES key"))
```
jmichelp: Now I see where all these `mut` come from :)
Can't this be done in the `new` method instead?
Semantically a `get_*` is a RO accessor so having a mutable object is not really intuitive. And this would remove pretty much all the `mut` at once.

src/ctap/pin_protocol.rs
```diff
@@ -0,0 +1,460 @@
...
+    ) -> Result<Vec<u8>, Ctap2StatusCode> {
+        encrypt(rng, self.get_aes_key(), plaintext, true)
+    }
+
+    fn decrypt(&mut self, ciphertext: &[u8]) -> Result<Vec<u8>, Ctap2StatusCode> {
+        decrypt(self.get_aes_key(), ciphertext, true)
+    }
+
+    fn verify(&mut self, message: &[u8], signature: &[u8]) -> Result<(), Ctap2StatusCode> {
+        verify_v2(self.get_hmac_key(), message, signature)
```
jmichelp: Why this calling indirection on both protocols instead of directly putting the code from the corresponding `verify_*` methods instead?
For encryption/decryption this makes sense from a factorisation point of view but that doesn't hold for `verify`

src/ctap/pin_protocol.rs
kaczmarczyck: We could remove the options around `aes_key` and `hmac_key` and initialize both right away. Downside is that we are slower in case we don't need both secrets, i.e. if there is an error before we even need the second crypto operation.
The lazy initialization is an optimization, up to you if we actually want it or not.

src/ctap/pin_protocol.rs
kaczmarczyck: Because `verify_pin_uv_auth_token` reuses them.

src/ctap/pin_protocol.rs
kaczmarczyck: Renamed.

src/ctap/pin_protocol.rs
kaczmarczyck: Renamed.

src/ctap/client_pin.rs
kaczmarczyck: This values are not customizable and tests break if we change them because the crypto does not match anymore. I updated the documentation to be more precise about internal and external constraints.

src/ctap/pin_protocol.rs
jmichelp: Right, I missed that part.

src/ctap/pin_protocol.rs
jmichelp: Another option would be to always compute and not cache the secrets as objet properties.
Not sure in "real life scenarios" which one is worse than the other though.

src/ctap/pin_protocol.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+    if verify_hmac_256::<Sha256>(key, message, array_ref![signature, 0, 32]) {
+        Ok(())
+    } else {
+        Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID)
+    }
+}
+
+pub struct SharedSecretV1 {
+    common_secret: [u8; 32],
```
ia0, line 199: Shouldn't we add a lazy `crypto::aes256::EncryptionKey::new(common_secret)`? It looks like both `encrypt` and `decrypt` compute it. But maybe those 2 functions are never called in the same CTAP command.

src/ctap/pin_protocol.rs
kaczmarczyck: I'd definitely not want to re-compute the secrets, as they are used multiple times, i.e. if you encrypt and decrypt in one command. In "real life", we usually need both AES and HMAC key. So for me, it's between the implementation detail choices inside `SharedSecret`, if we want to compute lazy or not.

ia0: Note that we could also use `RefCell` ([example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d2cb848d65a719347e8c0fcc8673c8cc)) to get RO-like interface to the cost of some runtime borrow check. So it's at least 3 options:
1. Recompute as needed. Pro: RO-like, No RAM usage. Con: Large CPU usage
2. Use RefCell. Pro: RO-like. Con: Small CPU usage, Small RAM usage, Returns Ref<T> instead of &T (same usage though)
3. Use `&mut`. Pro: Returns &T. Con: RW-like, Smaller CPU usage, Smaller RAM usage.

src/ctap/pin_protocol.rs
kaczmarczyck: At least for Pin protocol V2, we could make `aes_key` the encryption key (instead of the corresponding secret) without any overhead. For V1, it doubles the storage requirements while adding minimal overhead in CPU (some shifts and XORs). But it's definitely an option, WDYT?

src/ctap/pin_protocol.rs
kaczmarczyck: I don't see the difference between 2 and 3? I thought this is about implementing `SharedSecretV2` as

```rust
pub struct SharedSecretV2 {
    aes_key: [u8; 32],
    hmac_key: [u8; 32],
}
```

instead of 

```rust
pub struct SharedSecretV2 {
    handshake: [u8; 32],
    aes_key: Option<[u8; 32]>,
    hmac_key: Option<[u8; 32]>,
}
```

?

src/ctap/pin_protocol.rs
ia0: I agree that for V1 it doubles the memory usage (about 33 bytes), but it saves a call to `crypto::aes256::EncryptionKey::new` when both `encrypt` and `decrypt` are used right? Not sure how costly this function is. If its CPU usage is less than the memory usage of caching the result (this is subjective), then I agree it's not worth it. Just asking.

src/ctap/pin_protocol.rs
ia0: No the difference is more like:

```rust
pub struct SharedSecretV2 {
    handshake: [u8; 32],
    aes_key: RefCell<Option<[u8; 32]>>,  // Uses ~4 bytes
    hmac_key: RefCell<Option<[u8; 32]>>, // Same here
}

impl SharedSecretV2 {
    fn new(handshake: [u8; 32]) -> SharedSecretV2 {
        SharedSecretV2 {
            handshake,
            aes_key: RefCell::new(None),
            hmac_key: RefCell::new(None),
        }
    }
    // Notice that:
    // - Input is &self not &mut self
    // - Output is Ref<[u8; 32]> not &[u8; 32] but this doesn't change how the result is used
    fn get_aes_key(&self) -> Ref<[u8; 32]> {
        // Does a few instructions for each borrow call.
        if self.aes_key.borrow().is_none() {
            *self.aes_key.borrow_mut() =
                hkdf_empty_salt_256::<Sha256>(&self.handshake, b"CTAP2 AES key"));
        }
        Ref::map(self.aes_key.borrow(), |x| x.as_ref().unwrap())
    }
    // Similarly for get_hmac_key
}
```

instead of the current code.

ia0: We can actually save a call to `borrow` in the reuse case:

```rust
let mut cache = self.aes_key.borrow();
if cache.is_none() {
    drop(cache);  // Important otherwise panics at runtime.
    *self.aes_key.borrow_mut() = ...;
    cache = self.aes_key.borrow();
}
Ref::map(cache, ...)
```

kaczmarczyck: The current version is hopefully on a sweet spot we can agree on:
- no `mut`
- no recomputation of crypto, except for a rare conversion from encryption key to decryption key, if we do multiple decrypts in one command (very rare)
- very little RAM overhead (32B for V1, no overhead for V2, making them the same size)
- we only compute unused secrets if we fail early, which shouldn't affect usability

src/ctap/pin_protocol.rs
kaczmarczyck: Done.

ia0 (commented):
LGTM

jmichelp (approved):

PR #293, opened 2021-03-12
PIN protocol V2 in ClientPin

kaczmarczyck:
Successor for #292. This PR adds PIN protocol V2 to `client_pin.rs`, with the following main contributions:

- a new `enum` type for `PinUvAuthProtocol`, usage will be extended in a later PR
- updating the HMAC secret extension to work with PIN protocol V2
- improved testability in `SharedSecret`
- passing the protocol version into ClientPin subcommands, and making them work with V1 and V2
- a complete rework of ClientPin tests, to make them generalize more easily to both versions

Especially the unit test rework makes this PR huge, so all required changes for V2 outside of the main ClientPin command will be in the next PR. The ease the review, the tests are:
- more readable
- matching the public interface
- a functional superset of the former state (adding `reset` and everything for V2)

### DIFF
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -20,6 +20,7 @@ use super::rng256::Rng256;
 
 pub const NBYTES: usize = int256::NBYTES;
 
+#[derive(Clone)]
 pub struct SecKey {
     a: NonZeroExponentP256,
 }
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -20,7 +20,6 @@ use super::rng256::Rng256;
 
 pub const NBYTES: usize = int256::NBYTES;
 
-#[derive(Clone)]
 pub struct SecKey {
     a: NonZeroExponentP256,
 }
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -142,7 +142,7 @@ impl ClientPin {
 
     /// Computes the shared secret for the given version.
     fn get_shared_secret(
-        &mut self,
+        &self,
         pin_uv_auth_protocol: PinUvAuthProtocol,
         key_agreement: CoseKey,
     ) -> Result<Box<dyn SharedSecret>, Ctap2StatusCode> {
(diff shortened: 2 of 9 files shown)


src/ctap/client_pin.rs
```diff
@@ -133,12 +114,42 @@ impl ClientPin {
...
+    ) -> &mut PinProtocol {
+        match pin_uv_auth_protocol {
+            PinUvAuthProtocol::V1 => &mut self.pin_protocol_v1,
+            PinUvAuthProtocol::V2 => &mut self.pin_protocol_v2,
+        }
+    }
+
+    /// Computes the shared secret for the given version.
+    fn get_shared_secret(
+        &mut self,
```
jmichelp: `get_pin_protocol` doesn't require `mut self` so I guess that `get_shared_secret` should be also fine without a `mut self`.

libraries/crypto/src/ecdh.rs
```diff
@@ -20,6 +20,7 @@ use super::rng256::Rng256;
 
 pub const NBYTES: usize = int256::NBYTES;
 
+#[derive(Clone)]
```
jmichelp: I wonder if this is not going to be problematic when we'll move to hardware crypto.
On hardware the secret key is not directly accessible. So the fact that you need to clone it now might become an issue later.

src/ctap/data_formats.rs
```diff
@@ -367,16 +368,21 @@ impl TryFrom<cbor::Value> for GetAssertionHmacSecretInput {
...
+                4 => pin_uv_auth_protocol,
             } = extract_map(cbor_value)?;
         }
 
         let key_agreement = CoseKey::try_from(ok_or_missing(key_agreement)?)?;
         let salt_enc = extract_byte_string(ok_or_missing(salt_enc)?)?;
         let salt_auth = extract_byte_string(ok_or_missing(salt_auth)?)?;
+        let pin_uv_auth_protocol = pin_uv_auth_protocol
+            .map(PinUvAuthProtocol::try_from)
+            .unwrap_or(Ok(PinUvAuthProtocol::V1))?;
```
ia0: Those 2 lines could be `.map_or(Ok(V1), try_from)?`.

ia0 (commented):
Only went quickly through it, LGTM.

libraries/crypto/src/ecdh.rs
kaczmarczyck: It is actually only convenience for tests. We don't do conditional derive anymore, that's way it's not obvious from the code. Used here:

```rust

    #[cfg(test)]
    pub fn new_test(
        key_agreement_key: crypto::ecdh::SecKey,
        pin_uv_auth_token: [u8; PIN_TOKEN_LENGTH],
    ) -> ClientPin {
        ClientPin {
            pin_protocol_v1: PinProtocol::new_test(key_agreement_key.clone(), pin_uv_auth_token),
            pin_protocol_v2: PinProtocol::new_test(key_agreement_key, pin_uv_auth_token),
            ...
```


I could
- pass in 2 different keys
- ignore since tests run in software

? It doesn't matter anywhere if these are identical or not. Maybe for some later tests, I'll want different secret so I can check that you can't cross-combine protocols. So I'll just go with the former and remove the `Clone`?

kaczmarczyck: Clone removed, it's also a bit stricter now, and prepared for more tests.

src/ctap/client_pin.rs
kaczmarczyck: Good catch, though I wonder why I don't get a warning?

src/ctap/data_formats.rs
kaczmarczyck: Changed here and in another unrelated occasion.

ia0 (approved):

src/ctap/mod.rs
```diff
@@ -2320,7 +2320,8 @@ mod test {
         let mut rng = ThreadRng256 {};
         let key_agreement_key = crypto::ecdh::SecKey::gensk(&mut rng);
         let pin_uv_auth_token = [0x88; 32];
-        let client_pin = ClientPin::new_test(key_agreement_key, pin_uv_auth_token);
+        let client_pin =
+            ClientPin::new_test(key_agreement_key, pin_uv_auth_token, PinUvAuthProtocol::V1);
```
jmichelp, line 2324: I don't see any test using `ClientPin::new_test(..., ..., PinUvAuthProtocol::V2)`.
Is it WAI?

kaczmarczyck: > Especially the unit test rework makes this PR huge, so all required changes for V2 outside of the main ClientPin command will be in the next PR.

Tests in `client_pin.rs` are general, everything else is unchanged as much as possible.

Issue #176, opened 2020-10-05
Enable Rustfmt's wrap_comments option


gendx:
As pointed out in https://github.com/google/OpenSK/pull/165#discussion_r499035911, we currently allow comments that exceed the maximum line length normally allowed by Rustfmt.

We could enable the [`wrap_comments` option](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#wrap_comments) in `rustfmt.toml` to make sure the comments don't exceed this maximum length, and are wrapped properly. This however requires a few things.
- The `wrap_comments` option is currently unstable (rust-lang/rustfmt#3347), so we have to decide if we want to commit to using it.
- Enabling it will require some manual fixing of a few badly formatted comments, e.g. including ASCII art.

kaczmarczyck:
As indicated, I expected this to automatically checked. No strong opinion however, where are ASCII art examples that would need refactoring?

gendx:
> As indicated, I expected this to automatically checked. No strong opinion however, where are ASCII art examples that would need refactoring?

It'd be easier to have an exhaustive list while reviewing code, but for a start `src/ctap/ctap1.rs` has some tables (not sure if it should be best described as "ASCII art" or "markdown").

kaczmarczyck:
Just wanted to know if you have an intuition about how much we have to change. But it seems like our code would be nicer with correct length comments? But I'd say it's low priority.

kaczmarczyck:
We use standard formatting and clippy options.

PR #294, opened 2021-03-13
Update the documentation to use linking by name

ia0:
See https://doc.rust-lang.org/stable/rustdoc/linking-to-items-by-name.html

### DIFF
--- a/libraries/persistent_store/src/lib.rs
+++ b/libraries/persistent_store/src/lib.rs
@@ -27,7 +27,7 @@
 //!
 //! The store is flash-efficient in the sense that it uses the storage [lifetime](Store::lifetime)
 //! efficiently. For each page, all words are written at least once between erase cycles and all
-//! erase cycles are used. However, not all written words are user content: lifetime is also
+//! erase cycles are used. However, not all written words are user content: Lifetime is also
 //! consumed with metadata and compaction.
 //!
 //! The store is extendable with other entries than key-values. It is essentially a framework
--- a/libraries/persistent_store/src/lib.rs
+++ b/libraries/persistent_store/src/lib.rs
@@ -12,191 +12,191 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO(ia0): Add links once the code is complete.
+// The documentation is easier to read from a browser:
+// - Run: cargo doc --document-private-items --features=std
+// - Open: target/doc/persistent_store/index.html
+
 //! Store abstraction for flash storage
 //!
 //! # Specification
 //!
-//! The store provides a partial function from keys to values on top of a storage
-//! interface. The store total capacity depends on the size of the storage. Store
-//! updates may be bundled in transactions. Mutable operations are atomic, including
-//! when interrupted.
+//! The [store](Store) provides a partial function from keys to values on top of a
+//! [storage](Storage) interface. The store total [capacity](Store::capacity) depends on the size of
+//! the storage. Store [updates](StoreUpdate) may be bundled in [transactions](Store::transaction).
+//! Mutable operations are atomic, including when interrupted.
 //!
-//! The store is flash-efficient in the sense that it uses the storage lifetime
-//! efficiently. For each page, all words are written at least once between erase
-//! cycles and all erase cycles are used. However, not all written words are user
-//! content: lifetime is also consumed with metadata and compaction.
+//! The store is flash-efficient in the sense that it uses the storage [lifetime](Store::lifetime)
+//! efficiently. For each page, all words are written at least once between erase cycles and all
+//! erase cycles are used. However, not all written words are user content: lifetime is also
+//! consumed with metadata and compaction.
 //!
-//! The store is extendable with other entries than key-values. It is essentially a
-//! framework providing access to the storage lifetime. The partial function is
-//! simply the most common usage and can be used to encode other usages.
+//! The store is extendable with other entries than key-values. It is essentially a framework
+//! providing access to the storage lifetime. The partial function is simply the most common usage
+//! and can be used to encode other usages.
 //!
 //! ## Definitions
 //!
-//! An _entry_ is a pair of a key and a value. A _key_ is a number between 0
-//! and 4095. A _value_ is a byte slice with a length between 0 and 1023 bytes (for
-//! large enough pages).
+//! An _entry_ is a pair of a key and a value. A _key_ is a number between 0 and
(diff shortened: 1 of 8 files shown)


libraries/persistent_store/src/lib.rs
```diff
@@ -12,299 +12,297 @@
...
+//! the storage. Store [updates](StoreUpdate) may be bundled in [transactions](Store::transaction).
+//! Mutable operations are atomic, including when interrupted.
 //!
-//! The store is flash-efficient in the sense that it uses the storage lifetime
-//! efficiently. For each page, all words are written at least once between erase
-//! cycles and all erase cycles are used. However, not all written words are user
-//! content: lifetime is also consumed with metadata and compaction.
+//! The store is flash-efficient in the sense that it uses the storage [lifetime](Store::lifetime)
+//! efficiently. For each page, all words are written at least once between erase cycles and all
+//! erase cycles are used. However, not all written words are user content: lifetime is also
```
kaczmarczyck: nit: Colon -> Comma, or "lifetime" should have a capital L.

libraries/persistent_store/src/lib.rs
```diff
@@ -12,299 +12,297 @@
...
-//! store is either updated (as if the operation succeeded) or left unchanged (as if
-//! the operation did not occur). If the store is left unchanged, lifetime may still
-//! be consumed.
+//! -   Given a set of independent updates, [`Store::transaction`] applies the sequence of updates.
+//! -   Given a threshold, [`Store::clear`] removes all entries with a key greater or equal to the
+//!     threshold.
+//! -   Given a length in words, [`Store::prepare`] makes one step of compaction unless that many
+//!     words can be written without compaction. This operation has no effect on the store but may
+//!     still mutate its storage. In particular, the store has the same capacity but a possibly
+//!     reduced lifetime.
```
kaczmarczyck, line 64: Out of curiosity: This looks like the line length limit was something 80 < x < 100 before, and is now 100? #176 suggests that we might want to unify this sometime.

libraries/persistent_store/src/lib.rs
```diff
@@ -12,299 +12,297 @@
...
-//! last page of their window, with `I` the initial window:
+//! We want to show that the next N - 1 compactions won't move the tail past the last page of their
+//! window, with I the initial window:
 //!
-//! ```text
-//! forall 1 <= i <= N - 1, t_{I + i} <= (I + i + N - 1) * Q
-//! ```
+//! |                  |            |   |                     |
+//! | ----------------:| ----------:|:-:|:------------------- |
+//! | ∀(1 ≤ i ≤ N - 1) | t\_{I + i} | ≤ | (I + i + N - 1) × Q |
```
kaczmarczyck, line 218: These vertical lines are not aligned. I wonder if we want that.

libraries/persistent_store/src/lib.rs
```diff
@@ -12,299 +12,297 @@
...
-//! page.
+//! One step of compaction advances the tail by how many words were used in the first page of the
+//! window with the last entry possibly overlapping on the next page.
 //!
-//! ```text
-//! forall j, t_{j + 1} = t_j + |h_{j + 1}| - |h_j| + 1
-//! ```
+//! |    |            |   |                                      |
+//! | --:| ----------:|:-:|:------------------------------------ |
+//! | ∀j | t\_{j + 1} | = | t\_j + \|h\_{j + 1}\| - \|h\_j\| + 1 |
```
kaczmarczyck, line 227: Same here.

kaczmarczyck (dismissed):

libraries/persistent_store/src/lib.rs
ia0: Done.

libraries/persistent_store/src/lib.rs
ia0: Yes indeed, I don't know why it was 80 before. The Rust default is 100 and this is what our `rustfmt.toml` file thus implicitly specifies. I think it was 80 because I first wrote this doc in a markdown file where the line length was 80, then copied it here without rewrapping everything.

libraries/persistent_store/src/lib.rs
ia0: Weird, for me they are. But I know I didn't do it at other places where it would make the lines very wide.

libraries/persistent_store/src/lib.rs
ia0: Same, this particular table is aligned for me.

libraries/persistent_store/src/lib.rs
```diff
@@ -12,299 +12,297 @@
...
-//! t_{I + N - 1} = t_I + |h_{I + N - 1}| - |h_I| + N - 1
-//! |t_{I + N - 1}| - |h_{I + N - 1}| = |t_I| - |h_I| // Compaction preserves capacity.
-//! |h_{I + N - 1}| - |t_I| <= h_{I + N - 1} - t_I
-//! ```
+//!
+//! |                                         |   |                                              |        |
+//! | ---------------------------------------:|:-:|:-------------------------------------------- |:------ |
+//! | t\_{I + N - 1}                          | = | t\_I + \|h\_{I + N - 1}\| - \|h\_I\| + N - 1 |        |
+//! | \|t\_{I + N - 1}\| - \|h\_{I + N - 1}\| | = | \|t\_I\| - \|h\_I\|   | Compaction preserves capacity |
+//! | \|h\_{I + N - 1}\| - \|t\_I\|           | ≤ | h\_{I + N - 1} - t\_I |                               |
```
ia0, line 277: This one is the first table which is not aligned. But people should not read those tables here, they should use rustdoc, so it shouldn't hurt.

libraries/persistent_store/src/lib.rs
kaczmarczyck: alignment

This is what I see. (Image updated after wrong upload.)

ia0: Oh I see, this is because you don't have a monospace font that supports UTF-8.

kaczmarczyck: Oh true, the `∀` is actually pretty wide.

libraries/persistent_store/src/lib.rs
kaczmarczyck: Yeah, this one was more obviously intended! LGTM

kaczmarczyck (approved):

PR #296, opened 2021-03-22
PIN token state with timeouts

kaczmarczyck:
Successor of #295 to implement all bigger changes in the ClientPin command.

The biggest change is that PIN permissions and tokens now have a timeout. `token_state.rs` encapsulates this behaviour, reusing specification terminology. This file can be extended later if we want rolling timers or support internal UV.
Other than that, `TokenState` is used and unit tested.

### DIFF
--- a/src/ctap/token_state.rs
+++ b/src/ctap/token_state.rs
@@ -1,0 +1,277 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::ctap::client_pin::PinPermission;
+use crate::ctap::status_code::Ctap2StatusCode;
+use crate::ctap::timed_permission::TimedPermission;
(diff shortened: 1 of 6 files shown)


src/ctap/token_state.rs
```diff
@@ -0,0 +1,277 @@
...
+
+/// Implements pinUvAuthToken state from section 6.5.2.1.
+///
+/// The userPresent flag is omitted as the only way to set it to true is
+/// built-in user verification. Therefore, we never cache user presence.
+///
+/// This implementation does not use a rolling timer.
+pub struct PinUvAuthTokenState {
+    // Relies on the fact that all permissions are represented by powers of two.
+    permissions_set: u8,
```
jmichelp, line 38: If this is a bitfield, I wonder if we couldn't reuse some of the boilerplate that exists in Tock for the registers

jmichelp (approved):

src/ctap/token_state.rs
kaczmarczyck: In the specification, `permissions` is an "Unsigned Integer" described as
> Bitfield of permissions. If present, MUST NOT be 0. See § 6.5.5.7 Operations to Obtain a pinUvAuthToken.

This implementation tries to hide the bitfield as an implementation detail, allowing read access only through the `PinPermission` enum. Therefore, I went with the easiest and most direct implementation. Is this the crate you talk about?
[https://crates.io/crates/tock-registers](https://crates.io/crates/tock-registers)
I'd be worried about mixing register semantics in.

jmichelp: Yes, this crate.
But that was just food for thoughts because it's made to deal with bitfields in a nice way. Nothing blocking the merge here or I wouldn't have approved the PR ;-)

kaczmarczyck: Alright, thanks!

PR #297, opened 2021-03-23
Output parameters for CTAP2.1

kaczmarczyck:
Implements the last missing output parameters for CTAP2.1.

Also fixes a test that randomly failed with a chance of 1/256. It was introduced lately in #293, so I'm fixing it before anyone notices the fail :)

### DIFF
--- a/src/ctap/response.rs
+++ b/src/ctap/response.rs
@@ -139,6 +139,8 @@ pub struct AuthenticatorGetInfoResponse {
     // Missing response fields as they are only relevant for internal UV:
     // - 0x11: preferredPlatformUvAttempts
     // - 0x12: uvModality
+    // Add them when your hardware supports any kind of user verification within
+    // the boundary of the device, e.g. fingerprint or built-in keyboard.
     pub certifications: Option<BTreeMap<String, i64>>,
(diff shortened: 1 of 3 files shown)


src/ctap/response.rs
```diff
@@ -135,7 +136,12 @@ pub struct AuthenticatorGetInfoResponse {
     pub firmware_version: Option<u64>,
     pub max_cred_blob_length: Option<u64>,
     pub max_rp_ids_for_set_min_pin_length: Option<u64>,
+    // Missing response fields as they are only relevant for internal UV:
```
jmichelp, line 139: nit: I found the "internal" here a bit confusing even though that's the term used in FIDO Registry. But I can't think about an ideal adjective. Maybe "built-in" would more reflect that UV is fully executed within the authenticator boundaries?

jmichelp (dismissed):

src/ctap/response.rs
kaczmarczyck: I added an explanation below with instructions and examples.

I already use "internal UV" in multiple places already. Generally, I expect a reader to consult the specification.

jmichelp (approved):

PR #298, opened 2021-03-29
Separate file crypto wrappers, starting with AES-CBC

kaczmarczyck:
This change wants to make it easier to switch between different crypto backends. The new file `crypto_wrapper.rs` implements the interface that CTAP needs, and uses our software implementation as the backend. The implementation is identical to what was added for `pin_protocol.rs`, and simply moved. This allows reusing it in key wrapping, make the function much more readable.

This PR does not change the actual key wrapping functionality. Changes in key wrapping can potentially invalidate all existing server-side credentials, therefore I added some extra tests to ensure backwards compatibility. The first commit of the change contains those tests against the old implementation. The second commit removes them. This is just a precautionary measure. Old credentials can still be decrypted, and vice versa.

### DIFF
--- a/src/ctap/crypto_wrapper.rs
+++ b/src/ctap/crypto_wrapper.rs
@@ -23,12 +23,12 @@ pub fn aes256_cbc_encrypt(
     rng: &mut dyn Rng256,
     aes_enc_key: &crypto::aes256::EncryptionKey,
     plaintext: &[u8],
-    has_iv: bool,
+    embeds_iv: bool,
 ) -> Result<Vec<u8>, Ctap2StatusCode> {
     if plaintext.len() % 16 != 0 {
         return Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER);
     }
-    let iv = if has_iv {
+    let iv = if embeds_iv {
         let random_bytes = rng.gen_uniform_u8x32();
         *array_ref!(random_bytes, 0, 16)
     } else {
@@ -40,7 +40,7 @@ pub fn aes256_cbc_encrypt(
         blocks.push(*array_ref!(block, 0, 16));
     }
     cbc_encrypt(aes_enc_key, iv, &mut blocks);
-    let mut ciphertext = if has_iv { iv.to_vec() } else { vec![] };
+    let mut ciphertext = if embeds_iv { iv.to_vec() } else { vec![] };
     ciphertext.extend(blocks.iter().flatten());
     Ok(ciphertext)
 }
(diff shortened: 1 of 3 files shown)


src/ctap/crypto_wrapper.rs
```diff
@@ -0,0 +1,102 @@
...
+    Ok(blocks.iter().flatten().cloned().collect::<Vec<u8>>())
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crypto::rng256::ThreadRng256;
+
+    #[test]
+    fn test_encrypt_decrypt_with_iv() {
```
jmichelp, line 84: Here we're probably lacking a test to ensure that CBC encryption is actually CBC and using the IV correctly.
The 2 tests here are only testing that with the same parameters, `f(f^-1(x))=x`

src/ctap/crypto_wrapper.rs
```diff
@@ -0,0 +1,102 @@
...
+use alloc::vec::Vec;
+use crypto::cbc::{cbc_decrypt, cbc_encrypt};
+use crypto::rng256::Rng256;
+
+/// Wraps the AES256-CBC encryption to match what we need in CTAP.
+pub fn aes256_cbc_encrypt(
+    rng: &mut dyn Rng256,
+    aes_enc_key: &crypto::aes256::EncryptionKey,
+    plaintext: &[u8],
+    has_iv: bool,
```
jmichelp: CBC always need an IV. So this boolean is actually NULL IV vs. random IV.
Why don't we create the IV in the caller and pass it as an arg? The API would be easier to understand.
It would also make the API more consistent with `aes256_cbc_decrypt`.

That being said, when we'll move to protected implementations, this abstraction will likely change.

kaczmarczyck: The output with `has_iv = false` is not equivalent to passing in a zero IV. We only prepend the IV if we don't use the implicit zeros. The caller could of course drop the unused IV from the result, the required changes are:
- For `has_iv = false` use cases, pass in a zero IV and remove it from the output.
- For `has_iv = true` use cases, generate the random IV and move it into the function.

This makes it easier to get function calls wrong IMO. I agree that the API is not great for general crypto. It's supposed to be tailored for CTAP use cases. WDYT?

jmichelp: Ok, I was missing this part indeed.
What about `embeds_iv` or something alike?

kaczmarczyck: I can rename it!
We could simplify the interface and remove the RNG from the parameter list by using an `iv: Option<[u8; 16]>`. It still means the caller has to correctly add entropy themselves, but it would make the encrypt deterministic. On the other hand, the decrypt still only needs a `bool` to know whether the IV is part of the input `Vec` or not, which makes it a bit inconsistent.

src/ctap/crypto_wrapper.rs
kaczmarczyck: I added some tests for the IV and RNG. All additions to the raw `cbc_encrypt` should now be properly tested.

Outside of that, since this is a wrapper, I would assume I don't have to repeat all core tests for AES and CBC, as they are already part of the crypto library?

src/ctap/crypto_wrapper.rs
kaczmarczyck: Renaming done. I currently prefer the `bool` over the `Option`, but let me know if you prefer the latter.

jmichelp (approved):

Issue #177, opened 2020-10-06
OpenSK panic in corpus testing


mingxguo27:
1. Track kernel panic: some "sound" inputs in the corpus generated by the gray-box fuzzer caused a kernel crash when tested on hardware.
2. OpenSK's main currently has 5 explicit panics (at [L57](https://github.com/google/OpenSK/blob/master/src/main.rs#L57), [L96](https://github.com/google/OpenSK/blob/master/src/main.rs#L96), [L134](https://github.com/google/OpenSK/blob/master/src/main.rs#L134), [L209](https://github.com/google/OpenSK/blob/master/src/main.rs#L209) and [L353/355](https://github.com/google/OpenSK/blob/master/src/main.rs#L353)), consider keeping/changing the behaviour for fuzzing sake.

mingxguo27:
How can I assign it to myself?

mingxguo27:
The input file that caused a crash can be found [here](https://github.com/mingxguo27/CTAP2-test-tool/commit/dba46bba7d9528469575bfa2a2dc12526e29e0d7) and the debug logging retrieved over RTT can be found in this [doc](https://docs.google.com/document/d/16rCB5MpAMgOtInhY4WRsugbMQzKnpHVeVzZZyjSjTzs/edit)

PR #303, opened 2021-04-08
CBOR maps use Vec instead of BTreeMap

kaczmarczyck:
We only access CBOR maps by iterating over them. They also have to be sorted, but that is asserted when reading, and as we control the inputs, we can output the elements in the correct order (we do sort before writing though, to make sure). Therefore, maps can be represented internally by `Vec<Key, Value>`.

This change intends to reduce the binary size and RAM usage of OpenSK.
- binary size: The functions using `BTreeMap` have been the largest when analyzed.
- RAM: When investigating #177, we noticed that `BTreeMap` allocates lots of memory.

This changes removes all occurances of BTreeMap. Binary size in optimized mode is reduced from 130.5 KiB to 126.9 KiB as output by:
```sh
echo 'opt-level = "z"' >> Cargo.toml
```

This PR also cleans up some test code:
- `cbor_array_vec` is only used when actually needing to take a `vec`, not for `cbor_array_vec[vec![...]]
- arguments are sorted correctly (so that removing the `sort_by` call in `cbor::writer` could be removed)

All changes made are (impactful) implementation details, but we don't change any behaviour.


ia0:
Could you provide from which number the binary size is reduced from in the description? (both for base reference and history) Ex: Binary size is reduced from XXX.X KiB to 126.9 KiB. Thanks!

jmichelp (commented):
Overall LGTM.
One question: CBOR being canonical, I guess that there's unicity on the keys in a map. Do we have a test for that and isn't an implementation based on `Vec<>` problematic here (i.e. we would need to sort then check for duplicate keys)?

kaczmarczyck:
True! Dedup and test added.

The new binary size is 127.3 KiB, and I edited above correctly state the prior size.

libraries/cbor/src/writer.rs
```diff
@@ -289,6 +290,52 @@ mod test {
...
+
+    #[test]
+    fn test_write_map_duplicates() {
+        let unique_map = cbor_map! {
+            0 => "a",
+            -1 => "c",
+            b"a" => "e",
+            "c" => "h",
+        };
+        let duplicate_map = cbor_map! {
```
jmichelp: I think map duplicate keys are invalid CBOR. Otherwise it can't be canonical :)
This should be a rejected input.
Especially here, your test case has the same values for each key but appearing twice but if values were different, there would be no way figuring out which one is correct.

kaczmarczyck: Just making sure this is the failure behaviour we want, as it changes how we treat these cases. Generally, I am happy to introduce a failure here. Before, we simply overwrote and ignored duplicates.

This is the writer, so basically, `sort_by` and `dedup_by` are always no-ops. Sorting still makes sense to me, as the API is much nicer if we don't require sorted input (and should be fast for sorted input). Expecting dedup'ed input is reasonable, and has been an implicit property when we received `BTreeMap`. Silently ignoring duplicates best-effort means we catch bugs later, but helps ignore cases where we write the same key/value pair twice.

WDYT?

kaczmarczyck: I pushed a commit to change the behaviour, anticipating that this is what we will converge to. :)

libraries/cbor/src/macros.rs
```diff
@@ -608,30 +595,26 @@ mod test {
     }
 
     #[test]
-    fn test_cbor_map_btree_empty() {
-        let a = cbor_map_btree!(BTreeMap::new());
-        let b = Value::Map(BTreeMap::new());
+    fn test_cbor_map_collection_empty() {
+        let a = cbor_map_collection!(BTreeMap::new());
```
ia0: Why not `Vec::new()`?

libraries/cbor/src/values.rs
```diff
@@ -183,6 +183,22 @@ where
     }
 }
 
+impl From<Vec<(KeyType, Value)>> for Value {
+    fn from(map: Vec<(KeyType, Value)>) -> Self {
+        Value::Map(map)
+    }
+}
+
+impl From<BTreeMap<KeyType, Value>> for Value {
```
ia0: Is this temporary until all BTreeMap occurrences are removed or are there places where we plan to keep using BTreeMap?

libraries/cbor/src/writer.rs
```diff
@@ -56,8 +56,14 @@ impl<'a> Writer<'a> {
                     }
                 }
             }
-            Value::Map(map) => {
-                self.start_item(5, map.len() as u64);
+            Value::Map(mut map) => {
+                map.sort_by(|a, b| a.0.cmp(&b.0));
+                let map_len = map.len();
+                map.dedup_by(|a, b| a.0.eq(&b.0));
```
ia0, line 62: You can use `dedup_by_key(|x| &x.0)`. Same for `sort_by_key`.

libraries/cbor/src/writer.rs
```diff
@@ -56,8 +56,14 @@ impl<'a> Writer<'a> {
...
+                if map_len != map.len() {
+                    return false;
+                }
```
ia0, line 65: We are changing the previous behavior here. Before when using `cbor_map!` we would keep the last key-value pair for a given key. Now, it is unspecified. We could specify it using a stable sort and reversing the slice before deduping to preserve the previous behavior. But we probably don't care, so just mentioning. Another difference is that before we did not return false for duplicates. It was silently ignored. But there seems to be another discussion somewhere else about that. This is just to point out the 2 differences.

src/ctap/mod.rs
```diff
@@ -1033,26 +1032,29 @@ where
...
-        let mut options_map = BTreeMap::new();
-        options_map.insert(String::from("rk"), true);
-        options_map.insert(
-            String::from("clientPin"),
-            self.persistent_store.pin_hash()?.is_some(),
-        );
-        options_map.insert(String::from("up"), true);
-        options_map.insert(String::from("pinUvAuthToken"), true);
-        options_map.insert(String::from("largeBlobs"), true);
+        let mut options = vec![];
```
ia0, line 1035: It's more common to write `Vec::new()` in those cases, but why not already put all the elements between line 1042 and 1055?

src/ctap/response.rs
```diff
@@ -385,17 +384,17 @@ mod test {
             ResponseData::AuthenticatorGetAssertion(get_assertion_response).into();
         let expected_cbor = cbor_map_options! {
             0x01 => cbor_map! {
-                "type" => "public-key",
                 "id" => vec![0x2D, 0x2D, 0x2D, 0x2D],
+                "type" => "public-key",
```
ia0, line 388: I don't understand why the macro input is sometimes reordered. Because it's not always sorted so it doesn't seem to be that.

ia0 (commented):
LGTM (wasn't planning to review it, it's just that when giving an improvement it's good to have both old and new value to understand the impact)

libraries/cbor/src/writer.rs
kaczmarczyck: I was expecting that to work, too. But (see i.e. [here](https://github.com/rust-lang/rust/issues/34162)) there are lifetime issues, therefore the hack.

ia0: Wow interesting. Wasn't aware of this issue. That's quite problematic.

libraries/cbor/src/macros.rs
kaczmarczyck: We can do both, if we support initializing from `BTreeMap` and `Vec`. See below.

libraries/cbor/src/values.rs
kaczmarczyck: I already removed all calls using `BTreeMap`. I didn't see a reason why our library API shouldn't support initializing from `BTreeMap`. Since it's unused, it doesn't affect binary size.

libraries/cbor/src/writer.rs
kaczmarczyck: My other response is [here](https://github.com/google/OpenSK/pull/303#discussion_r610458792). I think this issue is new because from a library perspective, the input format simply didn't allow duplicates, as it was a `BTreeMap`. Therefore, I wouldn't see that as changing previous behaviour, but making a choice for some new behaviour. if that makes sense?

src/ctap/mod.rs
kaczmarczyck: We are using both `Vec::new()` and `vec![]` in our code often enough that I wouldn't want to homogenize in this PR. The reason for this initialization is that `"ep"` is the smallest key and only conditionally included. I could alternatively use `insert(0, ("ep", ...))` later.

src/ctap/response.rs
kaczmarczyck: It should always be sorted. If there is something missing, please let me know. Maybe it is just the non-intuitive way CBOR works. Its sorting rules are:

> If the major types are different, the one with the lower value in numerical order sorts earlier.
> If two keys have different lengths, the shorter one sorts earlier;
> If two keys have the same length, the one with the lower value in (byte-wise) lexical order sorts earlier.

libraries/cbor/src/values.rs
ia0: Ok I see, it's from the point of view of CBOR as a library that is not only used by OpenSK. Sounds good to me, although the library is still optimized for OpenSK use-case.

libraries/cbor/src/writer.rs
ia0: Yes it makes sense but I think it's only one part of the story. As I see it (only listing the relevant parts for this discussion):
- There is `cbor::write(value: cbor::Value, ...)` that is part of the public API. This function takes a `cbor::Value` which is an open data-type so changing from `BTreeMap` to `Vec` is an API change. As such an invariant that was enforced by the data-type is now removed. So we are free to specify the new behavior. That's your point.
- There is `cbor_map!{ k1 => v1, k2 => v2, ... }` that builds a `cbor::Value` out of a list of key-values which may contain duplicates and don't need to be sorted. The behavior was to silently use the last value for each duplicate key. It is also implicitly sorted due to the data-structure. My point is that this function changes behavior with the same API, so we are not really choosing a new behavior, but changing the existing one. I don't say we shouldn't change the behavior, I just say we should be aware that we are changing it and agree that's our choice. Ideally we should document the `cbor_map!` macro which currently doesn't have any documentation.

src/ctap/response.rs
ia0: Oh I see. Yes I didn't know that CBOR sorts strings differently than lexicographical order (which is the common sorting). Sounds good.

src/ctap/mod.rs
ia0: Sounds good for `Vec::new()` vs `vec![]`, it's just surprising.

Regarding order I first thought it was not important but apparently sorted input is important? Is there a reason other than latency? Because I don't think we should worry for sorting a few elements. Also it seems dangerous because this is a runtime requirement that can't be easily checked statically (typically with this kind of `if`-conditions).

kaczmarczyck: It is not important, but I sorted them during implementation and decided it's nice to keep. In general, inputs to `cbor_map!` etc do not have to be sorted.

ia0: I see, sounds good.

libraries/cbor/src/writer.rs
kaczmarczyck: The macros definitely need documentation. Added for maps and arrays. The literal types might also need some work later.

ia0: Oh my bad, I forgot the fact that we reject duplicate keys now. So the difference in behavior in `cbor_map!` is not observable for non-error cases. Nice!

ia0 (approved):

jmichelp (approved):

PR #306, opened 2021-04-13
Remove KeyType from CBOR

kaczmarczyck:
Refactors of our CBOR library to not use a separate type for map keys, but allows all major types as map keys.

This simplifies the implementation, as we don't split off `KeyType` from `Value`. Since our CBOR encodings need to be canonical, the map keys have to be sorted. We add `Array` and `Map` to our `Ord` implementation. Most other changes are trivial, removing all occurances of `KeyType` and `Value::KeyValue`.

There are differences between [CTAP2 CBOR](https://fidoalliance.org/specs/fido-v2.1-rd-20210309/fido-client-to-authenticator-protocol-v2.1-rd-20210309.html#message-encoding) and the [RFC](https://tools.ietf.org/html/rfc8949). But since CTAP only uses integers and strings as map keys, they practically don't need consideration. The implementation therefore may need to be tweaked if CTAP ever uses arrays or maps as map keys.

The binary size reduction in optimized mode is a small plus, down from 127.4 KiB to 127.1 KiB. Calculate with:
```sh
echo 'opt-level = "z"' >> Cargo.toml
```

### DIFF
--- a/libraries/cbor/src/values.rs
+++ b/libraries/cbor/src/values.rs
@@ -12,31 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::writer::write;
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
 use core::cmp::Ordering;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug)]
 pub enum Value {
-    KeyValue(KeyType),
-    Array(Vec<Value>),
-    Map(Vec<(KeyType, Value)>),
-    // TAG is omitted
-    Simple(SimpleValue),
-}
-
-// The specification recommends to limit the available keys.
-// Currently supported are both integer and string types.
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub enum KeyType {
     Unsigned(u64),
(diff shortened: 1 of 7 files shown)


libraries/cbor/src/values.rs
```diff
@@ -67,37 +70,22 @@ impl Value {
 
     pub fn type_label(&self) -> u8 {
         match self {
-            Value::KeyValue(key) => key.type_label(),
+            Value::Unsigned(_) => 0,
```
jmichelp, line 75: Wouldn't it make sense here to use an enum or something alike for the labels?
This would also make clearer the calls to `self.start_item(...)` in the CBOR writer.

kaczmarczyck: Good catch with the writer. It should use the existing `type_label` function. Fixed. This function should be inlined later into the enum declaration using the still unstable enum discriminant.

jmichelp: Ok then we're good as it is I think.
Replacing the numeric values for the type by an enum will improve readability in the CBOR reader and avoid potential mistakes if new types need to be added. This can be deferred until the feature gets more stable.

jmichelp (approved):

PR #321, opened 2021-06-09
Fix persistent store unit testing

### LINKED ISSUE #320: Testing the persistent store library in github workflows now fails
## Expected Behavior

Tests should pass if the code is left unchanged by a PR

## Actual Behavior

Tests fail because the toolchain is not properly set up.

1. Create a PR which doesn't affect the persistent store.

## Specifications

- Version: N/A
- Platform: N/A

jmichelp:
Fixes #320

### DIFF
--- a/.github/workflows/persistent_store_test.yml
+++ b/.github/workflows/persistent_store_test.yml
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: true
 
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


ia0:
Looks like `NoneError` was removed recently: https://github.com/rust-lang/rust/issues/85614. This fix will only postpone the issue until we update our compiler. I'll take a look how to fix this with the new `try_trait_v2` (or just work-around if not possible).

kaczmarczyck (approved):

jmichelp:
Dropping this PR (workaround) in favor of #322  (proper fix)

PR #322, opened 2021-06-09
Stop using try_trait in persistent_store

### LINKED ISSUE #320: Testing the persistent store library in github workflows now fails

ia0:
Fixes #320.

### DIFF
--- a/.github/workflows/persistent_store_test.yml
+++ b/.github/workflows/persistent_store_test.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Build Persistent store library
         uses: actions-rs/cargo@v1
         with:
-          command: test
+          command: build
           args: --manifest-path libraries/persistent_store/Cargo.toml --release --target=thumbv7em-none-eabi
 
       - name: Unit testing of Persistent store library (release mode)
--- a/.github/workflows/persistent_store_test.yml
+++ b/.github/workflows/persistent_store_test.yml
@@ -12,20 +12,33 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          target: thumbv7em-none-eabi
 
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
-          override: true
+
+      - name: Build Persistent store library
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --manifest-path libraries/persistent_store/Cargo.toml --release --target=thumbv7em-none-eabi
 
       - name: Unit testing of Persistent store library (release mode)
         uses: actions-rs/cargo@v1
         with:
+          toolchain: nightly
           command: test
-          args: --manifest-path libraries/persistent_store/Cargo.toml --release --features std
+          args: --manifest-path libraries/persistent_store/Cargo.toml --release --features=std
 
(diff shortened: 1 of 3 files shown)


.github/workflows/persistent_store_test.yml
```diff
@@ -12,20 +12,33 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - uses: actions-rs/toolchain@v1
```
jmichelp: I wasn't seeing it but now we have twice the toolchain setup.

.github/workflows/persistent_store_test.yml
```diff
@@ -12,20 +12,33 @@ jobs:
...
+        with:
+          target: thumbv7em-none-eabi
 
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
-          override: true
```
jmichelp: By default (i.e. when `override` is not set to true) the action will respect the `rust-toolchain` file. Therefore it shouldn't be necessary to specify we want nightly.
But this requires the git submodules to be checked-out (which you added on this PR)

.github/workflows/persistent_store_test.yml
```diff
@@ -12,20 +12,33 @@ jobs:
...
+      - name: Build Persistent store library
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --manifest-path libraries/persistent_store/Cargo.toml --release --target=thumbv7em-none-eabi
 
       - name: Unit testing of Persistent store library (release mode)
         uses: actions-rs/cargo@v1
         with:
+          toolchain: nightly
```
jmichelp: Is it necessary? The `rust-toolchain` file points to a specific nightly already. Do you need to override with a more recent one? Because this would have the effect that we unit-test the library with a different toolchain than the one we build against.

.github/workflows/persistent_store_test.yml
ia0: My understanding is that we install:
- the frozen compiler with the prod target
- the most recent nightly (to detect breakage early)

And we can only do this with 2 toolchain "actions".

.github/workflows/persistent_store_test.yml
ia0: I think it's better to check that the code compiles with both versions which is why I keep nightly.

.github/workflows/persistent_store_test.yml
ia0: I see 3 options:
- test only with frozen compiler
- test only with nightly
- test with both

I think they should be all equivalent (as long as the code compiles on both versions). First I wanted to do a matrix but it's not as simple because we can't name the rust-toolchain toolchain (in a maintainable way). But I can try again if we prefer to do everything (build + test) on both versions.

ia0:
Actually I just remembered why I was only using nightly, it's because opensk build already checks that the library builds. So it's enough to just run the tests.

jmichelp (approved):

jmichelp:
Don't forget to also issue a PR to the bugfix branch :)

ia0:
Actually, neither the bugfix nor the stable branch have this issue. It was introduced by #238 which was only on the develop branch.

PR #333, opened 2021-06-25
CBOR annotations for crates.io

kaczmarczyck:
This is a small fix to add some annotations to `Cargo.toml`. The default `README.md` seems not to be read by crates.io for our project.

### DIFF
--- a/libraries/cbor/Cargo.toml
+++ b/libraries/cbor/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sk-cbor"
-version = "0.1.0"
+version = "0.1.1"
 authors = [
   "Fabian Kaczmarczyck <kaczmarczyck@google.com>",
   "Guillaume Endignoux <guillaumee@google.com>",
@@ -14,5 +14,9 @@ homepage = "https://github.com/google/OpenSK"
 repository = "https://github.com/google/OpenSK"
 keywords = ["cbor", "serialization", "no_std"]
 categories = ["encoding"]
+readme = "README.md"
+
+[badges]
+maintenance = { status = "passively-maintained" }
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


libraries/cbor/Cargo.toml
```diff
@@ -14,5 +14,9 @@ homepage = "https://github.com/google/OpenSK"
 repository = "https://github.com/google/OpenSK"
 keywords = ["cbor", "serialization", "no_std"]
 categories = ["encoding"]
+readme = "README.md"
+
+[badges]
```
ia0, line 19: Badges are deprecated now: https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section. Instead we should add them in the README directly (or write some text).

jmichelp: I see that as complementary.
The badges section is not mentioned as being deprecated. Only the automatic display is.
I guess that the maintenance status from the manifest is still being used by, for example, `cargo audit`.

I can see a few badges which may be nice to put in our readme.md file though:
Number of downloads:
`[crates.io](https://crates.io/crates/sk-cbor)`

Latest published version:
`[crates.io](https://crates.io/crates/sk-cbor)`

Latest doc version:
`[docs.rs](https://docs.rs/sk-cbor)`

License:
`[License](https://crates.io/crates/sk-cbor)`

Regarding maintenance badges the only thing I could find is only a self-declarative boolean:
`[Maintenance](https://crates.io/crates/sk-cbor)`

jmichelp (dismissed):

libraries/cbor/Cargo.toml
kaczmarczyck: Good ideas, I added them.

kaczmarczyck:
After your approval, I moved the badges up so their position matches other projects (e.g. OpenSK).

jmichelp (approved):

ia0 (approved):

PR #334, opened 2021-06-25
Kernel minimization

kaczmarczyck:
This PR adds a patch to tock to remove 4 unused capsules for all boards we deploy to:

- `ble_advertising_driver::BLE`
- `ieee802154::RadioDriver`
- `temperature::TemperatureSensor`
- `nonvolatile_storage_driver::NonvolatileStorage` (was only in dev kit)

This reduces the binary size of the kernel significantly:
| board   | before | after |
|---------|--------|-------|
| dev kit | 139 kB | 98 kB |
| dongle  | 131 kB | 98 kB |
| mdk     | 126 kB | 94 kB |


jmichelp:
I think at this point, it would make more sense to create our own boards and leave the ones from Tock upstream untouched.
We already have a `boards/` directory for the Makerdiary dongle.
It's a much bigger PR but it also reduces at least one patch (for the storage).

kaczmarczyck:
So you want the all changes of `boards/nordic/` in patches to be moved / merged into one that creates new boards in `boards/nordic`, like an `opensk_nrf52840dk` etc? This would be a bigger patch refactoring. Might also mean that changes to persistent storage or USB need to update more different patches than before.
@jmichelp

kaczmarczyck:
Okay, I added our own 2 boards as copies from the dk and dongle. This PR is now pretty big, but compared to the first commit, doesn't do that much:

1. The boards `nrf52840dk` and `nrf52840_dongle` are copied and added to nordic/boards.
2. The (old) patch 03 and the (new) patch 08 are applied on top of these copied boards.
3. All other patches are kept as-is and applied to the new boards.
4. The deploy script, tests and documentation is changed to use boards `nrf52840dk_opensk` and `nrf52840_dongle_opensk`.

boards/nordic/nrf52840dk_opensk/src/main.rs
```diff
@@ -0,0 +1,396 @@
...
+        F: FnOnce(Option<&dyn kernel::Driver>) -> R,
+    {
+        match driver_num {
+            capsules::console::DRIVER_NUM => f(Some(self.console)),
+            capsules::gpio::DRIVER_NUM => f(Some(self.gpio)),
+            capsules::alarm::DRIVER_NUM => f(Some(self.alarm)),
+            capsules::led::DRIVER_NUM => f(Some(self.led)),
+            capsules::button::DRIVER_NUM => f(Some(self.button)),
+            capsules::rng::DRIVER_NUM => f(Some(self.rng)),
+            capsules::analog_comparator::DRIVER_NUM => f(Some(self.analog_comparator)),
```
jmichelp, line 158: non blocking comment: we don't use this neither. Should you need to shrink even more, you can take it away

boards/nordic/nrf52840dk_opensk/src/main.rs
```diff
@@ -0,0 +1,396 @@
...
+            .finalize(());
+
+    // Setup the console.
+    let console = components::console::ConsoleComponent::new(board_kernel, uart_mux).finalize(());
+    // Create the debugger object that handles calls to `debug!()`.
+    components::debug_writer::DebugWriterComponent::new(uart_mux).finalize(());
+
+    let rng = components::rng::RngComponent::new(board_kernel, &nrf52840::trng::TRNG).finalize(());
+
+    nrf52840::spi::SPIM0.configure(
```
jmichelp, line 318: SPI and and analog_comparator could also go away if you need to shrink even more. And this also applies to other boards (but on the dongle, SPI should already be gone).

patches/tock/01-persistent-storage.patch
```diff
@@ -1,8 +1,8 @@
-diff --git a/boards/nordic/nrf52840_dongle/src/main.rs b/boards/nordic/nrf52840_dongle/src/main.rs
```
jmichelp, line 1: This patch could be heavily simplified because only the last chunk needs to be applied on the stock kernel.
It could be pre-applied on the `boards/` directory with this PR.
But let's keep it for a future PR.

jmichelp (approved):

patches/tock/01-persistent-storage.patch
kaczmarczyck: Got it.

boards/nordic/nrf52840dk_opensk/src/main.rs
kaczmarczyck: Okay, will do when required.

PR #338, opened 2021-07-02
Storage locations to prepare upgrades

kaczmarczyck:
This patch adds the kernel changes required to access a potential second partition in the storage. All it requires is adding more `StorageLocations` to the board's `main.rs`.
I can merge this patch with the persistent storage if desired, but this should be easier to review.

### DIFF
--- a/patches/tock/07-upgrade-partitions.patch
+++ b/patches/tock/07-upgrade-partitions.patch
@@ -1,0 +1,92 @@
+diff --git a/boards/nordic/nrf52840_dongle_opensk/src/main.rs b/boards/nordic/nrf52840_dongle_opensk/src/main.rs
+index 051943867..e38898c3b 100644
+--- a/boards/nordic/nrf52840_dongle_opensk/src/main.rs
++++ b/boards/nordic/nrf52840_dongle_opensk/src/main.rs
+@@ -64,6 +64,7 @@ static mut PROCESSES: [Option<&'static dyn kernel::procs::ProcessType>; NUM_PROC
+ static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 1] = [kernel::StorageLocation {
+     address: 0xC0000,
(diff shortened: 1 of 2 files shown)


ia0 (approved):
This is a good idea to introduce those `StorageType`. We get good code factorization. The only thing that could be concerning is that this is too high level and too much tailored to OpenSK. Maybe a lower level types or bit flags like ReadOnly, ReadWrite, etc would be more generic and easier to understand without OpenSK knowledge. But I'm fine to revisit this in the future when it is needed (like merging the patch upstream).

EDIT: We actually need the application to control the type. So scratch the above proposal. If we want something on that direction then we need 2 fields: the permission and the "name"/"type" which is opaque for the kernel.

kaczmarczyck:
> The storage locations are currently readable by all processes due to lack of stable app id.

Currently, with the above limitation, the kernel wouldn't need to know about any storage types. I could simply introduce them as a concept in the application, and just have a syscall to add a storage location to the MPU. Which would revert how they are implemented right now: "Defined in kernel, read from app" became "Defined in app, written to kernel".

Not sure if we could have the proper per-app protections this way? Read the app ID from who syscalls, and assign the memory region to them on a first come first serve basis?

Or do you propose to have the storage locations in the kernel, and apps can claim them for themselves? That seems like we'd need them defined twice?

ia0:
Oh sorry, when I said application I essentially meant board. I wasn't worried about multi-application because there's no solution for that yet. I was worried about having `METADATA`, `STORE`, etc. defined in the kernel instead of the board (the kernel would just see some `usize` or whatever opaque identifier). It wouldn't change much from the current PR, just moving the enum to the board definition and only using `usize` in the kernel. That way the kernel is generic over boards (and not only OpenSK boards). But that's not really important, the current PR is alright.

jmichelp (approved):

PR #340, opened 2021-07-05
Interface to syscalls to partition location types

kaczmarczyck:
Following #338 , this implements an interface to the new types. This PR allows to check if an address is a partition or metadata address, and write the corresponding page.

### DIFF
--- a/src/embedded_flash/buffer_upgrade.rs
+++ b/src/embedded_flash/buffer_upgrade.rs
@@ -1,0 +1,74 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::helper::ModRange;
+use super::upgrade_storage::UpgradeStorage;
+use alloc::boxed::Box;
+use persistent_store::{StorageError, StorageResult};
+
+const PARTITION_START: usize = 0x20000;
+const PARTITION_LENGTH: usize = 0x40000;
+const METADATA_LENGTH: usize = 0x1000;
+
+pub struct BufferUpgradeStorage {
+    /// Content of the partition storage.
+    partition: Box<[u8]>,
+
+    /// Content of the metadata storage.
+    metadata: Box<[u8]>,
+}
+
+impl BufferUpgradeStorage {
+    pub fn new() -> StorageResult<BufferUpgradeStorage> {
+        Ok(BufferUpgradeStorage {
+            partition: vec![0xff; PARTITION_LENGTH].into_boxed_slice(),
+            metadata: vec![0xff; METADATA_LENGTH].into_boxed_slice(),
+        })
+    }
+}
+
+impl UpgradeStorage for BufferUpgradeStorage {
+    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]> {
+        let partition_range = ModRange::new(PARTITION_START, self.partition.len());
+        if partition_range.contains_range(&ModRange::new(offset, length)) {
+            Ok(&self.partition[offset - PARTITION_START..][..length])
+        } else {
+            Err(StorageError::OutOfBounds)
+        }
+    }
+
+    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()> {
+        let partition_range = ModRange::new(PARTITION_START, self.partition.len());
+        if partition_range.contains_range(&ModRange::new(offset, data.len())) {
+            self.partition[offset - PARTITION_START..][..data.len()].copy_from_slice(data);
+            Ok(())
+        } else {
+            Err(StorageError::OutOfBounds)
+        }
+    }
+
+    fn read_metadata(&self) -> StorageResult<&[u8]> {
+        Ok(&self.metadata[..])
+    }
+
+    fn write_metadata(&mut self, data: &[u8]) -> StorageResult<()> {
+        if data.len() <= METADATA_LENGTH {
+            self.metadata.copy_from_slice(&[0xff; METADATA_LENGTH]);
+            self.metadata[..data.len()].copy_from_slice(data);
+            Ok(())
+        } else {
+            Err(StorageError::NotAligned)
+        }
+    }
+}
--- a/src/embedded_flash/buffer_upgrade.rs
+++ b/src/embedded_flash/buffer_upgrade.rs
@@ -17,7 +17,6 @@ use super::upgrade_storage::UpgradeStorage;
 use alloc::boxed::Box;
 use persistent_store::{StorageError, StorageResult};
 
-const PARTITION_START: usize = 0x20000;
 const PARTITION_LENGTH: usize = 0x40000;
 const METADATA_LENGTH: usize = 0x1000;
 
@@ -40,22 +39,26 @@ impl BufferUpgradeStorage {
 
 impl UpgradeStorage for BufferUpgradeStorage {
     fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]> {
-        let partition_range = ModRange::new(PARTITION_START, self.partition.len());
+        let partition_range = ModRange::new(0, self.partition.len());
         if partition_range.contains_range(&ModRange::new(offset, length)) {
-            Ok(&self.partition[offset - PARTITION_START..][..length])
+            Ok(&self.partition[offset..][..length])
         } else {
             Err(StorageError::OutOfBounds)
         }
     }
 
     fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()> {
-        let partition_range = ModRange::new(PARTITION_START, self.partition.len());
+        let partition_range = ModRange::new(0, self.partition.len());
         if partition_range.contains_range(&ModRange::new(offset, data.len())) {
-            self.partition[offset - PARTITION_START..][..data.len()].copy_from_slice(data);
+            self.partition[offset..][..data.len()].copy_from_slice(data);
             Ok(())
         } else {
             Err(StorageError::OutOfBounds)
         }
+    }
+
+    fn partition_length(&self) -> usize {
+        PARTITION_LENGTH
     }
 
     fn read_metadata(&self) -> StorageResult<&[u8]> {
@@ -68,7 +71,7 @@ impl UpgradeStorage for BufferUpgradeStorage {
             self.metadata[..data.len()].copy_from_slice(data);
             Ok(())
         } else {
-            Err(StorageError::NotAligned)
+            Err(StorageError::OutOfBounds)
         }
     }
 }
--- a/src/embedded_flash/helper.rs
+++ b/src/embedded_flash/helper.rs
@@ -22,10 +22,9 @@ use persistent_store::{StorageError, StorageResult};
 ///
 /// The returned slice contains the interval `[start, start+length)`.
 ///
-/// # Preconditions
+/// # Errors
 ///
-/// - The passed in slices must not overlap.
-/// - The requested slice must fit entirely within a single one of the slices.
+/// Returns [`StorageError::OutOfBounds`] if the range is not within exactly one slice.
 pub fn find_slice<'a>(
     slices: &'a [&'a [u8]],
     mut start: usize,
@@ -54,7 +53,7 @@ pub fn is_aligned(block_size: usize, address: usize) -> bool {
 
 /// A range implementation using start and length.
 ///
-/// The range is treated as the interval `[start, start + length[`.
+/// The range is treated as the interval `[start, start + length)`.
 /// All objects with length of 0, regardless of the start value, are considered empty.
 pub struct ModRange {
     start: usize,
@@ -114,7 +113,7 @@ impl ModRange {
         new_length.map(|l| ModRange::new(self.start, l))
     }
 
-    /// Returns whether the given address is inside the range.
+    /// Returns whether the given value is inside the range.
     pub fn contains(&self, x: usize) -> bool {
         // We want to check the 2 following inequalities:
         // (1) `start <= x`
@@ -124,7 +123,9 @@ impl ModRange {
         self.start <= x && x - self.start < self.length
     }
 
-    /// Returns whether the given range is fully contained, i.e. their cut equals the parameter.
+    /// Returns whether the given range is fully contained.
+    ///
+    /// Mathematically, we calculate whether: `self ∩ range = range`.
     pub fn contains_range(&self, range: &ModRange) -> bool {
         if range.is_empty() {
             return true;
@@ -134,6 +135,9 @@ impl ModRange {
     }
 
     /// Returns an iterator for all contained numbers that are divisible by the modulus.
+    ///
+    /// Be aware that `usize::MAX` is a special case for simplicity. It is never in the returned
+    /// iterator, even when divisible by `modulus` and contained in the range.
     pub fn aligned_iter(&self, modulus: usize) -> impl Iterator<Item = usize> {
         let remainder = self.start % modulus;
         // Saturating first and limit in case they are usize::MAX.
--- a/src/embedded_flash/helper.rs
+++ b/src/embedded_flash/helper.rs
@@ -118,19 +118,13 @@ impl ModRange {
     }
 
     /// Returns an iterator for all contained numbers that are divisible by the modulus.
-    ///
-    /// Be aware that `usize::MAX` is a special case for simplicity. It is never in the returned
-    /// iterator, even when divisible by `modulus` and contained in the range.
     pub fn aligned_iter(&self, modulus: usize) -> impl Iterator<Item = usize> {
-        let remainder = self.start % modulus;
-        // Saturating first and limit in case they are usize::MAX.
-        let first = if remainder == 0 {
-            self.start
-        } else {
-            (self.start - remainder).saturating_add(modulus)
-        };
-        let limit = self.start.saturating_add(self.length);
-        (first..limit).step_by(modulus)
+        (self.start..=usize::MAX)
+            .take(self.length)
+            // Skip the minimum number of elements to align.
+            .skip((modulus - self.start % modulus) % modulus)
+            // Only return aligned elements.
+            .step_by(modulus)
     }
 }
 
--- a/src/embedded_flash/helper.rs
+++ b/src/embedded_flash/helper.rs
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// For compiling with std outside of tests.
+#![cfg_attr(feature = "std", allow(dead_code))]
+
+use core::iter::Iterator;
 use persistent_store::{StorageError, StorageResult};
 
 /// Reads a slice from a list of slices.
@@ -49,6 +53,9 @@ pub fn is_aligned(block_size: usize, address: usize) -> bool {
 }
 
 /// A range implementation using start and length.
+///
+/// The range is treated as the interval `[start, start + length[`.
+/// All objects with length of 0, regardless of the start value, are considered empty.
 pub struct ModRange {
     start: usize,
     length: usize,
@@ -57,25 +64,54 @@ pub struct ModRange {
 impl ModRange {
     /// Returns a new range of given start and length.
     ///
-    /// If the largest contained address would overflow the address space, return an empty range.
-    pub fn new(start: usize, length: usize) -> Self {
-        if start.checked_add(length - 1).is_none() {
-            return Self::new_empty();
+    /// The stored length can be shorter than what is passed in, so that the range does not include
+    /// values bigger than `usize::MAX`.
+    pub fn new(start: usize, mut length: usize) -> ModRange {
+        if length > 0 {
+            length = start.saturating_add(length - 1) - start + 1;
         }
         ModRange { start, length }
     }
 
     /// Create a new empty range.
-    pub fn new_empty() -> Self {
-        ModRange {
-            start: 0,
-            length: 0,
-        }
+    pub fn new_empty() -> ModRange {
+        ModRange::new(0, 0)
+    }
+
+    /// Returns the start of the range.
+    pub fn start(&self) -> usize {
+        self.start
+    }
+
+    /// Returns the length of the range.
+    pub fn length(&self) -> usize {
+        self.length
     }
 
     /// Returns whether this range contains any addresses.
     pub fn is_empty(&self) -> bool {
         self.length == 0
+    }
+
+    /// Returns the disjoint union with the other range, if is consecutive.
+    ///
+    /// Appending empty ranges is not possible.
+    /// Appending to the empty range returns the other range.
+    pub fn append(&self, other: ModRange) -> Option<ModRange> {
+        if self.is_empty() {
+            return Some(other);
+        }
+        if other.is_empty() {
+            return None;
+        }
+        if self.start >= other.start {
+            return None;
+        }
+        if self.length != other.start - self.start {
+            return None;
+        }
+        let new_length = self.length.checked_add(other.length);
+        new_length.map(|l| ModRange::new(self.start, l))
     }
 
     /// Returns whether the given address is inside the range.
@@ -86,6 +122,28 @@ impl ModRange {
         // However, the second one may overflow written as is. Using (1), we rewrite to:
         // (3) `x - start <= length`
         self.start <= x && x - self.start < self.length
+    }
+
+    /// Returns whether the given range is fully contained, i.e. their cut equals the parameter.
+    pub fn contains_range(&self, range: &ModRange) -> bool {
+        if range.is_empty() {
+            return true;
+        }
+        let end = range.start + (range.length - 1);
+        self.contains(range.start) && self.contains(end)
+    }
+
+    /// Returns an iterator for all contained numbers that are divisible by the modulus.
+    pub fn aligned_iter(&self, modulus: usize) -> impl Iterator<Item = usize> {
+        let remainder = self.start % modulus;
+        // Saturating first and limit in case they are usize::MAX.
+        let first = if remainder == 0 {
+            self.start
+        } else {
+            (self.start - remainder).saturating_add(modulus)
+        };
+        let limit = self.start.saturating_add(self.length);
+        (first..limit).step_by(modulus)
     }
 }
 
@@ -130,19 +188,49 @@ mod tests {
     }
 
     #[test]
+    fn mod_range_parameters() {
+        let range = ModRange::new(200, 100);
+        assert_eq!(range.start(), 200);
+        assert_eq!(range.length(), 100);
+        let range = ModRange::new(usize::MAX, 2);
+        assert_eq!(range.start(), usize::MAX);
+        assert_eq!(range.length(), 1);
+        let range = ModRange::new(usize::MAX - 3, 7);
+        assert_eq!(range.start(), usize::MAX - 3);
+        assert_eq!(range.length(), 4);
+        assert_eq!(ModRange::new_empty().length(), 0);
+    }
+
+    #[test]
     fn mod_range_is_empty() {
-        assert!(!ModRange::new(0x200, 0x100).is_empty());
-        assert!(ModRange::new(0x200, 0).is_empty());
+        assert!(!ModRange::new(200, 100).is_empty());
+        assert!(ModRange::new(200, 0).is_empty());
         assert!(ModRange::new_empty().is_empty());
-        assert!(ModRange::new(usize::MAX, 2).is_empty());
+        assert!(!ModRange::new(usize::MAX, 2).is_empty());
+    }
+
+    #[test]
+    fn mod_range_append() {
+        let range = ModRange::new(200, 100);
+        let new_range = range.append(ModRange::new(300, 400)).unwrap();
+        assert!(new_range.start() == 200);
+        assert!(new_range.length() == 500);
+        assert!(range.append(ModRange::new(299, 400)).is_none());
+        assert!(range.append(ModRange::new(301, 400)).is_none());
+        assert!(range.append(ModRange::new(200, 400)).is_none());
+        let empty_append = ModRange::new_empty()
+            .append(ModRange::new(200, 100))
+            .unwrap();
+        assert!(empty_append.start() == 200);
+        assert!(empty_append.length() == 100);
     }
 
     #[test]
     fn mod_range_contains() {
-        let start = 0x200;
-        let length = 0x100;
+        let start = 200;
+        let length = 100;
         let range = ModRange::new(start, length);
-        assert!(!range.contains(0x300));
+        assert!(!range.contains(300));
         for i in start..start + length {
             assert!(range.contains(i));
         }
@@ -154,6 +242,41 @@ mod tests {
         }
         assert!(!ModRange::new_empty().contains(0));
         assert!(ModRange::new(usize::MAX, 1).contains(usize::MAX));
-        assert!(!ModRange::new(usize::MAX, 2).contains(usize::MAX));
+        assert!(ModRange::new(usize::MAX, 2).contains(usize::MAX));
+    }
+
+    #[test]
+    fn mod_range_contains_range() {
+        let range = ModRange::new(200, 100);
+        assert!(!range.contains_range(&ModRange::new(199, 100)));
+        assert!(!range.contains_range(&ModRange::new(201, 100)));
+        assert!(!range.contains_range(&ModRange::new(199, 99)));
+        assert!(!range.contains_range(&ModRange::new(202, 99)));
+        assert!(!range.contains_range(&ModRange::new(200, 101)));
+        assert!(range.contains_range(&ModRange::new(200, 100)));
+        assert!(range.contains_range(&ModRange::new(200, 99)));
+        assert!(range.contains_range(&ModRange::new(201, 99)));
+        assert!(ModRange::new_empty().contains_range(&ModRange::new_empty()));
+        assert!(ModRange::new(usize::MAX, 1).contains_range(&ModRange::new(usize::MAX, 1)));
+        assert!(ModRange::new(usize::MAX, 2).contains_range(&ModRange::new(usize::MAX, 2)));
+    }
+
+    #[test]
+    fn mod_range_aligned_iter() {
+        let mut iter = ModRange::new(200, 100).aligned_iter(100);
+        assert_eq!(iter.next(), Some(200));
+        assert_eq!(iter.next(), None);
+        let mut iter = ModRange::new(200, 101).aligned_iter(100);
+        assert_eq!(iter.next(), Some(200));
+        assert_eq!(iter.next(), Some(300));
+        assert_eq!(iter.next(), None);
+        let mut iter = ModRange::new(199, 100).aligned_iter(100);
+        assert_eq!(iter.next(), Some(200));
+        assert_eq!(iter.next(), None);
+        let mut iter = ModRange::new(201, 99).aligned_iter(100);
+        assert_eq!(iter.next(), None);
+        let mut iter = ModRange::new(usize::MAX - 16, 20).aligned_iter(16);
+        assert_eq!(iter.next(), Some(0xf_fff_fff_fff_fff_ff0));
+        assert_eq!(iter.next(), None);
     }
 }
--- a/src/embedded_flash/helper.rs
+++ b/src/embedded_flash/helper.rs
@@ -62,13 +62,7 @@ pub struct ModRange {
 
 impl ModRange {
     /// Returns a new range of given start and length.
-    ///
-    /// The stored length can be shorter than what is passed in, so that the range does not include
-    /// values bigger than `usize::MAX`.
-    pub fn new(start: usize, mut length: usize) -> ModRange {
-        if length > 0 {
-            length = start.saturating_add(length - 1) - start + 1;
-        }
+    pub fn new(start: usize, length: usize) -> ModRange {
         ModRange { start, length }
     }
 
@@ -113,16 +107,6 @@ impl ModRange {
         new_length.map(|l| ModRange::new(self.start, l))
     }
 
-    /// Returns whether the given value is inside the range.
-    pub fn contains(&self, x: usize) -> bool {
-        // We want to check the 2 following inequalities:
-        // (1) `start <= x`
-        // (2) `x < start + length`
-        // However, the second one may overflow written as is. Using (1), we rewrite to:
-        // (3) `x - start <= length`
-        self.start <= x && x - self.start < self.length
-    }
-
     /// Returns whether the given range is fully contained.
     ///
     /// Mathematically, we calculate whether: `self ∩ range = range`.
@@ -130,8 +114,10 @@ impl ModRange {
         if range.is_empty() {
             return true;
         }
-        let end = range.start + (range.length - 1);
-        self.contains(range.start) && self.contains(end)
+        range.is_empty()
+            || (self.start <= range.start
+                && range.length <= self.length
+                && range.start - self.start <= self.length - range.length)
     }
 
     /// Returns an iterator for all contained numbers that are divisible by the modulus.
@@ -196,12 +182,6 @@ mod tests {
         let range = ModRange::new(200, 100);
         assert_eq!(range.start(), 200);
         assert_eq!(range.length(), 100);
-        let range = ModRange::new(usize::MAX, 2);
-        assert_eq!(range.start(), usize::MAX);
-        assert_eq!(range.length(), 1);
-        let range = ModRange::new(usize::MAX - 3, 7);
-        assert_eq!(range.start(), usize::MAX - 3);
-        assert_eq!(range.length(), 4);
         assert_eq!(ModRange::new_empty().length(), 0);
     }
 
@@ -227,26 +207,6 @@ mod tests {
             .unwrap();
         assert!(empty_append.start() == 200);
         assert!(empty_append.length() == 100);
-    }
-
-    #[test]
-    fn mod_range_contains() {
-        let start = 200;
-        let length = 100;
-        let range = ModRange::new(start, length);
-        assert!(!range.contains(300));
-        for i in start..start + length {
-            assert!(range.contains(i));
-        }
-        for i in start - length..start {
-            assert!(!range.contains(i));
-        }
-        for i in start + length..start + 2 * length {
-            assert!(!range.contains(i));
-        }
-        assert!(!ModRange::new_empty().contains(0));
-        assert!(ModRange::new(usize::MAX, 1).contains(usize::MAX));
-        assert!(ModRange::new(usize::MAX, 2).contains(usize::MAX));
     }
 
     #[test]
--- a/src/embedded_flash/helper.rs
+++ b/src/embedded_flash/helper.rs
@@ -12,10 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use core::ops::RangeInclusive;
 use persistent_store::{StorageError, StorageResult};
 
-/// Returns a slice from a list of slices that contains the interval [start, start+length).
+/// Reads a slice from a list of slices.
+///
+/// The returned slice contains the interval `[start, start+length)`.
+///
+/// # Preconditions
+///
+/// - The passed in slices must not overlap.
+/// - The requested slice must fit entirely within a single one of the slices.
 pub fn find_slice<'a>(
     slices: &'a [&'a [u8]],
     mut start: usize,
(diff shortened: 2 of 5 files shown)


src/embedded_flash/mod.rs
```diff
@@ -47,6 +54,38 @@ mod test {
         };
         Storage::new(store, options)
     }
+
+    fn is_page_aligned(x: usize) -> bool {
+        x & (PAGE_SIZE - 1) == 0
+    }
+
+    /// Mock implementation of a partition, only tests page alignment.
+    pub struct Partition {}
```
ia0: You can as well have `pub struct Partition;` and then write `Partition` instead of `Partition {}`. This is a little bit less awkward.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
     Err(StorageError::OutOfBounds)
 }
 
+struct PartitionSlice {
+    pub ptr: usize,
+    pub len: usize,
```
ia0: Why are those fields `pub`?

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
     Err(StorageError::OutOfBounds)
 }
 
+struct PartitionSlice {
```
ia0: The name is a bit confusing because this is also used for the METADATA storage location and not only the PARTITION ones. What about renaming to `StorageLocation`?

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+        // However, the second one may overflow written as is. Using (1), we rewrite to:
+        // (3) `address - ptr <= len`
+        self.ptr <= address && address - self.ptr <= self.len
+    }
+}
+
+pub struct Partition {
+    page_size: usize,
+    partition_locations: Vec<PartitionSlice>,
+    metadata_location: PartitionSlice,
```
ia0: After the `StorageLocation` renaming, those could simply be named `partitions` and `metadata` since the types would be `Vec<StorageLocation>` and `StorageLocation`.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+        self.metadata_location.contains(page_address)
+    }
+
+    pub fn rewrite_page(&self, page_ptr: usize, value: &mut [u8]) -> StorageResult<()> {
+        let value_len = value.len();
+        if !is_aligned(self.page_size, page_ptr) || !is_aligned(self.page_size, value_len) {
+            return Err(StorageError::NotAligned);
+        }
+
+        let shared_memory = syscalls::allow(DRIVER_NUMBER, allow_nr::WRITE_SLICE, value);
```
ia0: Let's move this between ERASE and WRITE? Because it's only used by write.

Also, let's have a `TODO(https://github.com/tock/tock/issues/1274): Use read-only slices`. Or even better, let's factorize with `write_slice` and `erase_page` above.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+        self.partition_locations
+            .iter()
+            .any(|storage_location| storage_location.contains(page_address))
+    }
+
+    pub fn is_page_in_metadata(&self, page_address: usize) -> bool {
+        self.metadata_location.contains(page_address)
+    }
+
+    pub fn rewrite_page(&self, page_ptr: usize, value: &mut [u8]) -> StorageResult<()> {
```
ia0: Let's take `value: &[u8]` to have a meaningful API. Tock is currently broken in that regard, but they should get read-only slices in 2.0.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+        // We want to check the 2 following inequalities:
+        // (1) `ptr <= address`
+        // (2) `address <= ptr + len`
+        // However, the second one may overflow written as is. Using (1), we rewrite to:
+        // (3) `address - ptr <= len`
+        self.ptr <= address && address - self.ptr <= self.len
+    }
+}
+
+pub struct Partition {
```
ia0: This name also is a bit confusing. This looks like `NonStoreStorageLocations` or better `UpgradeLocations`.

We could maybe even rename the enum to: STORE (or PERSISTENT_STORE), UPGRADE_PARTITION, UPGRADE_METADATA.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+            return Err(StorageError::CustomError);
+        }
+        let mut partition_locations = Vec::new();
+        let mut metadata_location = None;
+        for i in 0..memop(memop_nr::STORAGE_CNT, 0)? {
+            let storage_type = memop(memop_nr::STORAGE_TYPE, i)?;
+            match storage_type {
+                storage_type::STORE => continue,
+                storage_type::PARTITION | storage_type::METADATA => (),
+                _ => return Err(StorageError::CustomError),
```
ia0: This is not future-proof. We may decide to add new storage location types to the board that are not for the persistent store or for upgradability. Let's just do as you did for the `SyscallStorage` and `continue` unless the type is recognized.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+                storage_type::PARTITION => partition_locations.push(PartitionSlice {
+                    ptr: storage_ptr,
+                    len: storage_len,
+                }),
+                // We only expect one page of metadata.
+                storage_type::METADATA => {
+                    if metadata_location.is_none() {
+                        metadata_location = Some(PartitionSlice {
+                            ptr: storage_ptr,
+                            len: storage_len,
```
ia0: The partition slice can be created before the match to avoid code duplication. Also why not store the slices as `&[u8]` as done in the `SyscallStorage`?

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+        // (2) `address <= ptr + len`
+        // However, the second one may overflow written as is. Using (1), we rewrite to:
+        // (3) `address - ptr <= len`
+        self.ptr <= address && address - self.ptr <= self.len
+    }
+}
+
+pub struct Partition {
+    page_size: usize,
+    partition_locations: Vec<PartitionSlice>,
```
ia0: In practice do we really have more than one storage location for the other partition? For the store it's the case because it is split on 2 banks. But I don't think it's the case for upgradability.

src/embedded_flash/syscall.rs
```diff
@@ -116,14 +127,6 @@ impl SyscallStorage {
         }
         Ok(syscall)
     }
-
-    fn is_word_aligned(&self, x: usize) -> bool {
-        x & (self.word_size - 1) == 0
```
ia0, line 121: I would call `is_aligned(self.word_size, x)` here such that there is less call sites to check for the preconditions. Functions with preconditions should be very rarely directly called to simplify proofs.

I would also introduce `is_page_aligned` for the `UpgardeStorage` object. It's usually a bad idea to have a function taking 2 arguments of the same type because we don't know which one is which. In particular, for `is_aligned` the expected order could easily be the opposite of the current one because it's semantically `x % m == 0` so it's easier to read with `is_aligned(x, m)`. We just use a function for optimization purposes.

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+            .any(|storage_location| storage_location.contains(page_address))
+    }
+
+    pub fn is_page_in_metadata(&self, page_address: usize) -> bool {
+        self.metadata_location.contains(page_address)
+    }
+
+    pub fn rewrite_page(&self, page_ptr: usize, value: &mut [u8]) -> StorageResult<()> {
+        let value_len = value.len();
+        if !is_aligned(self.page_size, page_ptr) || !is_aligned(self.page_size, value_len) {
```
ia0: We don't check that we're writing to one of the valid storage locations?

src/embedded_flash/syscall.rs
```diff
@@ -210,6 +213,117 @@ fn find_slice<'a>(
...
+struct PartitionSlice {
+    pub ptr: usize,
+    pub len: usize,
+}
+
+impl PartitionSlice {
+    pub fn contains(&self, address: usize) -> bool {
+        // We want to check the 2 following inequalities:
+        // (1) `ptr <= address`
+        // (2) `address <= ptr + len`
```
ia0: This is wrong, we don't want to match the "one past" page. This should be `address < ptr + len`. Same below.

ia0 (commented):
The comments are not well organized so I'll try to give an overview here:
- The naming is not clear to me. It looks like often "partition" is used to mean "upgrade" (feature).
- There's quite some duplication and it looks like this could be factorized to improve maintenance. This might require having a function returning both `SyscallStorage` and `UpgradeStorage` and passing the first to `PersistentStore::new` instead of letting it call `new_storage`. (In that case we might want to crash for unknown storage location types.)
- It looks like having some `Upgrade` trait similar to `Storage`, implemented by `UpgradeStorage` would also make the prod vs test cleaner and force to have a more obvious API too.

ia0:
I can also suggest an alternative PR if my comments are not clear (it's probably easier to see the code).

src/embedded_flash/mod.rs
kaczmarczyck: Done

src/embedded_flash/syscall.rs
kaczmarczyck: Removed for `RangeInclusive`, see below. In case we like the struct more than the new type, since this moved to `helper.rs`, it would still need to be either `pub` or get a (trivial) `new`.

src/embedded_flash/syscall.rs
kaczmarczyck: I agree the name is not chosen well, which probably makes the code harder to understand as well. This is not supposed to be a slice, but a range. So I'd rename it to `UpgradeRange` (or something similar) or, as in the latest commit, replace it with an actual `RangeInclusive`.

Downside of this move (and reason I had my own implementation before) is that `Range`s operate on `start` and `end`, whereas I need `start` and `length`. This comes with some integer overflow problems, that the new solution has to work around. I implemented it so you can see both, please let me know what you think.

src/embedded_flash/syscall.rs
kaczmarczyck: Done

src/embedded_flash/syscall.rs
kaczmarczyck: Factorized to have a `erase_page_op` and `write_slice_op` to be called from both structs.

src/embedded_flash/syscall.rs
kaczmarczyck: With the refactoring, this is automatic as it uses your `syscalls::raw::allow` calls.

Is the reason you don't go with the non-raw `syscalls::allow` just so you can avoid the `&mut`?

src/embedded_flash/syscall.rs
kaczmarczyck: `UpgradeLocations` it is.

src/embedded_flash/syscall.rs
kaczmarczyck: Idea was that this fails if we add a new storage type and forget to adapt the implementation here. So we can't NOT notice. But I guess not having to change the code for unrelated changes is more important.

Done

src/embedded_flash/syscall.rs
kaczmarczyck: I don't want to do the `unsafe` calls to create the slices, as I don't need them.

Done for the first part of the comment.

src/embedded_flash/syscall.rs
kaczmarczyck: Yes, we do! The MPU has some restrictions that may need us to have 2 smaller storage locations.

src/embedded_flash/syscall.rs
kaczmarczyck: I reverted to have `is_page_aligned` and `is_word_aligned` everywhere, and only use `is_aligned` in their implementations.

src/embedded_flash/syscall.rs
kaczmarczyck: True, and the reason this actually passed is because the tests in this file were never run (including existing tests). I added
```
#[cfg(any(test, not(feature = "std")))]
mod helper;
```
and moved all tested helpers there.

kaczmarczyck: Also, the problematic code is removed if we keep the `RangeInclusive`. If you like the self-implemented `Range`, consider it fixed.

src/embedded_flash/syscall.rs
kaczmarczyck: The caller has to check `is_page_in_partition` and `is_page_in_metadata`. Duplicating the check inside `rewrite_page` wastes some time, but prevents accidental writes to PersistentStore locations. Trying to write to any other locations is generally prevented by the MPU.
We could also decide to make the API more specific and have `rewrite_metadata` and `rewrite_partition`. We then wouldn't need to pass in the metadata address and could let callers only specify a partition offset instead of an absolute address. (And add more features, like writing a dirty bit into the metadata when writing to partitions.) Questions is if we want to handle this here or in the app.

kaczmarczyck:
The new commit should clear up most questions, I hope. For me, there are 2 open questions:
- Keep the `RangeInclusive` or have `create_range` return the custom `UpgradeRange`?
- Change the API here to make it more specific to the upgrades or keep it more general and open?

Maybe we'd need to decide all caller functionality to choose in the latter option.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,118 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use core::ops::RangeInclusive;
+use persistent_store::{StorageError, StorageResult};
+
+/// Returns a slice from a list of slices that contains the interval [start, start+length).
```
ia0: Let's also mention that the requested range must fit within one of the slices in the list and cannot overlap.

ia0: Let's either use backticks around non-English stuff or avoid them in the title (e.g. `Reads a range from a list of slices`).

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,118 @@
...
+            break;
+        }
+        return Ok(&slice[start..][..length]);
+    }
+    Err(StorageError::OutOfBounds)
+}
+
+/// Checks whether the address is aligned with the block_size.
+///
+/// Requires block_size to be a power of two.
```
ia0: Let's use backticks for non-English (because symbols like `_` have meaning in Markdown). Same for title, but it's probably better to avoid non-English in title.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,118 @@
...
+pub fn is_aligned(block_size: usize, address: usize) -> bool {
+    address & (block_size - 1) == 0
```
ia0, line 51: We can add `debug_assert!(block_size.is_power_of_two());`

src/embedded_flash/mod.rs
```diff
@@ -30,13 +32,20 @@ mod prod {
 #[cfg(not(feature = "std"))]
 pub use self::prod::{new_storage, Storage};
 
-/// Storage definition for testing.
+/// Partition definition for production.
+#[cfg(not(feature = "std"))]
+pub use self::syscall::UpgradeLocations;
```
ia0: Let's move this to the `prod` module and `pub use` it on the same line as `new_storage` and `Storage`. The idea being to have one place with all `no std` and one with all `std` stuff. You can also remove the `pub use SyscallStorage` line 21 and inline it line 26. I don't know why it's like that, it's probably a leftover I did not clean.

src/embedded_flash/mod.rs
```diff
@@ -47,6 +56,37 @@ mod test {
...
+            }
+            Ok(())
+        }
+    }
 }
 #[cfg(feature = "std")]
 pub use self::test::{new_storage, Storage};
+
+#[cfg(feature = "std")]
+pub use self::test::UpgradeLocations;
```
ia0: Let's move this `pub use` to line 89 such that the main structure of this module is:
- a `prod` module with associated `pub use`
- a `test` module with associated `pub use`
- both module (and thus `pub use`) satisfy the same interface (thus same names are `pub use`)

I think this is cleaner.

src/embedded_flash/mod.rs
```diff
@@ -47,6 +56,37 @@ mod test {
...
     }
+
+    fn is_page_aligned(x: usize) -> bool {
+        x & (PAGE_SIZE - 1) == 0
+    }
+
+    /// Mock implementation of a partition, only tests page alignment.
+    pub struct UpgradeLocations;
+
+    impl UpgradeLocations {
```
ia0: Why is this not a trait?

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// Returns a range object.
+///
+/// If the range contains indices outside of usize, it returns an empty range.
+#[allow(clippy::reversed_empty_ranges)]
+pub fn create_range(ptr: usize, len: usize) -> RangeInclusive<usize> {
```
ia0: It looks like what you want is:

```rust
struct ModRange {
    start: usize,  // multiple of step
    length: usize,  // multiple of step
    step: usize,  // power of 2
}
impl ModRange {
    fn contains(&self, x: usize) -> bool {
        is_aligned(self.step, x) && start <= x && x - start < length
    }
}
```

src/embedded_flash/syscall.rs
ia0: > Is the reason you don't go with the non-raw `syscalls::allow` just so you can avoid the `&mut`?

Yes.

src/embedded_flash/syscall.rs
ia0: It's usually a good idea to avoid preconditions (in particular if they are not checked). In particular when we don't see the call-sites because we can't check that they are satisfied.

There is no need to have duplicate checks. If the function is safe, then there's no need to check before calling it, it would just return an error.

It sounds like having the API you propose would be better (and is much less low-level like `rewrite_page`):
```rust
trait UpgradeStorage {
    fn read_metadata(&self) -> StorageResult<&[u8]>;
    // error if page.len() != PAGE_SIZE
    fn write_metadata(&mut self, page: &[u8]) -> StorageResult<()>;
    // error if page_index >= num_partition_pages
    fn read_partition(&self, page_index: usize) -> StorageResult<&[u8]>;
    // error if page_index >= num_partition_pages || page.len() != PAGE_SIZE
    fn write_partition(&mut self, page_index: usize, page: &[u8]) -> StorageResult<()>;
    fn num_partition_pages(&self) -> usize;
}
```

src/embedded_flash/syscall.rs
```diff
@@ -59,6 +63,37 @@ fn memop(nr: u32, arg: usize) -> StorageResult<usize> {
     }
 }
 
+fn write_slice_op(ptr: usize, value: &[u8]) -> StorageResult<()> {
```
ia0: Why not simply call those `write_slice` and `erase_page`? The names shouldn't conflict because the other ones are methods.

src/embedded_flash/syscall.rs
```diff
@@ -157,83 +192,88 @@ impl Storage for SyscallStorage {
...
-    mut start: usize,
-    length: usize,
-) -> StorageResult<&'a [u8]> {
-    for slice in slices {
-        if start >= slice.len() {
-            start -= slice.len();
-            continue;
+pub struct UpgradeLocations {
+    page_size: usize,
+    partitions: Vec<RangeInclusive<usize>>,
```
ia0: It looks to me that this could simply be:
```rust
pub struct UpgradeLocations {
    page_size: usize,
    partition_pages: Vec<PageRange>,
    metadata_address: usize,
}
struct PageRange {
    address: usize,
    num_pages: usize,
}
```

ia0 (commented):
Yes, I believe the API should be higher level and specific to upgradability. This is less error-prone. In particular it should be a `trait` with the documentation in the `trait`. We could even have a real fake implementation for tests (where we actually write to RAM).

src/embedded_flash/helper.rs
kaczmarczyck: Done

src/embedded_flash/helper.rs
kaczmarczyck: Done

src/embedded_flash/helper.rs
kaczmarczyck: Done

src/embedded_flash/mod.rs
kaczmarczyck: I was probably trying a bit too hard to keep the style in the file. Should be much nicer now.

src/embedded_flash/mod.rs
kaczmarczyck: Done

src/embedded_flash/mod.rs
kaczmarczyck: I added a trait, see the general comment.

src/embedded_flash/helper.rs
kaczmarczyck: My new implementation relies much less on page borders. I added a `ModRange` (renaming possible) to have as much code as possible tested.

src/embedded_flash/syscall.rs
kaczmarczyck: I saw memop and thought it's similar in spirit. `_op` removed.

src/embedded_flash/syscall.rs
kaczmarczyck: I thought more about what I actually want to support, including our team discussion. With the new higher-level API, I made both `partition` and `metadata` one range. I think we can assume that a firmware image will be consecutive, so it shouldn't be a problem for `partition`. And I wanted to keep `metadata` extensible, so it is a range too. This simplifies the implementation a bit, too.

src/embedded_flash/syscall.rs
kaczmarczyck: I implemented something similar, see the general explanation for details.

src/embedded_flash/syscall.rs
kaczmarczyck: I went back to a custom range struct for testability.

kaczmarczyck:
I took a bit longer to think more about the high-level API. Some major differences:
- The firmware partition will be consecutive, so we should use that.
- The metadata might grow in the future.
- We should be able to write slices smaller than a page, in case the transport does not support that much data in one packet.

Here is the new trait implementation:
```rust
/// Accessors to storage locations used for upgrading from a CTAP command.
pub trait UpgradeStorage {
    /// Reads a slice of the partition, if within bounds.
    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]>;

    /// Writes the given data to the given offset address, if within bounds of the partition.
    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()>;

    /// Reads the metadata location.
    fn read_metadata(&self) -> StorageResult<&[u8]>;

    /// Writes the given data into the metadata location.
    fn write_metadata(&mut self, data: &[u8]) -> StorageResult<()>;
```

The new interface is independent of pages, and allows writing and reading across page borders. For efficiency, the implementation makes the assumption that writes are consecutive, so you erase a page on the first write to it only. But, a client will get feedback about differences in written and expected data through a hash check over their passed write data. (That also helps with random write errors.) The are still absolute addresses for partitions (not page-numbers, as you proposed) to reduce the chance of a calling client writing a firmware A to partition B or vice versa.

I tried to keep the implementation in `syscall.rs` as simple as possible, so must code that fiddles with indices moved to `helper.rs` to be properly unit tested. To stick to the persistent storage style, the other new files would need to go to a crate in `libraries`, but I kept them here for simplicity first.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+use core::iter::Iterator;
+use persistent_store::{StorageError, StorageResult};
+
+/// Reads a slice from a list of slices.
+///
+/// The returned slice contains the interval `[start, start+length)`.
+///
+/// # Preconditions
+///
+/// - The passed in slices must not overlap.
```
ia0: Why? It looks totally fine for the slices to overlap. E.g.:

```rust
let x: Vec<i32> = (0..100).collect();
find_slice(&[&x[3..10], &x[5..32]], 4, 2) // is totally valid
```

But you probably mean something else.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+/// - The requested slice must fit entirely within a single one of the slices.
```
ia0: This is not a precondition but an error.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// Requires `block_size` to be a power of two.
+pub fn is_aligned(block_size: usize, address: usize) -> bool {
+    debug_assert!(block_size.is_power_of_two());
+    address & (block_size - 1) == 0
+}
+
+/// A range implementation using start and length.
+///
+/// The range is treated as the interval `[start, start + length[`.
```
ia0: Sometimes intervals are written `[a, b)` (line 23) and sometimes `[a, b[`. I guess we should be consistent. I think English uses mostly the `[a, b)` notation.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+            return None;
+        }
+        if self.length != other.start - self.start {
+            return None;
+        }
+        let new_length = self.length.checked_add(other.length);
+        new_length.map(|l| ModRange::new(self.start, l))
+    }
+
+    /// Returns whether the given address is inside the range.
```
ia0: leftover: address -> value

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn contains(&self, x: usize) -> bool {
+        // We want to check the 2 following inequalities:
+        // (1) `start <= x`
+        // (2) `x < start + length`
+        // However, the second one may overflow written as is. Using (1), we rewrite to:
+        // (3) `x - start <= length`
+        self.start <= x && x - self.start < self.length
+    }
+
+    /// Returns whether the given range is fully contained, i.e. their cut equals the parameter.
```
ia0: I understand the first sentence, but not what is after `i.e.`.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Returns an iterator for all contained numbers that are divisible by the modulus.
+    pub fn aligned_iter(&self, modulus: usize) -> impl Iterator<Item = usize> {
+        let remainder = self.start % modulus;
+        // Saturating first and limit in case they are usize::MAX.
+        let first = if remainder == 0 {
+            self.start
+        } else {
+            (self.start - remainder).saturating_add(modulus)
+        };
+        let limit = self.start.saturating_add(self.length);
```
ia0: This looks wrong. If MAX is in the range, it will never be returned. For example the range [MAX - 3, MAX] with modulus 3 should return both MAX - 3 and MAX (assuming MAX is either 2^32 - 1 or 2^64 - 1).

You probably want `self.start.saturating_add(self.length - 1)` and `first..=last` below. You'll need to deal with empty ranges before and thus return `Box<dyn Iterator<Item = usize>>`.

src/embedded_flash/helper.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /// Returns an iterator for all contained numbers that are divisible by the modulus.
+    pub fn aligned_iter(&self, modulus: usize) -> impl Iterator<Item = usize> {
+        let remainder = self.start % modulus;
+        // Saturating first and limit in case they are usize::MAX.
+        let first = if remainder == 0 {
+            self.start
+        } else {
+            (self.start - remainder).saturating_add(modulus)
```
ia0: This looks wrong but is hidden by the bug below. If the addition saturates you don't get something congruent to 0. If the first value is outside `usize` range, then the iteration should be considered empty.

src/embedded_flash/mod.rs
```diff
@@ -12,31 +12,41 @@
...
-pub use self::prod::{new_storage, Storage};
+pub use self::prod::{new_storage, Storage, UpgradeLocations};
 
-/// Storage definition for testing.
+/// Definitions for testing.
 #[cfg(feature = "std")]
 mod test {
+    use super::buffer_upgrade::BufferUpgradeStorage;
+
+    const PAGE_SIZE: usize = 0x1000;
```
ia0: Let's put this back inside `new_storage` if it's not used? I don't see any reason it should be shared.

src/embedded_flash/upgrade_storage.rs
```diff
@@ -0,0 +1,30 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use persistent_store::StorageResult;
+
+/// Accessors to storage locations used for upgrading from a CTAP command.
+pub trait UpgradeStorage {
+    /// Reads a slice of the partition, if within bounds.
+    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]>;
```
ia0, line 26: Is the length of the partition a constant? If yes let's document it. Otherwise let's add a function to know it.

```rust
fn partition_length(&self) -> usize;
```

src/embedded_flash/upgrade_storage.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Writes the given data to the given offset address, if within bounds of the partition.
+    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()>;
+
+    /// Reads the metadata location.
+    fn read_metadata(&self) -> StorageResult<&[u8]>;
+
+    /// Writes the given data into the metadata location.
+    fn write_metadata(&mut self, data: &[u8]) -> StorageResult<()>;
```
ia0, line 50: Let's add a comment whether the data should cover the full length of the metadata (i.e. what is returned by `read`) or not.

src/embedded_flash/buffer_upgrade.rs
```diff
@@ -0,0 +1,74 @@
...
+        Ok(BufferUpgradeStorage {
+            partition: vec![0xff; PARTITION_LENGTH].into_boxed_slice(),
+            metadata: vec![0xff; METADATA_LENGTH].into_boxed_slice(),
+        })
+    }
+}
+
+impl UpgradeStorage for BufferUpgradeStorage {
+    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]> {
+        let partition_range = ModRange::new(PARTITION_START, self.partition.len());
```
ia0: I'm confused, why isn't `PARTITION_START` 0? I thought the offset argument was with respect to the beginning of the partition.

Also a whole range of the partition is not reachable because the length should be `self.partition.len() + PARTITION_START` then, which should be a hint that something's wrong.

src/embedded_flash/buffer_upgrade.rs
```diff
@@ -0,0 +1,74 @@
...
+        Ok(&self.metadata[..])
+    }
+
+    fn write_metadata(&mut self, data: &[u8]) -> StorageResult<()> {
+        if data.len() <= METADATA_LENGTH {
+            self.metadata.copy_from_slice(&[0xff; METADATA_LENGTH]);
+            self.metadata[..data.len()].copy_from_slice(data);
+            Ok(())
+        } else {
+            Err(StorageError::NotAligned)
```
ia0: Why `NotAligned` instead of `OutOfBounds`? The syscall implementation actually return `OutOfBounds`.

src/embedded_flash/syscall.rs
```diff
@@ -157,83 +192,127 @@ impl Storage for SyscallStorage {
...
-        if start >= slice.len() {
-            start -= slice.len();
-            continue;
+impl UpgradeStorage for SyscallUpgradeStorage {
+    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]> {
+        if self
+            .partition
+            .contains_range(&ModRange::new(offset, length))
+        {
+            Ok(unsafe { core::slice::from_raw_parts(offset as *mut u8, length) })
```
ia0: Ok so looking at this it really looks like the API for `read_partition` is that the offset is not an offset but an absolute position. This should be documented and possible renamed as `address`.

Or maybe this is a bug and the intent was:
```rust
if length <= partition.length && offset <= partition.length - length {
    Ok(from_raw_parts(partition.start + offset, length))
}
```

src/embedded_flash/syscall.rs
```diff
@@ -157,83 +192,127 @@ impl Storage for SyscallStorage {
...
-        if start + length > slice.len() {
-            break;
+    }
+
+    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()> {
+        if self
+            .partition
+            .contains_range(&ModRange::new(offset, data.len()))
+        {
+            // Erases all pages that have their first byte in the writte range.
```
ia0: typo: writte

src/embedded_flash/syscall.rs
ia0: Why is this `*mut u8` instead of `*const u8`? Is it some cast issue?

src/embedded_flash/syscall.rs
```diff
@@ -157,83 +192,127 @@ impl Storage for SyscallStorage {
...
+            // Since we expect calls in order, we don't want to erase half-written pages.
+            for address in ModRange::new(offset, data.len()).aligned_iter(self.page_size) {
+                erase_page(address, self.page_size)?;
+            }
+            write_slice(offset, data)
```
ia0: So this is not where we check that we only write on top of `0xff`s. Is it higher in the call stack?

Otherwise we could check `read_partition(offset, data.len())?.iter().all(|x| x == 0xff)`.

src/embedded_flash/syscall.rs
```diff
@@ -157,83 +192,127 @@ impl Storage for SyscallStorage {
...
+                storage_type::PARTITION => {
+                    locations.partition = locations
+                        .partition
+                        .append(range)
+                        .ok_or(StorageError::NotAligned)?
+                }
+                storage_type::METADATA => {
+                    locations.metadata = locations
+                        .metadata
+                        .append(range)
```
ia0, line 258: This assumes you get the locations in a deterministic and correct order from the kernel. We know this is true because we specify them in the right order. But let's mention it in the documentation.

src/embedded_flash/upgrade_storage.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use persistent_store::StorageResult;
+
+/// Accessors to storage locations used for upgrading from a CTAP command.
+pub trait UpgradeStorage {
+    /// Reads a slice of the partition, if within bounds.
+    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]>;
+
+    /// Writes the given data to the given offset address, if within bounds of the partition.
+    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()>;
```
ia0, line 35: From seeing the implementations of this trait, it looks like `offset` is not an offset but an address. We should rename it to either `address`, `ptr`, or any name that is absolute.

ia0, line 35: From seeing the implementations of this trait, it looks like `offset` is not an offset but an address. We should rename it to either `address`, `ptr`, or any name that is absolute.

ia0 (commented):
Overall it looks good. The trait is helping a lot to understand how this will be used.

> To stick to the persistent storage style, the other new files would need to go to a crate in libraries

I don't think this is needed. The upgrade mechanism is specific to OpenSK and it probably won't get fuzzed. So there's no benefit in moving it to its own crate.

src/embedded_flash/helper.rs
kaczmarczyck: Maybe I can just remove that comment, but the documentation doesn't precisely cover the `OutOfBounds`. Just wondered when looking at your implementation: What happens if our storage actually had 2 slices and we try to write to overlapping memory:

[0x0000..0x1000) [0x1000..0x2000)
                [0x0F00..0x1100)

src/embedded_flash/helper.rs
kaczmarczyck: It's supposed say that we want to know whether
> self ∩ range = range
I'll move it out of the title and just use the equation. We use non-ASCII somewhere else already, right?

src/embedded_flash/helper.rs
kaczmarczyck: Okay, I should have documented that I ignore MAX, motivated by application, i.e. no page will start at MAX. It should be okay to add that in the documentation that MAX is not returned, instead of changing the return type?

src/embedded_flash/helper.rs
kaczmarczyck: This should be clarified by my above comment? I basically don't care about `usize::MAX`. And if the addition saturation, the returned iterator is empty, so it can't return anything that is not congruent to 0.

src/embedded_flash/syscall.rs
kaczmarczyck: We do check higher up in the call stack by reading back the slice and checking if it matches our expectation of what was writing. But adding the check here allows to forward-fix and special case the first page to erase and re-write. In case something case wrong between the erase and rewrite of old data, it is kind of unexpected behaviour from an outside look, but it might make sense if we expect the caller to do something like:

- Try to greedy write slice by slice.
- If a slice write fails, retry the same slice.
- If the retry also fails, restart the whole process.

So how about I add the `all(|x| x == 0xff)`, and if it fails,
- read the first page
- erase the first page
- write the saved page until `offset`

?

src/embedded_flash/helper.rs
ia0: > the documentation doesn't precisely cover the OutOfBounds

I think `OutOfBounds` is correctly documented with:

```
# Errors

Returns `OutOfBounds` if the range is not within exactly one slice.
```

Because either it's outside all of them (usual meaning of out of bounds) or it's overlapping multiple slices (reusing the error name and factorizing the code, because users probably don't care, they don't expect errors).

> What happens if our storage actually had 2 slices and we try to write to overlapping memory

I'm not sure I see the link with this function. This function only reads from a slice of slices. There is no write.

src/embedded_flash/helper.rs
ia0: Ah ok, I didn't know we could say "cut" for "intersection". Makes sense. Yeah the mathematical notation is easier to understand.

> We use non-ASCII somewhere else already, right

I guess. As long as it's in comments/documentation I guess it's fine.

src/embedded_flash/syscall.rs
ia0: I'm not sure it's worth the complexity. I think for embedded it's ok to say: first failure -> fail the process -> let the user retry or  debug.

If we check higher-up that we read what we wrote, then all good.

src/embedded_flash/helper.rs
ia0: Actually looking at this, it looks quite complicated for what it does. I'm not sure it's worth dealing with MAX that much because we will never go close to it (embedded devices won't get 4G of flash before they get to 64 bits).

It looks like only doing the arithmetic in a way to avoid overflow should be enough:
- `new` would not truncate
- `append` would only saturate the length
- EDIT: removed iteration formula, it still needs to be done manually to avoid internal for loops.

What do you think?

src/embedded_flash/helper.rs
ia0: Yes, I see why it's not a problem if MAX is not a possible value.

src/embedded_flash/helper.rs
kaczmarczyck: Truncation in `new` allows the other functions to call the standard operators, like 
```
let end = range.start + (range.length - 1);
```

in `contains_range`. Saturation and checking in `append` isn't much of a difference to me, just fails a bit earlier if you have a typo. I tried to build it robust enough to handle external input, expecting offsets to be passed through CTAP directly from the client.

ia0: `contains_range` can be written more simply without any assumptions (and then `contains` removed because we don't need it):

```rust
self.start <= other.start && other.length <= self.length &&
    other.start - self.start <= self.length - other.length
```

Yes, I agree we should never trust CTAP input. But I don't think too many checks/invariants are helping, because they complexify the code all the time, for a debugging that might never happen and wouldn't be that much more complicated without (printf debugging). I think most important is that the code is kept simple (thus readable and probably also extensible, because there is less code that is not obvious whether it is actually needed or just here for debugging). Debugging asserts can be added because they are easy to filter out. If extra code for robustness against the language is needed (to prevent against bugs in the compiler or the underlying hardware), then yes we have to add that but only if it's worth it.

src/embedded_flash/helper.rs
kaczmarczyck: I was wondering how this eventuality is tested. But it's unrelated to the PR, and I changed the error documentation.

src/embedded_flash/helper.rs
kaczmarczyck: Done

src/embedded_flash/helper.rs
kaczmarczyck: Okay, done!

src/embedded_flash/helper.rs
kaczmarczyck: Done

src/embedded_flash/helper.rs
kaczmarczyck: Moved

src/embedded_flash/helper.rs
kaczmarczyck: Should be documented!

src/embedded_flash/mod.rs
kaczmarczyck: True, I forgot. Done.

src/embedded_flash/upgrade_storage.rs
kaczmarczyck: I added `partition_length`, and no `partition_offset` or anything like that, since the indexing is now actually offset by the partition start.

src/embedded_flash/upgrade_storage.rs
kaczmarczyck: Done

src/embedded_flash/buffer_upgrade.rs
kaczmarczyck: `PARTITION_START` is removed now.

src/embedded_flash/buffer_upgrade.rs
kaczmarczyck: Not sure how that happened, but changed.

src/embedded_flash/syscall.rs
kaczmarczyck: Changed to `const`, just a copy&paste leftover.

src/embedded_flash/syscall.rs
kaczmarczyck: Done

src/embedded_flash/syscall.rs
kaczmarczyck: Okay, then I'll keep this write simple for now and add features if we notice usability issues or change requirements.

src/embedded_flash/syscall.rs
kaczmarczyck: Documented

src/embedded_flash/upgrade_storage.rs
kaczmarczyck: It's an `offset` now!

kaczmarczyck: Duplicate comment, closed.

src/embedded_flash/helper.rs
kaczmarczyck: I added an `is_empty()` check because empty ranges should always be contained in anything. Otherwise, I agree to the code complexity! `contains` is removed, and therefore `new` can be simplified.

src/embedded_flash/helper.rs
```diff
@@ -114,26 +107,23 @@ impl ModRange {
...
+    /// Returns whether the given range is fully contained.
+    ///
+    /// Mathematically, we calculate whether: `self ∩ range = range`.
     pub fn contains_range(&self, range: &ModRange) -> bool {
         if range.is_empty() {
             return true;
         }
-        let end = range.start + (range.length - 1);
-        self.contains(range.start) && self.contains(end)
+        range.is_empty()
```
ia0, line 114: Leftover: You probably forgot to remove the `if` above.

src/embedded_flash/helper.rs
```diff
@@ -114,26 +107,23 @@ impl ModRange {
...
+            || (self.start <= range.start
+                && range.length <= self.length
+                && range.start - self.start <= self.length - range.length)
     }
 
     /// Returns an iterator for all contained numbers that are divisible by the modulus.
+    ///
+    /// Be aware that `usize::MAX` is a special case for simplicity. It is never in the returned
+    /// iterator, even when divisible by `modulus` and contained in the range.
```
ia0: I was curious to see what it would cost to get an implementation that supports MAX. It's not that bad.

```rust
let skip = (modulus - self.start % modulus) % modulus;
if self.start > MAX - skip || self.length < skip {
    (0..=0).take(0).step_by(1) // needs the same type as below
} else {
    (self.start + skip..=MAX).take(self.length - skip).step_by(modulus)
}
```

src/embedded_flash/syscall.rs
```diff
@@ -212,6 +212,9 @@ pub struct SyscallUpgradeStorage {
 impl SyscallUpgradeStorage {
     /// Provides access to the other upgrade partition and metadata if available.
     ///
+    /// The implementation assumes that storage locations returned by the kernel through
+    /// `memop_nr::STORAGE_*` calls are deterministic and in order.
```
ia0: nit: Actually I realize now that "in order" implies "deterministic", so we could change into "in address space order" or something like that and drop "deterministic".

ia0 (dismissed):
All good on my side to submit as is.

src/embedded_flash/syscall.rs
kaczmarczyck: Done

src/embedded_flash/helper.rs
kaczmarczyck: Oops

src/embedded_flash/helper.rs
kaczmarczyck: Not bad, but I keep the simpler implementation I think!

ia0: I actually have found the most readable and simple version:

```rust
// Iterate over all elements.
(self.start..=MAX)
    .take(self.length)
    // Skip the minimum number of elements to align.
    .skip((modulus - self.start % modulus) % modulus)
    // Only return aligned elements.
    .step_by(modulus)
```

ia0 (dismissed):

ia0 (approved):

PR #354, opened 2021-08-02
Update store latency numbers

ia0:
The main difference comes from c03605aa0ccecaeef339b05a85adcac77dd40465 (#311) where we started using -Oz instead of -O. A similar difference in latency is probably expected in crypto and cbor too.

### DIFF
--- a/examples/store_latency.rs
+++ b/examples/store_latency.rs
@@ -131,8 +131,8 @@ fn main() {
     //
     // | Pages | Overwrite | Length    | Boot     | Compaction | Insert  | Remove |
     // | ----- | --------- | --------- | -------  | ---------- | ------  | ------ |
-    // | 3     | no        | 50 words  | 2.0 ms   | 132.8 ms   | 4.3 ms  | 1.2 ms |
-    // | 20    | no        | 50 words  | 7.8 ms   | 135.7 ms   | 9.9 ms  | 4.0 ms |
-    // | 3     | yes       | 1 word    | 19.6 ms  | 90.8 ms    | 4.7 ms  | 2.3 ms |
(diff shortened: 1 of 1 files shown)


jmichelp:
Would it be worth tracking these as part of Github workflows by having the values in a machine-friendly format and compare that a PR doesn't result in latency changes by +/-5%?

ia0:
Yes, but how can we run on a board? (Also we would need to consider throttling the tests if we don't want the board to die too fast, depending on the current tests throughput.)

I'll do another PR later so that we can simply copy/paste the results from RTT output. Right now the format is different.

jmichelp:
Right... I forgot these numbers had to pulled out from a real device and can't be emulated :(

ia0:
That said, we could also add benchmarks that we run in `std` environment on host machine. It should still be a good estimate (for the variations, not the absolute value). That might be useful and Rust has easy support for that with `cargo bench`. (We would still need to find good solutions for the github action to not be flaky since the numbers may easily vary depending on the load of the container running the bench.)

jmichelp:
If we add such workflow it should only report by adding comment in the PR with the variations but never block a merge.
For flakiness, if it runs on emulation, running it 5 times and averaging should be good.

kaczmarczyck:
Orthogonal to the discussion on moving this to a workflow, this PR in itself is an improvement in documentation, so I'll approve. I didn't notice any practical latency problems so far.

kaczmarczyck (approved):

ia0:
Yes, whether we want host benchmarks is indeed orthogonal. We can address it later. And indeed, the latency change is probably not human-noticeable. I just wanted to update the numbers because I'm making the flash syscall API async and wanted to see the latency impact.

PR #358, opened 2021-08-03
ECDSA signatures and public keys in CTAP

kaczmarczyck:
Adds access to ECDSA `Signature`s and `PubKey`s. Adds `CoseSignature` to show usage of the crypto changes, and will be used itself for upgradability later.

The `Debug` and `PartialEq` implementations can be derived instead with compiler version 1.47.

Also some constants and imports are more consistent.

### DIFF
--- a/libraries/crypto/src/ecdsa.rs
+++ b/libraries/crypto/src/ecdsa.rs
@@ -264,7 +264,7 @@ impl PubKey {
     /// Verifies if the data's hash matches its signature.
     ///
     /// This function is not a constant time implementation, and does not resist side channel
-    /// attacks. Only use if all data invovled is public knowledge.
+    /// attacks. Only use if all data involved is public knowledge.
     pub fn verify_hash_vartime(&self, hash: &[u8; NBYTES], sign: &Signature) -> bool {
         let m = ExponentP256::modn(Int256::from_bin(hash));
 
--- a/libraries/crypto/src/ecdsa.rs
+++ b/libraries/crypto/src/ecdsa.rs
@@ -21,11 +21,9 @@ use super::rng256::Rng256;
 use super::{Hash256, HashBlockSize64Bytes};
 use alloc::vec;
 use alloc::vec::Vec;
-#[cfg(test)]
-use arrayref::array_mut_ref;
 #[cfg(feature = "std")]
-use arrayref::array_ref;
-use arrayref::mut_array_refs;
+use arrayref::array_mut_ref;
+use arrayref::{array_ref, mut_array_refs};
 use core::marker::PhantomData;
 
 pub const NBYTES: usize = int256::NBYTES;
@@ -166,6 +164,8 @@ impl SecKey {
 }
 
 impl Signature {
+    pub const BYTES_LENGTH: usize = 2 * int256::NBYTES;
+
     pub fn to_asn1_der(&self) -> Vec<u8> {
         const DER_INTEGER_TYPE: u8 = 0x02;
         const DER_DEF_LENGTH_SEQUENCE: u8 = 0x30;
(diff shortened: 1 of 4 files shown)


libraries/crypto/src/ecdsa.rs
```diff
@@ -252,12 +257,8 @@ impl PubKey {
         self.p.gety().to_int().to_bin(y);
     }
 
-    #[cfg(feature = "std")]
```
jmichelp, line 255: overall comment regarding all methods which previously were std only: these functions were not previously meant to be run in the firmware and were not part of the "best-effort contant time".
if they are now used in a critical path, they may require to be rewritten.

src/ctap/data_formats.rs
```diff
@@ -817,6 +820,87 @@ impl TryFrom<CoseKey> for ecdh::PubKey {
...
+            .debug_struct("CoseSignature")
+            .field("algorithm", &self.algorithm)
+            .field("bytes", &self.bytes.to_vec())
+            .finish()
+    }
+}
+
+impl PartialEq for CoseSignature {
+    fn eq(&self, other: &CoseSignature) -> bool {
+        self.algorithm == other.algorithm && self.bytes[..] == other.bytes[..]
```
jmichelp, line 864: Do we need constant timing for sig verification?

src/ctap/data_formats.rs
```diff
@@ -817,6 +820,87 @@ impl TryFrom<CoseKey> for ecdh::PubKey {
...
+        destructure_cbor_map! {
+            let {
+                "alg" => algorithm,
+                "signature" => bytes,
+            } = extract_map(cbor_value)?;
+        }
+
+        let algorithm = SignatureAlgorithm::try_from(ok_or_missing(algorithm)?)?;
+        let bytes = extract_byte_string(ok_or_missing(bytes)?)?;
+        if bytes.len() != 64 {
```
jmichelp: should this be checked against `ecdsa::Signature::BYTES_LENGTH` for consistency?

src/ctap/data_formats.rs
kaczmarczyck: On the line you marked, this `PartialEq` implementation is only used in tests. We dropped the `cfg_attr` from trait implementations as the compiler optimized them out anyway.

libraries/crypto/src/ecdsa.rs
kaczmarczyck: Good pointing that out. In the changes in this PR, all data that is handled should be public information. All from / to casts are generally not sensitive, as they deal with public keys or signatures that are part of a CTAP command. The call to `verify_*` also is something we want users to have all data for to replicate. In our case, it's about verifying that a firmware was correctly signed. A user could deploy OpenSK with a modified crypto implementation that has obvious side channels, and run the public data through it.

src/ctap/data_formats.rs
kaczmarczyck: It should, I just forgot to change the number here. Changed.

libraries/crypto/src/ecdsa.rs
jmichelp: Ok.
But if you think there's some of these functions that are sensitive and which end now being `pub`, it might be worth adding a comment to remember that these functions are only used for test or non-sensitive stuff.

src/ctap/data_formats.rs
jmichelp: Ok

jmichelp (dismissed):

libraries/crypto/src/ecdsa.rs
```diff
@@ -193,35 +193,40 @@ impl Signature {
...
 }
 
 impl PubKey {
     #[cfg(feature = "with_ctap1")]
     const UNCOMPRESSED_LENGTH: usize = 1 + 2 * int256::NBYTES;
 
+    /// Creates a new PubKey from its coordinates on the elliptic curve.
+    pub fn from_coordinates(x: &[u8; NBYTES], y: &[u8; NBYTES]) -> Option<PubKey> {
+        PointP256::new_checked_vartime(Int256::from_bin(x), Int256::from_bin(y))
+            .map(|p| PubKey { p })
```
gendx, line 231: IIRC you may be able to just write `.map(PubKey)`. But maybe this doesn't apply because `PubKey` is a struct rather than a tuple.

src/ctap/data_formats.rs
```diff
@@ -817,6 +820,87 @@ impl TryFrom<CoseKey> for ecdh::PubKey {
...
+        ecdsa::PubKey::from_coordinates(&x_bytes, &y_bytes)
+            .ok_or(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)
+    }
+}
+
+/// Data structure for receiving a signature.
+///
+/// See https://datatracker.ietf.org/doc/html/rfc8152#appendix-C.1.1 for reference.
+///
+/// TODO derive Debug and PartialEq with compiler version 1.47
```
gendx, line 845: Is this resolved? Or is the current compiler version still below that?

src/ctap/data_formats.rs
```diff
@@ -1814,6 +1899,56 @@ mod test {
...
+            "alg" => -1, // unused algorithm
+            "signature" => bytes,
+        };
+        let cose_signature = CoseSignature::try_from(cbor_value).unwrap();
+        let created_signature = crypto::ecdsa::Signature::try_from(cose_signature);
+        // Can not compare directly, since ecdsa::Signature does not implement Debug.
+        assert_eq!(
+            created_signature.err(),
+            Some(Ctap2StatusCode::CTAP2_ERR_UNSUPPORTED_ALGORITHM)
+        );
```
gendx, line 1937: Nit: `assert_eq!(foo, Err(Some(...)))` would be more idiomatic I think. Also, why is this returning a `Result<..., Option<Ctap2StatusCode>>` instead of simply a `Result<..., Ctap2StatusCode>`?

libraries/crypto/src/ecdsa.rs
gendx: +1. In this context the message being signed (the firmware) is public, so it's OK. But in general there might be cases where we'd want verification to be constant time. So I'd recommend documenting the functions as such, and/or renaming them with a `_vartime` suffix.

src/ctap/data_formats.rs
kaczmarczyck: We are currently on 1.45

src/ctap/data_formats.rs
kaczmarczyck: I need the `.err()` because `ecdsa::Signature` implements neither `Debug` nor `PartialEq`, so the whole `Result` can not be compared against. This is also where the `Option` is introduced, as `err()` turns a `Result<T, E>` into `Option<E>`.

libraries/crypto/src/ecdsa.rs
kaczmarczyck: The direct `map` doesn't work, but might if it had a suitable `PubKey::new`. I'd rather not implement one though.

libraries/crypto/src/ecdsa.rs
kaczmarczyck: The `verify_*` has a `vartime` suffix, and all from/to bytes functions should not need it.

I added documentation for all of them accordingly.

libraries/crypto/src/ecdsa.rs
```diff
@@ -257,6 +261,10 @@ impl PubKey {
         self.p.gety().to_int().to_bin(y);
     }
 
+    /// Verifies if the data's hash matches its signature.
+    ///
+    /// This function is not a constant time implementation, and does not resist side channel
+    /// attacks. Only use if all data invovled is public knowledge.
```
jmichelp: thanks for adding doc :)

typo: involved

kaczmarczyck: Thanks, fixed.

jmichelp (dismissed):

src/ctap/data_formats.rs
gendx: Got it!

src/ctap/data_formats.rs
```diff
@@ -1814,6 +1899,56 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_cose_signature_wrong_signature_length() {
```
gendx, line 1941: It would be interesting to test a shorter signature length as well, to check that it doesn't trigger any out-of-bounds read.

gendx (dismissed):

src/ctap/data_formats.rs
kaczmarczyck: Done

jmichelp (approved):

gendx (approved):

PR #359, opened 2021-08-04
adds storage access for upgrade keys

kaczmarczyck:
Implements storage access for the public keys for upgrading. Getter and Setter and currently dead code, but will be used in the following PRs.

This PR introduces an identifier, and the app is expected to know its layout in advance to have these identifiers properly set. Public keys are bound to specific partitions, and binaries only work when compiled for the correct address. This design means to reduce the chance of accidentally flashing the wrong firmware. It does come with a quirk of having these constants that need to match the board settings for the current and the other partition.

### DIFF
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use super::status_code::Ctap2StatusCode;
+use crate::embedded_flash;
 use alloc::string::String;
 use alloc::vec::Vec;
 use arrayref::array_ref;
@@ -1159,6 +1160,28 @@ impl From<CredentialManagementSubCommandParameters> for cbor::Value {
     }
 }
 
+/// Translates a the partition identifier for an upgrade to A or B.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(test, derive(IntoEnumIterator))]
+#[repr(u32)]
+pub enum UpgradeIdentifier {
+    A = embedded_flash::PARTITION_A_IDENTIFIER,
+    B = embedded_flash::PARTITION_B_IDENTIFIER,
+}
+
+impl TryFrom<cbor::Value> for UpgradeIdentifier {
+    type Error = Ctap2StatusCode;
+
+    fn try_from(cbor_value: cbor::Value) -> Result<Self, Ctap2StatusCode> {
+        let identifier = extract_unsigned(cbor_value)?;
+        match identifier as u32 {
+            embedded_flash::PARTITION_A_IDENTIFIER => Ok(UpgradeIdentifier::A),
+            embedded_flash::PARTITION_B_IDENTIFIER => Ok(UpgradeIdentifier::B),
+            _ => Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER),
+        }
+    }
+}
+
 pub(super) fn extract_unsigned(cbor_value: cbor::Value) -> Result<u64, Ctap2StatusCode> {
     match cbor_value {
         cbor::Value::Unsigned(unsigned) => Ok(unsigned),
@@ -2197,5 +2220,26 @@ mod test {
         assert!(PublicKeyCredentialSource::try_from(cbor_false!()).is_err());
         assert!(PublicKeyCredentialSource::try_from(cbor_array!(false)).is_err());
         assert!(PublicKeyCredentialSource::try_from(cbor_array!(b"foo".to_vec())).is_err());
+    }
+
+    #[test]
+    fn test_from_into_upgrade_identifier() {
+        let cbor_upgrade_identifier: cbor::Value =
+            cbor_int!(embedded_flash::PARTITION_A_IDENTIFIER as i64);
+        assert_eq!(
+            UpgradeIdentifier::try_from(cbor_upgrade_identifier),
+            Ok(UpgradeIdentifier::A)
+        );
+        let cbor_upgrade_identifier: cbor::Value =
+            cbor_int!(embedded_flash::PARTITION_B_IDENTIFIER as i64);
+        assert_eq!(
+            UpgradeIdentifier::try_from(cbor_upgrade_identifier),
+            Ok(UpgradeIdentifier::B)
+        );
+        let cbor_upgrade_identifier: cbor::Value = cbor_int!(0x12345678);
+        assert_eq!(
+            UpgradeIdentifier::try_from(cbor_upgrade_identifier),
+            Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)
+        );
     }
 }
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use super::status_code::Ctap2StatusCode;
-use crate::embedded_flash;
 use alloc::string::String;
 use alloc::vec::Vec;
 use arrayref::array_ref;
@@ -1160,28 +1159,6 @@ impl From<CredentialManagementSubCommandParameters> for cbor::Value {
     }
 }
 
-/// Translates a the partition identifier for an upgrade to A or B.
-#[derive(Clone, Copy, Debug, PartialEq)]
-#[cfg_attr(test, derive(IntoEnumIterator))]
-#[repr(u32)]
-pub enum UpgradeIdentifier {
-    A = embedded_flash::PARTITION_A_IDENTIFIER,
-    B = embedded_flash::PARTITION_B_IDENTIFIER,
-}
-
-impl TryFrom<cbor::Value> for UpgradeIdentifier {
-    type Error = Ctap2StatusCode;
-
-    fn try_from(cbor_value: cbor::Value) -> Result<Self, Ctap2StatusCode> {
-        let identifier = extract_unsigned(cbor_value)?;
-        match identifier as u32 {
-            embedded_flash::PARTITION_A_IDENTIFIER => Ok(UpgradeIdentifier::A),
-            embedded_flash::PARTITION_B_IDENTIFIER => Ok(UpgradeIdentifier::B),
-            _ => Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER),
-        }
-    }
-}
-
 pub(super) fn extract_unsigned(cbor_value: cbor::Value) -> Result<u64, Ctap2StatusCode> {
     match cbor_value {
         cbor::Value::Unsigned(unsigned) => Ok(unsigned),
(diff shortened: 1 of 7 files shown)


src/embedded_flash/upgrade_storage.rs
```diff
@@ -48,4 +48,9 @@ pub trait UpgradeStorage {
     ///
     /// Returns [`StorageError::OutOfBounds`] if the data is too long to fit the metadata storage.
     fn write_metadata(&mut self, data: &[u8]) -> StorageResult<()>;
+
+    /// Returns an identifier for these upgrade locations.
```
ia0: This is not clear to me. What are "these" upgrade locations? As I understand it we have the following distinct locations: Partition A, Partition B, Metadata. Does this mean "the current running partition" or "the next writable partition".

src/embedded_flash/upgrade_storage.rs
```diff
@@ -48,4 +48,9 @@ pub trait UpgradeStorage {
...
+    ///
+    /// We currently use the offset of the partition.
```
ia0: I guess by offset you mean address?

src/ctap/storage.rs
```diff
@@ -570,6 +570,42 @@ impl PersistentStore {
         Ok(self.store.insert(key::AAGUID, aaguid)?)
     }
 
+    /// Returns the upgrade public key if defined for the identifier.
+    // TODO remove allow
+    #[allow(dead_code)]
+    pub fn upgrade_public_key(
```
ia0: Can't we prefix the functions with `_` and remove it at first usage? It's easier not to forget because it's local information (in the name). This is more fail-proof because otherwise we might use them and forget to remove the `allow(dead_code)` and later when we stop using the function accidentally or deliberately we are not notified.

src/ctap/data_formats.rs
```diff
@@ -1159,6 +1160,28 @@ impl From<CredentialManagementSubCommandParameters> for cbor::Value {
     }
 }
 
+/// Translates a the partition identifier for an upgrade to A or B.
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(test, derive(IntoEnumIterator))]
+#[repr(u32)]
+pub enum UpgradeIdentifier {
+    A = embedded_flash::PARTITION_A_IDENTIFIER,
+    B = embedded_flash::PARTITION_B_IDENTIFIER,
```
ia0: Why is this the partition addresses? This doesn't look portable across boards. I think we want to use 0 and 1 here and return 0 or 1 in the UpgradeLocations trait. The board defines which location is A or B (this can be deduced from their index in the board).

src/embedded_flash/upgrade_storage.rs
kaczmarczyck: There is one metadata location per partition. And you only have access to the partition you are not currently on yourself. "these" are the ones in `self`. Maybe the later comments will also help clear things up.

src/ctap/data_formats.rs
kaczmarczyck: To make it board generic, I will need to add a syscall that reads the address of the current image, and use this together with the writable partition's `identifier` as `PARTITION_A_IDENTIFIER` and `PARTITION_B_IDENTIFIER` (with the smaller of the 2 being A, the larger number B).
That is, if we want to keep the property that firmware signature are bound to the address they write to. If that's the case, we will want to configure a device with both upgrade public keys and associate them with the address, meaning we know both identifiers whatever we boot into.
Else, we can make things easier and just associate 0 and 1. (We'd have to somehow learn whether we are on A or B in a board generic way though, still.)
I could also just use 0 and 1 for partition numbers, and include the firmware address in the signed over data, instead of making it the accessor for the public ECDSA key.

WDYT?

src/embedded_flash/upgrade_storage.rs
kaczmarczyck: Properly called address now.

kaczmarczyck:
[My branch](https://github.com/kaczmarczyck/OpenSK/commit/7bc40088eeaf3bfc05c8f1f342d16cb95f3819ee) contains all changes for reference. The tools are experimental and for personal use, as they contain some changes that are moving into `boards`. Also some tests are missing. Still, it should contain all App functionality and works, when flashed together with the bootloader.

src/ctap/data_formats.rs
kaczmarczyck: The identifier is now read using the apps own location, comparing that to the partition location, and returning a 0/1 enum.

src/ctap/storage.rs
kaczmarczyck: Done, even though it is slightly more work, as the `allow(dead_code)` is transitive for all subfunctions that now also need to get underscored.

src/ctap/storage.rs
```diff
@@ -570,6 +570,38 @@ impl PersistentStore {
...
+            .map(|k| {
+                _deserialize_cose_key(&k).ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)
+            })
+            .transpose()
+    }
+
+    /// Sets an upgrade public key.
+    ///
+    /// Overwrites an existing key for the given identifier.
+    pub fn _set_upgrade_public_key(
```
ia0, line 592: When looking at how this is used in `process_vendor_configure` I'm not sure I understand the final user experience. Should the user call configure twice to set both public keys? Somehow I have the impression it would be better to have a single persistent key for both public keys. The same way we do master keys and pin properties. Because we should probably always set both keys at the same time. For reading we just return the correct one or return the struct and let the caller do what they want. What do you think?

src/embedded_flash/mod.rs
```diff
@@ -19,6 +19,12 @@ mod helper;
 mod syscall;
 mod upgrade_storage;
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum UpgradeIdentifier {
+    A = 0,
+    B = 1,
```
ia0, line 25: Do we actually need the `= 0` and `= 1`? We never serialize this or anything.

src/ctap/storage.rs
ia0: Oh sorry, I forgot that this was recursive :-( If you were confident you would not forget to remove the `allow(dead_code)` because it's already part of a future commit we could have temporarily kept them.

src/embedded_flash/upgrade_storage.rs
ia0: Yeah, I was missing the fact that each "side" (A/B) has its own metadata.

ia0:
For future use, this link is better to check the final diff: https://github.com/google/OpenSK/compare/develop...kaczmarczyck:upgrade. You only pointed to a commit and not a branch. I assume your final branch is `upgrade`.

src/embedded_flash/mod.rs
kaczmarczyck: I will be matched against its integer representation when parsing the [configure command input](https://github.com/kaczmarczyck/OpenSK/blob/96202434661fd57b9ea1220549a6fba3066e0f03/src/ctap/command.rs#L591). This is how the client stores a public key either for partition A or for partition B. Depending on the comment above, this might not be needed if we always configure both public keys right away.

kaczmarczyck:
The other commits between `develop` and `upgrade` belong to my `remote-debugging` branch that removes all touch and replug requirements, for convenience in development. I can rebase when I next push changes and then link to the correct diff.

src/embedded_flash/mod.rs
jmichelp: IMO rotating the signing keys shouldn't be done through a vendor command but part of the firmware update process: you push a new key set (both A and B), signed respectively by their previous ones.
The initial key set should be part of the bootloader installation.

src/ctap/storage.rs
kaczmarczyck: I was thinking that client's just configure twice. But aren't those independent questions?
- 1 or 2 persistent keys
- storage API gets and sets 1 or 2 keys

I think for persistent keys, having 2 is better since we can't easily predict the length of their storage representation, so we'd need a CBOR map. CBOR maps need destructuring, and that costs binary size.

I don't have a use case in mind where the getter needs to return both public keys.

Your point with the setter is fair, you will probably always want both configured. I think that also depends on how we will deal with device lock down, and if we allow lockdown without upgradability (for users that rather want to reduce attack surface). One other use case I can think of outside of configure will be when a future firmware rotates the public keys, i.e. when the associated private keys are too old or compromised. That will probably also overwrite both.

src/embedded_flash/mod.rs
kaczmarczyck: I just added my comment mentioning key rotation in parallel. Seems like this needs more discussion, as we might have to converge on how to bootstrap first. If we don't intend to bootstrap the public keys with a CTAP command, we can as well bake them into the app. I scheduled a discussion on that.

src/ctap/storage.rs
ia0: I see, the keys are not constant length. But why use a map instead of an array? I'm not sure the binary size should influence too much the design (in particular for things like persistent storage key partition that can't be easily changed), but it's indeed a compromise.

ia0:
It's ok, I'm not planning to read the whole diff between `develop` and `upgrade` I just want to search for usages of introduced APIs. No need to remove those commits, rebase, or anything particular to clean up the diff.

src/ctap/storage.rs
kaczmarczyck: Or an array, yes. But the cleanest design would probably be to map `UpgradeIdentifier` to public key. That was my first implementation, and I saved around 1 kB with the current approach IIRC. And since in our case a few kB may mean that OpenSK deploys on a certain hardware, I'd give it more weight.

ia0: 1kB is due to some rounding? that's huge. I guess the cbor map implementations are quite big. I don't think arrays would be that bad. I'm not sure a map is cleaner because the fact that we have exactly 2 sides A/B is hard-coded in the whole upgrade design.

src/ctap/storage.rs
```diff
@@ -570,6 +570,38 @@ impl PersistentStore {
         Ok(self.store.insert(key::AAGUID, aaguid)?)
     }
 
+    /// Returns the upgrade public key if defined for the identifier.
+    pub fn _upgrade_public_key(
```
gendx, line 574: IIRC a function with an underscore prefix is equivalent to implicitly adding the `allow(dead_code)` attribute. Is that intended here?

ia0: Yes, see [this discussion](https://github.com/google/OpenSK/pull/359#discussion_r682787364) for the rationale. Now that I know it's recursive, I think the best approach is to use `warn(dead_code)`. This should only be allowed in `develop` branch. And the warning during compilation reminds that there is some work in progress. And if the work ends up being deprecated, then we know we can delete the function.

kaczmarczyck:
We decided to take a different entirely and have the upgrade key in `key_materials`. I'll close this PR and create a new one after the kernel version bump.

PR #360, opened 2021-08-05
Add tool for applying/saving Tock patches

ia0:
Also moving existing maintainers-only tools to the new maintainers directory, such that users don't get confused by them.


maintainers/patches
```diff
@@ -0,0 +1,88 @@
...
+}
+
+apply() {
+  git submodule -q update third_party/tock
+  ( cd third_party/tock
+    git reset -q --hard
+    git clean -qfxd
+    rsync -a ../../boards .
+    commit '00-boards'
+    for file in ../../patches/tock/*; do
```
kaczmarczyck, line 93: The `setup-submodules` script uses
```
for p in ../../patches/tock/[0-9][0-9]-*.patch; do
```

Might be good to be consistent?

kaczmarczyck (dismissed):
Curious how you use it to change 1 specific patch? First `apply`, then manually walk around with git, apply changes and rebase the rest, then last checkout HEAD and `save`?

maintainers/patches
```diff
@@ -0,0 +1,88 @@
...
+      patch -sp1 < $file
```
jmichelp: FYI I was using `patch` at the beginning. But it appears that the `patch` command is stricter than the `git apply` equivalent and several times I had `patch` failing when `git apply` worked.

maintainers/patches
```diff
@@ -0,0 +1,88 @@
...
+      commit "${${file#../../patches/tock/}%.patch}"
```
jmichelp: Without using zsh specific syntax, you can use: `commit "$(basename $file .patch)"`
This should work on Linux and OSX

maintainers/patches
```diff
@@ -0,0 +1,88 @@
...
+    rm ../../patches/tock/00-boards.patch
+    rsync -a --delete boards ../..
+    git reset -q --hard
+    git clean -qfxd
+    git checkout -q $top
+  ) || exit
+  success 'Saved the Tock submodule to the patches.'
+}
+
+[[ $(basename $PWD) == OpenSK ]] || fail 'Not running from OpenSK directory.'
```
jmichelp: Here it enforces a name for the local directory where you checkout the repo. Not sure if it's ideal.
I can't think about a neat way to check for this unfortunately beside ensuring there is a `.git/` directory combined with `git config --get remote.origin.url | grep -i opensk` not failing.

maintainers/patches
```diff
@@ -0,0 +1,88 @@
+#!/bin/zsh
+
```
jmichelp, line 3: Don't you want to `set -e` at the beginning to stop the script at the first error?

maintainers/patches
```diff
@@ -0,0 +1,88 @@
...
+    for file in $(git format-patch $root); do
+      sed -n '/^diff/,$p' $file | head -n-3 > ../../patches/tock/${file#*-}
+    done
+    git clean -qfxd
+    top=$(get_head)
+    git checkout -q $root
+    rm -r boards
+    patch -sp1 < ../../patches/tock/00-boards.patch
+    rm ../../patches/tock/00-boards.patch
+    rsync -a --delete boards ../..
```
jmichelp: Could be nice to check that `rsync` is installed at the very beginning. It's rarely installed by default.

maintainers/patches
```diff
@@ -0,0 +1,88 @@
...
+  success 'Applied the patches to the Tock submodule.'
+}
+
+save() {
+  local root=$(get_root)
+  ( cd third_party/tock
+    [[ -z "$(git status -s)" ]] || fail 'The Tock submodule is not clean.'
+    rm ../../patches/tock/*.patch
+    for file in $(git format-patch $root); do
+      sed -n '/^diff/,$p' $file | head -n-3 > ../../patches/tock/${file#*-}
```
jmichelp: That's the only line that is ZSH specific. I'm not sure what the input/output are for this substitution so I can't give you a POSIX equivalent.
The rest of the script should be portable between zsh and bash AFAICT.

maintainers/patches
ia0: Done. I usually do it but hesitated in this case because sometimes commands return a failing status for normal operation. But I don't think it's the case here (at least not in my experiments).

maintainers/patches
ia0: I don't think it's beneficial. I prefer this script to be more generic to how files are named such that if we change our nomenclature we don't break / need to update this tool.

maintainers/patches
ia0: Ok, I didn't observe any issue. It's maybe when patch don't apply very cleanly. I changed to `git apply` anyway because it should still work the same. The only difference we get are warnings about trailing spaces due to the `.jlink` files being DOS style (i.e. they use `\r\n` instead of `\n`). But I shut down the warnings which should be good enough.

maintainers/patches
ia0: Thanks, I wasn't aware of `basename` taking a suffix.

maintainers/patches
ia0: It's actually not zsh specific. It still works with bash (at least on linux). The problem with the other line was that it was chaining/nesting such constructs which bash doesn't support.

maintainers/patches
ia0: I ended using `cp -a` instead. We don't really need `rsync`, I just use it because that's what I usually use when copying directories.

maintainers/patches
ia0: I decided to go with another heuristic which should be generic enough, since this is really just to give some helpful message instead of some weird bugs.

maintainers/patches
```diff
@@ -42,44 +27,99 @@ get_head() {
...
+
+  6. Either continue repeating steps 2 to 5, or return to the normal state:
+
+    ./setup.sh
+EOF
+  exit 0
+}
+
 apply() {
-  git submodule -q update third_party/tock
```
jmichelp: Just double checking: is the `git checkout` below replacing the `git submodule update` here?

maintainers/patches
kaczmarczyck: I don't have a strong preference either way, I just thought it might be a source for confusion if scripts behaves differently. Counter argument would be a case like wanting to store a `README.md`? How about both scripts take all `.patch` files?

maintainers/patches
ia0: Yes, the reason is that if the submodule is dirty, then the `git submodule update` fails. This seems to be ok if there is nothing to update which is why it works from a normal state (i.e. after setup.sh). But I would like `apply` to work it all cases such that it's easy to restore to an edit state. We could argue that we can always to `reset && setup && apply` but that's more costly. Actually I should probably add a reset command (in addition to apply and save) to have a faster setup.sh. I'll take a look into that.

maintainers/patches
ia0: Oh I see, indeed one could use this script to rename patches in a way that would not work with setup.sh anymore. Good point. I would be more in favor of using all patches because what would the patches not satisfying the format be for? They would be confusing I think.

jmichelp: The reason the filenames for patches start with numbers is simply to ensure they are applied in a specific consistant order.

ia0: Yes we rely on the fact that pathname expansion is in alphabetical order but we don't need the patch names to start with exactly 2 digits and a dash. This may be modified in the future (if needed) and the scripts should still work the same.

jmichelp (approved):

PR #361, opened 2021-08-05
Explicitly limit the fuzzer input length

ia0:
This is to avoid timeouts in oss-fuzz.

We either want to also backport to stable, or just wait until we fuzz develop.


jmichelp (dismissed):

ia0:
I actually decided to increase the limit from 10k to 100k, because on that particular example 10k was too fast while 100k takes 16 seconds which is still ok. And I prefer to get an oss-fuzz report that a test exceeds timeout than have to remember to increase the limit if I don't get a report. This is easier to adjust the limit correctly.

jmichelp (approved):

PR #362, opened 2021-08-05
Make flash syscall interface async

ia0:
This is to support chips that only provide async flash APIs. For chips that only provide blocking flash APIs, the same trick as is used in this PR for the nrf52 chip can be used (using deferred calls).


patches/tock/02-usb.patch
```diff
@@ -215,8 +215,8 @@ index ebee8704b..6a2f8715b 100644
              kernel::ipc::DRIVER_NUM => f(Some(&self.ipc)),
              _ => f(None),
          }
-@@ -371,6 +388,21 @@ pub unsafe fn reset_handler() {
-         )
+@@ -377,6 +394,21 @@ pub unsafe fn reset_handler() {
+             .expect("no deferred call slot available for nvmc"),
```
jmichelp, line 219: Why is USB impacted if we don't change the reference commit for the kernel?

jmichelp (commented):
According to the workflows, you forgot to update our own boards (under `boards/`) in addition to those under `third_party/tock/`

patches/tock/02-usb.patch
ia0: This is just a context line. The `+` is actually part of the file not part of the diff. It's like for the line numbers that get updated for the patch to be applied perfectly.

ia0:
Yes indeed, I forgot the other boards. I've updated them and also moved the diffs out of the persistent storage patch to the initial fil, now that we don't use the Tock boards at all. This was already the case for the mdk board actually.

patches/tock/02-usb.patch
jmichelp: I still read `no deferred call slot available for nvmc` for a patch about USB.
I doubt this was there before ;)

ia0: Yes it wasn't there before this PR. The usb patch is applied after the boards are copied and after the persistent storage patch. So with this PR, when applying the usb patch, this line is actually present.

jmichelp: Gotcha.

jmichelp (approved):

kaczmarczyck (approved):
LGTM

If all changes in our own boards are supposed to move out of the patches, I can deal with mine later. Currently, mine is still comprehensive, including all changes to boards and kernel.

ia0:
Indeed:

```
% grep -rl src/main patches/tock 
patches/tock/07-upgrade-partitions.patch
patches/tock/06-firmware-protect.patch
patches/tock/02-usb.patch
```

Do you want to do the move for those files because you're the only one touching at least one of them? (This could be done next time you touch your patch.) Otherwise I can do it when you think the time is right.

Regarding this particular PR, I'll wait to test the dongle before submitting. I only tested the devkit.

kaczmarczyck:
Yes, I thought I'll just do it when I touch the patches!

ia0:
Tested dongle and dev-kit with webauthn (register, login, board reset, login).

PR #365, opened 2021-08-10
Remove elf2tab dev-dependency

### LINKED ISSUE #364: RUSTSEC-2021-0080: Links in archive can create arbitrary directories
> Links in archive can create arbitrary directories

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `tar`                      |
| Version             | `0.4.35`                   |
| URL                 | [https://github.com/alexcrichton/tar-rs/issues/238](https://github.com/alexcrichton/tar-rs/issues/238) |
| Date                | 2021-07-19                         |

When unpacking a tarball that contains a symlink the `tar` crate may create
directories outside of the directory it&#39;s supposed to unpack into.

The function errors when it&#39;s trying to create a file, but the folders are
already created at this point.

```rust
use std::{io, io::Result};
use tar::{Archive, Builder, EntryType, Header};

fn main() -&gt; Result&lt;()&gt; {
    let mut buf = Vec::new();

    {
        let mut builder = Builder::new(&amp;mut buf);

        // symlink: parent -&gt; ..
        let mut header = Header::new_gnu();
        header.set_path(&quot;symlink&quot;)?;
        header.set_link_name(&quot;..&quot;)?;
        header.set_entry_type(EntryType::Symlink);
        header.set_size(0);
        header.set_cksum();
        builder.append(&amp;header, io::empty())?;

        // file: symlink/exploit/foo/bar
        let mut header = Header::new_gnu();
        header.set_path(&quot;symlink/exploit/foo/bar&quot;)?;
        header.set_size(0);
        header.set_cksum();
        builder.append(&amp;header, io::empty())?;

        builder.finish()?;
    };

    Archive::new(&amp;*buf).unpack(&quot;demo&quot;)
}
```

This issue was discovered and reported by Martin Michaelis (@mgjm).

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2021-0080.html) for additional details.

ia0:
We don't use it anymore. Not sure when we used to use it. Let's see if any workflow fails.

Fixes #364

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ with_ctap1 = ["crypto/with_ctap1"]
 with_nfc = ["libtock_drivers/with_nfc"]
 
 [dev-dependencies]
-elf2tab = "0.6.0"
 enum-iterator = "0.6.0"
 
 [build-dependencies]


kaczmarczyck (approved):

Cargo.toml
```diff
@@ -30,7 +30,6 @@ with_ctap1 = ["crypto/with_ctap1"]
 with_nfc = ["libtock_drivers/with_nfc"]
 
 [dev-dependencies]
-elf2tab = "0.6.0"
 enum-iterator = "0.6.0"
```
jmichelp, line 33: And do we use this one? :)

kaczmarczyck: It's part of some tests. Basically doing something like
for variant in enum
  convert from and to CBOR
  assert_eq!

jmichelp (approved):

PR #368, opened 2021-08-12
Make sure the end offset doesn't overflow

ia0:
We used to only check that the length was not too big but didn't check that the starting offset wasn't too big. We want the end offset to not overflow.

Not sure which error code should be returned though.

Bug found by oss-fuzz.


kaczmarczyck (approved):
Thanks for the fix!

Issue #155, opened 2020-09-21
Investigate why the modifications on Tock regarding the MPU makes OpenSK panic


jmichelp:
## Expected Behavior

OpenSK should not panic when using the flash storage

## Actual Behavior

When moving to a more recent kernel version, it appears that the storage causes a panic because of the changes in the MPU configuration.
We 

## Steps to Reproduce the Problem

1. Use updated kernel/patches from #150 
1. Don't apply the patch `patches/tock/05-mpu-fix.patch`
1. Compile and deploy OpenSK firmware with the flag `--clear-storage`
1. kernel panic should happen immediately after the board resets.

## Specifications

- Version: N/A
- Platform: N/A

ia0:
The issue has been introduced by https://github.com/tock/tock/pull/1873. I've [commented](https://github.com/tock/tock/pull/1873#issuecomment-696217269) to get a tracking bug. That PR introduces a new concept, `KernelMpuConfig`, which is not yet implemented on any of the architectures, but necessary for a fix at least for Cortex-M.

ia0:
This bug has been triaged and fixed upstream: https://github.com/tock/tock/pull/2120. Do we want to proactively sync the kernel again to remove the MPU patch?

jmichelp:
The Time HIL also landed, potentially fixing other issues we have. Let's wait a bit more before a new sync.

kaczmarczyck:
Fixed with the kernel upgrade.

PR #374, opened 2021-08-16
Bump Tock kernel version

jmichelp:
No issue fixed here. Just upgrading to a more recent kernel.

Starting this PR, all the files in `boards/` come already with the necessary patches applied.

Also the PR template got updated to better reflect how the PR was tested.

-   [x] Local tests pass (running `run_desktop_tests.sh`)
-   [ ] Tested against boards
    -   [x] Nordic nRF52840 DK
    -   [ ] Nordic nRF52840 Dongle (JTAG programmed)
    -   [x] Nordic nRF52840 Dongle (DFU programmed)
    -   [ ] Makerdiary nRF52840 MDK USB Dongle
-   [x] Appropriate changes to README are included in PR


jmichelp:
As part of the tests on real hardware, I need to check if this new kernel version removes the need of `patches/04-mpu-fix.patch` and therefore closes issue #155

jmichelp:
Also: need to assess wether we should also bump `libtock-rs` as part of this PR.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,5 +2,10 @@ Fixes #<issue_number_goes_here>
 
 > It's a good idea to open an issue first for discussion.
 
-- [ ] Tests pass
-- [ ] Appropriate changes to README are included in PR
\ No newline at end of file
+-   [ ] Local tests pass (running `run_desktop_tests.sh`)
```
ia0, line 5: Do we still need this now that we have the github worflows?

ia0 (commented):
The dev kit panics for me. I can see the app is loaded, but I'm not sure it starts. And I get so "kernel stack overflow". But I couldn't debug in gdb because I couldn't reset the board while gdb is attached and break somewhere before the issue.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -2,5 +2,10 @@ Fixes #<issue_number_goes_here>
...
+-   [ ] Tested against boards
```
ia0, line 6: Should we specify what to test? What I do is register, login, reset board (or unplug dongle), login. This permits to test a bit of CTAP and a bit of storage.

.github/workflows/boards_build.yml
```diff
@@ -18,9 +18,8 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: "true"
-      - uses: actions-rs/toolchain@v1
-        with:
-          target: thumbv7em-none-eabi
+      - name: Install Rust toolchain
+        run: rustup show
```
ia0, line 22: This doesn't really install the toolchain right? It just prints the current installed toolchain and which one is effective currently (subject to rust-toolchain). It's deploy.py that installs the toolchain.

Same comment in all other workflows.

deploy.py
```diff
@@ -432,14 +436,23 @@ def create_tab_file(self, binaries):
...
+          ["nm", "--print-size", "--radix=x", app_file])
+      for line in nm.splitlines():
+        if "STACK_MEMORY" in line:
+          required_stack_size = int(line.split(" ", maxsplit=2)[1], 16)
+          stack_sizes.append(required_stack_size)
 
     elf2tab_args.extend([
-        "--stack={}".format(STACK_SIZE), "--app-heap={}".format(APP_HEAP_SIZE),
-        "--kernel-heap=1024", "--protected-region-size=64"
+        "--stack={}".format(max(stack_sizes)),
```
ia0: When can we have more than one? Why not warn if `len(stack_sizes) != 1`?

ia0 (commented):
When debugging `--opensk` on `--board=nrf52840dk_opensk` the process hard faults because of [this check](https://github.com/tock/tock/blob/e916444378a3e683cde5b24f2cc166e5f056ff8c/arch/cortex-m/src/syscall.rs#L183). At that point `state.psp = 0x20023fe0` which seems reasonable and `app_brk = 0x20020c00` which seems wrong. I don't know how the default app break is decided, but we probably need it to cover at least the stack.

.github/workflows/boards_build.yml
jmichelp: With the new `rust-toolchain` toml format, it actually installs the required toolchain if it's not present before displaying what's installed 🤷

ia0: Oh wow I wasn't aware of this magic. That's a bit counter-intuitive though :-)

src/main.rs
```diff
@@ -47,6 +47,8 @@ use libtock_drivers::timer::Timer;
 use libtock_drivers::timer::Timestamp;
 use libtock_drivers::usb_ctap_hid;
 
+libtock_core::stack_size! {0x4000}
```
kaczmarczyck, line 50: This is a bigger stack than before, or are units different?

jmichelp: Nope, exactly the same as currently stated in `deploy.py` AFAICT

kaczmarczyck: Hmm, I might have confused it with the example stack size from `layout.ld`.

.github/PULL_REQUEST_TEMPLATE.md
kaczmarczyck: A solution outside the scope of this PR would be to add a compiler flag for a "test mode". This test mode removes all user presence checks, like I currently do in my [remote debugging branch](https://github.com/kaczmarczyck/OpenSK/tree/remote-debugging). Then you just run the Test Tool against all board in test mode. This should cover all your use cases.

Downsides are that it takes a few minutes per board in wall time. Also, the Test Tool would need a touchless mode that omits all tests that would fail. Therefore, a short term solution would be to just ask to check webauthn.io or have a short Python script that sends a register and login?

.github/PULL_REQUEST_TEMPLATE.md
kaczmarczyck: I doesn't necessarily improve test coverage, but it does help advertising this feature to new contributors. Otherwise, they'd always have to wait for workflow feedback. That's why I'd keep it, WDYT?

deploy.py
jmichelp: We can have more than one compiled app/example in the TAB file when we'll start to compile for multiple architectures (armv7, armv8 and risc-v come to mind).
But the stack size should remain the same and using a `set()` instead of a `list()` and warning if we have more than one entry would make sense.

jmichelp:
I forgot to mention: I checked the source code and patch 4 (`04-mpu-fix.patch`) is not required anymore and can be safely removed from the list.

kaczmarczyck:
I removed it, thanks for checking!

deploy.py
kaczmarczyck: It's a set now and prints an error if different stack sizes are detected.

.github/PULL_REQUEST_TEMPLATE.md
ia0: I usually locally test my change only and rely on the workflow for the rest. But yeah we can keep it if it's a useful feature for new contributors.

deploy.py
ia0: Oh I missed the outer for loop over archs. SGTM

deploy.py
```diff
@@ -309,7 +304,16 @@ def checked_command_output(self, cmd, env=None, cwd=None):
   def update_rustc_if_needed(self):
     target_toolchain_fullstring = "stable"
     with open("rust-toolchain", "r", encoding="utf-8") as f:
-      target_toolchain_fullstring = f.readline().strip()
+      content = f.readlines()
+      if len(content) == 1:
+        # Old format, only the build is stored
+        target_toolchain_fullstring = f.readline().strip()
```
jmichelp: I see I made a mistake here.
This should be `content[0].strip()` instead of `f.readline().strip()`. The whole file has already been read in line 307.

deploy.py
```diff
@@ -432,14 +436,25 @@ def create_tab_file(self, binaries):
...
       dest_file = os.path.join(self.tab_folder, "{}.elf".format(arch))
       shutil.copyfile(app_file, dest_file)
       elf2tab_args.append(dest_file)
+      # extract required stack size directly from binary
+      nm = self.checked_command_output(
+          ["nm", "--print-size", "--radix=x", app_file])
+      for line in nm.splitlines():
+        if "STACK_MEMORY" in line:
+          required_stack_size = int(line.split(" ", maxsplit=2)[1], 16)
+          if stack_sizes and required_stack_size not in stack_sizes:
```
jmichelp: Why not simply adding to the set and, outside of the for loop, issue an error if `len(stack_sizes) != 1`?

deploy.py
```diff
@@ -432,14 +436,25 @@ def create_tab_file(self, binaries):
...
+            error("Detected different stack sizes across tab files.")
+          stack_sizes.add(required_stack_size)
 
     elf2tab_args.extend([
-        "--stack={}".format(STACK_SIZE), "--app-heap={}".format(APP_HEAP_SIZE),
-        "--kernel-heap=1024", "--protected-region-size=64"
+        "--stack={}".format(max(stack_sizes)),
```
jmichelp: Now that `stack_sizes` is a set and the above checks ensure that its size is 1, you can replace `max(stack_sizes)` by `stack_sizes.pop()`

docs/install.md
```diff
@@ -261,13 +261,15 @@ the board in order to see your OpenSK device on your system.
...
 our tool, `deploy.py` also supports other methods:
 
-*   OpenOCD: `./deploy.py --board=nrf52840_dongle --opensk --programmer=openocd`
-*   pyOCD: `./deploy.py --board=nrf52840_dongle --opensk --programmer=pyocd`
-*   Nordic DFU: `./deploy.py --board=nrf52840_dongle --opensk
+*   OpenOCD: `./deploy.py --board=nrf52840_dongle_opensk --opensk
+    --programmer=openocd`
+*   pyOCD: `./deploy.py --board=nrf52840_dongle_opensk --opensk
+    --programmer=pyocd`
+*   Nordic DFU: `./deploy.py --board=nrf52840_dongle_opensk --opensk
```
jmichelp: This one should be using another board, right?
Like `nrf52850_dfu_dongle` or something

kaczmarczyck: Right, that one is special indeed.

deploy.py
kaczmarczyck: Works too, changed!

deploy.py

deploy.py
kaczmarczyck: True, that's better. Done

jmichelp (commented):
LGTM (note that being the original author, I can't approve this PR)

Regarding testing: it seems my Makerdairy dongle's bootloader died. I would need to crack the case open and reflash a bootloader or at least reprogram UICR.

kaczmarczyck (approved):

kaczmarczyck:
I can formally approve, so please merge if you are happy with it as is!

PR #375, opened 2021-08-27
Reduce the store fuzzer input length

ia0:
We still hit timeouts about 4% of the time some days (other days there are 0%, not sure why).

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -308,7 +308,7 @@ class OpenSKInstaller:
 
   def update_rustc_if_needed(self):
     target_toolchain_fullstring = "stable"
-    with open("rust-toolchain", "r") as f:
+    with open("rust-toolchain", "r", encoding="utf-8") as f:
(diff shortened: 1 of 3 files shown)


jmichelp (dismissed):

deploy.py
```diff
@@ -308,7 +308,7 @@ def checked_command_output(self, cmd, env=None, cwd=None):
 
   def update_rustc_if_needed(self):
     target_toolchain_fullstring = "stable"
-    with open("rust-toolchain", "r") as f:
+    with open("rust-toolchain", "r", encoding='utf-8') as f:
```
jmichelp: Thanks for fixing it.
For consistency across the file (I think there might be a check in pylint or flake), please use double quotes.

ia0: Ok, done. I missed that fact.

jmichelp (approved):

PR #378, opened 2021-09-11
Configure the number of pages in the board

ia0:
This permits to have portable apps. The OpenSK boards decides where and how big the storage should be. The OpenSK app simply uses whatever the boards (through the kernel) provides.

The name `unallocated_size` comes from [this argument](https://github.com/tock/tock/blob/e916444378a3e683cde5b24f2cc166e5f056ff8c/arch/cortex-m/src/mpu.rs#L389) name, and is needed to satisfy the MPU on Nordic.


patches/tock/01-persistent-storage.patch
```diff
@@ -415,7 +415,7 @@ index c52754be3..ae6a58341 100644
 +                .mpu()
 +                .allocate_region(
 +                    storage_location.address as *const u8,
-+                    storage_location.size,
++                    storage_location.unallocated_size,
```
kaczmarczyck: Let's compare this to giving the MPU `size` instead of `unallocated_size`. It would probably round up to a power of 2, right? Any problem with that I'm missing?

I assume you don't want the extra space be allocated by anything else, so we will always be able to increase the number of pages? It does add more freedom to enforce that intention. But is it worth the extra complexity? Other use cases of `kernel::StorageLocation` probably won't use that feature.

deploy.py
```diff
@@ -94,7 +94,7 @@
             app_ldscript="nrf52840_layout.ld",
             app_address=0x40000,
             storage_address=0xC0000,
-            storage_size=0x40000,
+            storage_size=0x14000,
```
kaczmarczyck, line 97: Just noting I don't understand the intention behind `size` and `unallocated_size`, so I'm not sure why not to erase the unallocated section. Probably answered together with the other comment.

kaczmarczyck (commented):
Good idea to do, just curious about the `unallocated_size`.

patches/tock/01-persistent-storage.patch
ia0: The reason I added `unallocated_size` is because using `size` doesn't work. There is already a `size` argument and we use `size` there. The `unallocated_size` argument is to tell the MPU that it's fine to allocate up to that size which is needed because the MPU is not able to allocate small regions near the end of the flash.

The `unallocated_size` is not used to prevent that part of the flash from being allocated by other things. This is just a workaround for the MPU. There is indeed no need to reserve that space.

deploy.py
ia0: Yes we could erase for the whole `unallocated_size` (which for now is correct because it's not used by anything else) but it would be a waste of erase cycles since we don't use it (unless whatever erases those pages doesn't do it if they are already erased).

Ideally `deploy.py` should read those values from the board instead of hard-coding them.

patches/tock/01-persistent-storage.patch
kaczmarczyck: And we actually need to pass that number down as part of `StorageLocation`? We can't just round `size` up to the next power of 2?

deploy.py
jmichelp: It might be easier to do it the other way around: define the values in `deploy.py` which then passes them to the boards through environment variables.
It would require to adapt the `build.rs` files to make sure we rebuild in case the values have changed since last compile time.

patches/tock/01-persistent-storage.patch
ia0: Right, we could assume the `unallocated_size` being the smallest number such that the allocation works:
- PRO: we don't specify the unallocated_size
- CON: we don't specify the unallocated_size so we don't know what region is actually used
- PRO: We don't touch the MPU code.
- CON: We change the MPU code.

I'm fine to try that if we think it's better.

deploy.py
ia0: Yes that would work too.

patches/tock/01-persistent-storage.patch
kaczmarczyck: Why touch MPU code? It would be replacing `unallocated_size` here in `kernel/src/process.rs` with the rounded up `size`?
The argument of not knowing the region used is technically true for the app too, since the MPU rounds up its region size too. And the actual app size is even more implicit (even nondeterministic to some small degree?).

I like the cleaner board definition. Maybe more freedom to assign regions will be required for some boards, but we can always increase complexity if necessary?

deploy.py
kaczmarczyck: I will look into that when I work on A/B boards. Happy with both solutions for the `storage_size`, saving cycles makes sense.

patches/tock/01-persistent-storage.patch
ia0: > Why touch MPU code? It would be replacing unallocated_size here in kernel/src/process.rs with the rounded up size?

The logic is a bit complicated and I'm not sure rounding up always works. But I'll try and if it works let's do that. This is just a workaround for Nordic and we probably won't change those values often.

> The argument of not knowing the region used is technically true for the app too, since the MPU rounds up its region size too. And the actual app size is even more implicit (even nondeterministic to some small degree?).

But it doesn't do it outside of the flash already use by the binary. This is why elf2tab already does the rounding up. In our case, we would round outside the requested size. (Note that it's probably possible to improve the MPU code, quickly looking through it, it seems to miss the opportunity to use a smaller subregion size when possible.)

> I like the cleaner board definition. Maybe more freedom to assign regions will be required for some boards, but we can always increase complexity if necessary?

￼Yes, I'm fine with the hack in process.rs to use some heuristics to make it work for Nordic. We'll need to test it on other boards too.

kaczmarczyck: We might need a better MPU implementation down the road when we want more precise control over regions in the future, i.e. on boards with less flash. So hack for Nordic might be what we want for now. Sounds good.

kaczmarczyck (approved):

PR #379, opened 2021-09-15
Upgrade signing key generation

kaczmarczyck:
This PR adds the upgrade signing key, including:
- generate in `setup.sh`
- encoding in `build.rs`
- import in `ctap/key_material.rs`
- mention in `install.md`

On top, it adds one missing function to get the partition offset to the interface, and cleans up some unnecessary code after the compiler version jump.
The key is currently unused in `key_material.rs`. For the bigger picture, please check my [upgrade branch](https://github.com/kaczmarczyck/OpenSK/tree/upgrade).


build.rs
```diff
@@ -34,4 +39,29 @@ fn main() {
     content.truncate(36);
     let aaguid = Uuid::parse_str(&content).unwrap();
     aaguid_bin_file.write_all(aaguid.as_bytes()).unwrap();
+
+    // COSE encoding the public key, then write it out.
+    let file = File::open("crypto_data/opensk_upgrade_pub.pem").unwrap();
```
ia0: The file name could be shared in a constant with the println above.

tools/gen_key_materials.sh
```diff
@@ -88,6 +93,17 @@ generate_crypto_materials () {
       -sha256
   fi
 
+  if [ ! -f "${opensk_upgrade}" ]
```
jmichelp: Do we want to generate this key pair anyway in case the `force_generate` flag is set?

src/embedded_flash/buffer_upgrade.rs
```diff
@@ -57,6 +57,10 @@ impl UpgradeStorage for BufferUpgradeStorage {
         }
     }
 
+    fn partition_address(&self) -> usize {
+        0x60000
```
jmichelp, line 61: Is this address going to be generic enough for all boards?

build.rs
```diff
@@ -34,4 +39,29 @@ fn main() {
...
+    let (pem, _bytes_read) = Pem::read(std::io::BufReader::new(file)).unwrap();
```
jmichelp: I would rather use OpenSSL to manipulate and extract these more reliably.
If you look in the history (e.g. https://github.com/google/OpenSK/blob/6c363981c4b8b7a09716095e2b93043ef2a35c8a/build.rs), I used to do it to validate all keys/certificate.

src/embedded_flash/buffer_upgrade.rs
kaczmarczyck: Generic in what sense? As in, if you have two physical partitions, the address is a tuple `(partition_name, address)`? If that's the case, we can always adjust to other boards later, if we need it?

jmichelp: Generic in the way that the hardware abstraction is supposed to come from the kernel. An app should be able to work on any board provided all the syscalls it requires to work are implemented in the underlying kernel.
So I'm wondering about the relevance of a hard-coded address in an app.

src/ctap/key_material.rs
```diff
@@ -14,6 +14,9 @@
 
 pub const ATTESTATION_PRIVATE_KEY_LENGTH: usize = 32;
 pub const AAGUID_LENGTH: usize = 16;
+pub const _UPGRADE_PUBLIC_KEY_LENGTH: usize = 77;
 
 pub const AAGUID: &[u8; AAGUID_LENGTH] =
     include_bytes!(concat!(env!("OUT_DIR"), "/opensk_aaguid.bin"));
+pub const _UPGRADE_PUBLIC_KEY: &[u8; _UPGRADE_PUBLIC_KEY_LENGTH] =
```
ia0, line 21: Last time, after  discussion, I realized that `#[warn(dead_code)]` (and not `#[allow(dead_code)]`) might be more convenient (no need to change names). As you want.

ia0 (dismissed):
I also took a quick look at the `upgrade` branch. I guess you split the partition location in 2 because of the MPU? I actually realized I could probably do the same for the storage (and avoid the MPU hack :D). I'll try next time I have time.

src/ctap/key_material.rs
kaczmarczyck: I tried that, and realized that we error on warning. So it's either `allow` or renaming. And in this case, it wasn't too much work to rename.

ia0: Oh right this is not great. We might want to remove that rule in `develop` branch.

build.rs
kaczmarczyck: Done

tools/gen_key_materials.sh
kaczmarczyck: Probably makes sense, yes. The upgrade key should be tied to a AAGUID, and if that is regenerated, then the upgrade key should be as well.

build.rs
kaczmarczyck: I now parse the file using OpenSSL. Makes sense given that OpenSSL is also used earlier in the stack. It's different from your code, given that I parse an uncompressed public key instead of a private key. PTAL

kaczmarczyck:
@ia0 Yes, the split is to satisfy the MPU. The ugly regex script will make room for new boards eventually though. :)

src/embedded_flash/buffer_upgrade.rs
kaczmarczyck: Oh, then we should be good. This is the test implementation, so `upgrade_locations` work in unit test in a `std` environment. The actual implementation is in the parallel file:
```
    fn partition_address(&self) -> usize {
        self.partition.start()
    }
```
And that values is read from syscalls.

ia0 (approved):

jmichelp (approved):

PR #380, opened 2021-09-15
Revert the MPU hack by splitting the store in 2

ia0:
Essentially reverts kernel/src/process.rs to its state before moving the NUM_PAGES configuration in the board rather than the app.


kaczmarczyck (approved):
The only downside I see is that we have a limit of 8 regions, and `STORAGE_LOCATIONS` will need 5 of them. But from the logs I've seen so far, this should still be okay.
If we need to, we could make the new layout with upgrades a breaking change, and use either 16 or 32 pages, if latency allows.

ia0:
Oh I see, I didn't consider this issue that we are limited to 8 regions. But we should still be fine right?
- Region 0 is for the RAM.
- Region 1 is for the flash.
- Region 2 and 3 are for the store.
- Region 4, 5, and 6 are for the upgrade mechanism (2 partitions + 1 metadata).

kaczmarczyck:
Yes, absolutely. My test runs with upgradability definitely had open regions. Just wanted to bring it to attention.

PR #381, opened 2021-09-16
Upgrade commands

kaczmarczyck:
Implements the commands to upgrade OpenSK. While this PR is pretty large, the most (only interesting) changes are on the actual command logic in `mod.rs`. The other changes are:
- renamed the `Vendor` command to `VendorConfigure`, since we now have more than one vendor command
- added input parameter types to `command.rs`
- added output parameter types to `response.rs`

A few comments:
1. The `Upgrade` command does not come with any way to continue an interrupted upgrade. We might want to add that, or keep 
this nice and simple approach. I'd keep that discussion outside of this PR.
2. The `UpgradeInfo` could be integrated into `GetInfo` in the future, but it does not have a vendor section, so I had to write a new command. We could have used a high parameter number from `GetInfo`, hoping it will not be used in the near future, but the new command was not much more expensive in binary size.
3. The binary size increase due to this PR is around 4 kB.
4. The `UpgradeInfo` is probably not final. There is some freedom in how we implement it, and what other parties need for it to be useful. Conceptually, we need some way to identify where the upgrade is written to, and this identifier will be transmitted to the upgrade server. A valid concern with the chosen approach is that vendors could misuse it to uniquely identify users. This is detectable and therefore not too bad, but might still make sense to avoid. We could piggy-back it into the firmware version or AAGUID too. All this depends on discussions with FIDO, and is therefore subject to change. The current implementation's purpose is to work in a proof of concept.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -153,14 +153,18 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
 /// Parses the metadata of an upgrade, and checks its correctness.
 ///
 /// Returns the hash over the upgrade, including partition and some metadata.
-/// The metadata layout is:
-/// | -- 32B upgrade hash (SHA256) -- | 8B other |
+/// The metadata consists of:
+/// - 32B upgrade hash (SHA256)
+/// -  4B timestamp (little endian encoding)
+/// -  4B partition address (little endian encoding)
+/// The upgrade hash is computed over the firmware image and all metadata,
+/// except the hash itself.
 fn parse_metadata(
     upgrade_locations: &UpgradeLocations,
     metadata: &[u8],
 ) -> Result<[u8; 32], Ctap2StatusCode> {
     const METADATA_LEN: usize = 40;
-    if metadata.len() < METADATA_LEN {
+    if metadata.len() != METADATA_LEN {
         return Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER);
     }
     // The hash implementation handles this in chunks, so no memory issues.
@@ -1281,9 +1285,9 @@ where
             upgrade_locations
                 .write_metadata(&data)
                 .map_err(|_| Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)?;
-            upgrade_locations
+            &upgrade_locations
                 .read_metadata()
-                .map_err(|_| Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)?
+                .map_err(|_| Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)?[..data.len()]
         };
         let written_hash = Sha256::hash(written_slice);
         if hash != written_hash {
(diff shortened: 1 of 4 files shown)


src/ctap/mod.rs
```diff
@@ -146,6 +150,52 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
...
+///
+/// Returns the hash over the upgrade, including partition and some metadata.
+/// The metadata layout is:
+/// | -- 32B upgrade hash (SHA256) -- | 8B other |
+fn parse_metadata(
+    upgrade_locations: &UpgradeLocations,
+    metadata: &[u8],
+) -> Result<[u8; 32], Ctap2StatusCode> {
+    const METADATA_LEN: usize = 40;
+    if metadata.len() < METADATA_LEN {
```
jmichelp: "<" or "!="?
Is the metadata field valid if it's 50 bytes?

src/ctap/mod.rs
```diff
@@ -146,6 +150,52 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
     }
 }
 
+/// Parses the metadata of an upgrade, and checks its correctness.
+///
+/// Returns the hash over the upgrade, including partition and some metadata.
+/// The metadata layout is:
+/// | -- 32B upgrade hash (SHA256) -- | 8B other |
```
jmichelp: other?
Isn't there a better way to describe the purpose of these 8 bytes?

src/ctap/mod.rs
kaczmarczyck: The logic as it is now leaves some freedom to either call with a whole page, or only call with a prefix. The page will than implicitly be overwritten with a 0xFF padding. We could change that to only accept exact slices. Maybe a tighter interface with less room for clients makes it more maintainable?

src/ctap/mod.rs
kaczmarczyck: I could go into more detail, but it's basically transparent to the application. Inside these 8B are
- 4B start address so the bootloader knows the firmware matches the location
- 4B timestamp
The exact usage of this "other" shouldn't matter from an application point of view. You want more detail in the comment?

src/ctap/mod.rs
kaczmarczyck: I liked the idea of a more precise input, and implemented it. PTAL

src/ctap/mod.rs
kaczmarczyck: The full metadata content is now described here.

src/ctap/mod.rs
```diff
@@ -3162,7 +3166,7 @@ mod test {
             .to_vec();
         signed_over_data.extend(&[0xFF; 8]);
         let signed_hash = Sha256::hash(&signed_over_data);
-        let mut metadata = data.clone();
+        let mut metadata = vec![0xFF; 40];
```
jmichelp: `s/40/METADATA_LEN/`

src/ctap/mod.rs
jmichelp: Makes it easier when reading code to understand why these 8 bytes are part of the hash computation.

src/ctap/mod.rs

jmichelp (approved):

ia0 (commented):
Quickly went through it. LGTM. Don't wait for me to submit.

kaczmarczyck:
Force-pushing, as the lints will be fixed in #382 .

PR #385, opened 2021-09-24
Storage panic from oss-fuzz

kaczmarczyck:
oss-fuzz found a panic ([bug report](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39194) is private until fixed).

Fixed and added tests for this issue, on more coverage.

### DIFF
--- a/src/embedded_flash/syscall.rs
+++ b/src/embedded_flash/syscall.rs
@@ -290,7 +290,7 @@ impl SyscallUpgradeStorage {
 impl UpgradeStorage for SyscallUpgradeStorage {
     fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]> {
         if length == 0 {
-            return Ok(&[]);
+            return Err(StorageError::OutOfBounds);
         }
         let address = self.partition.start() + offset;
         if self
@@ -305,7 +305,7 @@ impl UpgradeStorage for SyscallUpgradeStorage {
 
     fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()> {
         if data.is_empty() {
-            return Ok(());
+            return Err(StorageError::OutOfBounds);
         }
         let address = self.partition.start() + offset;
         let write_range = ModRange::new(address, data.len());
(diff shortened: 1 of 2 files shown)


src/embedded_flash/syscall.rs
```diff
@@ -289,6 +289,9 @@ impl SyscallUpgradeStorage {
 
 impl UpgradeStorage for SyscallUpgradeStorage {
     fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]> {
+        if length == 0 {
```
ia0, line 292: Shouldn't we check that the offset is valid (essentially offset <= partition.len())? And then wouldn't the rest of the code just work by itself?

src/embedded_flash/syscall.rs
```diff
@@ -301,14 +304,15 @@ impl UpgradeStorage for SyscallUpgradeStorage {
     }
 
     fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()> {
+        if data.is_empty() {
```
ia0, line 307: Same here.

src/embedded_flash/syscall.rs
kaczmarczyck: I was wondering if I should ignore empty slices (my current choice) or treat them as being present at `offset`. The rest of the code just works after that initial check in both cases, right?
To be honest, both choices should do the job. I don't see why a client would read or write empty slices. The fail outside of range could be a hacky partition check, but that's harder than just asking directly.
Any advantages of one over the other?

ia0: > The rest of the code just works after that initial check in both cases, right?

I don't know. In the current case it's not executed so it doesn't matter. But if it works it makes it possible to just check the offset and still run the code.

> I don't see why a client would read or write empty slices.

Then why not return an error instead of an empty slice?

> The fail outside of range could be a hacky partition check, but that's harder than just asking directly.

I don't understand what you want to say.

> Any advantages of one over the other?

From the 3 options:
- Early exit with error
- Early exit with empty slice (no validity check)
- Check offset validity

I think "early exit with empty slice" should not be an option because it can hide bad behaviors (invalid offset returning correct result). Between the 2 other options it's a matter of whether it's valid to read an empty slice or not.

kaczmarczyck: Okay, then I'll just disallow empty slices in general, for read and write.

src/embedded_flash/syscall.rs
kaczmarczyck: Changed to error.

ia0 (approved):

PR #386, opened 2021-09-27
Fix possible missing entries after multiple consecutive interrupted overflowing compactions

ia0:
Compaction may temporarily break the invariant that the tail of the store is within the window (i.e. tail <= head + virt_size). This is not a problem while the store is powered because compactions continue until the invariant is restored. However, if a compaction is interrupted (power is lost before 100ms), only that compaction is resumed at boot time and the invariant may stay broken. This change fixes this by continuing compacting until the invariant is restored.

This change also increases the window size. The previous value was too conservative (accounting for erase entries twice).

The consequences of this bug are that entries which header lies in the last K - 1 words of storage are hidden until the next mutable operation, where K is the number of consecutive overflowing compactions (an overflowing compaction is when the compacted page is full of non-deleted entries) from which the last is interrupted. In other words, there must be at least 2 consecutive overflowing compactions to hide an empty entry (e.g. force pin change, enterprise attestation enable, always UV), and 3 to hide an entry of length at most 4 bytes (e.g. pin retries, pin length, signature counter).

### DIFF
--- a/libraries/persistent_store/fuzz/src/store.rs
+++ b/libraries/persistent_store/fuzz/src/store.rs
@@ -53,6 +53,7 @@ pub fn fuzz(mut data: &[u8], debug: bool, stats: Option<&mut Stats>) {
                 driver.check().unwrap();
             }
             if fuzzer.debug {
+                println!("{:?}", driver.model().content());
                 println!("----------------------------------------------------------------------");
             }
         }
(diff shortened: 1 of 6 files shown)


kaczmarczyck (approved):
I understand that this change is backwards compatible, right? Or is there some unlikely state that would not load after the change?

ia0:
So there are 2 changes:
1. Continuing compaction: this is a bug fix and is backward compatible because it just changes what we do during recovery and doesn't touch the storage format
2. Increase the window size: this is also backward compatible but it changes the storage format because we can now store more things. Previous valid states are included in new valid states, but not the opposite. So reverting that part of the change would break. What do you think?

kaczmarczyck:
Sounds good. Why would we want to revert it? Also, while on develop, reverting should be okay.

ia0:
Yes I agree.

> Why would we want to revert it?

Just tried to be exhaustive.

PR #387, opened 2021-09-28
New boards with layouts for dual partition setups

kaczmarczyck:
Adds the dev board layouts for a setup with a bootloader and 2 partitions.

Most of the changes in this PR are slightly changed copies of existing code. The `nrf52840_layout_*.ld` files are similar to the existing `nrf52840_layout.ld`, and the boards are similar to the non-partitioned variant, too. All changes from the existing dev board definition to the new ones are documented in the respective `README.md`.

Outside of boards and layouts, this biggest change is in the deploy script. Changes include:
- some refactoring
- paddings are optional
- new metadata writer

I'm not sure about the less redundant definitions of `OpenSKBoard`, please let me know what you think, I can revert them if you like. To have the boards running, you need to also flash the matching bootloader. Other than that, the deployment didn't change, just specify the board name.


deploy.py
```diff
@@ -205,6 +196,15 @@ def assert_python_library(module):
            f"Try to run: pip3 install {module}"))
 
 
+def create_metadata(firmware_image, partition_address):
+  timestamp = struct.pack("<I", int(time.time()))
+  partition_start = struct.pack("<I", partition_address)
+  sha256_hash = hashlib.sha256()
+  sha256_hash.update(firmware_image + timestamp + partition_start)
```
jmichelp: You could call `update()` on each part instead of concatenating in memory.

deploy.py
```diff
@@ -550,6 +576,8 @@ def verify_flashed_app(self, expected_app):
     if self.args.programmer not in ("jlink", "openocd"):
       return False
     args = copy.copy(self.tockloader_default_args)
+    board_props = SUPPORTED_BOARDS[self.args.board]
+    setattr(args, "app_address", board_props.app_address)
```
jmichelp: Isn't it part of the default arguments with this PR?

jmichelp (commented):
That's a lot of code being duplicated.
I wonder if we couldn't do better by passing the arguments through env variables and tweak the `build.rs` and the `main.rs` to go back to fewer boards (devkit, dongle, mdk).

deploy.py

deploy.py
kaczmarczyck: True, I forgot to remove its occurence here.

kaczmarczyck:
I reused the `src/` from `nrf52840dk_opensk`. Now all 3 dev board definitions load the `STORAGE_LOCATIONS` from a generated `locations.rs`. I'm not sure about the benefit of completely removing board definitions. The cleanest solution for 3 different layouts should be 3 different `layout.ld` files. WDYT?

deploy.py
```diff
@@ -50,12 +53,18 @@
...
         # Flash address at which the kernel will be written
         "kernel_address",
-        # Set to None is padding is not required for the board.
+        # Set to None if padding is not required for the board.
         # This creates a fake Tock OS application that starts at the
         # address specified by this parameter (must match the `prog` value
         # specified on the board's `layout.ld` file) and will end at
         # `app_address`.
         "padding_address",
+        # If present, enforce that the firmware image equals this value.
```
jmichelp: nit: add "padding it with `0xff` bytes"

deploy.py
```diff
@@ -205,6 +196,17 @@ def assert_python_library(module):
            f"Try to run: pip3 install {module}"))
 
 
+def create_metadata(firmware_image, partition_address):
+  timestamp = struct.pack("<I", int(time.time()))
```
jmichelp: To avoid issues with timezones, maybe it's better to use UTC and therefore `datetime` instead of `time`.
`int(datetime.datetime.utcnow().timestamp())`

deploy.py
kaczmarczyck: Added!

deploy.py

jmichelp (approved):

PR #388, opened 2021-10-01
Introduce window size and use it to do #386 correctly

ia0:
After thinking about #386, I've found out that extending the virtual size is wrong and there's a better fix than continuing the compactions during recovery.

This PR is split in 3 commits:
1. Adds a test showing the issue with the extended virtual size.
2. Reverts the 3rd commit of #386 (the only non test-only commit).
3. Introduces the window size (with the same value as the extended virtual size of #386) and keeps the virtual size the same. Also updates the proofs to mention the window size. The proofs are essentially the same though. The only change compared to before #386 is that when reading the store content, we use the window size instead of the virtual size which means we can see the entries that would have been hidden otherwise.

As such this PR is much easily reviewed by looking at commits separately.

### DIFF
--- a/libraries/persistent_store/tests/store.rs
+++ b/libraries/persistent_store/tests/store.rs
@@ -46,7 +46,10 @@ fn interrupted_overflowing_compaction() {
             StoreInterruption::pure(1),
         ) {
             Ok((None, d)) => driver = d.power_on().unwrap(),
-            _ => unreachable!(),
+            _ => {
+                assert!(false);
+                return;
+            }
         }
     }
 }
@@ -114,7 +117,10 @@ fn full_compaction_with_max_prefix() {
             StoreInterruption::pure(1),
         ) {
             Ok((None, d)) => driver = d.power_on().unwrap(),
-            _ => unreachable!(),
+            _ => {
+                assert!(false);
+                return;
+            }
         }
     }
     check_lifetime(&mut driver, c + n - 1);
--- a/libraries/persistent_store/tests/store.rs
+++ b/libraries/persistent_store/tests/store.rs
@@ -1,7 +1,10 @@
-use persistent_store::{BufferOptions, StoreDriverOff, StoreInterruption, StoreOperation};
+use persistent_store::{
+    BufferOptions, StoreDriverOff, StoreDriverOn, StoreInterruption, StoreOperation,
+};
 
 #[test]
 fn interrupted_overflowing_compaction() {
+    let num_pages = 7;
     let options = BufferOptions {
         word_size: 4,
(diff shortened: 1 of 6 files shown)


libraries/persistent_store/tests/store.rs
```diff
@@ -39,13 +41,81 @@ fn interrupted_overflowing_compaction() {
...
-            delay: 0,
-            corrupt: Box::new(|old, new| old.copy_from_slice(new)),
-        };
-        match driver.partial_apply(StoreOperation::Prepare { length: 1 }, interruption) {
+        match driver.partial_apply(
+            StoreOperation::Prepare { length: 1 },
+            StoreInterruption::pure(1),
+        ) {
+            Ok((None, d)) => driver = d.power_on().unwrap(),
+            _ => unreachable!(),
```
kaczmarczyck: Is this `unreachable` the test, i.e. we test whether it is indeed unreachable? If so, wouldn't an `assert!(false)` be more descriptive?

kaczmarczyck (dismissed):

libraries/persistent_store/tests/store.rs
ia0: Yes this is only in the test. I used `assert!(false)` initially but the compiler doesn't realize that the control flow always stops as `assert!(false)` so it complains that `driver` is not defined when taking that branch. But I guess a solution would be to have `{ assert!(false); return; }`. What do you think?

kaczmarczyck: Interesting. I like that a bit more. Reading the code, I was wondering: What is this even testing? At least me personally, I parse for `assert` variants.

kaczmarczyck (approved):

libraries/persistent_store/tests/store.rs
ia0: Ah I see. So no, this test is not testing that we don't reach this statement. It's testing that there's no panic which are in many places (including outside the test). The most important ones are those in the loop, i.e. the unwrap on the line above (which was the one failing before the bug) and the assert false on this line (which encompasses 2 failure types, unwrapping the partial apply and making sure that when it's Ok there's still no error).

kaczmarczyck: So the `assert` is still more precise than the `unreachable` right? I'm happy as is, and I don't think we have to document which ones are the "important" unwraps. Therefore resolving.

ia0: It depends if we associate an additional semantics to assert than unreachable. They should be the same from compiler point of view, but it's true that in tests we are more used to see assert, so it might be less surprising.

PR #390, opened 2021-10-07
Remove useless retry-loop

ia0:
This was needed in the past although we would never entered the loop (calling unwrap directly would fail though). Not sure if the issue was on the kernel or the compiler.


kaczmarczyck (approved):
Strange, but good thing it is fixed.

PR #392, opened 2021-10-25
erase unused flash pages so hashes matches

kaczmarczyck:
The bootloader failed when unused firmware page were not erased. This can happen when firmware images become smaller or spurious data is still on flash.

The PR generalizes the `clear_stoarge` function to take address and size, and uses that before writing kernel and app to make sure the setup is correct.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -29,6 +29,7 @@ import shutil
 import struct
 import subprocess
 import sys
+from typing import Dict, List, Tuple
 
 import colorama
 from six.moves import input
@@ -159,7 +160,7 @@ SUPPORTED_BOARDS = {
 APP_HEAP_SIZE = 90000
 
 
-def get_supported_boards():
+def get_supported_boards() -> Tuple[str]:
   """Returns a tuple all valid supported boards."""
   boards = []
   for name, props in SUPPORTED_BOARDS.items():
@@ -169,28 +170,28 @@ def get_supported_boards():
   return tuple(set(boards))
 
 
-def fatal(msg):
+def fatal(msg: str):
   print(f"{colorama.Fore.RED + colorama.Style.BRIGHT}fatal:"
         f"{colorama.Style.RESET_ALL} {msg}")
   sys.exit(1)
 
 
-def error(msg):
+def error(msg: str):
   print(f"{colorama.Fore.RED}error:{colorama.Style.RESET_ALL} {msg}")
 
 
-def info(msg):
+def info(msg: str):
   print(f"{colorama.Fore.GREEN + colorama.Style.BRIGHT}info:"
         f"{colorama.Style.RESET_ALL} {msg}")
 
 
-def assert_mandatory_binary(binary):
-  if not shutil.which(binary):
-    fatal((f"Couldn't find {binary} binary. Make sure it is installed and "
+def assert_mandatory_binary(binary_name: str):
+  if not shutil.which(binary_name):
+    fatal((f"Couldn't find {binary_name} binary. Make sure it is installed and "
            "that your PATH is set correctly."))
 
 
-def assert_python_library(module):
+def assert_python_library(module: str):
   try:
     __import__(module)
   except ModuleNotFoundError:
@@ -198,7 +199,7 @@ def assert_python_library(module):
            f"Try to run: pip3 install {module}"))
 
 
-def create_metadata(firmware_image, partition_address):
+def create_metadata(firmware_image: bytes, partition_address: int) -> bytes:
   """Creates the matching metadata for the given firmware.
 
   The metadata consists of a timestamp, the expected address and a hash of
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -487,6 +487,24 @@ def write_binary(self, binary, address):
     except TockLoaderException as e:
       fatal(f"Couldn't write binary: {str(e)}")
 
+  def clear_flash(self, address, size):
+    if self.args.programmer == "none":
+      return 0
+    info("Erasing the persistent storage")
```
jmichelp: "Erasing flash pages" maybe?

deploy.py
```diff
@@ -510,6 +528,11 @@ def install_padding(self):
     info("Flashing padding application")
     self.write_binary(self.get_padding(), board_props.padding_address)
 
+  def clear_unused_pages(self, padding_address, padding_size, page_size):
+    cleared_size = padding_size - (padding_size % page_size)
```
jmichelp: Maybe we should start adding some documentation because `deploy.py` keeps growing.
Here I don't quite follow the logic, especially around `clear_address` computation.

deploy.py
```diff
@@ -530,13 +556,21 @@ def pad_to(binary, length):
...
     firmware_image = pad_to(kernel, kernel_size) + pad_to(app, app_size)
 
-    metadata = create_metadata(firmware_image, board_props.kernel_address)
+    # Clear unused pages so their content matches the expected padding.
+    kernel_pad_address = kernel_address + len(kernel)
+    kernel_pad_size = kernel_size - len(kernel)
+    self.clear_unused_pages(kernel_pad_address, kernel_pad_size, page_size)
+    app_pad_address = app_address + len(app)
+    app_pad_size = app_size - len(app)
+    self.clear_unused_pages(app_pad_address, app_pad_size, page_size)
```
ia0: Did you consider padding kernel and app instead? The way I understand it, there's a problem because if we flash a smaller kernel/app, then we don't erase the now-unused pages. The suggestion is to always flash a kernel/app of the same size (the whole region allocated to the kernel/app) and manually pad with `0xff`. For example old kernel goes from 0x1000 to 0x3000 and new kernel from 0x1000 to 0x2800, then we would pad the new kernel with 0xff from 0x2800 to 0x3000.

kaczmarczyck: I was considering that. In fact, we are generating the exact `firmware_image` inside `install_metadata`. We could simply call `write_binary` with it. A firmware update over CTAP will rely on this kind of firmware image generation.

I could do a major overhaul of deploy to always generate firmware images first, and do one full write. It should also work for examples. That could reduce deploy complexity.

The current approach uses `tockloader`s API more naturally (`install`), and might therefore:
- age better on non-upgradable boards with new `tockloader` versions
- generalize better to multi-app scenarios (again, only without upgradability)

@jmichelp also your thoughts?

jmichelp: I'm not sure I understand the link between Julien's suggestion and Fabian's answer 😅 

Julien suggest we always pad the produced binaries (either kernel or app) to their "max" size.
Fabian seems to suggest we produce a full binary containing {kernel, kernel_padding, app, app_padding}.

Am I missing something?

kaczmarczyck: Maybe I read too much into that comment, then. :)

deploy.py
kaczmarczyck: Reverted this change, therefore obsolete.

deploy.py
kaczmarczyck: This function was removed, but I added docstrings throughout deploy nonetheless!

deploy.py
kaczmarczyck: I rewrote the fix, so that we now pad the kernel to the correct size right after reading it from disk. That removes the erase complexity. It does rely on elf2tab correctly padding the app though (which should be the case if we don't reduce the binary size by 50%).

kaczmarczyck:
Also added type hints to satisfy pylint. A follow-up PR will also reduce some complexity in `deploy.py` by moving some metadata logic to the upgrade tool.

deploy.py
```diff
@@ -534,7 +612,8 @@ def pad_to(binary, length):
 
     kernel_size = board_props.app_address - board_props.kernel_address
```
ia0, line 613: Unrelated to this PR, but here we assume there's no padding application between kernel and app, while in `read_kernel` we do. It's probably expected, but just mentioning in case.

ia0 (commented):
Looks good from my side, but I skipped the type annotations and doc-strings. I just looked at the code.

deploy.py
```diff
@@ -300,6 +353,7 @@ def checked_command_output(self, cmd, env=None, cwd=None):
     return cmd_output.decode()
 
   def update_rustc_if_needed(self):
+    """Updates the Rust and install the necessary target toolchain."""
```
jmichelp: nit: installs

deploy.py
```diff
@@ -576,7 +657,8 @@ def clear_storage(self):
     fatal(f"Programmer {self.args.programmer} is not supported.")
 
   # pylint: disable=protected-access
-  def verify_flashed_app(self, expected_app):
+  def verify_flashed_app(self, expected_app: str) -> bool:
+    """Uses Tockloader check if an app of the expected name was written."""
```
jmichelp: nit: Uses Tockloader to check

jmichelp (dismissed):

jmichelp (approved):

PR #400, opened 2021-11-02
Upgrade tooling

kaczmarczyck:
This PR provides a script to upgrade an OpenSK over CTAP that was deployed with upgradability.

A typical series of commands would be:
```
./deploy.py --opensk --board=nrf52840dk_opensk_a
[...time passes, changes are made...]
./tools/perform_upgrade.sh nrf52840dk_opensk_b
```

This flow currently breaks if you reset between these 2 commands, since reset deletes the `crypto_data` that contains the upgrade private key necessary to sign the upgrade (as it should).

I'll update the README and documentation with the follow-up PR introducing the bootloader. You can't test this PR without flashing the bootloader first.

### DIFF
--- a/tools/perform_upgrade.sh
+++ b/tools/perform_upgrade.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Creates a signature key and configures the public key.
+# The device will not be locked down for testing purposes.
+# Generates the binary and upgrades OpenSK.
+# To be run from the OpenSK base path.
+
+set -e
+
+BOARD="$1"
+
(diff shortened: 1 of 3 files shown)


tools/perform_upgrade.sh
```diff
@@ -0,0 +1,27 @@
...
+# Generates the binary and upgrades OpenSK.
+# To be run from the OpenSK base path.
+
+set -e
+
+BOARD="$1"
+
+./deploy.py --board=$BOARD --opensk --programmer=none
+python -m tools.deploy_partition --board=$BOARD
+nrfjprog --reset --family NRF52
```
jmichelp: This should only work with the devkit because `nrfjprog` implies using a JLink debugger IIRC.

tools/perform_upgrade.sh
```diff
@@ -0,0 +1,27 @@
...
+# The device will not be locked down for testing purposes.
+# Generates the binary and upgrades OpenSK.
+# To be run from the OpenSK base path.
+
+set -e
+
+BOARD="$1"
+
+./deploy.py --board=$BOARD --opensk --programmer=none
+python -m tools.deploy_partition --board=$BOARD
```
jmichelp: nit: python3

tools/perform_upgrade.sh
kaczmarczyck: In the beginning, upgradability will only be supported on the devkit as a PoC, before we decide how to proceed with the other bootloaders.
I changed it to try `nrfjprog` for user comfort for those who have it installed, and ask for replug in case of failure.

tools/perform_upgrade.sh
kaczmarczyck: Done

jmichelp (approved):

PR #403, opened 2021-11-04
Make maintainers/patches more portable

ia0:
Some tools behave differently between mac and linux. This PR tries to use those tools in the intersection of where those tools behave similarly between mac and linux.

This is only tested on linux. Note that we currently don't roundtrip because the 3rd patch (additional boards) has some index difference.

### DIFF
--- a/maintainers/patches
+++ b/maintainers/patches
@@ -20,7 +20,7 @@ commit() {
 }
 
 get_root() {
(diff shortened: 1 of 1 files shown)


jmichelp (approved):

jmichelp:
Confirmed it works on mac

ia0:
Cool, thanks!

PR #404, opened 2021-11-05
Bootloader infrastructure and main logic

kaczmarczyck:
This is the first PR containing the bootloader code. Compared to the [full code](https://github.com/kaczmarczyck/OpenSK/tree/bootloader), this PR misses the SHA256 computation using the Cryptocell.

It includes the full setup and infrastructure, including:
- desktop checks
- workflow checks
- docs updates
- linker script
- deploy code
- main logic

The bootloader directory is self-sufficient, it's basically independent from OpenSK. Therefore, I decided to give it its own `rust-toolchain` file. They currently run with the same toolchain for convenience, but an OpenSK toolchain upgrade should not necessarily burden is with also upgrading the bootloader toolchain (even though we can, of course).

- [x] Tests pass

### DIFF
--- a/bootloader/.cargo/config
+++ b/bootloader/.cargo/config
@@ -1,7 +1,2 @@
 [target.thumbv7em-none-eabi]
 linker = "arm-none-eabi-gcc"
-
-rustflags = [
-  "-C", "link-arg=-Wl,-Tlink.x",
-  "-C", "link-arg=-nostartfiles",
-]
--- a/bootloader/.cargo/config
+++ b/bootloader/.cargo/config
@@ -1,0 +1,7 @@
+[target.thumbv7em-none-eabi]
+linker = "arm-none-eabi-gcc"
+
+rustflags = [
+  "-C", "link-arg=-Wl,-Tlink.x",
+  "-C", "link-arg=-nostartfiles",
+]
--- a/bootloader/README.md
+++ b/bootloader/README.md
@@ -1,0 +1,35 @@
+# OpenSK Bootloader
+
+This bootloader supports upgradability for OpenSK. Its functionality is to
+
+-   check images on A/B partitions,
+-   boot the most recent valid partition.
+
+## How to use
+
+The bootloader is built and deployed by OpenSK's `deploy.py`. If your board
+defines a metadata address, it is detected as an upgradable board and this
+bootloader is flashed to memory address 0.
+
+## How to debug
+
+The bootloader prints debug message over RTT when compiled in debug mode. Using
+`nrfjprog` for flashing and inspecting memory is recommended for debugging.
+
+```shell
+cargo build --target thumbv7em-none-eabi
+llvm-objcopy -O ihex target/thumbv7em-none-eabi/debug/bootloader \
+    target/thumbv7em-none-eabi/debug/bootloader.hex
+nrfjprog --program target/thumbv7em-none-eabi/debug/bootloader.hex \
+    --sectoranduicrerase -f nrf52 --reset
+```
+
+To read the debug messages, open two terminals for:
+
+```shell
+JLinkRTTLogger -device NRF52840_XXAA -if swd -speed 1000 -RTTchannel 0
+JLinkRTTClient
+```
+
+The first command also logs the output to a file. The second shows all output in
+real time.
--- a/bootloader/README.md
+++ b/bootloader/README.md
@@ -17,11 +17,12 @@ The bootloader prints debug message over RTT when compiled in debug mode. Using
 `nrfjprog` for flashing and inspecting memory is recommended for debugging.
 
 ```shell
-cargo build --target thumbv7em-none-eabi
+RUSTFLAGS="-C link-arg=-Wl,-Tlink.x -C link-arg=-nostartfiles" \
+    cargo build --target thumbv7em-none-eabi
 llvm-objcopy -O ihex target/thumbv7em-none-eabi/debug/bootloader \
     target/thumbv7em-none-eabi/debug/bootloader.hex
 nrfjprog --program target/thumbv7em-none-eabi/debug/bootloader.hex \
-    --sectoranduicrerase -f nrf52 --reset
+    --sectorerase -f nrf52 --reset
 ```
 
 To read the debug messages, open two terminals for:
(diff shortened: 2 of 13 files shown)


bootloader/.cargo/config
```diff
@@ -0,0 +1,7 @@
+[target.thumbv7em-none-eabi]
+linker = "arm-none-eabi-gcc"
+
+rustflags = [
+  "-C", "link-arg=-Wl,-Tlink.x",
+  "-C", "link-arg=-nostartfiles",
+]
```
jmichelp: You probably want to reuse the flag (`--remap-path-prefix`) we have in Tock to ensure reproducibility in the build. Otherwise the full path is being used.
Not a breaker but good for consistency :)

bootloader/rust-toolchain
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "nightly-2021-03-25"
```
jmichelp, line 2: if this isn't the same toolchain that we use for the firmware, be careful that it requires a "hack" in the github workflow to force its installation.

docs/install.md
```diff
@@ -54,6 +54,7 @@ following:
 *   python3 and pip (can be installed with the `python3-pip` package on Debian)
 *   the OpenSSL command line tool (can be installed and configured with the
     `libssl-dev` and `pkg-config` packages on Debian)
+*   `gcc-arm-none-eabi` from Debian packages if you want an upgradable board
```
jmichelp: why not using LLVM tools which are part of installed rust toolchain?
That's what Tock does.

bootloader/.cargo/config
kaczmarczyck: Sounds good. I introduced the flags in deploy, consistent to the other compilations.

docs/install.md
kaczmarczyck: For me, it looks like I always need to install something, either

1. `apt install gcc-arm-none-eabi`
2. `apt install llvm`
3. `cargo install cargo-binutils && rustup component add llvm-tools` (for calling `cargo objcopy --`)

In Tock, I find usage examples of options 1 and 2. So maybe the change in the latest commit is better?

bootloader/rust-toolchain
kaczmarczyck: I'd leave this to my future self to decide if I want to also upgrade the bootloader toolchain or change the workflows instead. The breaking workflows will be pretty hard to miss, so I think this can be left as is?

kaczmarczyck:
I'm waiting for the documentation PR to #408 land, and then modify this PR to add its documentation to the right places.

kaczmarczyck:
Rebase on top of the documentation changes, and added more context.

kaczmarczyck:
Maybe the documentation should live in `customization.md`. Or link to the bootloader README. I'm taking opinions on that!

bootloader/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The bootloader prints debug message over RTT when compiled in debug mode. Using
+`nrfjprog` for flashing and inspecting memory is recommended for debugging.
+
+```shell
+cargo build --target thumbv7em-none-eabi
+llvm-objcopy -O ihex target/thumbv7em-none-eabi/debug/bootloader \
+    target/thumbv7em-none-eabi/debug/bootloader.hex
+nrfjprog --program target/thumbv7em-none-eabi/debug/bootloader.hex \
+    --sectoranduicrerase -f nrf52 --reset
```
jmichelp: Is it really needed to erase UICR here?
Tock kernel may erase it again if the registers aren't set to what it expects according to the selected board.

deploy.py
```diff
@@ -384,6 +384,39 @@ def build_opensk(self):
...
+    info("Building bootloader")
+    rust_flags = [
+        f"--remap-path-prefix={os.getcwd()}=",
+        "-C",
+        "link-arg=-Wl,-Tlink.x",
+        "-C",
+        "link-arg=-nostartfiles",
+    ]
+    env = os.environ.copy()
+    env["RUSTFLAGS"] = " ".join(rust_flags)
```
jmichelp, line 399: I suggest to remove the flags from the `bootloader/.cargo/config` file to ensure they are only defined once.

docs/boards/nrf52840dk.md
```diff
@@ -48,3 +48,25 @@ There are 3 switches that need to be in the correct position:
...
+
+Afterwards, you can upgrade the other partition with
+
+```shell
+./tools/perform_upgrade.sh nrf52840dk_opensk_b
+./tools/perform_upgrade.sh nrf52840dk_opensk_a
+```
+
+respectively. You can only upgrade the partiion that is not currently running,
+so always alternative your calls to `perform_upgrade.sh`. Otherwise, this script
```
jmichelp: typo: alternate

docs/install.md
```diff
@@ -149,3 +150,23 @@ If your board is already flashed with Tock OS, you may skip installing it:
...
+We experiment with a new CTAP command to allow upgrading your device without
+access to its debugging port. For that purpose, the flash storage is split into
+4 parts:
+
+*   the bootloader to decide with partition to boot
+*   firmware partition A
+*   firmware partition B
+*   the persistent storage for credentials
+
+The storage is backwards compatible to non-upgradble boards. Deploying an
```
jmichelp: typos: backward and non-upgradable

docs/install.md
```diff
@@ -149,3 +150,23 @@ If your board is already flashed with Tock OS, you may skip installing it:
...
+upgradble board automatically installs the bootloader. Please keep in mind that
```
jmichelp: typo: upgradable

kaczmarczyck: Done

docs/install.md
kaczmarczyck: Done

docs/boards/nrf52840dk.md
kaczmarczyck: Done

deploy.py
kaczmarczyck: Good point. README is updated to match, also setting the RUSTFLAGS manually.

bootloader/README.md
kaczmarczyck: It's not, that was just a copy-paste. Changed to `--sectorerase`.

jmichelp (approved):

PR #408, opened 2021-11-15
New documentation for develop

kaczmarczyck:
This PR reworks the documentation. After adding lots of features and getting feedback from users, I thought it is time to make it simpler to navigate the project - for new developers, as well as returning users. It is probably easiest to review by looking at the [rendered markdown](https://github.com/kaczmarczyck/OpenSK/tree/documentation).

This PR working towards closing a few issues like #406 and #402. For follow up, some questions are still not answered, like how to use the advanced programmers. Finally, a modified version of the develop documentation should be going to stable for visibility.

PR #410, opened 2021-11-18
Silent certificate checks

kaczmarczyck:
This is a new feature supporting #408. I prints the configuration command when the deploy script detects that a certificate is missing.

To achieve that, it allows configuration to not require user touch, if it is used as read-only, like GetInfo. This should not hurt security or privacy guarantees, since the command will only be available locally. The same information could potentially be retrieved some other way, like with a silent U2F make call. The proposed way is cleaner though.

To be merged after #408, since it references its documentation.


ia0 (approved):

jmichelp (approved):
As a general comment: considering how the python scripts are getting more complicated over time, it might be time to consider refactoring things in a future PR. Maybe have a `py` directory with classes/functions that `deploy.py` and `configure.py` could simply import and use. We could even think about adding unit tests.

kaczmarczyck:
Certainly correct. The run method is a bit of a pain already.

Issue #406, opened 2021-11-10
Reset Security Key in Nordic nRF52840-dongle?


Lostsite:
1. Is there any way to remove all the stored data inside the Security Key (after successfully installed) without using JTAG or other devices connected? (Basically, I just want to know Can I reset the Security Key in Windows or in the Browser? (I can't reset it in windows, basically, it ask me to push the pushbuttons twice on the device, and that won't work).

2. Is there any way to erase all the data in Nordic nRF52840-dongle and reset it into the Factory Mode (set the storage like the new device I get from the factory).

3. If 2. is possible, is there any way I can do it through USB connections (without JTAG) (would `--programmer=nordicdfu` order still work?)

Lostsite:
Nevermind, the issue kind already been discribed in issue #179

kaczmarczyck:
I'm reopening this for myself, there seems to be a problem with our documentation. No action required from you. Thanks for exploring OpenSK!

Issue #402, opened 2021-11-04
Please update the Installation guide document about the Nordic nRF52840 Dongle no longer need a Segger J-Link JTAG probe


Lostsite:
Based on issue #47 , OpenSK now can directly write through the USB port, so, please remove the [Installation guide] (https://github.com/google/OpenSK/blob/stable/docs/install.md) JTAG probe connection part. 

>Why I don't do this by myself and add the pull request?

>Because I haven't tested the USB port, and don't know if the JTAG probe gives the extra functions or would the USB port always function as expected, so I send out this issue to see if anyone could run a full test and remove the JTAG probe connection part.

kaczmarczyck:
Hi! I have a few ideas about improving our documentation. I'll keep this until then, and thanks for your interest!

Issue #349, opened 2021-07-23
Lack of information in the Makerdiary guide, it is not straight forward 


epita-cs2:
## Expected Behavior
We would like to add our github repository for Makerdiary, that we feel it is straight forward and include all the needed information for the users.

## Actual Behavior
OpenSK repository work for Makerdiary usb dongle but we have to get the python code from Makerdiary repository. And their guide it is not straight forward and lacks information. 

## Steps to Reproduce the Problem

1. if possible to add our Github installation guide. https://github.com/epita-cs2/Implementation-OpenSK-on-Makerdiary-nrf52840-mdk-usb-dongle 
1. 
1.

## Specifications

- Version:
- Platform:

kaczmarczyck:
Due to inactivity on #347, I integrated some improvements into our documentation update. Thanks for pointing out the flaws!

Issue #312, opened 2021-05-02
What is the best method to deploy OpenSK on makerdiary nRF52840 USB dongle without loosing boot loader?


alishir:
Hello,
As you know makerdiary nRF52840 USB dongle ships with UF2 boot loader that enables deploying/upgrading new firmware without using any programmer. But according to the pull request #186 deploying current version of OpenSK will erase the boot loader. So my question is what is the best method to deploy the OpenSK without loosing boot loader.

I know that this approach have some side effects and security risks, but currently I want to do some development and do some tests.

jmichelp:
It should be safe to flash the dongle over DFU. I've done it and it works if the dongle is new.
The risk is if another firmware was previously used on the dongle and was using a different configuration than OpenSK. Then it may result is erasing the 2 registers.

The bootloader (both the Nordic one and the UF2 one) use code at the beginning of the flash called MBR and this is safe with regards to TockOS/OpenSK. But they also use storage for settings in the flash, which OpenSK doesn't know about. And after using the dongle for a while, it might be possible that this settings area got overwritten by OpenSK. That's why the PR isn't merged yet. We need to analyze the bootloader, the settings, and the layout of the flash to find a reliable solution that removes that risk and ensures that OpenSK firmware can be safely upgraded.

coelner:
To understand that right:
`./deploy.py --board=nrf52840_dongle_dfu --opensk --programmer=none` will generate a DFU compatible hex file, which can be flashed with the nrf programmer? 
In my case this works so far (for a regular dongle). The `_dfu` board selection differs at the kernel address, this is needed because of what? 

The programmer `nordicdfu` currently does not work because python3.9 is incompatible. Is this line needed to build dfu packages which can be used by nrfutil instead of nrf connect and the corresponding programmer?
`./nrfutil-linux pkg generate --hw-version=52 --sd-req=0 --application-version=1 --application=/home/user/OpenSK/target/nrf52840_dongle_dfu_merged.hex /home/user/OpenSK/target/nrf52840_dongle_dfu_merged.zip`

jmichelp:
> To understand that right:
> `./deploy.py --board=nrf52840_dongle_dfu --opensk --programmer=none` will generate a DFU compatible hex file, which can be flashed with the nrf programmer?

This will generate a IntelHEX file. Which then can be turned into a DFU package using Nordic tools


> In my case this works so far (for a regular dongle). The `_dfu` board selection differs at the kernel address, this is needed because of what?

[...]

PR #413, opened 2021-11-19
Ported documentation

kaczmarczyck:
This PR ports the documentation from #408 (plus the minor fix from #412) to stable. You can review it in 2 different ways:

- Read both commits together to see the end result.
- Read the second commit, that contains the difference compared to develop.

The first commit is a simple copy from develop. The diff that is up for review is much shorter this way, if you trust that the documentation on develop is better now :)

When bugfix is merged into stable, this PR will fix a decent number of open issues: #406, #402, #349, #312, #300 plus potentially #289 + #179 + #86


docs/install.md
```diff
@@ -38,48 +38,16 @@ haven't tested them on Windows and other platforms.
 
 ### Compiling the firmware
 
-If this is your first time installing OpenSK, please skip directly to
-[Initial setup](#Initial-setup). Else, see
-[Updating your setup](#Updating-your-setup) below.
+If you are switching branches or used an old version of OpenSK before, we have
+tools to help you migrate on our develop branch. You find more information on
```
jmichelp: isn't it "migrate to"?

kaczmarczyck: English is hard.

jmichelp (approved):

PR #415, opened 2021-11-22
Do not set the stack size outside prod

ia0:
MacOS apparently doesn't like it. For some reason Linux doesn't complain (it's probably deleted as dead code before it becomes a problem).

### DIFF
--- a/maintainers/patches
+++ b/maintainers/patches
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -e
 
+shopt -s nullglob
+
 PROGRAM="$0"
+MODULES=(tock libtock-rs)
 
 success() {
   echo -e "\e[1;32mDone:\e[m $1"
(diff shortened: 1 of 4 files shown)


ia0:
This PR needs testing on MacOS (both to make sure the patches script is portable and that the compilation issue is fixed).

```
git checkout ia0/test_stack
./maintainers/patches apply
./maintainers/patches save
./maintainers/patches restore
git status  # should be clean
cargo test --features=std  # should compile and work
```

To also make sure there is no issue with the MacOS cross-compilation toolchain, we need to test that we can build for prod:

```
./deploy.py --board=nrf52840dk_opensk --opensk --programmer=none
```

kaczmarczyck (approved):
Approval conditioned on successfully testing on Mac.

Do the scripts work if some module does not have any patches?

ia0:
> Do the scripts work if some module does not have any patches?

I've tested the following scenarios:
- Round-trip (apply, save, restore) when no patch
- Start from no patch, apply, create commit, save
- Start from one patch, apply, delete commit, save

They all work as expected on Linux.

dianamin:
it works on mac. 😄

PR #416, opened 2021-11-23
Source-based coverage

kaczmarczyck:
The coveralls workflow used gcov-based coverage, but the produced information was not precise enough to help improve coverage:

> However, since LLVM doesn’t know exactly how Rust code is structured, there’s a lot lost in the translation between Rust source and LLVM IR. [1]

Instead, we switch to source-based coverage. The generation and upload to coveralls is tested locally and on my branch [2]. Locally produced html looks reasonable and should appear on [3].

I did not include `examples/` in the report. They are our own code, but probably not meant to be tested.

[1] https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html
[2] https://coveralls.io/builds/44481040
[3] https://coveralls.io/github/google/OpenSK?branch=develop

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -39,7 +39,7 @@ jobs:
           RUSTFLAGS: "-Zinstrument-coverage"
           LLVM_PROFILE_FILE: "opensk-%p-%m.profraw"
       - name: Run grcov
-        run: grcov . --binary-path ./target/debug/ -s . -t lcov --ignore-not-existing -o ./lcov.info --ignore "/*" --ignore "examples/*" --ignore "third_party/*"
+        run: grcov . --binary-path ./target/debug/ --source-dir . --output-type lcov --ignore-not-existing ---output-path ./lcov.info --ignore "/*" --ignore "examples/*" --ignore "third_party/*"
       - uses: coverallsapp/github-action@1.1.3
         name: upload report to coveralls
         with:
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -38,8 +38,12 @@ jobs:
         env:
           RUSTFLAGS: "-Zinstrument-coverage"
           LLVM_PROFILE_FILE: "opensk-%p-%m.profraw"
+      - name: Branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: branch_name
       - name: Run grcov
-        run: grcov . --binary-path ./target/debug/ --source-dir . --output-type lcov --ignore-not-existing ---output-path ./lcov.info --ignore "/*" --ignore "examples/*" --ignore "third_party/*"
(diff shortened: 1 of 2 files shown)


.github/workflows/coveralls.yml
```diff
@@ -34,14 +36,13 @@ jobs:
...
           args: --features "with_ctap1,with_nfc,std" --no-fail-fast
         env:
-          CARGO_INCREMENTAL: '0'
-          RUSTFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Coverflow-checks=off -Cpanic=abort -Zpanic_abort_tests'
-          RUSTDOCFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Coverflow-checks=off -Cpanic=abort -Zpanic_abort_tests'
-      - uses: actions-rs/grcov@v0.1.5
-        id: coverage
+          RUSTFLAGS: "-Zinstrument-coverage"
+          LLVM_PROFILE_FILE: "opensk-%p-%m.profraw"
+      - name: Run grcov
```
jmichelp, line 41: You should be able to keep the `action-rs/grcov`. The parameters that are passed to the binary are stored in [.github/actions-rs/grcov.yml](https://github.com/google/OpenSK/blob/develop/.github/actions-rs/grcov.yml) file.

Otherwise please use long options. Short ones are meant for command line and may change between 2 releases. Long options are meant for scripts and should remain stable over time. And in this case, it might be worth looking if setting the `--vcs-branch` parameter is required to avoid the reports being mixed up by coveralls (might require some Github actions trickery to extract this piece of information though)

.github/workflows/coveralls.yml
```diff
@@ -26,22 +26,23 @@ jobs:
...
-          RUSTDOCFLAGS: '-Zprofile -Ccodegen-units=1 -Cinline-threshold=0 -Coverflow-checks=off -Cpanic=abort -Zpanic_abort_tests'
+          RUSTFLAGS: "-Zinstrument-coverage"
+          LLVM_PROFILE_FILE: "opensk-%p-%m.profraw"
       - uses: actions-rs/grcov@v0.1.5
         id: coverage
       - uses: coverallsapp/github-action@1.1.3
         name: upload report to coveralls
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          path-to-lcov: ${{ steps.coverage.outputs.report }}
```
jmichelp, line 46: This should be reverted too if you use the github action.

kaczmarczyck: No, that was just a test run to confirm that the GitHub Action only works for the gcov based flow. The run failed, and the [documentation](https://github.com/marketplace/actions/rust-grcov) only mentions the former method. The commit now simply expands the parameter names.

.github/workflows/coveralls.yml
kaczmarczyck: Seems that `--vcs-branch` is not necessary. Long options are in. The `grcov` action doesn't seem to work for our use case.

jmichelp (approved):

PR #417, opened 2021-11-26
Add CIFuzz integration

DavidKorczynski:
Signed-off-by: David Korczynski <david@adalogics.com>

This adds CI integration with CIFuzz. CIFuzz will run the OpenSK OSS-Fuzz fuzzers for a short period of time for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process.

PR #418, opened 2021-11-26
cifuzz: add integration

DavidKorczynski:
Signed-off-by: David Korczynski <david@adalogics.com>

This adds CI integration with CIFuzz. CIFuzz will run the OpenSK OSS-Fuzz fuzzers for a short period of time for each PR. The benefit is that regressions and bugs has a chance of being captured early in the development process.

Referencing: https://github.com/google/OpenSK/pull/417#issuecomment-979889594

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,28 @@
+name: CIFuzz
+on: 
+  pull_request:
+    branches:
+      - develop
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'opensk'
+        dry-run: false
+        language: rust
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'opensk'
+        fuzz-seconds: 600
+        dry-run: false
+    - name: Upload Crash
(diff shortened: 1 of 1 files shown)


jmichelp:
According to what I see in the logs, the `setup.sh` script isn't run before trying to build and therefore the submodules in `third_party/` aren't initialised for the build to happen.

DavidKorczynski:
I think it's because of this line in the log:
```
2021-11-26 11:39:34,096 - root - INFO - Running: docker run --rm --privileged -e FUZZING_ENGINE=libfuzzer -e ARCHITECTURE=x86_64 -e CIFUZZ=True -e SANITIZER=address -e FUZZING_LANGUAGE=rust -e OUT=/github/workspace/build-out --volumes-from ada0bfe1d13d gcr.io/oss-fuzz/opensk /bin/bash -c 'cd / && rm -rf /src/OpenSK/* && cp -r /github/workspace/storage/OpenSK /src && cd - && compile'.
```

Essentially, CIFuzz deletes the /src/OpenSK folder after having run the commands in `Dockerfile`, which means that the effects of `setup.sh` command from here: https://github.com/google/oss-fuzz/blob/d44ba395f78b7716dc51c9ce76c1a6a8a433122c/projects/opensk/Dockerfile#L20-L23 
will essentially be overwritten. 

I have submitted this issue earlier https://github.com/google/oss-fuzz/issues/6755 so let me just ping in there

One quick fix would be to change the set up on the OSS-Fuzz side to move running of `setup.sh` and `fuzzing_setup.sh` from the `Dockerfile` in the `build.sh` instead.

DavidKorczynski:
Could we try and retrigger the CI? I think this should work now following solving https://github.com/google/oss-fuzz/issues/6755

jmichelp (approved):

PR #407, opened 2021-11-15
fixes the serde version

kaczmarczyck:
Desktop tests and workflows started failing with new nightly features in `serde_json` due to [this release](https://github.com/serde-rs/json/releases/tag/v1.0.70). To fix it, we either have to upgrade to a new nightly compiler or fix the serde_json version to <= 1.0.69.

PR #419, opened 2021-11-30
Forward desktop test fixes

kaczmarczyck:
This commit fixes clippy warnings of nightly compilers until 2021-11-25. This helps users who experiment with versions newer than our current 2021-03-25, and also can't hurt out style. Additionally, the second commit is a fix for `run_desktop_tests.sh` that sometimes fails with a similar problem as #407.

I decided to ignore some of the clippy warnings, instead of renaming the enum variants. This is to stay closer to names in the specification, e.g. command "authenticatorMakeCredential".

### DIFF
--- a/examples/store_latency.rs
+++ b/examples/store_latency.rs
@@ -54,7 +54,7 @@ unsafe fn boot_store(erase: bool) -> Store<Storage> {
 
 #[derive(Debug)]
 struct StorageConfig {
-    page_size: usize,
+    _page_size: usize,
     num_pages: usize,
 }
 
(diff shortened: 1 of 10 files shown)


examples/store_latency.rs
```diff
@@ -54,15 +54,15 @@ unsafe fn boot_store(erase: bool) -> Store<Storage> {
 
 #[derive(Debug)]
 struct StorageConfig {
-    page_size: usize,
+    _page_size: usize,
```
ia0: If this is not used, let's remove it. I didn't realize it got unused when refactoring last time.

ia0 (dismissed):

examples/store_latency.rs
kaczmarczyck: Done

ia0 (approved):

PR #421, opened 2022-01-07
Add track_caller for flex_unwrap in libtock-drivers

ia0:
This permits to get useful panic information. See [documentation](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute).

Binary size change (text size only since the rest didn't change):
| deploy flags | before | after | diff |
| --- | --- | --- | --- |
| `--board=nrf52840dk_opensk --opensk --panic-console` | 136428 | 137020 | +592 |
| `--board=nrf52840dk_opensk --opensk` | 120264 | 120264 | 0 |

So this probably only has impact for development/debugging and this is actually something useful for development/debugging.

### DIFF
--- a/third_party/libtock-drivers/src/result.rs
+++ b/third_party/libtock-drivers/src/result.rs
@@ -15,6 +15,7 @@ pub type TockResult<T> = Result<T, TockError>;
 // This trait allows to flexibly use `Result::unwrap` or `Option::unwrap` and is configured to do
 // so depending on the `debug_ctap` feature.
 pub trait FlexUnwrap<T> {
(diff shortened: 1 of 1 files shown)


kaczmarczyck (approved):
Have you see [this comment on binary size](https://github.com/rust-lang/rust/issues/70579#issuecomment-942556822) from Tock? Sounds like we should at least check binary size of changes like these, just in case.

Otherwise approving.

ia0:
No I wasn't aware of it. I did the measurements and updated the description. It doesn't impact production so I'll commit.

PR #422, opened 2022-01-11
Replaced Libtock driver clock with embedded_time::Clock

shilingwangggg:
Prototype for the concept. Please have a look, thank you.

### DIFF
--- a/src/clock.rs
+++ b/src/clock.rs
@@ -81,3 +81,69 @@ impl embedded_time::Clock for TestClock {
         Ok(embedded_time::Instant::new(0))
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use embedded_time::duration::{Milliseconds, Seconds};
+
+    #[test]
+    fn test_checked_add() {
+        let clock = TestClock::new();
+        let now = clock.try_now().unwrap();
+        assert_eq!(
+            now.checked_add(Seconds::new(1u64)),
+            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as u64))
+        );
+        assert_eq!(
+            now.checked_add(Seconds::new(1u64)),
+            now.checked_add(Milliseconds::new(1000u64))
+        );
+    }
+
+    #[test]
+    fn test_checked_add_overflow() {
+        assert_eq!(
+            CtapInstant::new(u64::MAX).checked_add(Seconds::new(1u64)),
+            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as u64 - 1u64))
+        );
+    }
+
+    #[test]
+    fn test_checked_add_error() {
+        assert!(CtapInstant::new(u64::MAX)
+            .checked_add(Seconds::new(u64::MAX / TEST_CLOCK_FREQUENCY_HZ as u64))
+            .is_none());
+        let now = TestClock::new().try_now().unwrap();
+        assert!(now.checked_add(Seconds::new(u64::MAX)).is_none());
+    }
+
+    #[test]
+    fn test_duration_since() {
+        let clock = TestClock::new();
+        let early = clock.try_now().unwrap();
+        let later = CtapInstant::new(1000u64);
+        assert_eq!(
+            later.checked_duration_since(&early).unwrap().integer(),
+            1000
+        );
+        assert_eq!(early.checked_duration_since(&later), None);
+    }
+
+    #[test]
+    fn test_duration_since_overflow() {
+        let early = CtapInstant::new(u64::MAX);
+        let later = CtapInstant::new(1000u64);
+        assert_eq!(
+            later.checked_duration_since(&early).unwrap().integer(),
+            1001
+        );
+        assert_eq!(early.checked_duration_since(&later), None);
+    }
+
+    #[test]
+    #[should_panic]
+    fn add_panic() {
+        let _ = CtapInstant::new(0) + Milliseconds(u64::MAX / 2 + 1);
+    }
+}
--- a/src/clock.rs
+++ b/src/clock.rs
@@ -1,0 +1,83 @@
+#[cfg(not(feature = "std"))]
+use alloc::fmt;
+use embedded_time::duration::Milliseconds;
+pub use embedded_time::Clock;
+#[cfg(not(feature = "std"))]
+use libtock_drivers::result::FlexUnwrap;
+
+#[cfg(not(feature = "std"))]
+pub struct LibtockClock<const CLOCK_FREQUENCY: u32>(libtock_drivers::timer::Timer<'static>);
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> LibtockClock<CLOCK_FREQUENCY> {
+    pub fn new() -> Self {
+        let boxed_cb = alloc::boxed::Box::new(libtock_drivers::timer::with_callback(|_, _| {}));
+        let timer = alloc::boxed::Box::leak(boxed_cb).init().flex_unwrap();
+        Self(timer)
+    }
+}
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> fmt::Debug for LibtockClock<CLOCK_FREQUENCY> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LibtockClock")
+            .field("CLOCK_FREQUENCY", &CLOCK_FREQUENCY)
+            .finish()
+    }
+}
+
+const KEEPALIVE_DELAY_MS: ClockInt = 100;
+pub const KEEPALIVE_DELAY: Milliseconds<ClockInt> = Milliseconds(KEEPALIVE_DELAY_MS);
+
+#[cfg(target_pointer_width = "32")]
+pub type ClockInt = u32;
+#[cfg(target_pointer_width = "64")]
+pub type ClockInt = u64;
+
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> embedded_time::Clock for LibtockClock<CLOCK_FREQUENCY> {
+    // TODO: Implement and use a 24-bits TimeInt for Nordic
+    type T = ClockInt;
+
+    const SCALING_FACTOR: embedded_time::fraction::Fraction =
+        <embedded_time::fraction::Fraction>::new(1, CLOCK_FREQUENCY);
+
+    fn try_now(&self) -> Result<embedded_time::Instant<Self>, embedded_time::clock::Error> {
+        let timer = &self.0;
+        let now = timer.get_current_clock().flex_unwrap();
+        Ok(embedded_time::Instant::new(now.num_ticks() as Self::T))
+    }
+}
+
+#[cfg(not(feature = "std"))]
+pub type CtapClock = LibtockClock<32768>;
+#[cfg(feature = "std")]
+pub type CtapClock = TestClock;
+
+pub fn new_clock() -> CtapClock {
+    CtapClock::new()
+}
+
+pub type CtapInstant = embedded_time::Instant<CtapClock>;
+
+#[cfg(feature = "std")]
+pub const TEST_CLOCK_FREQUENCY_HZ: u32 = 32768;
+
+#[cfg(feature = "std")]
+#[derive(Default, Clone, Copy, Debug)]
+pub struct TestClock;
+#[cfg(feature = "std")]
+impl TestClock {
+    pub fn new() -> Self {
+        TestClock
+    }
+}
+
+#[cfg(feature = "std")]
+impl embedded_time::Clock for TestClock {
+    type T = u64;
+    const SCALING_FACTOR: embedded_time::fraction::Fraction =
+        <embedded_time::fraction::Fraction>::new(1, TEST_CLOCK_FREQUENCY_HZ);
+
+    fn try_now(&self) -> Result<embedded_time::Instant<Self>, embedded_time::clock::Error> {
+        Ok(embedded_time::Instant::new(0))
+    }
+}
--- a/src/clock.rs
+++ b/src/clock.rs
@@ -89,40 +89,46 @@ mod test {
 
     #[test]
     fn test_checked_add() {
-        let clock = TestClock::new();
-        let now = clock.try_now().unwrap();
+        let now = CtapInstant::new(0);
         assert_eq!(
-            now.checked_add(Seconds::new(1u64)),
-            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as u64))
+            now.checked_add(Seconds::new(1 as ClockInt)),
+            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as ClockInt))
         );
         assert_eq!(
-            now.checked_add(Seconds::new(1u64)),
-            now.checked_add(Milliseconds::new(1000u64))
+            now.checked_add(Seconds::new(1 as ClockInt)),
+            now.checked_add(Milliseconds::new(1000 as ClockInt))
         );
     }
 
     #[test]
     fn test_checked_add_overflow() {
         assert_eq!(
-            CtapInstant::new(u64::MAX).checked_add(Seconds::new(1u64)),
-            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as u64 - 1u64))
+            CtapInstant::new(u64::MAX).checked_add(Seconds::new(1 as ClockInt)),
+            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as u64 - 1))
         );
     }
 
     #[test]
     fn test_checked_add_error() {
-        assert!(CtapInstant::new(u64::MAX)
-            .checked_add(Seconds::new(u64::MAX / TEST_CLOCK_FREQUENCY_HZ as u64))
+        assert!(CtapInstant::new(0)
+            .checked_add(Seconds::new(u64::MAX / TEST_CLOCK_FREQUENCY_HZ as ClockInt))
             .is_none());
-        let now = TestClock::new().try_now().unwrap();
-        assert!(now.checked_add(Seconds::new(u64::MAX)).is_none());
+        assert!(CtapInstant::new(0)
+            .checked_add(Seconds::new(
+                u64::MAX / TEST_CLOCK_FREQUENCY_HZ as ClockInt / 2
+            ))
+            .is_some());
+        assert!(CtapInstant::new(0)
+            .checked_add(Seconds::new(
+                u64::MAX / TEST_CLOCK_FREQUENCY_HZ as ClockInt / 2 + 1
+            ))
+            .is_none());
     }
 
     #[test]
     fn test_duration_since() {
-        let clock = TestClock::new();
-        let early = clock.try_now().unwrap();
-        let later = CtapInstant::new(1000u64);
+        let early = CtapInstant::new(0);
+        let later = CtapInstant::new(1000);
         assert_eq!(
             later.checked_duration_since(&early).unwrap().integer(),
             1000
@@ -133,7 +139,7 @@ mod test {
     #[test]
     fn test_duration_since_overflow() {
         let early = CtapInstant::new(u64::MAX);
-        let later = CtapInstant::new(1000u64);
+        let later = CtapInstant::new(1000);
         assert_eq!(
             later.checked_duration_since(&early).unwrap().integer(),
             1001
@@ -144,6 +150,7 @@ mod test {
     #[test]
     #[should_panic]
     fn add_panic() {
-        let _ = CtapInstant::new(0) + Milliseconds(u64::MAX / 2 + 1);
+        let _ =
+            CtapInstant::new(0) + Seconds(u64::MAX / TEST_CLOCK_FREQUENCY_HZ as ClockInt / 2 + 1);
     }
 }
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::super::clock::CtapInstant;
 use super::command::AuthenticatorClientPinParameters;
 use super::data_formats::{
     ok_or_missing, ClientPinSubCommand, CoseKey, GetAssertionHmacSecretInput, PinUvAuthProtocol,
@@ -32,7 +33,6 @@ use crypto::sha256::Sha256;
 use crypto::Hash256;
 #[cfg(test)]
 use enum_iterator::IntoEnumIterator;
-use libtock_drivers::timer::ClockValue;
 use subtle::ConstantTimeEq;
 
 /// The prefix length of the PIN hash that is stored and compared.
@@ -291,7 +291,7 @@ impl ClientPin {
         &mut self,
         env: &mut impl Env,
         client_pin_params: AuthenticatorClientPinParameters,
-        now: ClockValue,
+        now: CtapInstant,
     ) -> Result<AuthenticatorClientPinResponse, Ctap2StatusCode> {
         let AuthenticatorClientPinParameters {
             pin_uv_auth_protocol,
@@ -359,7 +359,7 @@ impl ClientPin {
         &mut self,
         env: &mut impl Env,
         mut client_pin_params: AuthenticatorClientPinParameters,
-        now: ClockValue,
+        now: CtapInstant,
     ) -> Result<AuthenticatorClientPinResponse, Ctap2StatusCode> {
         // Mutating client_pin_params is just an optimization to move it into
         // process_get_pin_token, without cloning permissions_rp_id here.
@@ -388,7 +388,7 @@ impl ClientPin {
         &mut self,
         env: &mut impl Env,
         client_pin_params: AuthenticatorClientPinParameters,
-        now: ClockValue,
+        now: CtapInstant,
     ) -> Result<ResponseData, Ctap2StatusCode> {
         let response = match client_pin_params.sub_command {
             ClientPinSubCommand::GetPinRetries => Some(self.process_get_pin_retries(env)?),
@@ -495,7 +495,7 @@ impl ClientPin {
     }
 
     /// Updates the running timers, triggers timeout events.
-    pub fn update_timeouts(&mut self, now: ClockValue) {
+    pub fn update_timeouts(&mut self, now: CtapInstant) {
         self.pin_uv_auth_token_state
             .pin_uv_auth_token_usage_timer_observer(now);
     }
@@ -568,9 +568,7 @@ impl ClientPin {
         };
         let mut pin_uv_auth_token_state = PinUvAuthTokenState::new();
         pin_uv_auth_token_state.set_permissions(0xFF);
-        const CLOCK_FREQUENCY_HZ: usize = 32768;
-        const DUMMY_CLOCK_VALUE: ClockValue = ClockValue::new(0, CLOCK_FREQUENCY_HZ);
-        pin_uv_auth_token_state.begin_using_pin_uv_auth_token(DUMMY_CLOCK_VALUE);
+        pin_uv_auth_token_state.begin_using_pin_uv_auth_token(CtapInstant::new(0));
         ClientPin {
             pin_protocol_v1: PinProtocol::new_test(key_agreement_key_v1, pin_uv_auth_token),
             pin_protocol_v2: PinProtocol::new_test(key_agreement_key_v2, pin_uv_auth_token),
@@ -586,10 +584,7 @@ mod test {
     use super::*;
     use crate::env::test::TestEnv;
     use alloc::vec;
-    use libtock_drivers::timer::Duration;
-
-    const CLOCK_FREQUENCY_HZ: usize = 32768;
-    const DUMMY_CLOCK_VALUE: ClockValue = ClockValue::new(0, CLOCK_FREQUENCY_HZ);
+    use embedded_time::duration::Milliseconds;
 
     /// Stores a PIN hash corresponding to the dummy PIN "1234".
     fn set_standard_pin(env: &mut TestEnv) {
@@ -822,7 +817,7 @@ mod test {
             power_cycle_state: Some(false),
         });
         assert_eq!(
-            client_pin.process_command(&mut env, params.clone(), DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params.clone(), CtapInstant::new(0)),
             Ok(ResponseData::AuthenticatorClientPin(expected_response))
         );
 
@@ -834,7 +829,7 @@ mod test {
             power_cycle_state: Some(true),
         });
         assert_eq!(
-            client_pin.process_command(&mut env, params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params, CtapInstant::new(0)),
             Ok(ResponseData::AuthenticatorClientPin(expected_response))
         );
     }
@@ -862,7 +857,7 @@ mod test {
             power_cycle_state: None,
         });
         assert_eq!(
-            client_pin.process_command(&mut env, params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params, CtapInstant::new(0)),
             Ok(ResponseData::AuthenticatorClientPin(expected_response))
         );
     }
@@ -882,7 +877,7 @@ mod test {
             create_client_pin_and_parameters(pin_uv_auth_protocol, ClientPinSubCommand::SetPin);
         let mut env = TestEnv::new();
         assert_eq!(
-            client_pin.process_command(&mut env, params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params, CtapInstant::new(0)),
             Ok(ResponseData::AuthenticatorClientPin(None))
         );
     }
@@ -915,14 +910,14 @@ mod test {
         let pin_uv_auth_param = shared_secret.authenticate(&auth_param_data);
         params.pin_uv_auth_param = Some(pin_uv_auth_param);
         assert_eq!(
-            client_pin.process_command(&mut env, params.clone(), DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params.clone(), CtapInstant::new(0)),
             Ok(ResponseData::AuthenticatorClientPin(None))
         );
 
         let mut bad_params = params.clone();
         bad_params.pin_hash_enc = Some(vec![0xEE; 16]);
         assert_eq!(
-            client_pin.process_command(&mut env, bad_params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, bad_params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP2_ERR_PIN_AUTH_INVALID)
         );
 
@@ -930,7 +925,7 @@ mod test {
             storage::decr_pin_retries(&mut env).unwrap();
         }
         assert_eq!(
-            client_pin.process_command(&mut env, params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP2_ERR_PIN_BLOCKED)
         );
     }
@@ -961,7 +956,7 @@ mod test {
         set_standard_pin(&mut env);
 
         let response = client_pin
-            .process_command(&mut env, params.clone(), DUMMY_CLOCK_VALUE)
+            .process_command(&mut env, params.clone(), CtapInstant::new(0))
             .unwrap();
         let encrypted_token = match response {
             ResponseData::AuthenticatorClientPin(Some(response)) => {
@@ -997,7 +992,7 @@ mod test {
         let mut bad_params = params;
         bad_params.pin_hash_enc = Some(vec![0xEE; 16]);
         assert_eq!(
-            client_pin.process_command(&mut env, bad_params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, bad_params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP2_ERR_PIN_INVALID)
         );
     }
@@ -1022,7 +1017,7 @@ mod test {
 
         assert_eq!(storage::force_pin_change(&mut env), Ok(()));
         assert_eq!(
-            client_pin.process_command(&mut env, params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP2_ERR_PIN_INVALID),
         );
     }
@@ -1055,7 +1050,7 @@ mod test {
         set_standard_pin(&mut env);
 
         let response = client_pin
-            .process_command(&mut env, params.clone(), DUMMY_CLOCK_VALUE)
+            .process_command(&mut env, params.clone(), CtapInstant::new(0))
             .unwrap();
         let encrypted_token = match response {
             ResponseData::AuthenticatorClientPin(Some(response)) => {
@@ -1091,21 +1086,21 @@ mod test {
         let mut bad_params = params.clone();
         bad_params.permissions = Some(0x00);
         assert_eq!(
-            client_pin.process_command(&mut env, bad_params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, bad_params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)
         );
 
         let mut bad_params = params.clone();
         bad_params.permissions_rp_id = None;
         assert_eq!(
-            client_pin.process_command(&mut env, bad_params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, bad_params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)
         );
 
         let mut bad_params = params;
         bad_params.pin_hash_enc = Some(vec![0xEE; 16]);
         assert_eq!(
-            client_pin.process_command(&mut env, bad_params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, bad_params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP2_ERR_PIN_INVALID)
         );
     }
@@ -1132,7 +1127,7 @@ mod test {
 
         assert_eq!(storage::force_pin_change(&mut env), Ok(()));
         assert_eq!(
-            client_pin.process_command(&mut env, params, DUMMY_CLOCK_VALUE),
+            client_pin.process_command(&mut env, params, CtapInstant::new(0)),
             Err(Ctap2StatusCode::CTAP2_ERR_PIN_INVALID)
         );
     }
@@ -1485,7 +1480,7 @@ mod test {
         let message = [0xAA];
         client_pin
             .pin_uv_auth_token_state
-            .begin_using_pin_uv_auth_token(DUMMY_CLOCK_VALUE);
+            .begin_using_pin_uv_auth_token(CtapInstant::new(0));
 
         let pin_uv_auth_token_v1 = client_pin
             .get_pin_protocol(PinUvAuthProtocol::V1)
@@ -1617,7 +1612,7 @@ mod test {
         params.permissions = Some(0xFF);
 
         assert!(client_pin
-            .process_command(&mut env, params, DUMMY_CLOCK_VALUE)
+            .process_command(&mut env, params, CtapInstant::new(0))
             .is_ok());
         for permission in PinPermission::into_enum_iter() {
             assert_eq!(
@@ -1634,7 +1629,7 @@ mod test {
             Ok(())
         );
 
-        let timeout = DUMMY_CLOCK_VALUE.wrapping_add(Duration::from_ms(30001));
+        let timeout = CtapInstant::new(0) + (Milliseconds::new(30001_u32));
         client_pin.update_timeouts(timeout);
         for permission in PinPermission::into_enum_iter() {
             assert_eq!(
@@ -1663,7 +1658,7 @@ mod test {
         params.permissions = Some(0xFF);
 
         assert!(client_pin
-            .process_command(&mut env, params, DUMMY_CLOCK_VALUE)
+            .process_command(&mut env, params, CtapInstant::new(0))
             .is_ok());
         for permission in PinPermission::into_enum_iter() {
             assert_eq!(
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -1629,7 +1629,7 @@ mod test {
             Ok(())
         );
 
-        let timeout = CtapInstant::new(0) + (Milliseconds::new(30001_u32));
+        let timeout = CtapInstant::new(0) + Milliseconds::new(30001_u32);
         client_pin.update_timeouts(timeout);
         for permission in PinPermission::into_enum_iter() {
             assert_eq!(
--- a/src/ctap/ctap1.rs
+++ b/src/ctap/ctap1.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::super::clock::CtapInstant;
 use super::apdu::{Apdu, ApduStatusCode};
 use super::CtapState;
 use crate::ctap::storage;
@@ -20,7 +21,6 @@ use alloc::vec::Vec;
 use arrayref::array_ref;
 use core::convert::Into;
 use core::convert::TryFrom;
-use libtock_drivers::timer::ClockValue;
 
 // For now, they're the same thing with apdu.rs containing the authoritative definition
 pub type Ctap1StatusCode = ApduStatusCode;
@@ -183,7 +183,7 @@ impl Ctap1Command {
         env: &mut impl Env,
         message: &[u8],
         ctap_state: &mut CtapState,
-        clock_value: ClockValue,
+        clock_value: CtapInstant,
     ) -> Result<Vec<u8>, Ctap1StatusCode> {
         if !ctap_state
             .allows_ctap1(env)
@@ -347,15 +347,9 @@ impl Ctap1Command {
 mod test {
     use super::super::{key_material, CREDENTIAL_ID_SIZE, USE_SIGNATURE_COUNTER};
     use super::*;
+    use crate::clock::TEST_CLOCK_FREQUENCY_HZ;
     use crate::env::test::TestEnv;
     use crypto::Hash256;
-
-    const CLOCK_FREQUENCY_HZ: usize = 32768;
-    const START_CLOCK_VALUE: ClockValue = ClockValue::new(0, CLOCK_FREQUENCY_HZ);
-    const TIMEOUT_CLOCK_VALUE: ClockValue = ClockValue::new(
-        (30001 * CLOCK_FREQUENCY_HZ as isize) / 1000,
-        CLOCK_FREQUENCY_HZ,
-    );
 
     fn create_register_message(application: &[u8; 32]) -> Vec<u8> {
         let mut message = vec![
@@ -400,15 +394,19 @@ mod test {
         let mut env = TestEnv::new();
         env.user_presence()
             .set(|_| panic!("Unexpected user presence check in CTAP1"));
-        let mut ctap_state = CtapState::new(&mut env, START_CLOCK_VALUE);
+        let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0_u64));
         storage::toggle_always_uv(&mut env).unwrap();
 
         let application = [0x0A; 32];
         let message = create_register_message(&application);
-        ctap_state.u2f_up_state.consume_up(START_CLOCK_VALUE);
-        ctap_state.u2f_up_state.grant_up(START_CLOCK_VALUE);
-        let response =
-            Ctap1Command::process_command(&mut env, &message, &mut ctap_state, START_CLOCK_VALUE);
+        ctap_state.u2f_up_state.consume_up(CtapInstant::new(0_u64));
+        ctap_state.u2f_up_state.grant_up(CtapInstant::new(0_u64));
+        let response = Ctap1Command::process_command(
+            &mut env,
+            &message,
+            &mut ctap_state,
+            CtapInstant::new(0_u64),
+        );
         assert_eq!(response, Err(Ctap1StatusCode::SW_COMMAND_NOT_ALLOWED));
     }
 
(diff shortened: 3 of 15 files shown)


src/ctap/timed_permissions_exp.rs
```diff
@@ -0,0 +1,217 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[cfg(feature = "with_ctap1")]
+use super::ClockInt;
+use embedded_time::duration::Milliseconds;
+use embedded_time::{Clock, Instant};
+
+#[derive(Debug)]
+pub enum TimedPermission<C: Clock> {
```
kaczmarczyck: I was wondering  if the `embedded_time` crate's [`Timer`](https://docs.rs/embedded-time/latest/embedded_time/timer/struct.Timer.html) should and can replace the data type eventually? It's a good first step to keep changes simple, just asking for a long-term goal.

src/ctap/timed_permissions_exp.rs
```diff
@@ -0,0 +1,217 @@
...
+
+impl<C: Clock> Copy for TimedPermission<C> {}
+
+impl<C: Clock> TimedPermission<C> {
+    pub fn waiting() -> TimedPermission<C> {
+        TimedPermission::Waiting
+    }
+
+    pub fn granted(now: Instant<C>, grant_duration: Milliseconds<C::T>) -> TimedPermission<C> {
+        TimedPermission::Granted(now.checked_add(grant_duration).unwrap())
```
kaczmarczyck: If you `unwrap` here, that means we panic when the clock wraps, I'd guess. In that case, I'd prefer if we rather not grant the permission or (even better) work around the wrapping clock.

src/ctap/mod.rs
```diff
@@ -61,9 +61,9 @@ use self::response::{
 };
 use self::status_code::Ctap2StatusCode;
 use self::storage::PersistentStore;
-use self::timed_permission::TimedPermission;
+use self::timed_permissions_exp::TimedPermission as TimedPermissionExp;
```
kaczmarczyck: What's the advantage of renaming?

src/ctap/mod.rs
```diff
@@ -79,13 +79,19 @@ use crypto::hmac::{hmac_256, verify_hmac_256};
...
+use embedded_time::duration::Milliseconds;
+use embedded_time::{Clock, Instant};
 #[cfg(feature = "debug_ctap")]
 use libtock_drivers::console::Console;
 use libtock_drivers::crp;
-use libtock_drivers::timer::{ClockValue, Duration};
 use sk_cbor as cbor;
 use sk_cbor::cbor_map_options;
 
+#[cfg(target_pointer_width = "32")]
```
kaczmarczyck: Does `usize` not work here?

src/ctap/mod.rs
```diff
@@ -105,12 +111,18 @@ const ED_FLAG: u8 = 0x80;
...
+const U2F_UP_PROMPT_TIMEOUT_MS: ClockInt = 10000;
 #[cfg(feature = "with_ctap1")]
-const U2F_UP_PROMPT_TIMEOUT: Duration<isize> = Duration::from_ms(10000);
+const U2F_UP_PROMPT_TIMEOUT: Milliseconds = Milliseconds(U2F_UP_PROMPT_TIMEOUT_MS);
 // TODO(kaczmarczyck) 2.1 allows Reset after Reset and 15 seconds?
-const RESET_TIMEOUT_DURATION: Duration<isize> = Duration::from_ms(10000);
-const STATEFUL_COMMAND_TIMEOUT_DURATION: Duration<isize> = Duration::from_ms(30000);
+const RESET_TIMEOUT_DURATION_MS: ClockInt = 10000;
+const RESET_TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(RESET_TIMEOUT_DURATION_MS);
+const STATEFUL_COMMAND_TIMEOUT_DURATION_MS: ClockInt = 30000;
```
kaczmarczyck: Why do we need both constants? So that we can annotate the integer with the correct type?

src/ctap/mod.rs
```diff
@@ -1342,8 +1355,19 @@ mod test {
     use cbor::{cbor_array, cbor_array_vec, cbor_map};
     use crypto::rng256::ThreadRng256;
 
-    const CLOCK_FREQUENCY_HZ: usize = 32768;
-    const DUMMY_CLOCK_VALUE: ClockValue = ClockValue::new(0, CLOCK_FREQUENCY_HZ);
+    const CLOCK_FREQUENCY_HZ: u32 = 32768;
+    #[derive(Default, Clone, Copy)]
+    pub struct TestClock();
```
kaczmarczyck: We could move all `Clock` implementations to a separate file, and this one conditionally compiles only for `#[cfg(test)]`. Then all files that need the `TestClock` can import it, and main is less cluttered.

src/main.rs
```diff
@@ -134,27 +153,29 @@ fn main() {
...
                         // TODO: handle this unexpected packet.
                     }
                 }
             }
         }
 
-        let now = timer.get_current_clock().flex_unwrap();
-        if let Some(wait_duration) = now.wrapping_sub(last_led_increment) {
+        let now = clock.try_now().unwrap();
+        use core::convert::TryInto;
```
kaczmarczyck: I think importing on the top of the file is cleaner.

src/main.rs
```diff
@@ -315,10 +336,10 @@ fn switch_off_leds() {
 
 fn check_user_presence(cid: ChannelID) -> Result<(), Ctap2StatusCode> {
     // The timeout is N times the keepalive delay.
-    const TIMEOUT_ITERATIONS: usize = ctap::TOUCH_TIMEOUT_MS as usize / KEEPALIVE_DELAY_MS as usize;
+    const TIMEOUT_ITERATIONS: u32 = ctap::TOUCH_TIMEOUT_MS / KEEPALIVE_DELAY_MS;
```
kaczmarczyck: Is this conditional on the architecture? So it should actually be u32 on a 32 bit machine etc.? Then it's closer in intention to the `usize`. Though the compiler doesn't accept that as easily.

src/ctap/hid/mod.rs
```diff
@@ -62,27 +63,8 @@ pub struct Message {
...
-#[allow(dead_code)]
-pub enum KeepaliveStatus {
-    Processing,
-    UpNeeded,
-}
-
-#[allow(dead_code)]
-// TODO(kaczmarczyck) disable the warning in the end
-impl CtapHid {
+pub struct CtapHidConsts;
```
kaczmarczyck: I wonder why this extra class is necessary?

kaczmarczyck (commented):
Thanks for the extensive PR! A few general comments:

- It it a pure port from one Clock library to another, it doesn't clean up some legacy files like `time_permission.rs`. The PR is big enough as is, so that's fine. Just wanted to point out there is some opportunity to prune.
- We should think if we can reasonably split this up before submission.
- Have you check binary sizes before and after? Try `cargo bloat --release --target=thumbv7em-none-eabi --features=with_ctap1` and please let me know the difference.
- The API of `CtapState` only changes in input types, it seems. Having a good API will be more important if we want a good library. Not necessarily to be tackled in this PR, but we should concern ourselves with that eventually.

kaczmarczyck:
I check binary size using our usual release flags.
```
export RUSTFLAGS="-C link-arg=-icf=all -C force-frame-pointers=no"
unset RUSTFLAGS
```

This PR adds 200 Bytes, which is within variance of these measurements. So almost free, very good.

src/main.rs
```diff
@@ -49,28 +52,42 @@ use libtock_drivers::usb_ctap_hid;
...
 
-const KEEPALIVE_DELAY_MS: isize = 100;
-const KEEPALIVE_DELAY: Duration<isize> = Duration::from_ms(KEEPALIVE_DELAY_MS);
+const KEEPALIVE_DELAY_MS: ClockInt = 100;
+const KEEPALIVE_DELAY: Milliseconds<ClockInt> = Milliseconds(KEEPALIVE_DELAY_MS);
+const KEEPALIVE_DELAY_TOCK: Duration<isize> = Duration::from_ms(KEEPALIVE_DELAY_MS as isize);
 const SEND_TIMEOUT: Duration<isize> = Duration::from_ms(1000);
 
+pub struct LibtockClock<'a, const CLOCK_FREQUENCY: u32>(libtock_drivers::timer::Timer<'a>);
+impl<const CLOCK_FREQUENCY: u32> embedded_time::Clock for LibtockClock<'_, CLOCK_FREQUENCY> {
```
ia0: Should this be in a separate `clock` module?

src/main.rs
```diff
@@ -49,28 +52,42 @@ use libtock_drivers::usb_ctap_hid;
...
+        Ok(embedded_time::Instant::new(now.num_ticks() as Self::T))
+    }
+}
+
 fn main() {
     // Setup the timer with a dummy callback (we only care about reading the current time, but the
     // API forces us to set an alarm callback too).
     let mut with_callback = timer::with_callback(|_, _| {});
     let timer = with_callback.init().flex_unwrap();
+    let clock = LibtockClock::<32768>(timer);
```
ia0: The clock frequency should be a configuration constant somewhere. Ideally, even `main` shouldn't be aware of it. For now I would hide this in a `clock` module exposing `pub fn new_clock() -> Box<dyn embedded_time::Clock>`. OpenSK shouldn't be aware of the clock implementation.

src/main.rs
```diff
@@ -49,28 +52,42 @@ use libtock_drivers::usb_ctap_hid;
...
 
     // Setup USB driver.
     if !usb_ctap_hid::setup() {
         panic!("Cannot setup USB driver");
     }
 
-    let boot_time = timer.get_current_clock().flex_unwrap();
+    let boot_time_instant = clock.try_now().unwrap();
```
ia0: Why rename suffix with `instant`? This information is already part of the type, no need to document it in the name. Also I would try to keep the diff small if possible.

src/main.rs
```diff
@@ -86,6 +103,7 @@ fn main() {
                 ButtonState::Released => (),
             };
         });
+
```
ia0: nit: Still in an effort to keep the diff small, we should not add this line unless it helps the current PR.

src/main.rs
```diff
@@ -49,28 +52,42 @@ use libtock_drivers::usb_ctap_hid;
 
 libtock_core::stack_size! {0x4000}
 
-const KEEPALIVE_DELAY_MS: isize = 100;
-const KEEPALIVE_DELAY: Duration<isize> = Duration::from_ms(KEEPALIVE_DELAY_MS);
+const KEEPALIVE_DELAY_MS: ClockInt = 100;
+const KEEPALIVE_DELAY: Milliseconds<ClockInt> = Milliseconds(KEEPALIVE_DELAY_MS);
+const KEEPALIVE_DELAY_TOCK: Duration<isize> = Duration::from_ms(KEEPALIVE_DELAY_MS as isize);
```
ia0: I like this idea of suffixing with `tock` all Tock-specific (and thus non-portable) code. This is a way to track how much we still need to clean to make CTAP a library.

src/main.rs
```diff
@@ -134,27 +153,29 @@ fn main() {
...
                     }
                 }
             }
         }
 
-        let now = timer.get_current_clock().flex_unwrap();
-        if let Some(wait_duration) = now.wrapping_sub(last_led_increment) {
+        let now = clock.try_now().unwrap();
+        use core::convert::TryInto;
+        if let Some(wait_duration) = now.checked_duration_since(&last_led_increment) {
```
ia0, line 150: I tried to look quickly in the `embedded_time` code and it seems they should support wrapping time like we need. But I would still add some tests in the `clock` module (that we need to create to abstract time) to make sure it works (it's also good to check for regressions when we update the crate version).

src/main.rs
```diff
@@ -187,15 +208,14 @@ fn main() {
 }
 
 #[cfg(feature = "debug_ctap")]
-fn print_packet_notice(notice_text: &str, timer: &Timer) {
-    let now = timer.get_current_clock().flex_unwrap();
-    let now_us = (Timestamp::<f64>::from_clock_value(now).ms() * 1000.0) as u64;
+fn print_packet_notice(notice_text: &str, clock: &LibtockClock) {
```
ia0: We shouldn't use `LibtockClock` here. The `clock` module should expose a `Clock` (implementing `embedded_time::Clock`). The implementation would be controlled by a board selection feature (but for now it can be `LibtockClock`).

src/main.rs
```diff
@@ -187,15 +208,14 @@ fn main() {
...
+    let now = clock.try_now().unwrap();
     writeln!(
         Console::new(),
         "{} at {}.{:06} s",
         notice_text,
-        now_us / 1_000_000,
-        now_us % 1_000_000
+        now / 1_000_000,
+        now % 1_000_000
```
ia0: Did you try to deploy on the dev-kit with `debug_ctap`?

```
./deploy.py --board=nrf52840dk_opensk --opensk --panic-console --debug
```

And check the output in RTT when register+login in webauthn.io?

src/ctap/hid/mod.rs
```diff
@@ -149,18 +151,16 @@ impl CtapHid {
     pub fn process_hid_packet<R, CheckUserPresence>(
         &mut self,
         packet: &HidPacket,
-        clock_value: ClockValue,
-        ctap_state: &mut CtapState<R, CheckUserPresence>,
+        clock_value: Instant<C>,
```
ia0: I think we should provide helpers. The `clock` module should expose `Instant` as a simple type. The implementation using the actual clock as argument.

ia0 (commented):
I didn't look much outside `main.rs` but overall comment is that we need to add a `clock` module with the following interface (should be extended if needed):

```rust
pub type Clock = LibtockClock;
pub type Instant = embedded_time::Instant<Clock>;
impl embedded_time::Clock for Clock { ... }
```

The idea being that all the code outside the `clock` module shouldn't have to know what is the actually clock implementation. They all use the type `Clock`. For convenience we can add `Instant` such that we don't need to write `Instant<Clock>` everywhere but just `Instant`, since we use a single clock.

This `clock` module could also have some tests that time wraps as expected.

src/ctap/hid/mod.rs
shilingwangggg: Because CtapHid depends on Generic C: Clock. However, some of the Consts are used in places where adding this C generic is not necessary. If we don't extract these consts out, then there will be more generics added to more classes and functions.

src/ctap/mod.rs
shilingwangggg: I revert the renaming.

src/ctap/timed_permissions_exp.rs
shilingwangggg: The long-term goal is to replace TimedPermission with Timer. However, that would mean a even bigger change than this one. I think the first PR can introduce the embedded_time to ctap and don't change the structure of code. Next step we remove TimedPermission and introduce Timer.

src/ctap/mod.rs
shilingwangggg: Yes. That is the goal. This PR will support both u32 and u64. Having an intermediate const is a way to make sure the duration is created with the correct type of int(which is important for the Clock)

src/main.rs
shilingwangggg: I didn't find an easy way to print out the Instant. Therefore in the second commit I just remove the time printing. I'll try to find a way to print the Instant and push another commit.

src/main.rs
```diff
@@ -187,25 +207,16 @@ fn main() {
...
-    writeln!(
-        Console::new(),
-        "{} at {}.{:06} s",
-        notice_text,
-        now_us / 1_000_000,
-        now_us % 1_000_000
-    )
-    .unwrap();
+fn print_packet_notice(notice_text: &str) {
+    writeln!(Console::new(), "{}", notice_text).unwrap();
```
kaczmarczyck: How hard is it to get timestamps back into the print?

src/main.rs
shilingwangggg: I changed it to ClockInt for consistency.

src/ctap/mod.rs
shilingwangggg: No. TimeInt doesn't implement usize: https://docs.rs/embedded-time/0.12.1/embedded_time/trait.TimeInt.html

src/main.rs
shilingwangggg: Returning a `Box<dyn embedded_time::Clock>`  to abstract away the clock frequency is unfortunately not possible because the trait `embedded_time::Clock` cannot be made into an object, because it contains this associated `const`: `const SCALING_FACTOR: Fraction;`.

src/ctap/timed_permissions_exp.rs
shilingwangggg: embedded_time's checked_add doesn't wrap. If it would overflow, it will return None. I don't know how should we deal with it. @ia0 It is also related to your comment about wrapping. What is your opinion on dealing with it?

src/ctap/mod.rs
shilingwangggg: Done.

src/main.rs
shilingwangggg: Done.

src/main.rs
shilingwangggg: Done. PTAL in the new commit.

src/main.rs
shilingwangggg: No reason for the renaming... Changed back.

src/main.rs
shilingwangggg: I hided it in the clock module added. PTAL.

src/ctap/hid/mod.rs
shilingwangggg: I think since we don't expose the Instant outside the lib and we have the plan of using embedded_time::Timer, now we don't need to make this change. If with embedded_time:Timer we still need Instant, I'll change it in another PR.

src/main.rs
ia0: Oh I wasn't aware of this limitation. The clock module should probably look like:

```rust
// Should implement embedded_time::Clock
pub type Clock = ...
pub fn new_clock() -> Clock { ... }
```

src/ctap/timed_permissions_exp.rs
ia0: I checked the [code](https://github.com/FluenTech/embedded-time/blob/83b09cb9c4e4c27fdc10e1089aab87b69a0436c9/src/instant.rs#L183) and `checked_add` actually wraps which is what we want. The error case is only when the duration is too large (which is why wrapping preserves ordering, but one shouldn't call `checked_add` multiple times though).

src/ctap/clock.rs
```diff
@@ -0,0 +1,48 @@
...
+pub struct LibtockClock<'a, const CLOCK_FREQUENCY: u32>(Timer<'a>);
+impl<'a, const CLOCK_FREQUENCY: u32> LibtockClock<'a, CLOCK_FREQUENCY> {
+    pub fn new(timer: Timer<'a>) -> Self {
+        Self(timer)
+    }
+}
+
+#[cfg(target_pointer_width = "32")]
+pub type ClockInt = u32;
+#[cfg(target_pointer_width = "64")]
```
ia0: When do we use `u64`? Is it for tests? If yes, I think it's fine for compilation to fail if `feature = "std"` is missing. Tests should run with `--features=std`.

src/ctap/clock.rs
```diff
@@ -0,0 +1,48 @@
...
+impl<const CLOCK_FREQUENCY: u32> embedded_time::Clock for LibtockClock<'_, CLOCK_FREQUENCY> {
+    type T = ClockInt;
+
+    const SCALING_FACTOR: embedded_time::fraction::Fraction =
+        <embedded_time::fraction::Fraction>::new(1, CLOCK_FREQUENCY);
+
+    fn try_now(&self) -> Result<embedded_time::Instant<Self>, embedded_time::clock::Error> {
+        let timer = &self.0;
+        let now = timer.get_current_clock().flex_unwrap();
+        Ok(embedded_time::Instant::new(now.num_ticks() as Self::T))
```
ia0: Ok, I'm just realizing we might have a problem on Nordic because the clock wraps at `0x00ffffff` instead of `0xffffffff`. I think `embedded_time` can handle this if we provide our own implementation of `TimeInt`. But given that the code is already [broken](https://github.com/google/OpenSK/blob/d47ca7fa5437e879b25f5b77e17f83118c4f1b51/third_party/libtock-drivers/src/timer.rs#L218) in libtock-drivers before this PR, I would say we can leave this to a future PR. For now, let's add a TODO near `ClockInt` or near `type T = ClockInt`:

```
// TODO: Implement and use a 24-bits TimeInt for Nordic
```

src/ctap/hid/mod.rs
```diff
@@ -124,19 +101,45 @@ impl CtapHid {
...
-    const TIMEOUT_DURATION: Duration<isize> = Duration::from_ms(100);
-    const WINK_TIMEOUT_DURATION: Duration<isize> = Duration::from_ms(5000);
+    const TIMEOUT_DURATION_MS: ClockInt = 100;
+    const TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(Self::TIMEOUT_DURATION_MS);
+    const WINK_TIMEOUT_DURATION_MS: ClockInt = 5000;
+    const WINK_TIMEOUT_DURATION: Milliseconds<ClockInt> =
+        Milliseconds(Self::WINK_TIMEOUT_DURATION_MS);
+}
+
+pub struct CtapHid<C: Clock> {
```
ia0: Now that we have a `clock` module exposing a `CtapClock` you don't need to abstract `CtapHid` over `Clock`. You can just directly use `CtapClock` anywhere you have `C`. This should reduce the diff considerably.

src/ctap/hid/mod.rs
ia0: I'm not sure what you mean by "not exposed outside the lib" but I agree this can be changed in another PR (which would just add `pub type CtapInstant = Instant<CtapClock>` to the `clock` module). But at least in this PR it should be `Instant<CtapClock>`. The code shouldn't be generic on the clock since for a given compilation the clock is unique. That's the idea of the `clock` module to make the first PR simple. Long-term, when we really make CTAP a library, we would need to find the right abstractions (probably a single `Hardware` trait exposing other types). In other words here is how I see the overall "CTAP as a library" project:

Step 1: For each hardware component, move it to a module and only use what the module exposes. This should be small refactoring and aims at having clear interfaces. A typical module would look like:
```rust
// What is the API for this component?
pub trait Interface { ... } // could be provided by a library (like embedded_time)
// The wrapper is just to make sure we don't use anything outside the trait. (this could be step 1.b)
pub struct Wrapper<T>(T);
impl<T: Interface> Interface for Wrapper<T> { ... }
// What is the type for this component?
pub type Component<...> = Wrapper<Impl<...>>;  // could be directly Impl<...> for step1.a
// Some way to build the component
pub fn new(...) -> Component<...> { ... }
```

Step 2: Now that we have clear APIs, make the code generic over those APIs. To avoid too many generic parameters, we could use a single trait:
```rust
trait Hardware {
    type Clock: clock::Interface;  // the rest of the clock module is useless
    type Usb: usb::Interface;
    ...
}
```
Some functions/types would just be abstract over `Hardware`:
```rust
struct Foo<H: Hardware> {
    instant: Instant<H::Clock>,
}
```
Some would take some components as arguments:
```rust
fn foo<H: Hardware>(clock: &mut H::Clock, ...) { ... }
```

src/ctap/clock.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+pub type CtapClock<'a> = LibtockClock<'a, 32768>;
```
ia0: The way I see it, this should be:
```rust
#[cfg(not(feature = "std"))]
pub type CtapClock<'a> = LibtockClock<'a, 32768>;
#[cfg(feature = "std")]
pub type CtapClock<'a> = TestClock;
```

Tests should use `CtapClock` as the rest of the code. That's the unique type for clock (for now the code shouldn't be generic).

src/ctap/clock.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[cfg(test)]
+pub const TEST_CLOCK_FREQUENCY_HZ: u32 = 32768;
+
+#[cfg(test)]
+#[derive(Default, Clone, Copy)]
+pub struct TestClock;
+
+#[cfg(test)]
```
ia0: I would use `cfg(feature = "std")` instead, because fuzzing is not considered "test" by Rust. So we use the "std" feature to mean "the code is built for testing or fuzzing and is not production code".

ia0 (commented):
I think it starts to get there. I've described in another comment how I see the chain of PRs to get CTAP as a library. I think we should first focus on getting the right interfaces but still keep the code monomorphic as it is currently. This permits to get smaller refactoring and should thus be more obvious that we are not changing the semantics. Once we have clear interfaces for all hardware components, then we can start making the code generic (which would be a bigger refactoring).

src/main.rs
shilingwangggg: Done.

src/ctap/hid/mod.rs
shilingwangggg: I see. I tried to remove the generics, but there is still a problem with the lifetime of the underlying timer (in case of ` no(std)` , that must be carried over in the user code. So should we keep a generics over that lifetime instead of the clock type?

src/ctap/clock.rs
shilingwangggg: As discussed with @kaczmarczyck in the beginning of the project, we expect the future crate can also be used for linux kernel as an example.

src/ctap/hid/mod.rs
shilingwangggg: See my comment above regarding lifetime issues.

src/ctap/clock.rs
shilingwangggg: Done

src/ctap/timed_permissions_exp.rs
shilingwangggg: Done, not granting in case of error, in new commit.

src/main.rs
shilingwangggg: We cannot name "our" clock `Clock` because this would lead to a conflict with the trait name `Clock`  from `embedded_time`, which is used by ` try_now`. So currently I re-export that trait in the module `clock` and name the clock implementation `CtapClock`. We could use `Clock` for this one and renamed the trait on re-export but that seems more cumbersome than necessary and prone to confusion for existing users of `embedded_time`.
Regarding the `new_clock` function, I added it. Note that it must take a timer with its lifetime as parameter, when used without `std`.

src/ctap/clock.rs
ia0: I think all your recent comments are related to the same issue. So I'll answer here with comments within how the `clock` module could look like (I'm omitting some parameters for readability):

```rust
#[cfg(not(feature = "std"))] // enabled for production (currently only Tock is supported)
mod prod {
    use libtock_drivers;
    pub struct LibtockClock(Timer);
    impl Libtock { pub fn new() { ... } }
    impl Clock for Libtock {
        type T = u32;  // ideally this should be a custom 24-bits integer
        ...
    }
}

#[cfg(feature = "std")] // enabled for testing and fuzzing
mod test {
    // We don't need anything from libtock for testing. We shouldn't even compile it.
    pub struct TestClock;
    impl Clock for TestClock { ... }
}

#[cfg(not(feature = "std"))]
pub type CtapClock = prod::LibtockClock;
#[cfg(feature = "std")]
pub type CtapClock = test::TestClock;
```

As you can see, `feature = "std"` means testing (including fuzzing which is a form of testing) and `not(feature = "std")` means production. We currently only support Tock for production so there's no need to have this `ClockInt = u64`. This would currently be dead-code and in the future irrelevant because hidden behind the hardware abstraction. In particular the CTAP library will only know of `H::Clock::T` and the user would be responsible for defining this type. In particular, in the code above, you can see the code in the `prod` and `test` module as if it were written by the user of the CTAP library. In one case (`prod`) they do `T = u32` or `T = u24` or if within linux they could do `T = u64`. In the other case (`test`) they do `T = u64`.

Does this make more sense?

src/main.rs
ia0: Yes, `Clock` was just an example. `CtapClock` is good.

src/ctap/clock.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+pub type CtapClock<'a> = LibtockClock<'a, 32768>;
+
+pub fn new_clock<'a>(timer: Timer<'a>) -> CtapClock {
+    CtapClock::new(timer)
+}
+
+//pub type CtapInstant<'a> = embedded_time::Instant<CtapClock<'a>>;
```
ia0: Mmm, I fear that this lifetime parameter of `Timer` might be problematic in the future. You might need to make some code parametric over this lifetime. If you do, please ping me with the locations. I'd like to take a look to see if there's a way to work around or factorize this effort for other hardware components.

src/main.rs
kaczmarczyck: Just saw your other comment, so this is a duplicate.

src/ctap/clock.rs
ia0: Here's a solution to remove the lifetime: https://github.com/ia0/OpenSK/commit/eb220320d4a21bb12a2df1fb21329ee4bb78e690. I think it's ok to go with this until the code is generic. At which point it might be easier to choose a different solution.

kaczmarczyck: If we plan to replace it eventually, that might be okay. Otherwise, is it necessary to introduce unsafe code?

ia0: Yes, any solution to hide the fact that the libtock clock needs a lifetime (which is just a consequence of its design asking a useless callback) is essentially unsafe somewhere. The 2 obvious solutions are:
1. Using static lifetime. This solution. I think the best solution given its simplicity.
2. Using `Pin`. The unsafe would be hidden under the hood though. But we could also hide the `unsafe` keyword under the hood for option 1 too. It's actually what we do for the storage:
    - No `unsafe` keyword but actually unsafe (I managed to get the English wrong in a single sentence, good job): https://github.com/google/OpenSK/blob/d47ca7fa5437e879b25f5b77e17f83118c4f1b51/src/ctap/storage.rs#L68
    - The actual `unsafe` keyword is much deeper: https://github.com/google/OpenSK/blob/d47ca7fa5437e879b25f5b77e17f83118c4f1b51/src/embedded_flash/syscall.rs#L164

Long term, all those "call only once" will disappear because the library won't create the hardware. The hardware will be created by the user and passed to the library. In embedded world it's pretty typical to have this "use once" constraint. The way it's done in embedded Rust is through a [`Peripherals::take`](https://docs.rs/nrf52840-hal/latest/nrf52840_hal/pac/struct.Peripherals.html#method.take) which only returns `Some` the first time and then `None`. We could also do that here if we want to get rid of the surfacing `unsafe` keyword. But this is only temporary until we have a library, so I'm not sure it's worth it.

shilingwangggg: Are we allowed to use the alloc crate? If so, we can leak instead of being unsafe:
```rust
pub fn new() -> Self {
       let boxed_cb = alloc::boxed::Box::new(timer::with_callback(|_, _| {}));
       let timer = alloc::boxed::Box::leak(boxed_cb).init().flex_unwrap();
       Self(timer)
}
```
Assuming we call it once, the leak doesn't matter, and if we call it more than once, at least we do not trash the memory. We could also track the number of time it is called and panic if it is called twice. WDYT?

ia0: Yes this is a safe option, sounds good. Note we could also use the [`once`](https://docs.rs/once/latest/once/macro.assert_has_not_been_called.html) crate if we want to check that it's called once.

src/ctap/hid/mod.rs
shilingwangggg: Now I removed the generics I can put them back. But rethinking on it, maybe it is not a bad idea to have them separate. WDYT?

src/main.rs
shilingwangggg: I updated the PR. PTAL.

src/ctap/hid/mod.rs
shilingwangggg: Done.

src/ctap/clock.rs
```diff
@@ -0,0 +1,79 @@
+#[cfg(not(feature = "std"))]
+use alloc::fmt;
+pub use embedded_time::Clock;
+#[cfg(not(feature = "std"))]
+use libtock_drivers::result::FlexUnwrap;
```
kaczmarczyck: I wonder if this file is at the correct location? Since it has a libtock dependency, shouldn't it be outside of `src/ctap`, so that this can eventually be free of everything Tock related?

src/ctap/client_pin.rs
```diff
@@ -569,6 +569,7 @@ impl ClientPin {
         key_agreement_key: crypto::ecdh::SecKey,
         pin_uv_auth_token: [u8; PIN_TOKEN_LENGTH],
         pin_uv_auth_protocol: PinUvAuthProtocol,
+        dummy_clock_value: CtapInstant,
```
kaczmarczyck: What's the reason to pass this in here, instead of passing a `CtapInstant::new(0)` to `begin_using_pin_uv_auth_token`?

src/ctap/hid/mod.rs
kaczmarczyck: What benefit do you have in mind? Otherwise, a smaller diff in this PR is welcome :)

I guess any future modifications of the nature would benefit from the separation, not sure if it's worth it, if it's just that?

src/ctap/clock.rs
```diff
@@ -0,0 +1,79 @@
...
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> fmt::Debug for LibtockClock<CLOCK_FREQUENCY> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LibtockClock")
+            .field("CLOCK_FREQUENCY", &CLOCK_FREQUENCY)
+            .finish()
+    }
+}
+
+#[cfg(target_pointer_width = "32")]
```
kaczmarczyck: Another comment on the ClockInt. If our `TestClock` does not actually use any system clock, but returns plain zero, we could actually have the same ClockInt in tests as we have in production. If I don't overlook anything, that comes with a few advantages:
- less complexity in this file
- tests are closer to production
- less diff in this PR

Would we still need to define constants like this?
```
    const TIMEOUT_DURATION_MS: ClockInt = 100;
    const TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(Self::TIMEOUT_DURATION_MS);
```
Or could that become a 1-liner?

src/ctap/timed_permission.rs
```diff
@@ -12,41 +12,53 @@
...
+
 impl TimedPermission {
     pub fn waiting() -> TimedPermission {
         TimedPermission::Waiting
     }
 
-    pub fn granted(now: ClockValue, grant_duration: Duration<isize>) -> TimedPermission {
-        TimedPermission::Granted(now.wrapping_add(grant_duration))
+    pub fn granted(now: CtapInstant, grant_duration: Milliseconds<ClockInt>) -> TimedPermission {
+        now.checked_add(grant_duration)
```
kaczmarczyck: This change might need more testing. We will probably end up with a custom ClockInt to solve all issues around this? Assuming that happens, is a `saturating_add` an option, so less calls to `granted` don't actually do anything? Wrapping clocks are not super rare events with 24 bit actually. A graceful way to handle wrapping would be nice. The implementation right now tries to work around that.

src/ctap/mod.rs
```diff
@@ -61,7 +63,7 @@ use self::response::{
 };
 use self::status_code::Ctap2StatusCode;
 use self::storage::PersistentStore;
-use self::timed_permission::TimedPermission;
+use self::timed_permission::TimedPermission as TimedPermissionExp;
```
kaczmarczyck: Can we revert this renaming?

src/ctap/clock.rs
ia0: I thought the same when the file was added. I also think it should be at the same level as `embedded_flash` (since both are just helpers to abstract hardware). But I didn't say anything since the whole module hierarchy is going to change anyway at the end (the `ctap` module will disappear and possibly become its own crate). But if it's easy to move the module now then +1 for doing it.

src/ctap/clock.rs
ia0: I would rewrite this module to have one sub-module per hardware instance (for now only `prod` and `test`, we could rename `prod` to `tock`) as in the previous comment. This will minimize the number of `#[cfg(...)]` and make the file easier to read and also easier to migrate to the final state. Because each sub-module will be moved to a separate file/module specific to its hardware instance.

src/main.rs
```diff
@@ -187,9 +193,14 @@ fn main() {
 }
 
 #[cfg(feature = "debug_ctap")]
-fn print_packet_notice(notice_text: &str, timer: &Timer) {
-    let now = timer.get_current_clock().flex_unwrap();
-    let now_us = (Timestamp::<f64>::from_clock_value(now).ms() * 1000.0) as u64;
+fn print_packet_notice(notice_text: &str, clock: &CtapClock) {
+    use core::convert::TryFrom;
+    use embedded_time::duration::Microseconds;
```
kaczmarczyck: Please move imports to the top of the file.

src/ctap/mod.rs
kaczmarczyck: What about this?
```
const RESET_TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(10000 as ClockInt);
```

shilingwangggg: Done.

src/ctap/hid/mod.rs
shilingwangggg: Done.

src/ctap/mod.rs
shilingwangggg: Done.

src/main.rs
shilingwangggg: Done

src/ctap/hid/mod.rs
```diff
@@ -98,21 +99,15 @@ impl CtapHid {
     const COMMAND_KEEPALIVE: u8 = 0x3B;
     const COMMAND_ERROR: u8 = 0x3F;
     // TODO: optional lock command
-    const COMMAND_LOCK: u8 = 0x04;
```
kaczmarczyck: Can be left in to reduce diff, will be refactored later!

src/ctap/timed_permission.rs
kaczmarczyck: Discussed offline, this file can stay as is. Testing should happen in clock.rs, and this file gets looked at later.

src/ctap/clock.rs
```diff
@@ -0,0 +1,79 @@
...
+pub type ClockInt = u32;
+#[cfg(target_pointer_width = "64")]
+pub type ClockInt = u64;
+
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> embedded_time::Clock for LibtockClock<CLOCK_FREQUENCY> {
+    // TODO: Implement and use a 24-bits TimeInt for Nordic
+    type T = ClockInt;
+
+    const SCALING_FACTOR: embedded_time::fraction::Fraction =
```
ia0: By the way, did you try having a custom clock API before trying embedded-time? I see multiple advantages:
- We have control over the API, so we can more easily work on it.
- Example: We can make it more dynamic (like Tock), because now we would essentially need custom features to set the correct clock frequency depending on which Tock board we run on.
- Example: We can make the `Instant::new` const which would reduce diff in tests by keeping a dummy constant.
- Example: We could work incrementally. First the clock API is a copy/paste of Tock but in this module and the Tock implementation is trivial forwarding. Then we can start to clean it up by removing Tock-specific problems like the lifetime.
- We could still provide an implementation bridging between embedded-time and our custom API, such that we get free support for embedded-time implementations.
- Because we're not bound to embedded-time, we can also do the same with other time libraries like fugit.

What do you think?

src/ctap/clock.rs
shilingwangggg: I moved clock.rs one level up.

src/ctap/clock.rs
shilingwangggg: I changed it into one-liner and also tried to reduce the size of the PR by using inline parameter for test CtapInstant

src/ctap/hid/mod.rs
```diff
@@ -170,18 +171,16 @@ impl CtapHid {
...
     #[cfg(feature = "with_ctap1")]
-    const CAPABILITIES: u8 = CtapHid::CAPABILITY_WINK | CtapHid::CAPABILITY_CBOR;
+    const CAPABILITIES: u8 = Self::CAPABILITY_WINK | Self::CAPABILITY_CBOR;
     #[cfg(not(feature = "with_ctap1"))]
-    const CAPABILITIES: u8 =
-        CtapHid::CAPABILITY_WINK | CtapHid::CAPABILITY_CBOR | CtapHid::CAPABILITY_NMSG;
+    const CAPABILITIES: u8 = Self::CAPABILITY_WINK | Self::CAPABILITY_CBOR | Self::CAPABILITY_NMSG;
 
+    const TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(100 as ClockInt);
+    const WINK_TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(5000 as ClockInt);
```
ia0, line 178: Typo? Those lines moved before their comment.

src/ctap/timed_permission.rs
```diff
@@ -12,41 +12,52 @@
...
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Debug)]
 pub enum TimedPermission {
     Waiting,
-    Granted(ClockValue),
+    Granted(CtapInstant),
+}
+
+impl Clone for TimedPermission {
```
ia0: I don't see no reason to not derive `Copy` and `Clone` like before. Is this a leftover of a previous constraint?

src/ctap/timed_permission.rs
ia0: Actually here I see only 2 valid options:
- Saturate (can be left for later)
- Panic (in practice we probably don't grant huge durations so I would suggest that for this PR)

In particular, this looks wrong to return `Waiting`.

Note that `checked_add` doesn't return `None` in case of wrapping (which is normal behavior) but in case the duration is so big you can't tell which instant precedes the other when you have only 2 instants.

src/ctap/timed_permission.rs
```diff
@@ -12,41 +12,52 @@
...
 
     // Checks if the timeout is not reached, false for differing ClockValue frequencies.
-    pub fn is_granted(&self, now: ClockValue) -> bool {
+    pub fn is_granted(&self, now: CtapInstant) -> bool {
         if let TimedPermission::Granted(timeout) = self {
-            if let Some(remaining_duration) = timeout.wrapping_sub(now) {
-                return remaining_duration > Duration::from_ms(0);
-            }
+            let remaining_duration = timeout.checked_duration_since(&now);
+            return remaining_duration.map_or(false, |duration| duration.integer() > 0);
```
ia0: This could really just be `timeout.checked_duration_since(&now).is_some()`. The reason the previous code had `> 0` was because duration was signed and the `wrapping_sub` was making it negative if timed out.

src/ctap/timed_permission.rs
```diff
@@ -12,41 +12,52 @@
...
     // Returns a new state for differing ClockValue frequencies.
-    pub fn check_expiration(self, now: ClockValue) -> TimedPermission {
+    pub fn check_expiration(self, now: CtapInstant) -> TimedPermission {
         if let TimedPermission::Granted(timeout) = self {
-            if let Some(remaining_duration) = timeout.wrapping_sub(now) {
-                if remaining_duration > Duration::from_ms(0) {
-                    return TimedPermission::Granted(timeout);
-                }
+            let remaining_duration = timeout.checked_duration_since(&now);
+            if remaining_duration.map_or(false, |duration| duration.integer() > 0) {
```
ia0: Same as above.

src/ctap/client_pin.rs
```diff
@@ -1699,7 +1724,7 @@ mod test {
             Ok(())
         );
 
-        let timeout = DUMMY_CLOCK_VALUE.wrapping_add(Duration::from_ms(30001));
+        let timeout = CtapInstant::new(0) + (Milliseconds::new(30001_u32));
```
ia0: Looks like there's too many parentheses.

src/ctap/ctap1.rs
```diff
@@ -399,15 +393,19 @@ mod test {
         let mut env = TestEnv::new();
         env.user_presence()
             .set(|_| panic!("Unexpected user presence check in CTAP1"));
-        let mut ctap_state = CtapState::new(&mut env, START_CLOCK_VALUE);
+        let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0_u64));
```
ia0: I don't see why we need to write `0_u64` here and not in other files. It seems to compile just fine with `0`.

ia0 (commented):
I finally went over all files. LGTM modulo comments.

src/ctap/ctap1.rs
```diff
@@ -745,12 +805,16 @@ mod test {
...
-        ctap_state.u2f_up_state.grant_up(START_CLOCK_VALUE);
-        let response =
-            Ctap1Command::process_command(&mut env, &message, &mut ctap_state, TIMEOUT_CLOCK_VALUE);
+        ctap_state.u2f_up_state.consume_up(CtapInstant::new(0_u64));
+        ctap_state.u2f_up_state.grant_up(CtapInstant::new(0_u64));
+        let response = Ctap1Command::process_command(
+            &mut env,
+            &message,
+            &mut ctap_state,
+            CtapInstant::new((30001 * TEST_CLOCK_FREQUENCY_HZ as u64) / 1000),
```
kaczmarczyck, line 753: With embedded_time, is there a way to write this like:
`CtapInstant::new(0) + Milliseconds::new(30001)`
for easier readability?

src/ctap/hid/receive.rs
```diff
@@ -577,13 +578,13 @@ mod test {
             Ok(None)
         );
         for seq in 0..0x7F {
-            timestamp += delay;
+            timestamp = timestamp + delay;
```
kaczmarczyck, line 581: Curious why this is necessary, would we need to send a PR to embedded_time to allow this?

src/ctap/timed_permission.rs
kaczmarczyck: Since I expect this type to go away sooner than later, I'm happy with any choice that doesn't show worse behavior on real hardware. So if what Julien says works, fine with me!

src/env/tock.rs
```diff
@@ -157,15 +157,15 @@ pub fn switch_off_leds() {
 }
 
 const KEEPALIVE_DELAY_MS: isize = 100;
-pub const KEEPALIVE_DELAY: Duration<isize> = Duration::from_ms(KEEPALIVE_DELAY_MS);
+pub const KEEPALIVE_DELAY_TOCK: Duration<isize> = Duration::from_ms(KEEPALIVE_DELAY_MS);
```
kaczmarczyck: Couldn't we reduce the diff by not renaming here?

src/ctap/timed_permission.rs
```diff
@@ -118,71 +129,79 @@ impl U2fUserPresenceState {
...
+    fn zero() -> CtapInstant {
+        CtapInstant::new(0)
+    }
+
+    fn big_positive() -> CtapInstant {
+        CtapInstant::new(u64::MAX / 1000 - 1)
+    }
+
+    fn small_negative() -> CtapInstant {
+        CtapInstant::new(u64::MIN / 1000 + 1)
```
kaczmarczyck: Adding to the above comment, this is pointless now and can be removed.

src/ctap/client_pin.rs
shilingwangggg: I removed it. I think it is a left over from the previous iteration when there were many generics.

src/env/tock.rs
shilingwangggg: The problem is in the main.rs, there are 2 places using KEEPALIVE_DELAY. With this PR, we will need one of them using duration of embedded_time and the other with tock duration. That is why I made the change here. I believe in the future PR, this will not be a problem.

src/ctap/hid/receive.rs
shilingwangggg: Compiler: cannot use `+=` on type `embedded_time::Instant<clock::TestClock>`

src/ctap/timed_permission.rs
shilingwangggg: It is not leftover from previous constraint. Without it, the compiler complains:
cannot move out of `self.hid.wink_permission` which is behind a mutable reference
  --> src/lib.rs:73:36
   |
73 |         self.hid.wink_permission = self.hid.wink_permission.check_expiration(now);
   |                                    ^^^^^^^^^^^^^^^^^^^^^^^^ move occurs because `self.hid.wink_permission` has type `TimedPermission`, which does not implement the `Copy` trait

src/ctap/hid/mod.rs
shilingwangggg: I think this comment is invalid after I rebased to the head of develop on Monday.

src/ctap/timed_permission.rs
kaczmarczyck: You don't have to manually implement Copy, though, try:
```
#[derive(Clone, Copy, Debug)]
pub enum TimedPermission {
```

ia0: Could you try this? https://github.com/shilingwangggg/OpenSK/compare/feature/embedded_time...ia0:shiling_copy_clone?expand=1

It works for me. If it fails for you, could you provide me the command you use?

shilingwangggg: It works with adding copy and clone to derive. @ia0 you're right it is a left over from the previous iteration with the generics.

src/main.rs
shilingwangggg: Added unittests in clock module

src/ctap/hid/mod.rs
shilingwangggg: Done.

src/ctap/timed_permission.rs
shilingwangggg: Done.

src/ctap/timed_permission.rs
shilingwangggg: Done.

src/ctap/client_pin.rs
shilingwangggg: Done.

src/ctap/ctap1.rs
shilingwangggg: Done.

src/ctap/timed_permission.rs
shilingwangggg: Deleted.

src/ctap/timed_permission.rs
```diff
@@ -163,38 +146,31 @@ mod test {
...
-        assert!(!u2f_state.is_up_needed(start_time.checked_add(request_duration()).unwrap()));
     }
 
     fn grant_up_timeout(start_time: CtapInstant) {
         let mut u2f_state = U2fUserPresenceState::new(request_duration(), presence_duration());
         assert!(!u2f_state.consume_up(start_time));
         assert!(u2f_state.is_up_needed(start_time));
         u2f_state.grant_up(start_time);
-        // The timeout excludes equality, so it should be over at this instant.
-        assert!(!u2f_state.consume_up(start_time.checked_add(presence_duration()).unwrap()));
```
kaczmarczyck: Curious why this assert (and the one above) became invalid? Is this because equality is treated differently? What if you checked something like `start_time.checked_add(Milliseconds::new(1001_u64))`?

src/clock.rs
```diff
@@ -81,3 +81,69 @@ impl embedded_time::Clock for TestClock {
...
+    fn test_checked_add_overflow() {
+        assert_eq!(
+            CtapInstant::new(u64::MAX).checked_add(Seconds::new(1u64)),
+            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as u64 - 1u64))
+        );
+    }
+
+    #[test]
+    fn test_checked_add_error() {
+        assert!(CtapInstant::new(u64::MAX)
```
ia0: This doesn't need to be MAX for the test and I think it's less confusing to use 0 (or now like you do below) instead, to make it obvious the problem is not an overflow, it's just the duration which is too big.

src/clock.rs
```diff
@@ -81,3 +81,69 @@ impl embedded_time::Clock for TestClock {
...
+            .checked_add(Seconds::new(u64::MAX / TEST_CLOCK_FREQUENCY_HZ as u64))
+            .is_none());
+        let now = TestClock::new().try_now().unwrap();
+        assert!(now.checked_add(Seconds::new(u64::MAX)).is_none());
```
ia0: This doesn't test anything new right? Maybe use `MAX / FREQ / 2 + 1` which should also fail and is more strict than the one above.

src/clock.rs
```diff
@@ -81,3 +81,69 @@ impl embedded_time::Clock for TestClock {
...
+    }
+
+    #[test]
+    fn test_duration_since() {
+        let clock = TestClock::new();
+        let early = clock.try_now().unwrap();
+        let later = CtapInstant::new(1000u64);
```
ia0: Do we need to suffix with `u64`? Isn't the compiler able to infer the type? I think it's more readable to drop types when it can't lead to errors. IDEs are here to show them if needed.

src/clock.rs
```diff
@@ -81,3 +81,69 @@ impl embedded_time::Clock for TestClock {
...
+        assert_eq!(
+            later.checked_duration_since(&early).unwrap().integer(),
+            1000
```
ia0, line 134: This test relies on `try_now` returning 0. I would either:
- Use `CtapInstant::new(0)` for `early`.
- Or better make `TestClock` support advancing time.

```rust
pub struct TestClock {
    now: u64,
};
impl TestClock {
    pub fn advance(&mut self, ticks: u64) {
        self.now = self.now.wrapping_add(ticks);
    }
}
... fn try_now(...) -> ... { ... Instant::new(self.now) ... }
```

src/clock.rs
```diff
@@ -81,3 +81,69 @@ impl embedded_time::Clock for TestClock {
...
+            later.checked_duration_since(&early).unwrap().integer(),
+            1001
+        );
+        assert_eq!(early.checked_duration_since(&later), None);
+    }
+
+    #[test]
+    #[should_panic]
+    fn add_panic() {
+        let _ = CtapInstant::new(0) + Milliseconds(u64::MAX / 2 + 1);
```
ia0: Do we agree that the `MAX / 2 + 1` is approximative here? In theory the correct bound should involve the frequency like you did in tests above. Ideally tests involving a bound should have 2 sides: one working and one failing. For example testing `Seconds(MAX / FREQ / 2)` and `Seconds(MAX / FREQ / 2 + 1)`.

src/ctap/timed_permission.rs
ia0: Same remark as in the clock module regarding bound checking. We should check both sides of the bound:
- `start_time + presence_duration()`
- `start_time + presence_duration() + epsilon()` (or `- epsilon()`)

src/clock.rs
shilingwangggg: I changed it into Seconds(Max/req/2 + 1)

src/clock.rs
shilingwangggg: I deleted all the u64

src/clock.rs
shilingwangggg: Changed that into max/freq/2 + 1

src/clock.rs
shilingwangggg: Changed it to use 0

src/clock.rs
```diff
@@ -0,0 +1,152 @@
...
+    use super::*;
+    use embedded_time::duration::{Milliseconds, Seconds};
+
+    #[test]
+    fn test_checked_add() {
+        let clock = TestClock::new();
+        let now = clock.try_now().unwrap();
+        assert_eq!(
+            now.checked_add(Seconds::new(1 as ClockInt)),
+            Some(CtapInstant::new(TEST_CLOCK_FREQUENCY_HZ as ClockInt))
```
ia0, line 95: This test also assumes that `try_now` returns 0. We can also change the `try_now` into `Instant(0)`.

src/clock.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[test]
+    fn test_checked_add_error() {
+        assert!(CtapInstant::new(0)
+            .checked_add(Seconds::new(u64::MAX / TEST_CLOCK_FREQUENCY_HZ as ClockInt))
+            .is_none());
+        assert!(CtapInstant::new(0)
+            .checked_add(Seconds::new(
+                u64::MAX / TEST_CLOCK_FREQUENCY_HZ as ClockInt / 2 + 1
+            ))
+            .is_none());
```
ia0, line 125: Let's copy/paste this assert and remove the `+1` and make sure `is_some`? (I think it's the case but I might be wrong.)

ia0 (commented):
LGTM. I'll do a full review once it's rebased and ready to submit.

src/ctap/ctap1.rs
shilingwangggg: Done.

src/clock.rs
shilingwangggg: Done.

src/clock.rs
shilingwangggg: I changed all the try_now into instant(0)

src/ctap/ctap1.rs
```diff
@@ -417,32 +411,36 @@ mod test {
...
-        ctap_state.u2f_up_state.grant_up(START_CLOCK_VALUE);
-        let response =
-            Ctap1Command::process_command(&mut env, &message, &mut ctap_state, START_CLOCK_VALUE);
+        ctap_state.u2f_up_state.consume_up(CtapInstant::new(0));
+        ctap_state.u2f_up_state.grant_up(CtapInstant::new(0));
+        let response = Ctap1Command::process_command(
+            &mut env,
+            &message,
+            &mut ctap_state,
+            CtapInstant::new(0_u64),
```
kaczmarczyck, line 433: nit: can probably just be 0 without `_u64`?

src/clock.rs
kaczmarczyck: The advancing clock might actually break tests outside of this module. I vote for postponing this for PR that also introduces the 24 bit time int. That will have to look at tests as well, so good fit. Let's keep this refactoring only.

`CtapInstant::new(0)` sounds good!

kaczmarczyck (dismissed):

kaczmarczyck:
Only nits for me.

src/clock.rs
```diff
@@ -0,0 +1,156 @@
...
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> LibtockClock<CLOCK_FREQUENCY> {
+    pub fn new() -> Self {
+        let boxed_cb = alloc::boxed::Box::new(libtock_drivers::timer::with_callback(|_, _| {}));
+        let timer = alloc::boxed::Box::leak(boxed_cb).init().flex_unwrap();
+        Self(timer)
+    }
+}
+#[cfg(not(feature = "std"))]
+impl<const CLOCK_FREQUENCY: u32> fmt::Debug for LibtockClock<CLOCK_FREQUENCY> {
```
ia0, line 19: In the list of PR we could send to embedded-time crate we should add this. The clock shouldn't need to implement Debug. This is a side effect of deriving Debug for embedded-time::Instant which adds that condition while actually what is needed is `Clock::T` to implement Debug.

The other PR was Fabian asking about missing `AddAssign` implementation for `Instant`.

Could you follow up on this? Either create issues to the crate (if they don't already exist) or implement the fix and send the PR.

src/ctap/timed_permission.rs
ia0: I see 2 immediate options before merging here:
- Use `TimedPermission::Granted(now.checked_add(grant_duration).unwrap())`
- Add `TODO: Should panic or saturate if the grant duration is too big.`

src/ctap/timed_permission.rs
```diff
@@ -118,71 +116,84 @@ impl U2fUserPresenceState {
...
+    fn request_duration() -> Milliseconds<u64> {
+        Milliseconds::new(1000_u64)
+    }
+
+    fn presence_duration() -> Milliseconds<u64> {
+        Milliseconds::new(1000_u64)
+    }
+
+    fn epsilon() -> Milliseconds<u64> {
+        Milliseconds::new(1_u64)
```
ia0: Nit: Those `_u64` could be removed. They are already part of the return type.

ia0 (dismissed):
There's really just the comment in `timed_permission` which I think is not great, but we can merge and fix the remaining details in a follow-up PR too. Which would simplify merge conflicts.

src/ctap/timed_permission.rs
shilingwangggg: Done.

src/clock.rs
shilingwangggg: I'll follow up on this.

kaczmarczyck (approved):

ia0 (approved):

PR #423, opened 2022-01-17
Restrict Sha256 to be used sequentially

ia0:
Also remove useless HashBlockSize64Bytes bound.

This is to prepare for hardware cryptography support which often only supports one hash at a time. This is also not really a big restriction. Almost all usages of Sha256 are through `Hash256::hash` which is sequential by definition (OpenSK is not single-threaded). Also, if concurrent usage is needed, on can just store the updates in a `Vec` first.

### DIFF
--- a/libraries/crypto/src/hmac.rs
+++ b/libraries/crypto/src/hmac.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{Hash256, HashBlockSize64Bytes};
+use super::Hash256;
 use arrayref::array_ref;
 use subtle::ConstantTimeEq;
 
@@ -21,7 +21,7 @@ const HASH_SIZE: usize = 32;
 
 pub fn verify_hmac_256<H>(key: &[u8], contents: &[u8], mac: &[u8; HASH_SIZE]) -> bool
 where
-    H: Hash256 + HashBlockSize64Bytes,
+    H: Hash256,
 {
     let expected_mac = hmac_256::<H>(key, contents);
     bool::from(expected_mac.ct_eq(mac))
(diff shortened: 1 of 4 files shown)


kaczmarczyck (approved):
I wonder if the module level description should have a specific disclaimer that this is not a general purpose crypto library. Before, there were no obvious downsides of using it for anything else but OpenSK.

libraries/crypto/src/hmac.rs
```diff
@@ -31,15 +31,15 @@ where
...
-    H: Hash256 + HashBlockSize64Bytes,
+    H: Hash256,
 {
     let expected_mac = hmac_256::<H>(key, contents);
     bool::from(array_ref![expected_mac, 0, 16].ct_eq(pin))
 }
 
 pub fn hmac_256<H>(key: &[u8], contents: &[u8]) -> [u8; HASH_SIZE]
 where
-    H: Hash256 + HashBlockSize64Bytes,
```
jmichelp, line 42: I thought that `HashBlockSize64Bytes` was defining `BLOCK_SIZE` which is needed for hmac implementation.

ia0:
> I wonder if the module level description should have a specific disclaimer that this is not a general purpose crypto library.

Good point. However there's currently no crate-level or module-level documentation for the relevant code. We could create one, but see the point below.

> Before, there were no obvious downsides of using it for anything else but OpenSK.

This crate is probably unusable outside OpenSK. It refers to our custom `libtock-drivers` by path. I doubt anyone would try to use since it would be much easier to use the crates providing the same features in crates.io (better documentation, better integration with cargo, etc).

What do you think? Should I still add some crate-level documentation that this is a custom crate for OpenSK?

libraries/crypto/src/hmac.rs
ia0: Maybe it was, but not in the git history then. `BLOCK_SIZE` was a constant since the initial commit.

jmichelp: Ok. Sad. It would have made sense ¯\_(ツ)_/¯

ia0: Yes, I was surprised this trait wasn't used. It definitely looked like some leftover.

kaczmarczyck:
I approved to express that I think it's nice to have, but no complaints if you just submit. It could also be part of an eventual PR that properly moves to `///` etc.

PR #424, opened 2022-01-18
HMAC: only 32 byte keys

kaczmarczyck:
The crypto library now only supports 32 byte HMAC keys. The CTAP protocol only ever calculates HMAC with this specific key length. That makes it easier to support hardware implementations of HMAC, if they only support specific key lengths as well.

### DIFF
--- a/libraries/crypto/src/hmac.rs
+++ b/libraries/crypto/src/hmac.rs
@@ -46,6 +46,13 @@ pub fn hmac_256<H>(key: &[u8; KEY_SIZE], contents: &[u8]) -> [u8; HASH_SIZE]
 where
     H: Hash256,
 {
+    H::hmac(key, contents)
+}
+
+pub(crate) fn software_hmac_256<H>(key: &[u8; KEY_SIZE], contents: &[u8]) -> [u8; HASH_SIZE]
+where
+    H: Hash256,
+{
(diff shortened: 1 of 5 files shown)


libraries/crypto/src/hmac.rs
```diff
@@ -101,30 +106,6 @@ mod test {
...
-    #[test]
-    fn test_hmac_sha256_empty() {
-        let mut buf = [0; 96];
-        buf[..64].copy_from_slice(&[0x5c; 64]);
-        buf[64..].copy_from_slice(&Sha256::hash(&[0x36; 64]));
-        assert_eq!(hmac_256::<Sha256>(&[], &[]), Sha256::hash(&buf));
-    }
-
-    #[test]
-    fn test_hmac_sha256_examples() {
```
ia0, line 113: Shouldn't we keep those examples with fixed-length key? (i.e. empty message and non-empty message)

ia0 (dismissed):

libraries/crypto/src/hmac.rs
kaczmarczyck: The empty message case is contained in the tese below, but I added it back with a different key, together with the fox, because everyone likes foxes. :)

ia0 (approved):

kaczmarczyck:
Pushing through unrelated OSS fuzz problem.

PR #425, opened 2022-01-21
Remove the software-specific Decrypt16BytesBlock and Encrypt16BytesBlock traits

ia0:
They assume blocks are encrypted/decrypted one at a time. To avoid one syscall per block, it is preferable to encrypt/decrypt the data at once.

### DIFF
--- a/libraries/crypto/src/cbc.rs
+++ b/libraries/crypto/src/cbc.rs
@@ -16,8 +16,13 @@ use super::util::{xor_block_16, Block16};
 use crate::aes256::{DecryptionKey, EncryptionKey};
 use core::convert::TryInto;
 
-pub fn cbc_encrypt(key: &EncryptionKey, mut iv: Block16, blocks: &mut [u8]) {
-    for block in blocks.chunks_mut(16) {
+/// Encrypts a byte slice.
+///
+/// # Panics
+///
+/// Panics if the byte slice is not a multiple of the block size (16 bytes).
+pub fn cbc_encrypt(key: &EncryptionKey, mut iv: Block16, bytes: &mut [u8]) {
+    for block in bytes.chunks_mut(16) {
         let block: &mut Block16 = block.try_into().unwrap();
         xor_block_16(block, &iv);
         key.encrypt_block(block);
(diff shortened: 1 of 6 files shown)


libraries/crypto/src/cbc.rs
```diff
@@ -13,24 +13,21 @@
...
-use super::{Decrypt16BytesBlock, Encrypt16BytesBlock};
+use crate::aes256::{DecryptionKey, EncryptionKey};
+use core::convert::TryInto;
 
-pub fn cbc_encrypt<K>(key: &K, mut iv: Block16, blocks: &mut [Block16])
-where
-    K: Encrypt16BytesBlock,
-{
-    for block in blocks {
+pub fn cbc_encrypt(key: &EncryptionKey, mut iv: Block16, blocks: &mut [u8]) {
```
gendx: Nit: Given that the input doesn't take a slice of blocks anymore but a slice of bytes, it might make sense to rename the `blocks` parameter as something like `bytes`, as the new API doesn't guarantee the constraint that this parameter is a multiple of 16-byte blocks.

It would also be worth documenting what the behavior should be if the input is not a multiple of 16 bytes (i.e. the "remainder" is left as is rather than encrypted, which might be surprising at best, and at worst could lead to a security issue).

libraries/crypto/src/cbc.rs
```diff
@@ -13,24 +13,21 @@
...
+    for block in blocks.chunks_mut(16) {
+        let block: &mut Block16 = block.try_into().unwrap();
```
gendx, line 26: It's a bit unfortunate that the new API potentially adds runtime checks rather than checking lengths at compile time. At least Rust nightly now has `slice::array_chunks` which would be useful here (https://doc.rust-lang.org/std/primitive.slice.html#method.array_chunks).

libraries/crypto/src/cbc.rs
gendx: IMO it would be more robust to have an API that returns a `Result` or `panic`s if the input is not a multiple of 16 bytes.

ia0: Renamed to `bytes`. The implementation already panics if the input is not a multiple of 16 bytes. Added documentation.

libraries/crypto/src/cbc.rs
ia0: The length was already checked at runtime before, but it was done in the `crypto_wrapper.rs`. Now the check is moved here and made a panic to preserve the signature. Could be made an error if needed.

Yes, there's a comment somewhere about `array_chunks` but it's still unstable.

kaczmarczyck (approved):
No further complaints. Feel free to push despite the failing test when comments are resolved.

PR #427, opened 2022-03-02
Introduce a trait to abstract the CTAP environment

ia0:
The end goal is to provide users with:
- the `Env` trait that they should implement
- the `Ctap` struct that they can use

The change is mostly boilerplate threading the environment through all functions needing it.

### DIFF
--- a/src/ctap/ctap1.rs
+++ b/src/ctap/ctap1.rs
@@ -13,14 +13,12 @@
 // limitations under the License.
 
 use super::apdu::{Apdu, ApduStatusCode};
-use super::hid::ChannelID;
-use super::status_code::Ctap2StatusCode;
 use super::CtapState;
+use crate::env::Env;
 use alloc::vec::Vec;
 use arrayref::array_ref;
 use core::convert::Into;
 use core::convert::TryFrom;
-use crypto::rng256::Rng256;
 use libtock_drivers::timer::ClockValue;
 
 // For now, they're the same thing with apdu.rs containing the authoritative definition
@@ -180,15 +178,12 @@ impl Ctap1Command {
     const VENDOR_SPECIFIC_FIRST: u8 = 0x40;
     const VENDOR_SPECIFIC_LAST: u8 = 0xBF;
 
-    pub fn process_command<R, CheckUserPresence>(
+    pub fn process_command(
+        env: &mut impl Env,
         message: &[u8],
-        ctap_state: &mut CtapState<R, CheckUserPresence>,
+        ctap_state: &mut CtapState,
         clock_value: ClockValue,
-    ) -> Result<Vec<u8>, Ctap1StatusCode>
-    where
-        R: Rng256,
-        CheckUserPresence: Fn(ChannelID) -> Result<(), Ctap2StatusCode>,
-    {
+    ) -> Result<Vec<u8>, Ctap1StatusCode> {
         if !ctap_state
             .allows_ctap1()
             .map_err(|_| Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?
@@ -204,7 +199,7 @@ impl Ctap1Command {
                 if !ctap_state.u2f_up_state.consume_up(clock_value) {
                     return Err(Ctap1StatusCode::SW_COND_USE_NOT_SATISFIED);
                 }
-                Ctap1Command::process_register(challenge, application, ctap_state)
+                Ctap1Command::process_register(env, challenge, application, ctap_state)
             }
 
             U2fCommand::Authenticate {
@@ -220,6 +215,7 @@ impl Ctap1Command {
                     return Err(Ctap1StatusCode::SW_COND_USE_NOT_SATISFIED);
                 }
                 Ctap1Command::process_authenticate(
+                    env,
                     challenge,
                     application,
                     key_handle,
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -15,9 +15,44 @@
...
+use crate::env::Env;
+use libtock_drivers::timer::ClockValue;
 
 pub mod ctap;
 pub mod embedded_flash;
+pub mod env;
 
-#[macro_use]
-extern crate arrayref;
+pub struct Ctap<E: Env> {
```
kaczmarczyck, line 32: I think there should be more doc strings somewhere, and it's probably here. I assume you wanted us to take a general look before you add them?

src/main.rs
```diff
@@ -315,7 +338,8 @@ fn switch_off_leds() {
 
 fn check_user_presence(cid: ChannelID) -> Result<(), Ctap2StatusCode> {
     // The timeout is N times the keepalive delay.
-    const TIMEOUT_ITERATIONS: usize = ctap::TOUCH_TIMEOUT_MS as usize / KEEPALIVE_DELAY_MS as usize;
+    const TIMEOUT_ITERATIONS: usize =
+        ctap2::ctap::TOUCH_TIMEOUT_MS as usize / KEEPALIVE_DELAY_MS as usize;
```
kaczmarczyck, line 313: This `ctap2` is the name from the `Cargo.toml` right?

I was wondering how confusing all the `ctap`s will be. There is the
- `ctap`: mod
- `Ctap`: new struct
- `ctap2`: lib name
- `ctap1` which is almost the same, and actually more on the level of CtapState than its child, really

And technically, the names are wrong. The lib should actually be `ctap` with a `Ctap` struct, containing `ctap1` and `ctap2` next to each other. Then there is also no `ctap2::ctap::ctap1` anymore.

I'll open extra comments for HID and CTAP1, but feel free to respond only at one place since it's all the same question really. Also not necessarily for this PR, but I think it's good to discuss goal and big picture here.

src/ctap/ctap1.rs
```diff
@@ -180,15 +178,12 @@ impl Ctap1Command {
     const VENDOR_SPECIFIC_FIRST: u8 = 0x40;
     const VENDOR_SPECIFIC_LAST: u8 = 0xBF;
 
-    pub fn process_command<R, CheckUserPresence>(
+    pub fn process_command(
+        env: &mut impl Env,
         message: &[u8],
-        ctap_state: &mut CtapState<R, CheckUserPresence>,
+        ctap_state: &mut CtapState,
```
kaczmarczyck, line 184: This function takes a CtapState, just like the `process_*` in `hid` does. I'm thinking this should eventually move. Unlike all other commands (MakeCredential etc.), this function is actually called directly from HID.

src/ctap/hid/mod.rs
```diff
@@ -146,16 +146,13 @@ impl CtapHid {
 
     // Process an incoming USB HID packet, and optionally returns a list of outgoing packets to
     // send as a reply.
-    pub fn process_hid_packet<R, CheckUserPresence>(
+    pub fn process_hid_packet(
```
kaczmarczyck, line 149: I'm thinking where everything should live. Not actionable for now, but I'm looking into the refactoring of HID right now. To let you know about my ideas: And since most of the code in this crate is actually independent of `CtapState` and `ctap1`, maybe there is a good way to move all of `hid` out of the `ctap` mod.

ia0: Yes, there's definitely a lot of refactoring that can/should be done. This PR only starts to incrementally address the high-level picture of the library.

src/main.rs
ia0: Yes this is definitely not for this PR and should be addressed at some point. When the crate was split between main and lib for fuzzing it was done half-way. In particular the modules were used by both main and lib. This PR fixes this. As such, when main is referring to the library, it has to use the crate name ctap2. I agree that in the end we probably want the library to be called opensk or opensk-ctap or something like that, and it would contain just what the user needs to know with the minimum amount of modules. I'm not even sure ctap2 and ctap1 should be visible to the user. Probably only as cargo features. But that's to be designed later.

src/lib.rs
ia0: The idea was more to get the code slowly in shape and add documentation once it starts to be stable. Are you thinking about something more than "CTAP implementation parameterized by its environment"? (In the future there will be examples. The main doc should probably be at crate-level though, explaining what is CTAP etc.)

src/ctap/ctap1.rs
ia0: Yes I was surprised to see those functions take a `CtapState`. I didn't try to see why yet, but that's indeed something that could be addressed during refactoring.

src/lib.rs
kaczmarczyck: SGTM
Most questions were not necessarily about this PR directly, but observations I wanted to share and address eventually.

src/main.rs
kaczmarczyck: Sounds good, leaving this open for reference.

src/ctap/ctap1.rs
kaczmarczyck: I think it's just the hierarchy of our current design. `CtapState` is to `hid` what `check_user_presence` is for `ctap`. The packet gets processed from receive to send, and the CTAP logic is called in between.

kaczmarczyck: The above was the answer for HID, for CTAP1, the `CtapState` mostly takes the role of `env`, but there is also `decrypt_credential_source` (and encrypt). These functions both don't depend on `CtapState`, but on `env` themselves, so the new design should seperate them.

kaczmarczyck (dismissed):
I'm happy, will add some other people FYI and potentially review.

src/lib.rs
ia0: Added some documentation.

ia0:
I've pushed a commit to already have the environment for Tock. This way it should be more stable where should code go (Tock-related stuff should only be in `env::tock` in the end) although we could still rename things, this is just for organization. I initially didn't do it because of `check_user_presence` but I think it's better to have it well organized and ugly (TockEnv parametrized over check_user_presence) than not organized closer to the final state from API point of view.

jmichelp (approved):

kaczmarczyck (approved):

PR #428, opened 2022-03-02
Make user feedback fields private

ia0:
(This is based on top of #427.)

PR #430, opened 2022-03-03
Move check_user_presence to env::tock

ia0:
(This is based on top of #428.)

This permits to make the binary (main) not rely on implementation details of the library. However, fuzzing still depends on it (and might always? that's a question for fuzzing refactoring).

### DIFF
--- a/src/env/tock.rs
+++ b/src/env/tock.rs
@@ -1,35 +1,36 @@
-use crypto::rng256::TockRng256;
-
-use crate::ctap::hid::ChannelID;
+use crate::ctap::hid::{ChannelID, CtapHid, KeepaliveStatus, ProcessedPacket};
 use crate::ctap::status_code::Ctap2StatusCode;
 use crate::env::{Env, UserPresence};
+use core::cell::Cell;
+#[cfg(feature = "debug_ctap")]
+use core::fmt::Write;
+use crypto::rng256::TockRng256;
+use libtock_core::result::{CommandError, EALREADY};
+use libtock_drivers::buttons::{self, ButtonState};
+#[cfg(feature = "debug_ctap")]
+use libtock_drivers::console::Console;
+use libtock_drivers::result::{FlexUnwrap, TockError};
+use libtock_drivers::timer::Duration;
+use libtock_drivers::{led, timer, usb_ctap_hid};
 
-pub struct TockEnv<CheckUserPresence: Fn(ChannelID) -> Result<(), Ctap2StatusCode>> {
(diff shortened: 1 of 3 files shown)


kaczmarczyck (commented):
I found the name of this PR a bit misleading, since it also adds public members to `src/env/tock.rs`, like `pub fn wink_leds`. I think it's fine to do that for now, but I was wondering what the general idea is for what should stay in `main.rs` and what moves over?

I assume that the reason these function are allowed in `env/tock.rs` is because we expect them to be part of env enventually?

ia0:
Good question, I probably should have added some comments.

> I was wondering what the general idea is for what should stay in main.rs and what moves over?

I think this is up to us, because by providing a main, we play the role of the user. The way I see the steps to use the library:
- (We provide the `Ctap` implementation)
- The user (or us) provide an `Env` implementation
- The user (or us) provide a "runtime" which "runs" a `Ctap` instance (I'm not sure exactly how this would look like)
- The user (or us) provide a binary which only does CTAP (this is our main)
- The user (or us) provide a script to deploy a CTAP binary (this is our deploy script)

My current opinion (very weak, I would need to experiment) is that the main should be as small as possible (one-liner), the runtime should also be as small as possible (some kind of event loop like now or some interrupt-driven runtime like RTIC).

To come back to the current concern, I'm not sure "user feedback" should be part of the environment, but it should probably be provided for convenience. Maybe we need to introduce a new concept for that. So for now I just went with putting them as free function in under the `env` module (since they go together with the actual type of environment, like `tock` or `test`).

> I assume that the reason these function are allowed in env/tock.rs is because we expect them to be part of env enventually?

Either that, or because they are Tock-related. But in that case, we could also have the following alternative. Instead of having `tock` and `test` under `env`. We have another concept (maybe `board`) which would contain `tock` and `test` and would provide an `Env` implementation as well as other useful (but possibly non-necessary) stuff.

kaczmarczyck (approved):
This seems to be the easiest and most convenient solution at the moment. To be more concrete on the points you make:

**main vs runtime**
We might want this eventually, but no need to change main further until we have a clear goal.

**move user interface**
I do feel like that's cleaner, but we should only move after we have a plan for the user interface API I think

**about user interface**
A general API should also cover use cases where (Wink and?) UP are not necessarily simply active for some time, but feel more like events. A desktop software implementation e.g. would have an event that pops open a window to ask for UP, and another event that notifies the user of the time out.

Nothing for this PR, should be good for now.

PR #431, opened 2022-03-03
Add Storage and UpgradeStorage to Env

ia0:
(This is based on top of #430.)

Also introduce the `api` module for the traits used by `Env`. Until now the traits were already defined somewhere else (crypto or store) but the upgrade storage trait is new. In theory `UserPresence` should be moved in `api` but given it's a temporary trait until we get clock, usb, button, and possible led, it's probably not worth moving to `api`.

Also remove the now merged `embedded_flash` module.


src/env/tock/mod.rs
```diff
@@ -12,18 +14,48 @@ use libtock_drivers::console::Console;
...
 
 impl TockEnv {
-    pub fn new() -> Self {
-        let rng = TockRng256 {};
-        TockEnv { rng }
+    /// Returns the unique instance of the Tock environment.
+    ///
+    /// This function returns `Some` the first time it is called. Afterwards, it repeatedly returns
+    /// `None`.
+    pub fn new() -> Option<Self> {
```
kaczmarczyck: Do we expect any caller to not `unwrap`, but actually use the fact that it's an `Option`? Or is the motivation a more general design?

src/env/tock/mod.rs
```diff
@@ -12,18 +14,48 @@ use libtock_drivers::console::Console;
...
+            upgrade_storage: false,
+        })
     }
 }
 
+/// Creates a new storage instance.
+///
+/// # Safety
+///
+/// It is probably technically memory-safe to hame multiple storage instances at the same time, but
```
kaczmarczyck: type "hame"

src/env/tock/mod.rs
```diff
@@ -12,18 +14,48 @@ use libtock_drivers::console::Console;
...
+/// for extra precaution we mark the function as unsafe. To ensure correct usage, this function
+/// should only be called if the previous storage instance was dropped.
+// This function is exposed for example binaries testing the hardware. This could probably be
```
kaczmarczyck: "exposed to"?

src/env/tock/mod.rs
```diff
@@ -41,6 +75,22 @@ impl Env for TockEnv {
...
     }
+
+    fn storage(&mut self) -> StorageResult<Self::Storage> {
+        assert_once(&mut self.storage);
+        unsafe { steal_storage() }
+    }
+
+    fn upgrade_storage(&mut self) -> StorageResult<Self::UpgradeStorage> {
+        assert_once(&mut self.upgrade_storage);
+        SyscallUpgradeStorage::new()
```
kaczmarczyck, line 85: I wonder where asymmetry between `storage` and `upgrade_storage` comes from. Is this because some examples use `storage`, but not (yet) `TockEnv`? Or is it related to this comment?
> cleaned up by having the persistent store return its storage

src/env/tock/mod.rs
```diff
@@ -12,18 +14,48 @@ use libtock_drivers::console::Console;
...
+/// Creates a new storage instance.
+///
+/// # Safety
+///
+/// It is probably technically memory-safe to hame multiple storage instances at the same time, but
+/// for extra precaution we mark the function as unsafe. To ensure correct usage, this function
+/// should only be called if the previous storage instance was dropped.
+// This function is exposed for example binaries testing the hardware. This could probably be
+// cleaned up by having the persistent store return its storage.
+pub unsafe fn steal_storage() -> StorageResult<SyscallStorage> {
```
kaczmarczyck, line 54: Not sure what the name implies. Does it mean that in case someone else called `steal_storage` before, their instance becomes unsuable?
Or is there some convention about `steal_` functions I don't know? If the only reason for this functions is to keep examples alive until they get ported to TockEnv, then the name is fine with me.

src/ctap/mod.rs
```diff
@@ -330,20 +330,21 @@ impl StatefulPermission {
 
 // This struct currently holds all state, not only the persistent memory. The persistent members are
 // in the persistent store field.
-pub struct CtapState {
-    persistent_store: PersistentStore,
+pub struct CtapState<E: Env> {
+    persistent_store: PersistentStore<E>,
```
kaczmarczyck, line 334: The PersistentStore line is the reason why we have to introduce `E: Env` here (and in `Ctap`), right? I wonder if there is a way around this.

ia0: Yes that's the reason.

The way around would be to have the persistent store part of the environment directly. Let me try to write a new PR with that idea (and include the fix for all open comments here that would still be applicable).

src/env/tock/mod.rs
ia0: This mimics how peripherals are taken when using `svd2rust`:
- Example: https://docs.rs/nrf52840-hal/latest/nrf52840_hal/pac/struct.Peripherals.html#method.take
- Documentation: https://docs.rs/svd2rust/0.21.0/svd2rust/#peripheral-api

I don't know yet how examples should look like. Also, I think I should be able to work around this function anyway at some point.

src/env/tock/mod.rs
ia0: Yes, it's because of the examples. If they use `TockEnv` and I make the PersistentStore return its storage (as the comment you selected), then the asymmetry should disappear.

src/env/tock/mod.rs
ia0: Probably :-) Changed.

src/env/tock/mod.rs
ia0: Fixed.

src/env/tock/mod.rs
ia0: Good question. Indeed it's probably better to panic here. In particular because I foresee that we may panic for other reasons during setup in this function.

ia0 (commented):
I addressed all comments but the one to move `PersistentStore` to the `Env` API, thus removing it from `CtapState`.

src/env/tock/mod.rs
kaczmarczyck: I assumed it's not necessarily a permanent solution. I don't hate it, so resolved.

src/ctap/mod.rs
ia0: I created #438 and would suggest merging that one instead of this one. I'll rebase once the one to be merged is identified to only rebase that one.

ia0:
This was merged as part of #438 (which is actually a rewrite, so not really merged).

PR #432, opened 2022-03-03
Adds SHA256 Cryptocell code to bootloader

kaczmarczyck:
The last piece to make the bootloader actually work. Advantage of a hardware SHA256 is the much faster boot time.

### DIFF
--- a/bootloader/src/bitfields.rs
+++ b/bootloader/src/bitfields.rs
@@ -1,4 +1,4 @@
-// Copyright 2020-2021 Google LLC
+// Copyright 2020-2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/bootloader/src/bitfields.rs
+++ b/bootloader/src/bitfields.rs
@@ -1,0 +1,118 @@
+// Copyright 2020-2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use tock_registers::register_bitfields;
+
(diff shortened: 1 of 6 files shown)


bootloader/src/bitfields.rs
```diff
@@ -0,0 +1,118 @@
+// Copyright 2020-2021 Google LLC
```
jmichelp: 2022 now :)

bootloader/src/crypto_cell.rs
```diff
@@ -0,0 +1,283 @@
+// Copyright 2019-2021 Google LLC
```
jmichelp: Same here

bootloader/src/crypto_cell.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn clear_data(&self) {
+        let mut ctx = self.hash_ctx.get();
+        ctx.iter_mut().for_each(|b| *b = 0);
+        self.hash_ctx.set(ctx);
+        self.hash_total_size.set(0);
+    }
+
+    /// Adds data to the current hash computation.
+    ///
+    /// You have to know in advance is this is going to be the last block, and indicate that
```
jmichelp: typo: advance if this is going

bootloader/src/registers.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright 2020-2021 Google LLC
```
jmichelp: Again :)

jmichelp (commented):
As we may need the Cryptocell code also in Tock, I wonder if this wouldn't make sense to make it a separate crate that we could import in both the bootloader and the kernel. Otherwise we'll end up with code duplication and different kinds of API to compute hashes.

kaczmarczyck:
Question is if you want the new crate now or submit and put the work in when / if we submit it to Tock?

This PR is simpler than what I think you had in mind for Tock, i.e. it's API doesn't allow concurrent calls. So I don't know if a simply move to a new crate would actually be that useful?

jmichelp:
We can merge it as-is at the moment to get the feature out.
And let's align with @ia0 to make it a separate crate that can bootstrap the work around hardware crypto support.

jmichelp (approved):

ia0:
> And let's align with @ia0 to make it a separate crate that can bootstrap the work around hardware crypto support.

Ack! (I don't have any ideas yet, and that might be related to ctap as a library too, so should definitely be in our radar).

kaczmarczyck:
Thanks!

PR #433, opened 2022-03-04
Add firmware protection to the environment

ia0:
(This is based on top of #431.)

After this PR, if I'm not mistaken, the only tock-related things in `src/ctap` are console- and clock-related.


src/env/test/mod.rs
```diff
@@ -38,11 +39,18 @@ impl UserPresence for TestUserPresence {
     }
 }
 
+impl FirmwareProtection for TestEnv {
+    fn lock(&mut self) -> bool {
+        true
```
kaczmarczyck, line 73: I don't know off the top of my hat, but does
`crp::set_protection(crp::ProtectionLevel::FullyLocked).is_ok()`
return false the second time you call it?

In that case, we could give the Test version a bool to store its state, and replicate that behavior.
Since workflows are green, we
1. either don't have a test for that (not sure if we should care)
2. or TockEnv always returns true and you can ignore this comment.

Looking at the [only place I think we test](https://github.com/google/OpenSK/blob/91ba2c375eeb31dfebde7bbba6cf95dc97181c12/src/ctap/mod.rs#L2944), it's option (1).

ia0: Interesting, I didn't see that `set_protection` returns an error if there's nothing to do. That's a curious design. The documentation for [`lockdown`](https://github.com/google/OpenSK/blob/f2496a8e6d71a4e838884996a1c9b62121f87df2/src/ctap/command.rs#L419) doesn't say that the user should set it at most once. And I think it would be surprising for the user to get `VENDOR_INTERNAL_ERROR` if they try to lock twice. So I'm fixing the tock environment implementation instead and keep the test implementation as is.

Thanks for finding this out!

kaczmarczyck: I agree that `VENDOR_INTERNAL_ERROR` is misleading. Maybe the original intention was that if you try to lock your device twice, you might have some wrong asusmptions about the state of your device that you shuld be made aware of. 

@jmichelp FYI as the original author of the vendor configure command, in case that was intentional and I forgot the reason. Otherwise LGTM from my side.

kaczmarczyck (dismissed):
LGTM for everything that is not in #431

src/env/test/mod.rs
jmichelp: I tried to think the driver interface as being generic enough to apply to different kind of chips. So under the hood (not exposed at the CTAP level), the driver allows different levels of protection (None, JtagDisabled, FullyLocked) with the idea that you can always upgrade the protection level but you can't downgrade. Returning `EALREADY` permits to notify that you're not trying to re-apply the same protection level (some chips may not permit it).
At the CTAP level, we don't really care whether we consider `EALREADY` an error or a success. At best, treating this as an error saves a syscall to distinguish between a locked and non-locked device.

More important error cases are when someone tries to lock a device that isn't fully programmed (cert + key) as this may create a non-working device.

I'll have to update the CRP logic anyway because the newer revisions of the Nordic chip changed the way you lock the device.

ia0: > At the CTAP level, we don't really care whether we consider EALREADY an error or a success.

That's the question. Currently we treat it as an internal error. So I guess you're fine with both considering it as a user error (what Fabian mentioned) or considering it as a way to make sure it's locked (what I suggests, i.e. operations are idempotent and can be repeated as no-ops).

> At best, treating this as an error saves a syscall to distinguish between a locked and non-locked device.

I'm not sure I see which syscall is saved since we do only one syscall in the configure command (not counting storage which we do before anyway).

> More important error cases are when someone tries to lock a device that isn't fully programmed (cert + key) as this may create a non-working device.

Yes, that's the current code (inside the ctap library) which I'm not changing.

> I'll have to update the CRP logic anyway because the newer revisions of the Nordic chip changed the way you lock the device.

I suspect this will be probably orthogonal if it's only Tock changes (and not the configure command of the ctap implementation).

jmichelp: > > At the CTAP level, we don't really care whether we consider EALREADY an error or a success.
> 
> That's the question. Currently we treat it as an internal error. So I guess you're fine with both considering it as a user error (what Fabian mentioned) or considering it as a way to make sure it's locked (what I suggests, i.e. operations are idempotent and can be repeated as no-ops).


Yes, both are fine.

> > At best, treating this as an error saves a syscall to distinguish between a locked and non-locked device.
> 
> I'm not sure I see which syscall is saved since we do only one syscall in the configure command (not counting storage which we do before anyway).

With the current implementation, you have 2 syscalls implemented for the protection level: `get` and `set`.
If you want to know if the key is already protected to a given level, you can call `get` and compare the returned value or, with the current implementation, you can call `set` and the error `EALREADY` tells you it's already done.
But that's a corner case.

> > I'll have to update the CRP logic anyway because the newer revisions of the Nordic chip changed the way you lock the device.
> 
> I suspect this will be probably orthogonal if it's only Tock changes (and not the configure command of the ctap implementation).

It should be Tock-only changes (only the driver file basically). But I'll have to check the new process in the datasheet. Maybe it requires to reboot the chip between flash ops which may make the change more complex.

ia0: > With the current implementation, you have 2 syscalls implemented for the protection level: get and set.
> If you want to know if the key is already protected to a given level, you can call get and compare the returned value or, with the current implementation, you can call set and the error EALREADY tells you it's already done.
> But that's a corner case.

Ah I see, makes sense. So the "treating this as an error" was for the syscall API, not the configure command. Sounds good.

kaczmarczyck (approved):

PR #434, opened 2022-03-04
Freeze cargo fuzz's version and add fuzz corpus to gitignore

hcyang-google:
Freeze cargo fuzz's version since newer versions require newer nightly toolchain.

### DIFF
--- a/fuzzing_setup.sh
+++ b/fuzzing_setup.sh
@@ -20,4 +20,4 @@ done_text="$(tput bold)DONE.$(tput sgr0)"
 set -e
 
 # Install cargo-fuzz library.
-cargo +stable install cargo-fuzz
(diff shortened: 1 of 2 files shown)


fuzzing_setup.sh
```diff
@@ -20,4 +20,4 @@ done_text="$(tput bold)DONE.$(tput sgr0)"
 set -e
 
 # Install cargo-fuzz library.
-cargo +stable install cargo-fuzz
+cargo +stable install cargo-fuzz --version 0.10.0
```
kaczmarczyck: I think there is a version 0.10.2 that also worked for me. Can you double check?

hcyang-google: Updated to 0.10.2

kaczmarczyck (approved):

PR #436, opened 2022-03-04
HID refactoring

kaczmarczyck:
This PR has 2 commits. The first
- introduces the `CtapHidCommand` enum
- splits off logic to parse `Message`s with a new `process_message` that contains core logic
- fixes the documentation

After this bigger cleanup, the second commit actually works on the goal of this PR: To have an API that separates all CTAP logic, and is the first step on the road to a pure HID library. It does so without changing the API. The plan is to move `process_hid_packet` (which is now very short) out of the library, and have two instances with different `process_message` functions later. This will also allow removing the `wink_permission` eventually.

If you rather review in 2 PRs, I'll split the second commit to a new PR (or I could also split commit 1 if necessary). In this case, consider this the overview.

Plan for future submissions is:
- Refactor the tests in `hid/mod.rs`, the new interface is much more test friendly.
- Do the aforementioned move to make this a pure library (possible multiple steps actually).
- Introduce the second USB transport.


src/ctap/hid/mod.rs
```diff
@@ -51,79 +88,84 @@ pub enum ProcessedPacket<'a> {
...
+/// To process a packet and receive the response, you can call `process_hid_packet`.
+/// If you want more control, you can also do the processing in steps:
+///
+/// 1.  `HidPacket` -> `Option<Message>`
+///     `parse_packet` assembles the message and preprocesses all pure HID commands and errors.
+/// 2.  `Option<Message>` -> `Message`
+///     If you didn't receive any message or preprocessing discarded it, stop.
+/// 3.  `Message` -> `Message`
+///     `process_message` handles all protocol interactions.
+/// 4.  `Message` -> `HidPacketIterator
```
ia0: Missing backtick. Did you try `cargo doc --document-private-items --features=std --open` to check the output? (there's also wrong indentation line 124)

ia0 (dismissed):

src/ctap/hid/mod.rs
kaczmarczyck: Fixed. Thanks for the pointer. I noticed the formatting didn't work the way I wanted it, so I changed it a bit. Also added more documentation for what I think is the forseeable public interface.

ia0 (dismissed):

ia0 (approved):

kaczmarczyck:
The fuzzing check passes locally, therefore I'll submit and fix that separately.

ia0:
> The fuzzing check passes locally, therefore I'll submit and fix that separately.

Note also that it's the CIFuzz job that's failing, not the "Cargo fuzz build" one, so it's working locally even for GitHub. It's the docker setup the issue.

PR #438, opened 2022-03-07
Add Store and UpgradeStorage to Env

ia0:
This is the counter-part of #431. The differences are:
- `Env` provides access to the `persistent_store::Store` instead of its underlying storage.
- The store and upgrade storage access in `Env` is through mutable references like other components.

This solution is better aligned with the rest of the APIs and also avoids some quirks due to storing the `Store` under `CtapState` thus leaking the environment to its type.


kaczmarczyck (approved):
I like this version better than #431. Depending on how fast #422 is finished, we should sync offline which to submit first.
Also I noticed a few functions in `src/ctap/mod.rs` that take a `self`, but don't actually need it anymore. Not to be fixed by this PR (I'll look into that soon I think), but it's a good sign that your latest patches made refactoring this large file easier!

PR #439, opened 2022-03-07
roll back fuzzing install script, documentation instead

kaczmarczyck:
The script `fuzzing_setup.sh` is used by OSS-Fuzz, and their workflow failed when we changed the cargo fuzz version. As an immediate fix for the workflow, I reverted that change and replaced it with documentation for how to run fuzzing as a user.

Since these changes only propagate after merging, the fuzzing workflow is not expected to pass with this PR.

FYI @hcyang-google

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -64,3 +64,32 @@ a few things you can personalize:
     *   Whether you want to use batch attestation.
     *   Whether you want to use signature counters.
     *   Various constants to adapt to different hardware.
+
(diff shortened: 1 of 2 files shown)


ia0 (dismissed):

kaczmarczyck:
Sorry, I forgot the markdown lint, the code blocks are now happy :)

ia0 (approved):

PR #440, opened 2022-03-08
Replace steal_storage with take_storage

ia0:
(This is a single commit based on top of #438.)


kaczmarczyck (approved):
Approval for the extra commit.

Issue #420, opened 2021-12-15
Ordering invariant does not hold between `cbor_map!` and `destructure_cbor_map!` on the develop branch


gendx:
## Expected Behavior

In the `develop` branch, the `cbor_map!` macro's documentation indicates that keys do not need to be sorted: https://github.com/google/OpenSK/blob/develop/libraries/cbor/src/macros.rs#L143.

It is therefore expected that a map built like:

```rust
let map = cbor_map! {
    1 => 10,
    2 => 20,
};
```

can be correctly used with `destructure_cbor_map!` as follows:

```rust
    #[test]
    fn test_destructure_unsorted_cbor_map() {
        let map = cbor_map! {
            2 => 20,
            1 => 10,
        };

        destructure_cbor_map! {
            let {
                1 => x1,
                2 => x2,
            } = extract_map(map);
        }

        assert_eq!(x1, Some(cbor_unsigned!(10)));
        assert_eq!(x2, Some(cbor_unsigned!(20)));
    }
```

## Actual Behavior

This test fails with:

```
failures:

---- macros::test::test_destructure_unsorted_cbor_map stdout ----
thread 'macros::test::test_destructure_unsorted_cbor_map' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some(Unsigned(10))`', src/macros.rs:667:9
```

This is because `destructure_cbor_map!` expects that both the destructured keys (left-hand-side of the arrow) and the keys in the map itself are sorted. The former constraint is well documented by `destructure_cbor_map!` (and checked via unit tests for lack of static assertions), but the latter does not hold if inputs are not sorted in `cbor_map!` (which just copies the key-value pairs into a vectors without doing any sorting).

### Affected code

1. #303 did change a bunch of `cbor_map!` expressions to actually have sorted keys in their inputs, so I don't think the OpenSK usage of `sk-cbor` is affected.
1. The `stable` branch shouldn't be affected, because it's still using the `BTreeMap` representation of map values (https://github.com/google/OpenSK/pull/303 wasn't applied).
1. A quick look at the dependents of the `sk-cbor` crate (https://crates.io/crates/sk-cbor/reverse_dependencies), as well as a global GitHub search (https://github.com/search?q=destructure_cbor_map&type=Code) shows that `destructure_cbor_map!` does not seem to be used outside of OpenSK.

So there doesn't seem to be code affected by this bug at the moment.

## Steps to Reproduce the Problem

[...]

PR #441, opened 2022-03-14
Generate valid structure for MakeCredential params

hcyang-google:
The fuzzers are not working well currently because the logic is to parse random bytes as cbor value and parse them as specific command format. If the format doesn't match, it returns and goes to the next round. With random inputs, this has ~0% probability of generating valid input, so we're only fuzzing the command parsing (ineffectively too, as the bytes can't be interpreted as cbor maps most of the time). We should generate valid inputs in each fuzzing round with high probability.

To do that, we need to generate valid cbor maps that can be parsed into specific parameters (like AuthenticatorMakeCredentialParameters). Here, I mostly take use of the arbitrary crate that is the default choice for Rust fuzzing. I decided to put the extra code in the ctap package itself behind a feature flag because otherwise we'll have to bring many extra dependency and make many types/functions public to implement it in the fuzz package. Even if we do that it's not enough, because we can't implement foreign traits on foreign types, so we'll likely have to rewrite our own arbitrary library as well. I feel that the way I add the fuzz-related code to the ctap library doesn't really complicates the code base so it should be fine. If you have better suggestions please tell me, let us first fix this CL to an acceptable format before we implement similar logic to other fuzz targets.

After this CL, we have 91.67% function coverage in ctap/hid/mod.rs (compared to 0%), and decently covers main logic in ctap/mod.rs for MakeCredential command processing. Main limitation of coverage is crypto-related checks, like we return an error if PubKeyCredentialsParams doesn't include ES256. These implementation-specific improvements can be added later.

Finally, I want to discuss our attitudes toward introducing probability in fuzzing. Theoretically as long as the inputs are reachable, we should let the fuzzer find its paths, and not choosing some specific paths for very high probability. But in practice we need to use probabilities or else fuzzing is very ineffective or even useless. I used 95% for valid structure, but for other cases (like whether PubKeyCredentialsParams include ES256), do we want to set higher probabilities for those "more interesting" paths too?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ subtle = { version = "2.2", default-features = false, features = ["nightly"] }
 # This import explicitly locks the version.
 serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 embedded-time = "0.12.1"
+arbitrary = { version = "0.4.7", features = ["derive"], optional = true }
 
 [features]
 debug_allocations = ["lang_items/debug_allocations"]
@@ -32,6 +33,7 @@ verbose = ["debug_ctap", "libtock_drivers/verbose_usb"]
 with_ctap1 = ["crypto/with_ctap1"]
 with_nfc = ["libtock_drivers/with_nfc"]
 vendor_hid = []
+fuzz = ["arbitrary", "std"]
 
 [dev-dependencies]
 enum-iterator = "0.6.0"
--- a/fuzz/fuzz_helper/Cargo.toml
+++ b/fuzz/fuzz_helper/Cargo.toml
@@ -11,5 +11,6 @@ embedded-time = "0.12.1"
 libtock_drivers = { path = "../../third_party/libtock-drivers" }
 crypto = { path = "../../libraries/crypto", features = ['std'] }
 sk-cbor = { path = "../../libraries/cbor" }
-ctap2 = { path = "../..", features = ['std'] }
+ctap2 = { path = "../..", features = ["fuzz"] }
 lang_items = { path = "../../third_party/lang-items", features = ['std'] }
+arbitrary = { version = "0.4.7", features = ["derive"] }
--- a/fuzz/fuzz_helper/src/lib.rs
+++ b/fuzz/fuzz_helper/src/lib.rs
@@ -16,17 +16,18 @@
 // `libtock_alloc_init` symbol.
 extern crate lang_items;
 
+use arbitrary::{Arbitrary, Unstructured};
 use arrayref::array_ref;
 use core::convert::TryFrom;
 use ctap2::clock::CtapInstant;
-use ctap2::ctap::cbor_read;
 use ctap2::ctap::command::{
     AuthenticatorClientPinParameters, AuthenticatorGetAssertionParameters,
-    AuthenticatorMakeCredentialParameters,
+    AuthenticatorMakeCredentialParameters, Command,
 };
 use ctap2::ctap::hid::{
     ChannelID, CtapHidCommand, HidPacket, HidPacketIterator, Message, MessageAssembler,
 };
+use ctap2::ctap::{cbor_read, Channel, CtapState};
 use ctap2::env::test::TestEnv;
 use ctap2::{Ctap, Transport};
 
@@ -164,6 +165,39 @@ pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
     }
     command.extend(data);
     process_message(&command, &mut ctap);
+}
+
+pub fn process_ctap_structured(data: &[u8], input_type: InputType) -> arbitrary::Result<()> {
+    let unstructured = &mut Unstructured::new(data);
+
+    let mut env = TestEnv::new();
+    let mut state = CtapState::new(&mut env, CtapInstant::new(0));
+
+    let command = match input_type {
+        InputType::CborMakeCredentialParameter => Command::AuthenticatorMakeCredential(
+            AuthenticatorMakeCredentialParameters::arbitrary(unstructured)?,
+        ),
+        InputType::CborGetAssertionParameter => {
+            unimplemented!()
+        }
+        InputType::CborClientPinParameter => {
+            unimplemented!()
+        }
+        InputType::Ctap1 => {
+            unimplemented!()
+        }
+    };
+
+    state
+        .process_parsed_command(
+            &mut env,
+            command,
+            Channel::MainHid(ChannelID::arbitrary(unstructured)?),
+            CtapInstant::new(0),
+        )
+        .ok();
+
+    Ok(())
 }
 
 // Splits the given data as HID packets and reassembles it, verifying that the original input message is reconstructed.
--- a/fuzz/fuzz_targets/fuzz_target_process_ctap2_make_credential_structured.rs
+++ b/fuzz/fuzz_targets/fuzz_target_process_ctap2_make_credential_structured.rs
@@ -1,0 +1,10 @@
+#![no_main]
+
+use fuzz_helper::{process_ctap_structured, InputType};
+use libfuzzer_sys::fuzz_target;
+
+// Fuzz inputs as CTAP2 make credential command parameters.
+// The inputs will used to construct arbitrary make credential parameters.
+fuzz_target!(|data: &[u8]| {
+    process_ctap_structured(data, InputType::CborMakeCredentialParameter).ok();
+});
--- a/libraries/cbor/src/values.rs
+++ b/libraries/cbor/src/values.rs
@@ -225,6 +225,12 @@ impl From<&str> for Value {
     }
 }
 
+impl From<Vec<Value>> for Value {
+    fn from(array: Vec<Value>) -> Self {
+        Value::Array(array)
+    }
+}
+
 impl From<Vec<(Value, Value)>> for Value {
     fn from(map: Vec<(Value, Value)>) -> Self {
         Value::Map(map)
--- a/src/ctap/data_formats.rs
+++ b/src/ctap/data_formats.rs
@@ -15,6 +15,8 @@
 use super::status_code::Ctap2StatusCode;
 use alloc::string::String;
 use alloc::vec::Vec;
+#[cfg(feature = "fuzz")]
+use arbitrary::Arbitrary;
 use arrayref::array_ref;
 use core::convert::TryFrom;
 use crypto::{ecdh, ecdsa};
@@ -28,6 +30,7 @@ const ES256_ALGORITHM: i64 = -7;
 
 // https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialrpentity
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct PublicKeyCredentialRpEntity {
     pub rp_id: String,
     pub rp_name: Option<String>,
@@ -70,6 +73,7 @@ impl From<PublicKeyCredentialRpEntity> for cbor::Value {
 
 // https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialuserentity
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct PublicKeyCredentialUserEntity {
     pub user_id: Vec<u8>,
     pub user_name: Option<String>,
@@ -117,6 +121,7 @@ impl From<PublicKeyCredentialUserEntity> for cbor::Value {
 
 // https://www.w3.org/TR/webauthn/#enumdef-publickeycredentialtype
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub enum PublicKeyCredentialType {
     PublicKey,
     // This is the default for all strings not covered above.
@@ -149,6 +154,7 @@ impl TryFrom<cbor::Value> for PublicKeyCredentialType {
 
 // https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialparameters
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct PublicKeyCredentialParameter {
     pub cred_type: PublicKeyCredentialType,
     pub alg: SignatureAlgorithm,
@@ -183,6 +189,7 @@ impl From<PublicKeyCredentialParameter> for cbor::Value {
 // https://www.w3.org/TR/webauthn/#enumdef-authenticatortransport
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(test, derive(IntoEnumIterator))]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub enum AuthenticatorTransport {
     Usb,
     Nfc,
@@ -219,6 +226,7 @@ impl TryFrom<cbor::Value> for AuthenticatorTransport {
 
 // https://www.w3.org/TR/webauthn/#dictdef-publickeycredentialdescriptor
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct PublicKeyCredentialDescriptor {
     pub key_type: PublicKeyCredentialType,
     pub key_id: Vec<u8>,
@@ -270,6 +278,7 @@ impl From<PublicKeyCredentialDescriptor> for cbor::Value {
 }
 
 #[derive(Clone, Debug, Default, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct MakeCredentialExtensions {
     pub hmac_secret: bool,
     pub cred_protect: Option<CredentialProtectionPolicy>,
@@ -388,6 +397,7 @@ impl TryFrom<cbor::Value> for GetAssertionHmacSecretInput {
 
 // Even though options are optional, we can use the default if not present.
 #[derive(Clone, Debug, Default, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct MakeCredentialOptions {
     pub rk: bool,
     pub uv: bool,
@@ -488,6 +498,7 @@ impl From<PackedAttestationStatement> for cbor::Value {
 }
 
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub enum SignatureAlgorithm {
     ES256 = ES256_ALGORITHM as isize,
     // This is the default for all numbers not covered above.
@@ -516,6 +527,7 @@ impl TryFrom<cbor::Value> for SignatureAlgorithm {
 #[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
 #[cfg_attr(test, derive(IntoEnumIterator))]
 #[allow(clippy::enum_variant_names)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub enum CredentialProtectionPolicy {
     /// The credential is always discoverable, as if it had no protection level.
     UserVerificationOptional = 0x01,
@@ -884,6 +896,7 @@ impl TryFrom<CoseSignature> for ecdsa::Signature {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub enum PinUvAuthProtocol {
     V1 = 1,
     V2 = 2,
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -532,7 +532,7 @@ impl CtapState {
     ///
     /// This function contains the logic of `parse_command`, minus all CBOR encoding and decoding.
     /// It should make command parsing easier to test.
-    fn process_parsed_command(
+    pub fn process_parsed_command(
         &mut self,
         env: &mut impl Env,
         command: Command,
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
+fuzz/artifacts
 fuzz/corpus
+fuzz/coverage
 target/
 Cargo.lock
 
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -49,6 +49,12 @@ test = false
 doc = false
 
 [[bin]]
+name = "fuzz_target_process_ctap2_make_credential_structured"
+path = "fuzz_targets/fuzz_target_process_ctap2_make_credential_structured.rs"
+test = false
+doc = false
+
+[[bin]]
 name = "fuzz_target_split_assemble"
 path = "fuzz_targets/fuzz_target_split_assemble.rs"
 test = false
--- a/src/ctap/command.rs
+++ b/src/ctap/command.rs
@@ -26,6 +26,8 @@ use super::status_code::Ctap2StatusCode;
 use super::{cbor_read, key_material};
 use alloc::string::String;
 use alloc::vec::Vec;
+#[cfg(feature = "fuzz")]
+use arbitrary::Arbitrary;
 use arrayref::array_ref;
 use core::convert::TryFrom;
 use sk_cbor as cbor;
@@ -155,6 +157,7 @@ impl Command {
 }
 
 #[derive(Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct AuthenticatorMakeCredentialParameters {
     pub client_data_hash: Vec<u8>,
     pub rp: PublicKeyCredentialRpEntity,


libraries/cbor/src/values.rs
```diff
@@ -225,6 +225,12 @@ impl From<&str> for Value {
     }
 }
 
+impl From<Vec<Value>> for Value {
```
hcyang-google, line 228: Is there specific reasons we don't have this before? I need to auto parse `Vec<Value>` in cbor_map! macro. If we have `From<Vec<(Value, Value)>>` for map I think we can have this for array too.

kaczmarczyck: My reasoning was that I'd rather have people explicitly choose what they want, so they don't create the wrong type accidentally. There are currently different types that rely on `Vec` in some form or another:

- ByteString (`Vec<u8>`)
- Array (`Vec<Value>`)
- Map (`Vec<(Value, Value>)`)

I'm not against it, if you think it improves the API, we should have it. Also I'm thinking about API changes to fix #420 , so we can also merge this as is and I'll consider your use case later.

Cargo.toml
```diff
@@ -22,6 +22,7 @@ subtle = { version = "2.2", default-features = false, features = ["nightly"] }
 # This import explicitly locks the version.
 serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 embedded-time = "0.12.1"
+arbitrary = { version = "0.4.7", features = ['derive'], optional = true }
```
kaczmarczyck: I noticed that we are using inconsistent string delimiters throughout our `.toml` files. We should probably only use " going forward?

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -136,6 +152,23 @@ pub fn process_ctap_any_type(data: &[u8]) {
 // invokes message splitting, packet processing at CTAP HID level and response assembling
 // using an initialized and allocated channel.
 pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
+    let mut unstructured = Unstructured::new(data);
+
+    let valid_data;
+    let use_valid_structure = unstructured
+        .int_in_range(1..=100)
+        .map_or(false, |num| num <= USE_VALID_STRUCTURE_PERCENTAGE);
```
kaczmarczyck: I assume this is still deterministic from a reproducibility point of view?

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -16,22 +16,26 @@
...
 use ctap2::ctap::hid::receive::MessageAssembler;
 use ctap2::ctap::hid::send::HidPacketIterator;
 use ctap2::ctap::hid::{ChannelID, CtapHidCommand, HidPacket, Message};
+use ctap2::ctap::{cbor_read, cbor_write};
 use ctap2::env::test::TestEnv;
 use ctap2::Ctap;
 
 const CHANNEL_BROADCAST: ChannelID = [0xFF, 0xFF, 0xFF, 0xFF];
 
+const USE_VALID_STRUCTURE_PERCENTAGE: i32 = 95;
```
kaczmarczyck: Is this a rule of thumb, or did you hyperparameter tune this a bit, or just a guess?

src/ctap/mod.rs
```diff
@@ -130,7 +132,7 @@ pub fn cbor_read(encoded_cbor: &[u8]) -> Result<cbor::Value, Ctap2StatusCode> {
         .map_err(|_e| Ctap2StatusCode::CTAP2_ERR_INVALID_CBOR)
 }
 
-fn cbor_write(value: cbor::Value, encoded_cbor: &mut Vec<u8>) -> Result<(), Ctap2StatusCode> {
+pub fn cbor_write(value: cbor::Value, encoded_cbor: &mut Vec<u8>) -> Result<(), Ctap2StatusCode> {
```
kaczmarczyck: You could call the `cbor::writer` directly in fuzzing. This is the CTAP specific function with a CTAP error message. That has the advantage that fuzzing can play with different input nesting depths (that should still be parsed correctly, even when they don't match our expectatations).

So this doesn't have to be public then!

src/ctap/data_formats.rs
```diff
@@ -15,6 +15,8 @@
 use super::status_code::Ctap2StatusCode;
 use alloc::string::String;
 use alloc::vec::Vec;
+#[cfg(feature = "fuzz")]
```
kaczmarczyck, line 18: I like the clear separation. Did you check binary size with the conditional compilation though? We realized that the compiler is smart enough to remove unused code. It might still make sense for readability.

@ia0 Opinions?

fuzz/fuzz_helper/src/lib.rs
hcyang-google: Yes, deterministic.

fuzz/fuzz_helper/src/lib.rs
hcyang-google: This is just a guess. I'm imagining something like HIGH_PROBABILITY, GOOD_PROBABILITY probs, to correspond to different levels of requirements we want. Like we "almost always" want the structure to be valid, and "prefer" the pub key cred params to include ES256. But I don't know good ways to determine their best values.

libraries/cbor/src/values.rs
ia0: Yes, I agree those conversions should be unambiguous given the input and output types (i.e. there's only one way to go from the input to the output).

src/ctap/data_formats.rs
ia0: There's 2 things to consider:
1. Only guarding code that would otherwise to be statically eliminated.
2. Depending non-optionally from the `arbitrary` crate.

I think it makes sense to only depend optionally from `arbitrary`. This permits to make it clearer what are the real dependencies. And as a consequence of this, all the `derive(Arbitrary)` need to be guarded by the presence of the `arbitrary` crate otherwise they won't compile. So all in all, I'm fine with the current choice.

src/ctap/data_formats_fuzz.rs
```diff
@@ -0,0 +1,98 @@
+use super::{
+    command::AuthenticatorMakeCredentialParameters,
+    data_formats::{
+        CredentialProtectionPolicy, MakeCredentialExtensions, MakeCredentialOptions,
+        PinUvAuthProtocol, PublicKeyCredentialDescriptor, PublicKeyCredentialParameter,
+        PublicKeyCredentialRpEntity, PublicKeyCredentialUserEntity,
+    },
+};
```
ia0: Our `rustfmt` doesn't enforce anything about [`imports_granularity`](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#imports_granularity) but maybe we should decide something? I think in practice we use `Module` so I would vote in favor of continuing this practice and maybe even enforcing it in the `rustfmt` configuration.

@kaczmarczyck What do you think?

src/ctap/mod.rs
hcyang-google: Done

Cargo.toml
hcyang-google: Done

src/ctap/data_formats_fuzz.rs
```diff
@@ -0,0 +1,98 @@
...
+use cbor::{cbor_map, cbor_map_options};
+use sk_cbor as cbor;
+
+impl From<PinUvAuthProtocol> for cbor::Value {
+    fn from(protocol: PinUvAuthProtocol) -> Self {
+        Self::from(protocol as u64)
+    }
+}
+
+pub trait ArbitraryCbor<'a>: Sized {
```
ia0: I don't understand why we need this trait compared to converting the output of `Arbitrary` on the given type to CBOR. (I'll try to discuss higher level questions like mutations somewhere else if I don't forget.)

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -136,6 +153,23 @@ pub fn process_ctap_any_type(data: &[u8]) {
...
+        .map_or(false, |num| num <= USE_VALID_STRUCTURE_PERCENTAGE);
+    let data = if use_valid_structure {
+        if let Some(valid_type) = gen_valid_type(&mut unstructured, input_type) {
+            valid_data = valid_type;
+            valid_data.as_slice()
+        } else {
+            data
+        }
+    } else {
+        data
```
ia0: We shouldn't use `data` here because it's biased. We should use `unstructured.take_rest()`.

ia0 (commented):
This PR raises good questions with non-obvious answers. Here are some high-level comments:
- The PR description mentions "0% probability with random inputs", but fuzzing is not random, it's coverage-guided. For example, fuzzers are able to reconstruct file formats like PDF without help.
- The PR description mentions coverage. How was this measured? And on which corpus? I'd suggest using the existing OSS-Fuzz corpus or building one with 24h to 7d of continuous sequential fuzzing.
- One way to help fuzzing is to provide examples or a starting corpus.
- The problem of converting the fuzzer input to valid input only a fraction of the time means the fuzzer input is not homogeneous and thus can't be easily reused/mutated. A better approach would be to have 2 fuzzing targets: one using the fuzzer input as raw input, and one using the fuzzer input to generate valid input. In some way, the first version is more end-to-end while the second is more like fuzzing the internal functions (we could actually fuzz them directly to avoid unnecessary processing which could accidentally restrict coverage).
- Regarding validity, there's also the questions of mutation. Do we want to slightly mutate the input such that it may also be invalid although it looks valid (e.g. reorder items in a map).

I guess the ultimate goal of this PR is to improve fuzzing coverage, so for that I think a prerequisite is to agree on what is coverage and how we measure it. For example, for the persistent store fuzzing, the coverage is behavior based (in particular it's not about code) and is measured with the example binary when the argument is a directory. There's also an unsubmitted `libraries/persistent_store/coverage.sh` script in the `fuzz_coverage` branch of my fork that measures code coverage. Should eventually make a PR of it.

libraries/cbor/src/values.rs
kaczmarczyck: Actually my list above was incorrectly formatted, I fixed it. It is not ambiguous, since there is no notion of tuples in CBOR, and u8 is different to Value. It might be tricky if we extend this to macros, but it's okay as is I think.
A user could try to call `.into()` on a vector of `Value`s of `u8` and expect it to become a ByteString, if they accidentally casted their array elements into `Value`. But since @hcyang-google found this more intuitive, we have real world evidence that this is better.

ia0: Ah I see. If the `Value` version of a `Vec<Value>` is unambiguously an array (including in the future), then SGTM.

src/ctap/data_formats_fuzz.rs
kaczmarczyck: Would have made it easier in this case to meet our style, just run `cargo fmt`. So if we care here, and I think we should, then we should merge #445 :)

kaczmarczyck:
I did run
```
cargo fuzz run fuzz_target_process_ctap2_make_credential
```
with and without this PR, and the coverage numbers do look much better after a few minutes, at least. It might of course also make it harder to find obscure CTAP bugs. But we do have different fuzz targets, and this one is specifically designed to test make credential behavior, so I like it!

I'm also curious about the 0% though. OSS-Fuzz did find logic bugs (#368 was one result), so maybe can you explain a bit more? +1 to @ia0's questions.

hcyang-google:
- The PR description mentions "0% probability with random inputs", but fuzzing is not random, it's coverage-guided. For example, fuzzers are able to reconstruct file formats like PDF without help.
- The PR description mentions coverage. How was this measured? And on which corpus? I'd suggest using the existing OSS-Fuzz corpus or building one with 24h to 7d of continuous sequential fuzzing.
- One way to help fuzzing is to provide examples or a starting corpus.

Answering these together. As fuzzer is coverage-guided, if you fuzz it long enough it will start to find interesting input and mutate more of those. But (arguably) if our fuzzer reaches interesting paths with pure random bytes, then it can have decent coverage in minutes without starting corpus, and might be more efficient on OSS-Fuzz too. This conclusion might not be correct, but the practice in our team is that we used to locally-run our new fuzzer for minutes and see if it has decent coverage before submitting that fuzzer. If not we fix the fuzzer logic so that it reaches interesting paths more easily. So seems that we have different ways of doing this, it's good that we can learn from your practices if OSS-fuzzing is efficient enough even without fine-tuning fuzzer's behavior. On the other hand making fuzzer reach interesting paths with pure random bytes seems to not have obvious disadvantages too; if it's efficient in minutes it should be efficient in days too.
I measured both cases' coverage with empty corpus and minutes of running on my workstation.

- The problem of converting the fuzzer input to valid input only a fraction of the time means the fuzzer input is not homogeneous and thus can't be easily reused/mutated. A better approach would be to have 2 fuzzing targets: one using the fuzzer input as raw input, and one using the fuzzer input to generate valid input. In some way, the first version is more end-to-end while the second is more like fuzzing the internal functions (we could actually fuzz them directly to avoid unnecessary processing which could accidentally restrict coverage).

I never thought of this before. So choosing type A of input with X% and type B of input with Y%... limits the fuzzers ability significantly? If that's the case I agree we should probably add a target. But the logic of this CL is mainly that this target is for MakeCredential, and I want it to reach the internals of MakeCredential command processing more easily.

- Regarding validity, there's also the questions of mutation. Do we want to slightly mutate the input such that it may also be invalid although it looks valid (e.g. reorder items in a map).

I'm thinking of letting the non-valid (original) path to discover all the cases where the input isn't valid. I'm not sure if we want to add more specific logic to manipulate input to reach possibly interesting paths. Now the logic is that one case is arbitrary inputs, and another is arbitrary valid inputs.

src/ctap/data_formats_fuzz.rs
hcyang-google: ArbitraryCbor is used on types which are not one-to-one corresponding to cbors. For example `MakeCredentialOptions` has two fields but is parsed from a cbor map with three. Also, some fields of the types have defaults in case the map doesn't contain that entry, but some fields are required to exist. If we generate arbitrary types and converting them back to CBOR, we miss some valid CBOR map structures that the type can be parsed from.

ia0:
Regarding the "raw input" vs "structured input" discussion, here is how I would compare both.

Advantages of raw vs structured:
- Simple (i.e. less bugs in the fuzzing code)
- 100% theoretical coverage

Advantages of structured vs raw:
- Faster effective coverage

Note that ideally we want to improve the coverage speed regardless of how much is already covered, i.e. the fuzzer is always able to regularly find new paths. This is different from both the theoretical coverage (when running with infinite time) and the initial coverage (when running for a few minutes).

[Here](https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md) is some documentation in case we want structure-aware fuzzing which could enable both faster effective coverage and 100% theoretical coverage, but to the price of being more complex. I don't think we should go this way unless we have evidence that we need it.

Regarding running the fuzzer for minutes instead of days, maybe random testing like [proptest](https://altsysrq.github.io/proptest-book/intro.html) would be more adapted for those usages. For me, the role of fuzzing is to theoretically cover all edge cases since no other testing technique permits that. We can still also use it for random testing, but that should be secondary.

Regarding splitting targets, [here](https://github.com/google/fuzzing/blob/master/docs/split-inputs.md#multiple-options) is documented that it is often the best approach. I think that if we want random testing, then it should be a separate target because the raw and structured input are very different and shouldn't be part of the same corpus.

This is still unclear to me which number you look at when comparing coverage. Is it the number in the output of the fuzzer? Or is it a separate command? I think those numbers (when comparing after at least one day of fuzzing) would help decide what we want to do.

src/ctap/data_formats_fuzz.rs
ia0: I see, the structures we use make deserializing from CBOR non-injective. And this trait permits to theoretically generate all CBORs that would successfully deserialize to any possible struct value.

Note though that in practice this is not the case. For example for `MakeCredentialOptions`, the trait implementation always sets all map keys while it would be valid to deserialize when some keys are missing since they have a default value.

hcyang-google:
Thanks, the information you provided is very helpful and I'll read them more thoroughly afterwards.
The only thing I want to clarify is that the structured fuzzer is not just random fuzzing, it still theoretically generates every valid input that can be parsed into MakeCredential command, and I thought the main purpose of this target is for MakeCredential internal command processing. But like you said, maybe we can directly fuzz inner functions for the command processing, so I'll see if I can skip the parsing part and directly fuzz the command processing. If that's the case, we don't need the ArbitraryCbor trait as we only need to generate the structure itself.

When comparing coverage, I used the llvm-cov command (https://llvm.org/docs/CommandGuide/llvm-cov.html). It generates a nice website view where I can see each file's function coverage and line coverage. I compared the function coverage and line coverage of the more relevant files (ctap/mod.rs, ctap/hid/mod.rs, ...).

ia0:
> maybe we can directly fuzz inner functions for the command processing

Yes, if that's possible I think it's the best. It should be simpler and clearer.

> I used the llvm-cov command

I see. I compared 5 minutes of fuzzing with your branch with the current corpus on OSS-Fuzz and the coverage of `process_make_credential` is the same except for `large_blob_key` which takes both `Some` (3 times) and `None` (114 times) in your branch while only `None` (418 times) in OSS-Fuzz. But there is much more that is not covered in both cases, in particular stuff that depends on the persistent state (enterprise attestation, PIN hash, PIN length, UV, credential exclude list, cred protect). I don't think those would get covered with the raw vs structured input discussion. We should just setup the persistent state based on the input too (which could be raw and structured, but it's an orthogonal discussion).

I think it would also be worth having some behavior coverage, e.g. how often is `rk` true, how often is `uv` true, etc.

src/ctap/data_formats_fuzz.rs
hcyang-google: Done.

src/ctap/data_formats_fuzz.rs
hcyang-google: > Note though that in practice this is not the case. For example for MakeCredentialOptions, the trait implementation always sets all map keys while it would be valid to deserialize when some keys are missing since they have a default value.

What trait implementation? I only see a trait from cbor::Value to MakeCredentialOptions, not reversed. I thought commands from the host are parsed as cbor, and then parsed as MakeCredentialOptions, so we can't guarantee that no host will send a cbor map that has some missing values right?

hcyang-google:
@ia0 fixed to only fuzz the CtapState command processing, with a separate fuzz target.

src/ctap/data_formats_fuzz.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn arbitrary_cbor(u: &mut Unstructured<'a>) -> arbitrary::Result<cbor::Value> {
+        let rk = bool::arbitrary(u)?;
+        let up = bool::arbitrary(u)?;
+        let uv = bool::arbitrary(u)?;
+
+        let cbor_map = cbor_map! {
+            "rk" => rk,
+            "up" => up,
+            "uv" => uv,
+        };
```
ia0: In reply to https://github.com/google/OpenSK/pull/441/files#r828940199:
> What trait implementation?

This implementation always sets all keys for the map. Ideally we want to use `Option<bool>::arbitrary()` and `cbor_option_map!`, or something equivalent.

src/ctap/data_formats_fuzz.rs
ia0: I'm replying in a comment below in this file.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -166,6 +181,39 @@ pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
...
+        InputType::CborMakeCredentialParameter => {
+            command.push(0x01);
+        }
+        InputType::CborGetAssertionParameter => {
+            command.push(0x02);
+        }
+        InputType::CborClientPinParameter => {
+            command.push(0x06);
+        }
+        _ => return,
```
ia0: Let's be exhaustive to not forget to add cases in the future. Otherwise it's easy to forget fuzzing a new command.

fuzz/fuzz_helper/Cargo.toml
```diff
@@ -11,5 +11,6 @@ embedded-time = "0.12.1"
 libtock_drivers = { path = "../../third_party/libtock-drivers" }
 crypto = { path = "../../libraries/crypto", features = ['std'] }
 sk-cbor = { path = "../../libraries/cbor" }
-ctap2 = { path = "../..", features = ['std'] }
+ctap2 = { path = "../..", features = ['fuzz'] }
 lang_items = { path = "../../third_party/lang-items", features = ['std'] }
+arbitrary = { version = "0.4.7", features = ['derive'] }
```
ia0: Let's start consistently using double quotes for modified line.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -166,6 +181,39 @@ pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
...
+    }
+    command.extend(valid_data);
+
+    state.process_command(
+        &mut env,
+        &command,
+        Channel::MainHid(CHANNEL_BROADCAST),
```
ia0: Shouldn't the channel be arbitrary too?

ia0 (commented):
I'm still not convinced we need the `ArbitraryCbor` trait (long-term). Short-term it's probably true because of how the library is designed. But already with #447 I think we could fuzz the new `CtapState::process_parsed_command` function directly and short-circuit CBOR logic.

But I think more importantly we need to decide what do we want to fuzz. As I understand the current state, we have a fuzz target for all commands taking CBOR and one fuzz target per command taking CBOR. This choice looks quite weird. I think we probably want something like:
- One or multiple fuzz targets at environment level, i.e. taking HID packets (the fuzzing code should not do any preprocessing on the fuzzer input like splitting in HID packets). Those targets are the ultimate targets since they fuzz based on user input.
- One or multiple fuzz targets at library level, i.e. taking stuff out of `Arbitrary`. Those targets are here to improve coverage. We could use their corpus to create better examples for the targets above (by serializing and splitting into packets).

What do you think?

Regarding the separate fuzz target, at first I thought we could reuse the existing one for "make credential" because it's kind of contained in target checking all commands, but the coverage of that one is actually very bad. So it's really useful to split by command. So the current split looks good to me.

src/ctap/data_formats_fuzz.rs
hcyang-google: Ok this is a mistake, I should put option there. But might remove this trait, replying in comment below.

hcyang-google:
Yes, with #447 I would prefer removing the trait and directly fuzz the command processing with parsed type. The interface is separated much better after it. I'll wait until it is merged to continue on this.

ia0:
> Yes, with #447 I would prefer removing the trait and directly fuzz the command processing with parsed type. The interface is separated much better after it. I'll wait until it is merged to continue on this.

Cool, I've merged #447. Feel free to merge/rebase this PR.

fuzz/fuzz_helper/src/lib.rs
hcyang-google: done

fuzz/fuzz_helper/Cargo.toml
hcyang-google: done

fuzz/fuzz_helper/src/lib.rs
hcyang-google: done

src/ctap/mod.rs
```diff
@@ -532,7 +532,7 @@ impl CtapState {
     ///
     /// This function contains the logic of `parse_command`, minus all CBOR encoding and decoding.
     /// It should make command parsing easier to test.
-    fn process_parsed_command(
+    pub fn process_parsed_command(
         &mut self,
```
hcyang-google, line 536: If we don't do this, we have to use this [crate](https://docs.rs/visibility/latest/visibility/attr.make.html) or make something similar ourselves to achieve conditional visibility. We might want to eventually do it otherwise we'll run into this problem again when fuzzing other private functions to increase coverage.

hcyang-google:
@ia0 can you check again if everything is fine? My only concern is how to deal with the visibility, which I put in the comment above.

fuzz/fuzz_targets/fuzz_target_process_ctap2_make_credential_structured.rs
```diff
@@ -0,0 +1,11 @@
+#![no_main]
+
+use fuzz_helper::{process_ctap_structured, InputType};
+use libfuzzer_sys::fuzz_target;
+
+// Fuzz inputs as CTAP2 make credential command parameters encoded in cbor.
+// The inputs will used to construct cbors that can be parsed into valid
+// make credential parameters.
```
ia0: We can update this comment now.

src/ctap/mod.rs
ia0: I don't think this is an issue. We already control the visibility of the whole ctap module here: https://github.com/google/OpenSK/blob/c3e1b5df5058e9e85e74b69262235c43d6ac5b0f/src/lib.rs#L49-L53.

ia0 (dismissed):
LGTM modulo the documentation. I think we can submit this and check the coverage of the OSS-Fuzz corpus on this new target in a week. In the meantime do you plan to add the same for other targets? I'm actually curious how a single structured target for all commands would perform regarding coverage. Maybe that's actually enough.

src/ctap/mod.rs
hcyang-google: Oh so ctap module is private except when testing, thus we can make public all methods we want to test. Thanks.

hcyang-google:
Fixed the comment. I'll add other fuzz targets in an upcoming PR.

ia0 (approved):

PR #442, opened 2022-03-14
Clean public HID interface

kaczmarczyck:
This PR further prepares the `hid` mod for its future move:

- Clean up the duplicate internal error data type.
- Refactor and add tests. Note that most don't rely on `CtapState` anymore with the new API.
- You can import from this module now directly, `send.rs` and `receive.rs` became implementation details.

There is only a fraction of the code that relies on `CtapState` now. Plan is to split all of that out, and have a clean `hid` mod afterwards.

### DIFF
--- a/src/ctap/hid/mod.rs
+++ b/src/ctap/hid/mod.rs
@@ -12,11 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub mod receive;
-pub mod send;
+mod receive;
+mod send;
 
+// Implementation details must be public for testing (in particular fuzzing).
+#[cfg(feature = "std")]
+pub use self::receive::MessageAssembler;
+#[cfg(not(feature = "std"))]
 use self::receive::MessageAssembler;
-use self::send::HidPacketIterator;
+pub use self::send::HidPacketIterator;
 use super::super::clock::{ClockInt, CtapInstant};
 #[cfg(feature = "with_ctap1")]
 use super::ctap1;
@@ -28,6 +32,8 @@ use alloc::vec;
 use alloc::vec::Vec;
 use arrayref::{array_ref, array_refs};
 use embedded_time::duration::Milliseconds;
+#[cfg(test)]
+use enum_iterator::IntoEnumIterator;
 
 pub type HidPacket = [u8; 64];
 pub type ChannelID = [u8; 4];
@@ -36,6 +42,7 @@ pub type ChannelID = [u8; 4];
 ///
 /// See section 11.2.9. of FIDO 2.1 (2021-06-15).
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(test, derive(IntoEnumIterator))]
 pub enum CtapHidCommand {
     Ping = 0x01,
     Msg = 0x03,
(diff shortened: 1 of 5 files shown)


kaczmarczyck:
The report of decreased coverage is unrelated btw. Some files are not correctly counted for some reason still, and the [last report](https://coveralls.io/builds/47323220) says:

```
+ 8.31% | src/ctap/hid/mod.rs
-37.48% | libraries/persistent_store/src/store.rs
```

src/ctap/hid/mod.rs
```diff
@@ -70,9 +75,40 @@ impl From<u8> for CtapHidCommand {
...
+    /// Unspecified error.
+    _Other = 0x7F,
+    /// This error is silently ignored.
+    UnexpectedContinuation,
+}
+
 /// Describes the structure of a parsed HID packet.
 ///
 /// A packet is either an Init or a Continuation packet.
+#[derive(Debug, PartialEq)]
```
ia0: Is it intended to not derive `Eq`? Looks like equality should be reflexive here.

src/ctap/hid/mod.rs
```diff
@@ -503,68 +504,25 @@ impl CtapHid {
             payload: response,
         }
     }
+
+    #[cfg(test)]
+    pub fn new_initialized() -> (CtapHid, ChannelID) {
```
ia0, line 509: Can this be defined in the `test` module? When there's a `test` module, I think it helps to put as much test-only code there as possible such that it's easy to know what is production code.

src/ctap/hid/mod.rs
```diff
@@ -713,35 +639,188 @@ mod test {
...
+        let mut response = CtapHid::split_message(message);
+        let mut expected_packet = [0x00; 64];
+        expected_packet[..8].copy_from_slice(&[0x12, 0x34, 0x56, 0x78, 0x90, 0x00, 0x01, 0xF2]);
+        assert_eq!(response.next(), Some(expected_packet));
+        assert_eq!(response.next(), None);
+    }
+
+    #[test]
+    fn test_keepalive() {
+        for status in [KeepaliveStatus::Processing, KeepaliveStatus::UpNeeded].iter() {
```
ia0: nit/fyi: it's possible to write `&status` here (so once) instead of `*status` at each usage occurrences below (when the type is copy).

ia0 (dismissed):

src/ctap/hid/mod.rs
kaczmarczyck: Ah right, this doesn't make sense from this PR's perspective. It will be used from the outside, when HID gets instantiated twice. I'll then use it for testing structs that instantiate HID.

ia0: Ah ok. SGTM then.

src/ctap/hid/mod.rs
kaczmarczyck: Better, fixed.

src/ctap/hid/mod.rs
kaczmarczyck: I think throughout `data_format.rs`, we always derive like `#[derive(Clone, Debug, PartialEq)]` because that is useful for writing tests, like calling `assert_eq`. We can derive more since it makes sense, this is the minimum set that we actually need though. Added `Clone` and `Eq`.

ia0: Tests probably rely on `==` to be reflexive which is only true with `Eq`. So technically the tests are wrong even if in practice they're not since the types are morally `Eq`, they just don't advertise it.

ia0 (approved):

PR #443, opened 2022-03-14
Separates HID command logic

kaczmarczyck:
This PR is mostly moving code around. All CTAP command related logic is moved out of `hid` into a new `MainHid` struct. Renaming is certainly possible, suggestions appreciated.
From now on, HID does not do full round trips from packet to packet anymore. You need one call for receiving, and one for sending. The new struct is taking over this responsibility.

Spoiler: There will be a second struct to contain a `CtapHid` object for the second usage page. It will also need a good name, "VendorHid"? It will be much smaller though. No CTAP1, no Wink.

### DIFF
--- a/src/ctap/hid/mod.rs
+++ b/src/ctap/hid/mod.rs
@@ -22,11 +22,6 @@ pub use self::receive::MessageAssembler;
 use self::receive::MessageAssembler;
 pub use self::send::HidPacketIterator;
 use super::super::clock::{ClockInt, CtapInstant};
-#[cfg(feature = "with_ctap1")]
-use super::ctap1;
-use super::status_code::Ctap2StatusCode;
-use super::timed_permission::TimedPermission;
-use super::CtapState;
 use crate::env::Env;
 use alloc::vec;
 use alloc::vec::Vec;
@@ -37,6 +32,8 @@ use enum_iterator::IntoEnumIterator;
 
 pub type HidPacket = [u8; 64];
 pub type ChannelID = [u8; 4];
+
+const CTAP2_ERR_VENDOR_INTERNAL_ERROR: u8 = 0xF2;
 
 /// CTAPHID commands
 ///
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2022 Google LLC
```
ia0: Shouldn't we write `2019-2022`?

src/ctap/hid/mod.rs
```diff
@@ -456,14 +380,14 @@ impl CtapHid {
...
-    fn split_message(message: Message) -> HidPacketIterator {
+    pub fn split_message(message: Message) -> HidPacketIterator {
         let cid = message.cid;
         HidPacketIterator::new(message).unwrap_or_else(|| {
             // The error payload is 1 <= 7609 bytes, so unwrap() is safe.
             HidPacketIterator::new(Message {
                 cid,
                 cmd: CtapHidCommand::Cbor,
-                payload: vec![Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR as u8],
+                payload: vec![CTAP2_ERR_VENDOR_INTERNAL_ERROR as u8],
```
ia0: I don't understand this change. What's the motivation behind it?

ia0 (dismissed):
"VendorHid" sounds good to me. Regarding renaming "MainHid" if we ever want to do it, it should be trivial with rust-analyzer.

src/lib.rs
kaczmarczyck: Ah I thought this file is from this year, but we actually had it for longer. Fixed here and in other touched files, and I also fixed some documentation that I noticed.

src/ctap/hid/mod.rs
kaczmarczyck: This was the only CTAP dependency and I wanted to make this module independent, so we can later move it around trivially.

... is what I thought, but I just double checked and realized that there is `MAX_MSG_SIZE` as well, that I overlooked. Since that would be a logic change to move it, I reverted the line in question and we can do the move later, if we want.

ia0: Sounds good.

ia0 (approved):

PR #446, opened 2022-03-15
Vendor HID

kaczmarczyck:
Introduces a second HID parser. In this PR still, the `CtapState` treats packets from both HIDs indiscriminately. So from here on, when we introduce the second HID page to the driver and use it in main, we should start seeing results.

The next PR will focus on processing the incoming channels differently. Afterwards, all that is left to do to actually make it useful is outside of the CTAP library.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ std = ["crypto/std", "lang_items/std", "persistent_store/std"]
 verbose = ["debug_ctap", "libtock_drivers/verbose_usb"]
 with_ctap1 = ["crypto/with_ctap1"]
 with_nfc = ["libtock_drivers/with_nfc"]
+with_vendor_hid = []
 
 [dev-dependencies]
 enum-iterator = "0.6.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ std = ["crypto/std", "lang_items/std", "persistent_store/std"]
 verbose = ["debug_ctap", "libtock_drivers/verbose_usb"]
 with_ctap1 = ["crypto/with_ctap1"]
 with_nfc = ["libtock_drivers/with_nfc"]
-with_vendor_hid = []
+vendor_hid = []
 
 [dev-dependencies]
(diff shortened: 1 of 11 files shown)


kaczmarczyck:
I also updated the workflows to add the new feature. It is worth noting that binary size increases by ~100B with the feature, and does not increase without the feature.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ std = ["crypto/std", "lang_items/std", "persistent_store/std"]
 verbose = ["debug_ctap", "libtock_drivers/verbose_usb"]
 with_ctap1 = ["crypto/with_ctap1"]
 with_nfc = ["libtock_drivers/with_nfc"]
+with_vendor_hid = []
```
ia0: Let's not propagate this [bad style](https://rust-lang.github.io/api-guidelines/naming.html#feature-names-are-free-of-placeholder-words-c-feature) and name the feature `vendor_hid`.

src/lib.rs
```diff
@@ -87,6 +99,12 @@ impl<E: Env> Ctap<E> {
                 self.hid
                     .process_hid_packet(&mut self.env, packet, now, &mut self.state)
             }
+            #[cfg(feature = "with_vendor_hid")]
+            Transport::VendorHid => {
+                self.vendor_hid
+                    .process_hid_packet(&mut self.env, packet, now, &mut self.state)
+            }
+            #[cfg(not(feature = "with_vendor_hid"))]
             Transport::VendorHid => HidPacketIterator::none(),
```
ia0: If we protect the code with a feature, shouldn't we also protect the transport enum value? It looks weird to have that enum value if it's not really used.

ia0 (dismissed):

src/lib.rs
kaczmarczyck: Done

Cargo.toml
kaczmarczyck: I only renamed the new feature. Also added the flag for deploy, even though it's not useful yet.

ia0 (approved):

PR #447, opened 2022-03-16
Channels logic

kaczmarczyck:
The vendor HID and channels introduced in #444 and #446 are now used. To do that, I refactored the `process_command` function heavily into:
- a small and slightly improved outer function for reading and writing CBOR
- an inner function that only calls command logic minus CBOR
- one function for all FIDO commands
- once function for all vendor commands

Stateful commands are now assigned to the channel, and their state gets cleared on timeout or channel switch. (Some reasoning is in the comments in code.) This is slightly more correct and simpler than before, where commands check timeouts individually. That was not really necessary, as a timed out stateful commands is equivalent to never having called the command to create the state.

I added a bunch of tests to check the new logic, and... found a bug where GetNextAssertion actually answers as GetAssertion. None of the test tools check this I believe. I decided to fix it right away, so the PR got a bit bigger. Can split up if desired.

Also I fix a tiny bug with the debug print macro.

### DIFF
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -22,7 +22,7 @@ use super::data_formats::{
 };
 use super::response::{AuthenticatorCredentialManagementResponse, ResponseData};
 use super::status_code::Ctap2StatusCode;
-use super::{StatefulCommand, StatefulPermission};
+use super::{Channel, StatefulCommand, StatefulPermission};
 use crate::ctap::storage;
 use crate::env::Env;
 use alloc::collections::BTreeSet;
@@ -137,13 +137,14 @@ fn process_get_creds_metadata(
 fn process_enumerate_rps_begin(
     env: &mut impl Env,
     stateful_command_permission: &mut StatefulPermission,
+    channel: Channel,
     now: CtapInstant,
 ) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
     let rp_set = get_stored_rp_ids(env)?;
     let total_rps = rp_set.len();
 
     if total_rps > 1 {
-        stateful_command_permission.set_command(now, StatefulCommand::EnumerateRps(1));
+        stateful_command_permission.set_command(now, StatefulCommand::EnumerateRps(1), channel);
     }
     // TODO https://github.com/rust-lang/rust/issues/62924 replace with pop_first()
     let rp_id = rp_set
@@ -174,6 +175,7 @@ fn process_enumerate_credentials_begin(
     stateful_command_permission: &mut StatefulPermission,
     client_pin: &mut ClientPin,
     sub_command_params: CredentialManagementSubCommandParameters,
+    channel: Channel,
     now: CtapInstant,
 ) -> Result<AuthenticatorCredentialManagementResponse, Ctap2StatusCode> {
     let rp_id_hash = sub_command_params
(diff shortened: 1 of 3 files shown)


kaczmarczyck:
I checked coverage locally, and `src/ctap/mod.rs` should have improved from 92% to 95% coverage. I ran the same command as the workflow.

src/ctap/credential_management.rs
```diff
@@ -269,7 +275,7 @@ pub fn process_credential_management(
         | (
             CredentialManagementSubCommand::EnumerateCredentialsGetNextCredential,
             Ok(StatefulCommand::EnumerateCredentials(_)),
-        ) => stateful_command_permission.check_command_permission(now)?,
```
ia0, line 272: Is this removed because wrong or redundant?

src/ctap/mod.rs
```diff
@@ -280,22 +284,40 @@ impl StatefulPermission {
...
+        // A) Always reset state (our choice).
+        // B) Only reset state if the new command is stateful.
+        // C) Keep state on all channels until timeout.
+        //
+        // If we wanted to switch to (B) or (C), we'd have to be very careful that the state does
+        // not go stale. For example, we keep credential keys that we expect to still exist.
+        // However, interleaving (stateless) commands could delete credentials or change the PIN,
+        // which chould make invalidate our access. Some read-only commands should be okay to run,
+        // but (A) is the safest and easiest solution.
+        if let Some(c) = self.channel.as_ref() {
```
ia0, line 310: nit/fyi: `&self.channel` should work too

ia0, line 310: nit/fyi: `&self.channel` should work too

src/ctap/mod.rs
```diff
@@ -486,94 +511,126 @@ impl CtapState {
...
+    fn process_parsed_command(
+        &mut self,
+        env: &mut impl Env,
+        command: Command,
+        channel: Channel,
+        now: CtapInstant,
+    ) -> Result<ResponseData, Ctap2StatusCode> {
+        // Correct behavior between CTAP1 and CTAP2 isn't defined yet. Just a guess.
+        #[cfg(feature = "with_ctap1")]
+        {
```
ia0, line 544: I don't think those braces are needed. They're probably a left-over.

ia0, line 544: I don't think those braces are needed. They're probably a left-over.

ia0 (approved):

ia0:
I don't know why Github merged 2 versions of my review. I deleted the old comments.

ia0:
I merged to unblock #441. The 2 comments were nits.

src/ctap/mod.rs
kaczmarczyck: That would need a `#![feature(stmt_expr_attributes)]`:

> error[E0658]: attributes on expressions are experimental

src/ctap/mod.rs
kaczmarczyck: Addressed in #449.

src/ctap/mod.rs
ia0: Interesting, this is probably a quirk of our setup. It should work: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=19206afa3cb252e5746d4efd0670435c. Let's wait until we have CTAP as a library to figure out where it comes from.

jmichelp: This would be the equivalent and it indeed fails to compile: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7f0c21feacafe80a8e5066a838fa701e

ia0: Oh I see, I didn't know assignment was an expression... that's a bit weird decision, in particular given it returns unit so the expression value isn't useful, only its side-effect. So the work-around would be: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aae7cb8e93d80cfd8f965e808a4677ba. Not sure it's worth it, both solutions are equivalently ugly.

kaczmarczyck: This would be much more confusing for me as a reader. I'd feel like I don't understand the purpose. With the current solution, if you don't know the purpose, you're just like: "probably a leftover"

ia0: For me in both cases I don't understand the purpose (which is why I suspected a left-over). You actually need to fix it to see the problem. Which means we probably need to document it.

Issue #448, opened 2022-03-25
Cannot RESET through Windows 10/11 sign-in options and Chrome settings on macOS


superskybird:
## Expected Behavior
FIDO2 security key can be RESET through
- Chrome Settings->Manage Security Keys (macOS)
chrome://settings/securityKeys?search=security+key
- Windows 10/11 Sign-in Options->Security Key

## Actual Behavior
show error.
Even though replug and press button inside 10 seconds.

ia0:
Thanks for the bug report!

I could actually reproduce on both Windows and Linux (Chrome).

Here is the trace for Linux after the replug:
```
Received message: Message { cid: [ff, ff, ff, ff], cmd: Init, payload: [18, 5b, 26, be, 24, 67, 4e, a7] }
Sending message: Message { cid: [ff, ff, ff, ff], cmd: Init, payload: [18, 5b, 26, be, 24, 67, 4e, a7, 00, 00, 00, 01, 02, 01, 00, 00, 05] }
Received message: Message { cid: [00, 00, 00, 01], cmd: Cbor, payload: [04] }
[...]

PR #449, opened 2022-03-28
Allow Make before Reset

kaczmarczyck:
Solves #448, tested on Linux. Needs to be confirmed for MacOS.

I addressed the nit from #447 as well.


ia0 (approved):
Should we also add a test in `CTAP2-test-tool`?

kaczmarczyck:
Good question. Not sure how temporary this fix is. Once 2.1 is final and all platforms detect 2.1 support, we could go back to being strict actually. But most likely, we will just keep it as is. So I might!

ia0:
Yeah, we should probably initially not enforce the outcome of the test, but just test it to document what "features" keys possess. This would be a (compatibility) checkbox whether the key accepts MakeCredential before Reset. And if the standard and client implementations decide to take a stance on it, we would force the outcome accordingly.

PR #450, opened 2022-03-28
Add documentation for #[cfg] work-around on expressions

ia0:
This follows from https://github.com/google/OpenSK/pull/447#discussion_r836405374.


kaczmarczyck (dismissed):
We do this more often, in `src/ctap/mod.rs` inside

- `process_get_info`
- `process_reset`

and at the end of `fn main()` of `main.rs`. In case you want to document all cases.

ia0:
> In case you want to document all cases

Good point, I didn't think there were more cases. I went over all cases of a `#[cfg]` on a block statement. Some of them are legitimate (i.e. there's multiple statements in the block, so it's obvious the block is here to factorize the `#[cfg]`), some of them were unnecessary (I fixed them), and only another one remained that I documented too.

kaczmarczyck (approved):

PR #452, opened 2022-04-04
Vendor HID for libtock-drivers

kaczmarczyck:
This PR adds Vendor HID outside of the CTAP crate, now libtock-drivers can be called with a `UsbInterface` to send, and returns one on receive. This will later be realized by sending and receiving another integer to encode that.

This way, we can receive on 2 interfaces, without making the code non-blocking.

@liamjm The changes that are left for the next PR are marked by TODOs:
- `send_or_recv_with_timeout_detail` has to actually use the interface in its command syscall,
- on receiving `subscribe_nr::callback_status::RECEIVED`, the syscall should return one more int for the interface it received the data on.

### DIFF
--- a/third_party/libtock-drivers/src/usb_ctap_hid.rs
+++ b/third_party/libtock-drivers/src/usb_ctap_hid.rs
@@ -100,7 +100,7 @@ pub fn recv_with_timeout(
     if let Some(SendOrRecvStatus::Received(interface)) = result {
         writeln!(
             Console::new(),
-            "Received packet = {:02x?} on interface {:?}",
+            "Received packet = {:02x?} on interface {}",
             buf as &[u8],
             interface as u8,
         )
@@ -143,7 +143,7 @@ pub fn send_or_recv_with_timeout(
     if let Some(SendOrRecvStatus::Received(received_interface)) = result {
         writeln!(
             Console::new(),
-            "Received packet = {:02x?} on interface {:?}",
+            "Received packet = {:02x?} on interface {}",
             buf as &[u8],
             received_interface as u8,
         )
--- a/third_party/libtock-drivers/src/usb_ctap_hid.rs
+++ b/third_party/libtock-drivers/src/usb_ctap_hid.rs
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2019-2022 Google LLC
 //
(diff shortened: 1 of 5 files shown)


third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -188,19 +97,36 @@ pub fn recv_with_timeout(
...
 
     #[cfg(feature = "verbose_usb")]
-    if let Some(SendOrRecvStatus::Received) = result {
-        writeln!(Console::new(), "Received packet = {:02x?}", buf as &[u8]).unwrap();
+    if let Some(SendOrRecvStatus::Received(interface)) = result {
+        writeln!(
+            Console::new(),
+            "Received packet = {:02x?} on interface {:?}",
+            buf as &[u8],
+            interface as u8,
```
ia0, line 105: Why `as u8` for `{:?}`? I see 2 options:
1. Keep as `UsbInterface` and debug-print with `{:?}`. This might increase the binary size.
2. Convert to `u8` and print with `{}`.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -211,11 +137,17 @@ pub fn send_or_recv_with_timeout(
...
 
     #[cfg(feature = "verbose_usb")]
-    if let Some(SendOrRecvStatus::Received) = result {
-        writeln!(Console::new(), "Received packet = {:02x?}", buf as &[u8]).unwrap();
+    if let Some(SendOrRecvStatus::Received(received_interface)) = result {
+        writeln!(
+            Console::new(),
+            "Received packet = {:02x?} on interface {:?}",
+            buf as &[u8],
+            received_interface as u8,
```
ia0, line 148: Same here.

ia0 (dismissed):

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: My intention was (2), just had a `:?` too much.

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: Done

ia0 (approved):

PR #453, opened 2022-04-05
Feature/timeint

shilingwangggg:
To discuss

### DIFF
--- a/src/api/clock.rs
+++ b/src/api/clock.rs
@@ -1,0 +1,223 @@
+use embedded_time::TimeInt;
+use embedded_time::fraction::Fraction;
+use core::fmt;
+use core::ops::{Add, Sub, Mul, Div, Rem};
+
+#[derive(Clone, Copy, PartialOrd, Ord, PartialEq, Eq, fmt::Debug)]
+pub struct U24(u32);
+impl U24 {
+    pub const unsafe fn new_unchecked(n: u32) -> Self {
+        // SAFETY: this is guaranteed to be safe by the caller.
+        Self(n)
+    }
+
+    pub const fn new(n: u32) -> Option<Self> {
+        if n <= 0xffffff {
+            Some(Self(n))
+        } else {
+            None
+        }
+    }
+
+    pub const fn new_wrap(n: u32) -> Self {
+        Self(n & 0xffffff)
+    }
+
+    pub const fn get(self) -> u32 {
+        self.0
+    }
+}
+
+impl fmt::Display for U24 {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl Add for U24 {
+    type Output = Self;
+    fn add(self, other: Self) -> Self {
+        let res = self.get().add(other.get());
+        debug_assert!(res <= 0xffffff);
+        Self::new_wrap(res)
+    }
+}
+
+impl Sub for U24 {
+    type Output = Self;
+    fn sub(self, other: Self) -> Self {
+        let res = self.get().sub(other.get());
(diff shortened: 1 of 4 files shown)


src/api/clock.rs
```diff
@@ -0,0 +1,52 @@
+use embedded_time::TimeInt;
+use embedded_time::fraction::Fraction;
+
+#[derive(Clone, Copy, PartialOrd, Ord, Eq, fmt::Debug)]
```
ia0: We shouldn't derive `PartialOrd` and `Ord`. If we need those traits, we should implement it ourselves because the derived implementation is wrong.

ia0: I'm surprised you can derive `Eq` without `PartialEq`. You should add `PartialEq`.

ia0 (commented):
To help development you can declare the module in `src/api/mod.rs`. This way the compiler will compile it (and that's also needed in the end).

src/api/clock.rs
shilingwangggg: Yes, my mistake.

shilingwangggg: I do not understand why the derived implementation is wrong. Using our creation functions (` new()`, `new_wrap()`) we are assured that only the 24 LSB of the u32 are non-zero. Therefore, it looks safe to me that `PartialOrd` and `Ord`  hold. Do I miss anything?

shilingwangggg:
I implemented all required traits on U24 to make it a valid `TimeInt`.

The user code of ` TimeInt`  requires quite some update too because we are currently using `TimeInt` in many const context. However, with `U24` we do not have a primitive type anymore, and therefore operators such as ` as`  are not available, ` std::ops` cannot be declared consts without using an [unstable feature](https://github.com/rust-lang/rust/issues/90080), and `from` is also non-consts.

Do you have any idea how should I proceed?

src/api/mod.rs
```diff
@@ -5,3 +5,4 @@
 
 pub mod firmware_protection;
 pub mod upgrade_storage;
+pub mod clock;
```
ia0, line 8: This should be sorted.

src/clock.rs
```diff
@@ -4,6 +4,8 @@ use embedded_time::duration::Milliseconds;
 pub use embedded_time::Clock;
 #[cfg(not(feature = "std"))]
 use libtock_drivers::result::FlexUnwrap;
+#[cfg(target_pointer_width = "32")]
```
ia0, line 7: I think we always want to use `U24` for now, in both test and prod.

src/api/clock.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl num::traits::WrappingAdd for U24 {
+    fn wrapping_add(&self, v: &Self) -> Self {
+        Self::new_wrap(self.get().wrapping_add(v.get()))
+    }
+}
+
+impl num::traits::WrappingSub for U24 {
+    fn wrapping_sub(&self, v: &Self) -> Self {
+        Self::new_wrap(self.get().wrapping_sub(v.get()))
```
ia0, line 164: We need to `& 0xffffff` here and in `WrappingAdd` too.

ia0 (commented):
I think this shows that we don't want to use `embedded-time` and use our own abstraction instead. I think this is a bug in `embedded-time` to ask `TimeInt` to implement `Integer` (etc) when it really should only use wrapping operations. I think they ask for that because they use the same representation for ticks and time AFAICT. And in our case we don't want that. We want to use non-wrapping `usize` for time (seconds, etc) and wrapping `u24` for ticks. I say non-wrapping because it's probably a bug to have wrapping seconds, while doing it at the ticks level is expected.

kaczmarczyck:
Picked up by #596

PR #458, opened 2022-04-12
Introduce Customization struct

hcyang-google:
* Introduce Customization struct including the customizations that
  control various behaviors.

* Expose Customization through a getter API in Env, and make the code
  that directly access the constants currently switch to accessing the
  customizations via Env.

* TockEnv's customization getter implementation directly returns the
  reference of the global DEFAULT_CUSTOMIZATION constant, so the
  constant values are still inlined and dead code won't be compiled.

* We'll add the customizations from global constants to the struct
  one-by-one, only MAX_MSG_SIZE in this commit.

Fixes #<issue_number_goes_here>

- [ ] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -473,8 +473,6 @@ class OpenSKInstaller:
     print("Testing invariants in customization.rs...")
     self.checked_command_output(
         ["cargo", "test", "--features=std", "--lib", "customization"])
-    self.checked_command_output(
-        ["cargo", "test", "--features=std", "--lib", "api::customization"])
 
   def generate_crypto_materials(self, force_regenerate: bool):
     """Calls a shell script that generates cryptographic material."""
--- a/deploy.py
+++ b/deploy.py
@@ -473,6 +473,8 @@ class OpenSKInstaller:
     print("Testing invariants in customization.rs...")
     self.checked_command_output(
         ["cargo", "test", "--features=std", "--lib", "customization"])
+    self.checked_command_output(
+        ["cargo", "test", "--features=std", "--lib", "api::customization"])
 
   def generate_crypto_materials(self, force_regenerate: bool):
     """Calls a shell script that generates cryptographic material."""
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -1,0 +1,62 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! This file contains all customizable constants.
+//!
+//! If you adapt them, make sure to run the tests before flashing the firmware.
+//! Our deploy script enforces the invariants.
+
+pub trait Customization {
+    /// Maximum message size send for CTAP commands.
+    ///
+    /// The maximum value is 7609, as HID packets can not encode longer messages.
+    /// 1024 is the default mentioned in the authenticatorLargeBlobs commands.
+    /// Larger values are preferred, as that allows more parameters in commands.
+    /// If long commands are too unreliable on your hardware, consider decreasing
+    /// this value.
+    fn max_msg_size(&self) -> usize;
+}
+
+#[derive(Clone)]
+pub struct CustomizationImpl {
+    pub max_msg_size: usize,
+}
+
+pub const DEFAULT_CUSTOMIZATION: CustomizationImpl = CustomizationImpl { max_msg_size: 7609 };
+
+impl Customization for CustomizationImpl {
+    fn max_msg_size(&self) -> usize {
+        self.max_msg_size
+    }
+}
+
+#[cfg(test)]
+fn is_valid(customization: &impl Customization) -> bool {
+    // Max message size must be between 1024 and 7609.
+    if customization.max_msg_size() < 1024 || customization.max_msg_size() > 7609 {
+        return false;
+    }
+    true
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_invariants() {
+        assert!(is_valid(&DEFAULT_CUSTOMIZATION));
+    }
+}
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -41,8 +41,8 @@ impl Customization for CustomizationImpl {
     }
 }
 
-#[cfg(test)]
-fn is_valid(customization: &impl Customization) -> bool {
+#[cfg(feature = "std")]
+pub fn is_valid(customization: &impl Customization) -> bool {
     // Max message size must be between 1024 and 7609.
     if customization.max_msg_size() < 1024 || customization.max_msg_size() > 7609 {
         return false;
--- a/src/ctap/customization.rs
+++ b/src/ctap/customization.rs
@@ -119,15 +119,6 @@ pub const ENTERPRISE_ATTESTATION_MODE: Option<EnterpriseAttestationMode> = None;
 /// VendorFacilitated.
 pub const ENTERPRISE_RP_ID_LIST: &[&str] = &[];
 
-/// Maximum message size send for CTAP commands.
-///
-/// The maximum value is 7609, as HID packets can not encode longer messages.
-/// 1024 is the default mentioned in the authenticatorLargeBlobs commands.
-/// Larger values are preferred, as that allows more parameters in commands.
-/// If long commands are too unreliable on your hardware, consider decreasing
-/// this value.
-pub const MAX_MSG_SIZE: usize = 7609;
-
 /// Sets the number of consecutive failed PINs before blocking interaction.
 ///
 /// # Invariant
@@ -252,8 +243,6 @@ mod test {
         } else {
             assert!(ENTERPRISE_RP_ID_LIST.is_empty());
         }
-        assert!(MAX_MSG_SIZE >= 1024);
-        assert!(MAX_MSG_SIZE <= 7609);
         assert!(MAX_PIN_RETRIES <= 8);
         assert!(MAX_CRED_BLOB_LENGTH >= 32);
         if let Some(count) = MAX_CREDENTIAL_COUNT_IN_LIST {
--- a/src/ctap/hid/mod.rs
+++ b/src/ctap/hid/mod.rs
@@ -235,7 +235,7 @@ impl CtapHid {
         packet: &HidPacket,
         clock_value: CtapInstant,
     ) -> Option<Message> {
-        match self.assembler.parse_packet(packet, clock_value) {
+        match self.assembler.parse_packet(env, packet, clock_value) {
             Ok(Some(message)) => {
                 debug_ctap!(env, "Received message: {:02x?}", message);
                 self.preprocess_message(message)
@@ -420,6 +420,7 @@ mod test {
 
     #[test]
     fn test_split_assemble() {
+        let mut env = TestEnv::new();
         for payload_len in 0..7609 {
             let message = Message {
                 cid: [0x12, 0x34, 0x56, 0x78],
@@ -430,7 +431,7 @@ mod test {
             let mut messages = Vec::new();
             let mut assembler = MessageAssembler::new();
             for packet in HidPacketIterator::new(message.clone()).unwrap() {
-                match assembler.parse_packet(&packet, CtapInstant::new(0)) {
+                match assembler.parse_packet(&mut env, &packet, CtapInstant::new(0)) {
                     Ok(Some(msg)) => messages.push(msg),
                     Ok(None) => (),
                     Err(_) => panic!("Couldn't assemble packet: {:02x?}", &packet as &[u8]),
--- a/src/ctap/hid/receive.rs
+++ b/src/ctap/hid/receive.rs
@@ -12,12 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::clock::CtapInstant;
-
-use super::super::customization::MAX_MSG_SIZE;
 use super::{
     ChannelID, CtapHid, CtapHidCommand, CtapHidError, HidPacket, Message, ProcessedPacket,
 };
+use crate::api::customization::Customization;
+use crate::clock::CtapInstant;
+use crate::env::Env;
 use alloc::vec::Vec;
 use core::mem::swap;
 
@@ -73,6 +73,7 @@ impl MessageAssembler {
     // packet was received.
     pub fn parse_packet(
         &mut self,
+        env: &mut impl Env,
         packet: &HidPacket,
         timestamp: CtapInstant,
     ) -> Result<Option<Message>, (ChannelID, CtapHidError)> {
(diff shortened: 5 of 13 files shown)


src/ctap/customization.rs
```diff
@@ -19,6 +19,15 @@
 
 use crate::ctap::data_formats::{CredentialProtectionPolicy, EnterpriseAttestationMode};
 
+#[derive(Clone)]
+pub struct Customization {
+    pub max_msg_size: usize,
+}
+
+pub const DEFAULT_CUSTOMIZATION: Customization = Customization {
+    max_msg_size: MAX_MSG_SIZE,
```
hcyang-google: @ia0 WDYT about the structure in this file? Should we keep the constants and comments in original place and use this style, or do you think we should put all comments and values directly inside this struct and delete the original consts?

src/ctap/hid/receive.rs
```diff
@@ -143,6 +143,7 @@ impl MessageAssembler {
 
     fn parse_init_packet(
         &mut self,
+        env: &impl Env,
```
hcyang-google: There's APIs like this that needs to access the constants but don't have env as parameter originally. This one can be modified to gain access to Env easily but I'm not sure if there's any other places that are more difficult to get Env.

hcyang-google:
@ia0 Should I add these customizations in n PRs? Or in one (or few) PRs each containing n commits?

src/lib.rs
```diff
@@ -88,6 +88,11 @@ impl<E: Env> Ctap<E> {
         &mut self.hid
     }
 
+    #[cfg(feature = "fuzz")]
```
hcyang-google: @ia0 I don't really like exposing this here. But otherwise when using assembler individually (like in fuzz helper) we'll have to pass another env as parameter. But I think we should use the same env. Any better ways?

src/ctap/customization.rs
kaczmarczyck: Inlining is okay with me.

Thinking of documentation, I noticed that the comments right now contain information about:
- What is this value?
- Why this default value?

Should be okay to just document the `DEFAULT_CUSTOMIZATION` like before, but just wanted to point out that we could be more precise here.

src/ctap/hid/receive.rs
kaczmarczyck: It shouldn't get harder to add than this.

FYI: In this particular instance actually, I was wondering if we can make the HID parser completely independent of CTAP. This constant is one of the very few dependencies left. But it's an independent question from this PR, so feel free to ignore :)

kaczmarczyck (commented):
Thanks for experimenting on this!

src/ctap/customization.rs
ia0: Good question. The way I see it:

```rust
// src/api/customization.rs
pub trait Customization {
    /// "Documentation of the constant"
    fn the_constant(&self) -> TheType;  // IMPORTANT: This is &self
    // ... and so on ...
}

pub struct Impl {
    pub the_constant: TheType,
    // ...
}

impl Customization for Impl {
    fn the_constant(&self) -> TheType { self.the_constant }
    // ...
}

pub const DEFAULT: Impl = Impl {
    the_constant: THE_DEFAULT,
    // ...
}

// src/env/mod.rs
pub trait Env {
    // ...
    type Customization: Customization;

    // ...
    fn customization(&mut self) -> &mut Customization;  // IMPORTANT: This is &mut self
    // I think we should always use `&mut env` and thus it's not a huge restriction to have &mut self here.
    // Also it's needed to modify the customization in tests. Because I think it's more convenient to just create
    // the Test Env without arguments, then modify it, and then use it. Rather than pass a customization at
    // creation time.
}
```

So `src/ctap/customization.rs` would disappear and be moved to `src/api`. Since you're doing this incrementally (constant by constant) which is good, this migration will occur incrementally.

src/ctap/hid/mod.rs
```diff
@@ -421,6 +421,7 @@ mod test {
     #[test]
     fn test_split_assemble() {
         for payload_len in 0..7609 {
+            let mut env = TestEnv::new();
```
ia0: Do we need to create this in the loop? I would have put it outside the loop.

src/ctap/hid/receive.rs
ia0: +1 it's not a problem to add `&mut impl Env` where needed. Also in the future (for clock) we might need to make it `&mut E` and have `<E: Env>` on the function or class (like `CtapState`).

src/env/test/mod.rs
```diff
@@ -113,4 +117,8 @@ impl Env for TestEnv {
     fn write(&mut self) -> Self::Write {
         TestWrite
     }
+
+    fn customization(&self) -> &Customization {
```
ia0: This will need to be `mut`.

src/lib.rs
ia0: I think this function makes sense even for normal tests so I would make it `std` instead of `fuzz`. The scenario I have in mind for tests is something like error injection:
- Create test env
- Create Ctap
- Do some stuff that use some part of the env is successful way
- Modify env to inject errors
- Do some stuff that use the same part of the env as before but this time should fail

ia0 (commented):
Looks good overall. I would just move the new API in `src/api` with the rest. Also I think we want a trait to both have flexibility in the future as well as permit mutation for tests but not prod.

src/ctap/customization.rs
hcyang-google: Looks good, except `fn customization(&mut self) -> &mut Customization;` won't work because we are returning a ref to the constant DEFAULT in `TockEnv`. Specifically customization can't be changed in prod.
So the trait method should be kept as `fn customization(&self) -> &Customization;`, but expose another method `customization_mut` in `TestEnv`. 

I didn't use trait at first because in my imagination that impls will always be just containing exact identical fields as the constants we want to access. But using trait makes sense too since some constants are dependent on each other and we can have less fields (than num of constants) in TestCustomization. This way we can enforce valid customization settings when fuzzing (if we want).

kaczmarczyck: We will have to enforce valid settings there. See the unit test in the original `customization.rs`, this has to be checked.

src/ctap/hid/mod.rs
hcyang-google: Done

src/env/test/mod.rs
hcyang-google: I exposed `customization_mut` in `impl TestEnv` now instead of in trait impl.

src/api/customization.rs
```diff
@@ -0,0 +1,54 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_invariants() {
+        assert!(DEFAULT_CUSTOMIZATION.max_msg_size >= 1024);
+        assert!(DEFAULT_CUSTOMIZATION.max_msg_size <= 7609);
```
ia0: Let's instead have a validation function if those invariants should hold for all customizations.

```rust
fn is_valid(customization: &impl Customization) -> bool {
    // explanation
    if ... { return false; }
    // explanation
    if ... { return false; }
    true
}
```

Then the test would just `assert!(is_valid(&DEFAULT_CUSTOMIZATION))`.

src/ctap/hid/receive.rs
ia0: I would still make this `&mut impl Env` instead of `&impl Env` even if we don't need it to be mutable. Because we don't have concurrent code and the environment is always threaded, so it's not an issue. And having it already mut would permit to not have to change the call-sites in tests when we'll need to access mutable parts of the environment (like debug printing).

src/env/test/mod.rs
```diff
@@ -27,6 +29,8 @@ impl core::fmt::Write for TestWrite {
     }
 }
 
+pub type TestCustomization = CustomizationImpl;
```
ia0: Let's inline this. We don't need to hide the actual type since we don't access it with just traits.

src/api/customization.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_invariants() {
```
kaczmarczyck, line 59: `deploy.py` runs the customization test to enforce that all constants are correctly set. Can you add this one as well?

https://github.com/google/OpenSK/blob/81996f650e7ddbb60461242e62b40cb0f467b183/deploy.py#L471

ia0 (dismissed):
Can you put in the PR description the binary size before and after (and with which flags)?

src/ctap/customization.rs
ia0: Right, I did not think of the problem returning a constant in `TockEnv`. The `customization_mut` sounds good.

ia0:
> Should I add these customizations in n PRs? Or in one (or few) PRs each containing n commits?

Let's merge this one first because it introduces the logic. For the rest, you can split as you judge appropriate. For example, you can migrate until you hit an obstacle, and create a PR before that obstacle and another one for that obstacle.

kaczmarczyck (commented):
LGTM, I'll let @ia0 approve. Please add the test in deploy.

src/api/customization.rs
kaczmarczyck: We can probably compile that function for `std` only?

ia0: What would be the motivation? I can see 3:
- Reduce compilation time. I'm not sure of the impact and if it's needed.
- Prevent users to use it outside tests. Makes sense. But not sure why this should be forbidden.
- Reduce binary size. This shouldn't be a concern if the function is not used.

kaczmarczyck: It conveys the intention that all of these are constants, so they should be tested during compile time or tests, not runtime.

What we initially wanted was static assertions, but IIRC that wasn't well supported back then. Now we have an additional use case, but runtime testing in prod is still a bad idea.

hcyang-google: Actually, we should compile that for cfg(test) only otherwise there will be dead code warning

src/api/customization.rs
hcyang-google: done

src/env/test/mod.rs
hcyang-google: done

hcyang-google:
Regarding binary size, the command I used is ` cargo bloat --release --target=thumbv7em-none-eabi --features=with_ctap1`.
Before this PR:
```
File  .text     Size            Crate Name
 6.8%  18.9%  22.1KiB            ctap2 ctap2::ctap::CtapState::process_fido_command
 4.2%  11.6%  13.6KiB        [Unknown] main
 3.9%  10.8%  12.7KiB            ctap2 ctap2::ctap::command::Command::deserialize
 0.5%   1.5%   1.8KiB              std core::slice::sort::recurse
 0.5%   1.5%   1.7KiB          sk_cbor sk_cbor::writer::Writer::encode_cbor
 0.5%   1.3%   1.6KiB            ctap2 ctap2::ctap::CtapState::assertion_response
 0.5%   1.3%   1.5KiB            ctap2 ctap2::ctap::storage::deserialize_credential
 0.4%   1.0%   1.2KiB          sk_cbor sk_cbor::reader::Reader::decode_complete_data_item
 0.3%   1.0%   1.1KiB            ctap2 ctap2::ctap::credential_management::get_stored_rp_ids
 0.3%   0.8%    1010B           crypto <&crypto::ec::montgomery::Montgomery as core::ops::arith::Mul>::mul
 0.3%   0.8%     918B           crypto crypto::ec::montgomery::Montgomery::square
 0.3%   0.7%     862B              std core::fmt::Formatter::pad
 0.2%   0.7%     820B            ctap2 <ctap2::ctap::data_formats::CoseKey as core::convert::TryFrom<sk_cbor::values::Value>>::try_from
 0.2%   0.6%     734B persistent_store persistent_store::store::Store<S>::transaction
 0.2%   0.6%     726B persistent_store persistent_store::fragment::get_handles
 0.2%   0.6%     704B            ctap2 <ctap2::ctap::data_formats::PublicKeyCredentialDescriptor as core::convert::TryFrom<sk_cbor::values::Value...
 0.2%   0.5%     656B           crypto crypto::ec::montgomery::Montgomery::reduce_degree
 0.2%   0.5%     652B           crypto crypto::ecdsa::SecKey::sign_rfc6979
 0.2%   0.5%     612B            ctap2 <ctap2::env::tock::TockEnv as ctap2::env::UserPresence>::check
 0.2%   0.5%     612B           crypto crypto::ec::point::PointP256::from_projective
16.0%  44.1%  51.7KiB                  And 936 smaller methods. Use -n N to show more.
36.3% 100.0% 117.1KiB                  .text section size, the file size is 323.0KiB
```
After this PR:
```
 File  .text     Size            Crate Name
 6.9%  18.9%  22.1KiB            ctap2 ctap2::ctap::CtapState::process_fido_command
 4.2%  11.6%  13.5KiB        [Unknown] main
 3.9%  10.8%  12.7KiB            ctap2 ctap2::ctap::command::Command::deserialize
 0.5%   1.5%   1.8KiB              std core::slice::sort::recurse
 0.5%   1.5%   1.7KiB          sk_cbor sk_cbor::writer::Writer::encode_cbor
 0.5%   1.3%   1.6KiB            ctap2 ctap2::ctap::CtapState::assertion_response
 0.5%   1.3%   1.5KiB            ctap2 ctap2::ctap::storage::deserialize_credential
 0.4%   1.0%   1.2KiB          sk_cbor sk_cbor::reader::Reader::decode_complete_data_item
 0.3%   1.0%   1.1KiB            ctap2 ctap2::ctap::credential_management::get_stored_rp_ids
 0.3%   0.8%    1010B           crypto <&crypto::ec::montgomery::Montgomery as core::ops::arith::Mul>::mul
 0.3%   0.8%     918B           crypto crypto::ec::montgomery::Montgomery::square
 0.3%   0.7%     862B              std core::fmt::Formatter::pad
 0.2%   0.7%     820B            ctap2 <ctap2::ctap::data_formats::CoseKey as core::convert::TryFrom<sk_cbor::values::Value>>::try_from
 0.2%   0.6%     734B persistent_store persistent_store::store::Store<S>::transaction
 0.2%   0.6%     726B persistent_store persistent_store::fragment::get_handles
 0.2%   0.6%     704B            ctap2 <ctap2::ctap::data_formats::PublicKeyCredentialDescriptor as core::convert::TryFrom<sk_cbor::values::Value...
 0.2%   0.5%     656B           crypto crypto::ec::montgomery::Montgomery::reduce_degree
 0.2%   0.5%     652B           crypto crypto::ecdsa::SecKey::sign_rfc6979
 0.2%   0.5%     612B            ctap2 <ctap2::env::tock::TockEnv as ctap2::env::UserPresence>::check
 0.2%   0.5%     612B           crypto crypto::ec::point::PointP256::from_projective
16.0%  44.1%  51.6KiB                  And 933 smaller methods. Use -n N to show more.
36.3% 100.0% 117.0KiB                  .text section size, the file size is 322.6KiB
```

By the way, how about adding a github work flow to analyze binary size increase? We have that in Chromium.

kaczmarczyck (dismissed):

src/api/customization.rs
ia0: That's weird if we get a dead-code warning. If it's a public function it should be fine. We might use it in fuzzing.

Sounds good for the `std` gate.

deploy.py
```diff
@@ -473,6 +473,8 @@ def _check_invariants(self):
     print("Testing invariants in customization.rs...")
     self.checked_command_output(
         ["cargo", "test", "--features=std", "--lib", "customization"])
+    self.checked_command_output(
+        ["cargo", "test", "--features=std", "--lib", "api::customization"])
```
ia0: I'm surprised we need this. The command above should also contain `api::customization` because it just greps anything in the test name. I would remove it if it's the case because we test twice.

kaczmarczyck: Ah I didn't read carefully. My intention was: Make sure that both tests are run before deploying. If no change is required for that, good with me.

hcyang-google: Fixed, need approve again.

src/api/customization.rs
hcyang-google: Done

kaczmarczyck (approved):

PR #459, opened 2022-04-14
Move more customization constants to new file

hcyang-google:
- Move DEFAULT_CRED_PROTECT to customization api

### DIFF
--- a/src/ctap/customization.rs
+++ b/src/ctap/customization.rs
@@ -17,13 +17,11 @@
 //! If you adapt them, make sure to run the tests before flashing the firmware.
 //! Our deploy script enforces the invariants.
 
-use crate::ctap::data_formats::{EnterpriseAttestationMode};
+use crate::ctap::data_formats::EnterpriseAttestationMode;
 
 // ###########################################################################
 // Constants for adjusting privacy and protection levels.
 // ###########################################################################
-
-
 
 /// Sets the initial minimum PIN length in code points.
 ///
--- a/src/ctap/customization.rs
+++ b/src/ctap/customization.rs
@@ -17,29 +17,13 @@
 //! If you adapt them, make sure to run the tests before flashing the firmware.
 //! Our deploy script enforces the invariants.
 
-use crate::ctap::data_formats::{CredentialProtectionPolicy, EnterpriseAttestationMode};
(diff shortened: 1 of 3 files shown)


src/ctap/customization.rs
```diff
@@ -17,29 +17,13 @@
 //! If you adapt them, make sure to run the tests before flashing the firmware.
 //! Our deploy script enforces the invariants.
 
-use crate::ctap::data_formats::{CredentialProtectionPolicy, EnterpriseAttestationMode};
+use crate::ctap::data_formats::{EnterpriseAttestationMode};
```
ia0: rustfmt doesn't remove those braces?

src/ctap/mod.rs
```diff
@@ -765,9 +765,12 @@ impl CtapState {
 
         let mut cred_protect_policy = extensions.cred_protect;
         if cred_protect_policy.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
-            < DEFAULT_CRED_PROTECT.unwrap_or(CredentialProtectionPolicy::UserVerificationOptional)
+            < env
+                .customization()
+                .default_cred_protect()
```
ia0: Let's maybe introduce a local variable for those 2 usages.

hcyang-google: Done

src/ctap/customization.rs
hcyang-google: Fixed

src/ctap/mod.rs
```diff
@@ -763,11 +763,13 @@ impl CtapState {
         env.user_presence().check(channel)?;
         self.client_pin.clear_token_flags();
 
+        let default_cred_protect = env.customization().default_cred_protect();
+
```
ia0: nit: I would remove this empty line. This variable is "part" of the following block which deals with cred protect.

ia0 (dismissed):

src/ctap/mod.rs
hcyang-google: Done. And I want to merge these first because the next one has some more refactors (introducing TestCustomization that makes array ref constants configurable).

ia0 (approved):

src/ctap/mod.rs
ia0: Sounds good.

PR #460, opened 2022-04-14
Move three dependent customizations into new file

hcyang-google:
* default_min_pin_length(_rp_ids) and max_rp_ids_length

* Did some backing store tricks to make the list configurable in
  TestCustomization.

I need some suggestions on implementing the TestCustomization's list field.

### DIFF
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -18,6 +18,8 @@
 //! Our deploy script enforces the invariants.
 
 use crate::ctap::data_formats::CredentialProtectionPolicy;
+use alloc::string::String;
+use alloc::vec::Vec;
 
 pub trait Customization {
     // ###########################################################################
@@ -66,7 +68,7 @@ pub trait Customization {
     ///
     /// Only the RP IDs listed in default_min_pin_length_rp_ids are allowed to read
     /// the minimum PIN length with the minPinLength extension.
-    fn default_min_pin_length_rp_ids(&self) -> &[&str];
+    fn default_min_pin_length_rp_ids(&self) -> Vec<String>;
 
     /// Enforces the alwaysUv option.
     ///
@@ -164,8 +166,11 @@ impl Customization for CustomizationImpl {
         self.default_min_pin_length
     }
 
-    fn default_min_pin_length_rp_ids(&self) -> &[&str] {
+    fn default_min_pin_length_rp_ids(&self) -> Vec<String> {
         self.default_min_pin_length_rp_ids
+            .iter()
+            .map(|s| String::from(*s))
+            .collect()
     }
 
     fn enforce_always_uv(&self) -> bool {
@@ -221,7 +226,6 @@ mod test {
     use super::*;
 
     #[test]
-    #[allow(clippy::assertions_on_constants)]
     fn test_invariants() {
         assert!(is_valid(&DEFAULT_CUSTOMIZATION));
     }
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -42,6 +42,32 @@ pub trait Customization {
     /// This can improve privacy, but can make usage less comfortable.
     fn default_cred_protect(&self) -> Option<CredentialProtectionPolicy>;
 
+    /// Sets the initial minimum PIN length in code points.
+    ///
+    /// # Invariant
+    ///
+    /// - The minimum PIN length must be at least 4.
+    /// - The minimum PIN length must be at most 63.
+    /// - default_min_pin_length_rp_ids() must be non-empty if MAX_RP_IDS_LENGTH is 0.
+    ///
+    /// Requiring longer PINs can help establish trust between users and relying
+    /// parties. It makes user verification harder to break, but less convenient.
+    /// NIST recommends at least 6-digit PINs in section 5.1.9.1:
+    /// https://pages.nist.gov/800-63-3/sp800-63b.html
+    ///
+    /// Reset reverts the minimum PIN length to this DEFAULT_MIN_PIN_LENGTH.
+    fn default_min_pin_length(&self) -> u8;
+
+    /// Lists relying parties that can read the minimum PIN length.
+    ///
+    /// # Invariant
+    ///
+    /// - default_min_pin_length_rp_ids() must be non-empty if MAX_RP_IDS_LENGTH is 0
+    ///
+    /// Only the RP IDs listed in default_min_pin_length_rp_ids are allowed to read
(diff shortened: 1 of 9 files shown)


src/env/test/customization.rs
```diff
@@ -0,0 +1,82 @@
...
+impl Customization for TestCustomization {
+    fn default_cred_protect(&self) -> Option<CredentialProtectionPolicy> {
+        self.default_cred_protect
+    }
+
+    fn default_min_pin_length(&self) -> u8 {
+        self.default_min_pin_length
+    }
+
+    fn default_min_pin_length_rp_ids(&self) -> &[&str] {
```
hcyang-google: @ia0 @kaczmarczyck is there better way of doing this? The current implementation is quite complicated and I feel it's unnecessary.

hcyang-google:
Add a test to verify that the list configuration implementation isn't broken.

src/env/test/customization.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::api::customization::{is_valid, DEFAULT_CUSTOMIZATION};
+
+    #[test]
+    #[allow(clippy::assertions_on_constants)]
+    fn test_invariants() {
```
kaczmarczyck, line 88: This should be 2 tests: One part is testing the current customization, and the other is testing your implementation.

src/env/test/customization.rs
ia0: Yes, I think we should change the `Customization` API:

```rust
// Current PR:
fn default_min_pin_length_rp_ids(&self) -> &[&str];
// Proposal:
fn default_min_pin_length_rp_ids(&self) -> Vec<String>;
```

The cost of creating the vector seems to already be present in `ctap::storage::min_pin_length_rp_ids()`. It's not as clear for `ctap::storage::set_min_pin_length_rp_ids` because if the RP id was already present in the function argument then we recreate it. But it's probably fine since the cost is comparable to the final result.

There are 2 alternatives:
```rust
// This isn't yet supported (but will in the future):
fn default_min_pin_length_rp_ids(&self) -> impl Iterator<Item = &str>;
// This would add rp_ids if they are not already there. It could either be called with an empty Vec to get the list, or with the argument to merge. This is essentially the union of both usages.
fn merge_default_min_pin_length_rp_ids(&self, rp_ids: &mut Vec<String>);
```

src/env/test/customization.rs
hcyang-google: fixed

src/api/customization.rs
```diff
@@ -50,27 +88,110 @@ pub trait Customization {
...
 impl Customization for CustomizationImpl {
     fn default_cred_protect(&self) -> Option<CredentialProtectionPolicy> {
         self.default_cred_protect
     }
 
+    fn default_min_pin_length(&self) -> u8 {
+        self.default_min_pin_length
+    }
+
+    fn default_min_pin_length_rp_ids(&self) -> Vec<String> {
```
hcyang-google, line 169: @ia0 Now we do this, so this customization (and ENTERPRISE_RP_ID_LIST) won't be treated as constants anymore. But I think it should be fine? There's very limited code path affected.

src/api/customization.rs
```diff
@@ -42,6 +44,42 @@ pub trait Customization {
     /// This can improve privacy, but can make usage less comfortable.
     fn default_cred_protect(&self) -> Option<CredentialProtectionPolicy>;
 
+    /// Sets the initial minimum PIN length in code points.
+    ///
+    /// # Invariant
+    ///
+    /// - The minimum PIN length must be at least 4.
+    /// - The minimum PIN length must be at most 63.
+    /// - default_min_pin_length_rp_ids() must be non-empty if MAX_RP_IDS_LENGTH is 0.
```
ia0: nit: `MAX_RP_IDS_LENGTH` is now `max_rp_ids_length` but we could as well do a pass on documentation after the code change if it's easier (there's a few other cases like this one below, I'm not commenting on them to minimize comments to those regarding design).

src/api/customization.rs
ia0: Indeed they look less like constants. But I would be a bit surprised if some optimizations would not trigger. The optimization we rely on is when we iterate on this list in `set_min_pin_length_rp_ids`. The whole loop should be eliminated. We could try to check that it's still the case. One possible trick is to call an extern function in the loop and see if the linker complains:

```rust
    for &rp_id in DEFAULT_MIN_PIN_LENGTH_RP_IDS.iter() {
        // Add the 2 following lines to test if it still links
        extern "C" { fn should_not_be_called(); }
        unsafe { should_not_be_called() };
        let rp_id = String::from(rp_id);
        if !min_pin_length_rp_ids.contains(&rp_id) {
            min_pin_length_rp_ids.push(rp_id);
        }
    }
```

hcyang-google:
@ia0 I discovered my previous tests about whether compiler notices env customization are constants were wrong. After using the extern function testing I found out that compiler doesn't know they are constants. It just turned out that these customization constants have no big effect at the current codebase yet. I'll investigate further whether we can solve this.

src/api/customization.rs
hcyang-google: The optimization won't trigger. I think it's because to eliminate the loop code the function has be able to be computed at compilation time. But we have some heap allocation like `String::from` in that function which make it impossible to do so?

ia0: I don't think the `String::from` should matter if the list is empty. But yeah, it's possible the constant propagation is not aggressive enough. In particular, if `collect` is not able to see that the iterator is empty, then we won't get the optimization.

src/api/customization.rs
hcyang-google: done. @ia0 are there more that needs to be improved in this CL or can this be approved and merged?

ia0: I wanted to test a bit the optimizations to see if there was a way to make it work with `Vec<String>` too. But actually ever `Vec<&'static str>` doesn't seem to work. Probably the `iter()` is too much.

ia0 (commented):
This is approved on my side, but since it's a big PR, I think we should also wait for @kaczmarczyck to take a look too.

kaczmarczyck (approved):
I had a comment 5 days ago, and left everything else to you :)
I think the most interesting decision was the change from static strings, and since you are aware, feel free to merge when you think the impact on performance and binary size is small enough.

PR #461, opened 2022-04-15
Make rng in TestEnv deterministic and seedable

hcyang-google:
@ia0 , I want to try take some bytes (a u64) in fuzz data in our process_* fuzz targets to seed the rng, but that will affect the corpus's effectiveness right? Because after taken the bytes the valid cbor formats in corpus might become invalid.

### DIFF
--- a/fuzz/fuzz_helper/src/lib.rs
+++ b/fuzz/fuzz_helper/src/lib.rs
@@ -128,25 +128,40 @@ fn process_message(data: &[u8], ctap: &mut Ctap<TestEnv>) {
 // Interprets the raw data as any ctap command (including the command byte) and
 // invokes message splitting, packet processing at CTAP HID level and response assembling
 // using an initialized and allocated channel.
-pub fn process_ctap_any_type(data: &[u8]) {
+pub fn process_ctap_any_type(data: &[u8]) -> arbitrary::Result<()> {
+    let mut unstructured = Unstructured::new(data);
+
+    let mut env = TestEnv::new();
+    env.rng()
+        .seed_rng_from_u64(u64::arbitrary(&mut unstructured)?);
+
+    let data = unstructured.take_rest();
     // Initialize ctap state and hid and get the allocated cid.
-    let mut ctap = Ctap::new(TestEnv::new(), CtapInstant::new(0));
+    let mut ctap = Ctap::new(env, CtapInstant::new(0));
     let cid = initialize(&mut ctap);
     // Wrap input as message with the allocated cid.
     let mut command = cid.to_vec();
     command.extend(data);
     process_message(&command, &mut ctap);
+    Ok(())
 }
 
 // Interprets the raw data as of the given input type and
 // invokes message splitting, packet processing at CTAP HID level and response assembling
 // using an initialized and allocated channel.
-pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) {
+pub fn process_ctap_specific_type(data: &[u8], input_type: InputType) -> arbitrary::Result<()> {
+    let mut unstructured = Unstructured::new(data);
+
+    let mut env = TestEnv::new();
+    env.rng()
+        .seed_rng_from_u64(u64::arbitrary(&mut unstructured)?);
+
+    let data = unstructured.take_rest();
     if !is_type(data, input_type) {
-        return;
+        return Ok(());
     }
     // Initialize ctap state and hid and get the allocated cid.
-    let mut ctap = Ctap::new(TestEnv::new(), CtapInstant::new(0));
+    let mut ctap = Ctap::new(env, CtapInstant::new(0));
     let cid = initialize(&mut ctap);
     // Wrap input as message with allocated cid and command type.
     let mut command = cid.to_vec();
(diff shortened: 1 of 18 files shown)


fuzz/fuzz_helper/src/lib.rs
```diff
@@ -189,6 +189,8 @@ pub fn process_ctap_structured(data: &[u8], input_type: InputType) -> arbitrary:
         }
     };
 
+    env.seed_rng_from_u64(u64::arbitrary(unstructured)?);
```
ia0: I would move this at the very top to be the first usage of `unstructured`. This would make the "fuzzing format" more stable because it would be 64 bits prefix followed by command input (which is variable sized).

src/env/test/mod.rs
```diff
@@ -70,6 +86,12 @@ impl TestEnv {
     pub fn customization_mut(&mut self) -> &mut CustomizationImpl {
         &mut self.customization
     }
+
+    pub fn seed_rng_from_u64(&mut self, state: u64) {
```
ia0: I would make this a method of `TestRng256` to get better organization, i.e. `env.rng().seed_from_u64(42)` instead of `env.seed_rng_from_u64(42)`.

hcyang-google: Done

hcyang-google:
Not quite sure why we're encountering this.

```
warning: unused import: `crate::env::Env`
   --> src/ctap/credential_management.rs:367:9
    |
367 |     use crate::env::Env;
    |         ^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
```

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -174,6 +174,8 @@ pub fn process_ctap_structured(data: &[u8], input_type: InputType) -> arbitrary:
     let mut env = TestEnv::new();
     let mut state = CtapState::new(&mut env, CtapInstant::new(0));
 
+    env.rng().seed_rng_from_u64(u64::arbitrary(unstructured)?);
```
ia0, line 191: Actually, we should actually move this right after the environment creation. The reason being that if CtapState uses `env.rng` we want it to use the seed from the fuzz input already.

The usual pattern for testing should always be:
```rust
// Create the test environment.
let mut env = TestEnv::new();
// Customize the test environment.
env.foo().set_bar();
// Use the test enviroment.
quux(&mut env);
// Possibly inject errors.
env.foo().next_bar(...);
// Continue using the test environment.
quux(&mut env);
```

ia0:
> Not quite sure why we're encountering this.

I don't know. I can't reproduce. Can you?

hcyang-google:
Fixed the unused warning, it only happens in a test I didn't modify, not sure how the unused import slipped through in the first place.

Also, should we add seed rng to other fuzzers? Or we don't want to break the existing corpus?

fuzz/fuzz_helper/src/lib.rs
hcyang-google: done

src/env/test/mod.rs
```diff
@@ -4,19 +4,39 @@ use crate::api::firmware_protection::FirmwareProtection;
...
 pub struct TestEnv {
-    rng: ThreadRng256,
+    rng: TestRng256,
     user_presence: TestUserPresence,
     store: Store<BufferStorage>,
     upgrade_storage: Option<BufferUpgradeStorage>,
     customization: CustomizationImpl,
 }
 
+pub struct TestRng256 {
```
kaczmarczyck, line 24: I was wondering before if RNG should be part of `crypto`, and this is another reason for separating it? So we don't have too many different places where we define some sort of RNG. Doesn't have to be right now, but I would like to hear your opinion.

(The other use case was a library that `crypto` was supposed to use, and that library needing RNG. Meaning a circular dependency because of RNG not being separated.)

kaczmarczyck (commented):
You can break as much as you like in fuzzing in order to improve it! It would be great if we can reuse the fuzzing corpus from the unstructured fuzzing later to replay it on other security keys though.

src/env/test/mod.rs
hcyang-google: I think this case doesn't serve as an additional reason to separate it because it's normal that fuzzer/test define its own implementation of some public interface together in a test folder. But I see no disadvantage of separating them too.

ia0: Yes, I agree that RNG should not be part of `crypto`.

ia0 (dismissed):

hcyang-google:
I seed the rng in all test targets, PTAL at the changes again, thanks!

kaczmarczyck (approved):

src/env/test/mod.rs
```diff
@@ -1,20 +1,42 @@
...
-    customization: CustomizationImpl,
+    customization: TestCustomization,
+}
+
+pub struct TestRng256 {
+    rng: StdRng,
+}
+
+impl TestRng256 {
+    pub fn seed_rng_from_u64(&mut self, state: u64) {
```
ia0, line 29: I missed the fact that this was `seed_rng_from_u64`. I would remove the `_rng` in your next PR since it's redundant.

PR #462, opened 2022-04-19
Workflow for cargo bloat

kaczmarczyck:
We often had to manually add binary size comparisons in PRs before, now we get a quick overview of what we usually care about automatically.

Important details:
- The binary is compiled with the most likely feature set: `with_ctap, vendor_hid`
- The reported size is the actual size of the app from Tock's perspective.

### DIFF
--- a/.github/workflows/cargo_bloat.yml
+++ b/.github/workflows/cargo_bloat.yml
@@ -3,10 +3,6 @@ on: pull_request
 
 jobs:
   cargo_bloat:
-    permissions:
-      contents: write 
-      issues: write
-      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       # Setup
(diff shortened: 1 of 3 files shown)


ia0 (dismissed):

.github/workflows/cargo_bloat.yml
```diff
@@ -0,0 +1,49 @@
...
+        working-directory: ./OpenSK_base
+        run: ./setup.sh
+      - name: Run bloat on base
+        working-directory: ./OpenSK_base
+        run: RUSTFLAGS="-C link-arg=-icf=all -C force-frame-pointers=no" cargo bloat --release --target=thumbv7em-none-eabi --features=with_ctap1,vendor_hid --crates >> "$GITHUB_WORKSPACE/.github/workflows/bloat_output_old.txt"
+
+      - name: Run output formatter
+        run: ./.github/workflows/bloat_formatter.sh bloat_output_new.txt bloat_output_old.txt bloat_comment.md
+      - name: Run workflow command
+        run: echo "::notice file=.github/workflows/cargo_bloat.yml,title=Binary size::$(cat .github/workflows/bloat_comment.md)"
```
ia0: nit: We actually don't need `bloat_comment.md`. We can echo directly from the script. But fine to keep like that and improve later.

Note also that doing stuff from the script directly permits to do more stuff:
- Change from notice to warning if size increases
- Do not notice if the size is the same.

ia0 (dismissed):

.github/workflows/cargo_bloat.yml
kaczmarczyck: I like the idea, trying it out now. I locally tested the script, I'll have to add an actual change to see something on this PR though. So please ignore the next commit, that is just to trigger the warning.

ia0 (approved):

PR #463, opened 2022-04-20
Move enterprise mode related customizations to new file

hcyang-google:


src/api/customization.rs
```diff
@@ -80,6 +80,48 @@ pub trait Customization {
...
+
+    /// Lists relying party IDs that can perform enterprise attestation.
+    ///
+    /// # Invariant
+    ///
+    /// - If the mode is VendorFacilitated, enterprise_attestation_mode() must be non-empty.
+    ///
+    /// This list is only considered if the enterprise attestation mode is
+    /// VendorFacilitated.
+    fn enterprise_rp_id_list(&self) -> Vec<String>;
```
hcyang-google, line 124: @ia0 unlike the previous one, this one does have overhead because the caller only needs to check whether the list contains a value. Should we do the backing store trick to expose a &[&str] interface? Or ignoring is fine?

kaczmarczyck: Usage might always change, right? So if we don't optimize before, the same argument should still apply?

kaczmarczyck (dismissed):
Let's wait for @ia0 to take a look.

src/api/customization.rs
```diff
@@ -206,6 +285,26 @@ pub fn is_valid(customization: &impl Customization) -> bool {
...
+
+    // enterprise_rp_id_list() should be non-empty in vendor facilitated mode, and empty otherwise.
+    if let Some(EnterpriseAttestationMode::VendorFacilitated) =
+        customization.enterprise_attestation_mode()
+    {
+        if customization.enterprise_rp_id_list().is_empty() {
+            return false;
+        }
+    } else if !customization.enterprise_rp_id_list().is_empty() {
+        return false;
```
ia0, line 312: This whole condition could be:

```rust
if matches!(customization.enterprise_attestation_mode(),
            Some(EnterpriseAttestationMode::VendorFacilitated)) ==
   customization.enterprise_rp_id_list().is_empty()
{
    return false;
}

File: src/api/customization.rs (hunk 1 above)
Comment (ia0) [reply src/api/customization.rs:124]: I think we should expose:

```rust
fn is_enterprise_rp_id(&self, rp_id: &str) -> bool;
```

If we need something else in the future, we will assess at that point. For the other `Vec<String>` we actually needed to iterate/merge its content.

src/api/customization.rs
hcyang-google: Done

src/api/customization.rs
hcyang-google: Done, but PTAL

kaczmarczyck: I understood the comment as "only expose `is_enterprise_rp_id`, remove `enterprise_rp_id_list`", but let's check with @ia0 to clarify.

ia0: Yes indeed, but I realize now that we need to check whether the list is empty in tests. So we still need either `enterprise_rp_id_list` or `is_enterprise_rp_id_list_empty` (or similar).

I guess the simplest is to add `#[cfg(feature = "std")]` to `enterprise_rp_id_list`.

hcyang-google: Yeah that's why I kept the enterprise_rp_id_list API. Now I changed it to only exposed in std.

ia0 (approved):
LGTM. I don't know if @kaczmarczyck wants to take another look, otherwise it's good to merge.

ia0:
Actually something to do is merge/rebase on top of `develop` and check the new binary size workflow comment from @kaczmarczyck . Ideally there's no binary size change, so no comment.

kaczmarczyck (approved):
100 byte extra is within normal variance.

PR #465, opened 2022-04-25
Enterprise attestation testing

kaczmarczyck:
Uses @hcyang-google 's PR #463 to add tests for enterprise attestation. Contributions outside of tests:

- Fixes one bug in MakeCredential.
- More permissions with the customization of platform managed enterprise attestation.
- Only allows activating enterprise attestation when the certificate is set, to prevent later vendor errors.

### DIFF
--- a/src/ctap/storage.rs
+++ b/src/ctap/storage.rs
@@ -577,9 +577,7 @@ pub fn enterprise_attestation(env: &mut impl Env) -> Result<bool, Ctap2StatusCod
 
 /// Marks enterprise attestation as enabled.
 pub fn enable_enterprise_attestation(env: &mut impl Env) -> Result<(), Ctap2StatusCode> {
-    if attestation_private_key(env).unwrap_or(None).is_none()
-        || attestation_certificate(env).unwrap_or(None).is_none()
-    {
+    if attestation_private_key(env)?.is_none() || attestation_certificate(env)?.is_none() {
         return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
     }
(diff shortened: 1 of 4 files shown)


kaczmarczyck:
I checked the coverage report to investigate the drop.

The file [libraries/persistent_store/src/store.rs](https://coveralls.io/builds/48533329/source?filename=libraries%2Fpersistent_store%2Fsrc%2Fstore.rs#L37) is now correctly included in the coverage calculation, instead of falsely counting as 100%.

However, the coverage increased in [src/ctap/mod.rs](https://coveralls.io/builds/48533329/source?filename=src%2Fctap%2Fmod.rs#L2057) by 0.86%, because the enterprise attestation code in MakeCredential is now tested!

src/ctap/storage.rs
```diff
@@ -577,6 +577,11 @@ pub fn enterprise_attestation(env: &mut impl Env) -> Result<bool, Ctap2StatusCod
 
 /// Marks enterprise attestation as enabled.
 pub fn enable_enterprise_attestation(env: &mut impl Env) -> Result<(), Ctap2StatusCode> {
+    if attestation_private_key(env).unwrap_or(None).is_none()
+        || attestation_certificate(env).unwrap_or(None).is_none()
```
ia0: Why unwrap? I don't think we want to hide storage errors.

```rust
if attestation_private_key(env)?.is_none() || attestation_certificate(env)?.is_none() {
```

kaczmarczyck: Ah yes, of course.

ia0 (approved):

PR #466, opened 2022-04-25
Speed up crypto library unit tests

kaczmarczyck:
Local testing (`run_desktop_tests.sh`) is unnecessarily slow, in part because of the slow unit tests in the crypto library. Some unit tests took over a minute to complete.

This PR reduces the number of iterations for these tests. No test should take longer than a second. `--report-time` helps finding slow tests. This PR is a 50x reduction in CPU load for these tests.

Before:
```
$ time cargo test --features=std  -- -Zunstable-options --report-time
...
real	2m10.315s
user	12m6.005s
sys	0m2.419s
```

After:
```
$ time cargo test --features=std  -- -Zunstable-options --report-time
...
real	0m5.148s
user	0m13.719s
sys	0m0.100s
```

### DIFF
--- a/libraries/crypto/src/ec/point.rs
+++ b/libraries/crypto/src/ec/point.rs
@@ -601,7 +601,7 @@ pub mod test {
     fn get_test_values_affine() -> Vec<PointAffine> {
         let mut values = Vec::new();
         for table in 0..2 {
-            for index in 0..15 {
+            for index in (0..15).step_by(3) {
                 values.push(precomputed(table, index));
             }
         }
@@ -1005,7 +1005,10 @@ pub mod test {
     fn test_scalar_base_mul_is_scalar_mul_generator() {
         let gen = precomputed(0, 0);
         // TODO: more scalars
-        for scalar in &super::super::exponent256::test::get_test_values() {
+        for scalar in super::super::exponent256::test::get_test_values()
+            .iter()
+            .step_by(3)
+        {
             assert_eq!(
                 PointProjective::scalar_base_mul(scalar),
                 gen.scalar_mul(scalar)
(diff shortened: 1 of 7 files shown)


kaczmarczyck:
Similar to #465 , the coverage report only changed in that it correctly handles [libraries/persistent_store/src/store.rs](https://coveralls.io/builds/48539836/source?filename=libraries%2Fpersistent_store%2Fsrc%2Fstore.rs#L37) now.

libraries/crypto/src/ec/point.rs
```diff
@@ -1005,7 +1005,10 @@ pub mod test {
     fn test_scalar_base_mul_is_scalar_mul_generator() {
         let gen = precomputed(0, 0);
         // TODO: more scalars
```
ia0, line 1007: Maybe remove those TODOs? Since the PR does the opposite.

libraries/crypto/src/sha256.rs
```diff
@@ -266,8 +266,8 @@ mod test {
         let hash = hex::decode("32beecb58a128af8248504600bd203dcc676adf41045300485655e6b8780a01d")
             .unwrap();
 
-        for i in 0..512 {
-            for j in i..512 {
+        for i in (0..512).step_by(16) {
+            for j in (i..512).step_by(16) {
```
ia0, line 270: Why step 16 here? The idea of the test is to test slices of different sizes. I would use 3 and 5 (or 13 and 17 if we want bigger numbers)

ia0 (commented):
I understand the motivation for this, but I'm not sure this is the right solution at the moment. Here are some thoughts:
- We want tests to be fast. This PR does it.
- We want to thoroughly test the crypto. This PR reverts some of this. It would be fine if we had fuzzing. (Those test targets are essentially fuzzing targets.)

I see 2 options if we want to address desktop tests running time:
- This PR followed by crypto fuzzing in a short time span.
- Remove crypto tests from desktop tests (the tests still run on github workflows), take time to add crypto fuzzing, finally this PR (or even just delete the tests that are subsumed by fuzzing, unit tests should be concrete cases, not loops).

kaczmarczyck:
Closed for #467.

PR #468, opened 2022-04-26
Added own timer.

shilingwangggg:
Create Timer trait for future use. Implement Timer using libtock core.

### DIFF
--- a/src/ctap/main_hid.rs
+++ b/src/ctap/main_hid.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::clock::{ClockInt, CtapInstant};
 #[cfg(feature = "with_ctap1")]
 use crate::ctap::ctap1;
 #[cfg(feature = "with_ctap1")]
@@ -20,18 +19,19 @@ use crate::ctap::hid::ChannelID;
 use crate::ctap::hid::{
     CtapHid, CtapHidCommand, CtapHidError, HidPacket, HidPacketIterator, Message,
 };
-use crate::ctap::{Channel, CtapState, TimedPermission};
+use crate::ctap::{Channel, CtapState};
 use crate::env::Env;
+use crate::timer::{Timer,LibtockAlarmTimer};
 use embedded_time::duration::Milliseconds;
 
 /// Implements the standard CTAP command processing for HID.
 pub struct MainHid {
     hid: CtapHid,
-    wink_permission: TimedPermission,
+    wink_permission: Option<LibtockAlarmTimer>,
 }
 
 impl MainHid {
-    const WINK_TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(5000 as ClockInt);
+    const WINK_TIMEOUT_DURATION: u32 = 5000_u32;
 
     /// Instantiates a HID handler for CTAP1, CTAP2 and Wink.
     pub fn new() -> Self {
@@ -42,7 +42,7 @@ impl MainHid {
             CtapHid::CAPABILITY_WINK | CtapHid::CAPABILITY_CBOR | CtapHid::CAPABILITY_NMSG;
 
         let hid = CtapHid::new(capabilities);
-        let wink_permission = TimedPermission::waiting();
+        let wink_permission = None;
         MainHid {
             hid,
             wink_permission,
@@ -54,11 +54,10 @@ impl MainHid {
         &mut self,
         env: &mut impl Env,
         packet: &HidPacket,
-        now: CtapInstant,
         ctap_state: &mut CtapState,
     ) -> HidPacketIterator {
-        if let Some(message) = self.hid.parse_packet(env, packet, now) {
-            let processed_message = self.process_message(env, message, now, ctap_state);
+        if let Some(message) = self.hid.parse_packet(env, packet) {
+            let processed_message = self.process_message(env, message, ctap_state);
             debug_ctap!(env, "Sending message: {:02x?}", processed_message);
             CtapHid::split_message(processed_message)
         } else {
@@ -71,11 +70,10 @@ impl MainHid {
         &mut self,
         env: &mut impl Env,
         message: Message,
-        now: CtapInstant,
         ctap_state: &mut CtapState,
     ) -> Message {
         // If another command arrives, stop winking to prevent accidential button touches.
-        self.wink_permission = TimedPermission::waiting();
+        self.wink_permission = None;
 
         let cid = message.cid;
         match message.cmd {
@@ -86,7 +84,7 @@ impl MainHid {
                 return CtapHid::error_message(cid, CtapHidError::InvalidCmd);
 
                 #[cfg(feature = "with_ctap1")]
-                match ctap1::Ctap1Command::process_command(env, &message.payload, ctap_state, now) {
+                match ctap1::Ctap1Command::process_command(env, &message.payload, ctap_state) {
                     Ok(payload) => MainHid::ctap1_success_message(cid, &payload),
                     Err(ctap1_status_code) => MainHid::ctap1_error_message(cid, ctap1_status_code),
                 }
@@ -97,7 +95,7 @@ impl MainHid {
                 // don't handle any other packet in the meantime.
                 // TODO: Send "Processing" type keep-alive packets in the meantime.
                 let response =
-                    ctap_state.process_command(env, &message.payload, Channel::MainHid(cid), now);
+                    ctap_state.process_command(env, &message.payload, Channel::MainHid(cid));
                 Message {
                     cid,
                     cmd: CtapHidCommand::Cbor,
@@ -107,8 +105,7 @@ impl MainHid {
             // CTAP 2.1 from 2021-06-15, section 11.2.9.2.1.
             CtapHidCommand::Wink => {
                 if message.payload.is_empty() {
-                    self.wink_permission =
-                        TimedPermission::granted(now, Self::WINK_TIMEOUT_DURATION);
+                    self.wink_permission = Some(LibtockAlarmTimer::start(Self::WINK_TIMEOUT_DURATION));
                     // The response is empty like the request.
                     message
                 } else {
@@ -121,13 +118,13 @@ impl MainHid {
     }
 
     /// Returns whether a wink permission is currently granted.
-    pub fn should_wink(&self, now: CtapInstant) -> bool {
-        self.wink_permission.is_granted(now)
+    pub fn should_wink(&self) -> bool {
+        self.wink_permission.is_some() && self.wink_permission.unwrap().has_elapsed().is_some()
     }
 
     /// Updates the timeout for the wink permission.
-    pub fn update_wink_timeout(&mut self, now: CtapInstant) {
-        self.wink_permission = self.wink_permission.check_expiration(now);
+    pub fn update_wink_timeout(&mut self) {
+        self.wink_permission = Some(LibtockAlarmTimer::start(Self::WINK_TIMEOUT_DURATION));
     }
 
     #[cfg(feature = "with_ctap1")]
@@ -161,7 +158,7 @@ mod test {
 
     fn new_initialized() -> (MainHid, ChannelID) {
         let (hid, cid) = CtapHid::new_initialized();
-        let wink_permission = TimedPermission::waiting();
+        let wink_permission = None;
         (
             MainHid {
                 hid,
@@ -174,7 +171,7 @@ mod test {
     #[test]
     fn test_process_hid_packet() {
         let mut env = TestEnv::new();
-        let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0));
+        let mut ctap_state = CtapState::new(&mut env);
         let (mut main_hid, cid) = new_initialized();
 
         let mut ping_packet = [0x00; 64];
@@ -184,7 +181,6 @@ mod test {
         let mut response = main_hid.process_hid_packet(
             &mut env,
             &ping_packet,
-            CtapInstant::new(0),
             &mut ctap_state,
         );
         assert_eq!(response.next(), Some(ping_packet));
@@ -194,7 +190,7 @@ mod test {
     #[test]
     fn test_process_hid_packet_empty() {
         let mut env = TestEnv::new();
-        let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0));
+        let mut ctap_state = CtapState::new(&mut env);
         let (mut main_hid, cid) = new_initialized();
 
         let mut cancel_packet = [0x00; 64];
@@ -204,7 +200,6 @@ mod test {
         let mut response = main_hid.process_hid_packet(
             &mut env,
             &cancel_packet,
-            CtapInstant::new(0),
             &mut ctap_state,
         );
         assert_eq!(response.next(), None);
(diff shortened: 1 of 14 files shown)


src/timer.rs
```diff
@@ -0,0 +1,72 @@
+trait Timer {
+    fn start(milliseconds: u32) -> Self;
+    fn has_elapsed(self) -> Option<Self>
+    where
+        Self: Sized;
```
ia0: Let's move this to the trait:

```rust
trait Timer: Sized {
    fn start(milliseconds: u32) -> Self;
    fn has_elapsed(self) -> Option<Self>;
}
```

src/timer.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub const GET_CLOCK_FREQUENCY: usize = 1;
+    pub const GET_CLOCK_VALUE: usize = 2;
+    pub const _STOP_ALARM: usize = 3;
+    pub const _SET_ALARM: usize = 4;
+}
+const DRIVER_NUMBER: usize = 0x00000;
+
+lazy_static! {
+    static ref CLOCK_FREQUENCY: usize =
+        syscalls::command(DRIVER_NUMBER, command_nr::GET_CLOCK_FREQUENCY, 0, 0).unwrap_or(0);
```
ia0: I don't think we need this to be constant. We can do the syscall when creating a timer. That's already the case in the current implementation.

src/timer.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+struct LibtockAlarmTimer {
+    start_tick: usize,
+    milliseconds: u32,
```
ia0: This should store the final tick.

```rust
struct LibtockAlarmTimer {
    expiration_tick: u32,  // or end_tick
}
```

src/timer.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+fn wrapping_sub_u24(lhs: usize, rhs: usize) -> usize {
+    lhs.wrapping_sub(rhs) & 0xffffff
+}
+
+impl Timer for LibtockAlarmTimer {
+    fn start(milliseconds: u32) -> Self {
+        let start_tick =
+            syscalls::command(DRIVER_NUMBER, command_nr::GET_CLOCK_VALUE, 0, 0).unwrap_or(0);
```
ia0: Here, we can also do the frequency syscall and compute the expiration/end tick. We should add a comment that we assume the clock is 24-bits.

```rust
let delta_tick = frequency * milliseconds / 1000;  // or something against overflow (see below)
assert!(delta_tick < 0x800000);
let end_tick = wrapping_add_u24(start_tick, delta_tick);
```

To compute `delta_tick` without overflow we could do things like:

```rust
// Assume frequency multiple of 100 and milliseconds multiple of 10.
let delta_tick = (frequency / 100) * (milliseconds / 10);
// Assume frequency multiple of 2. We already assert the result to be on 23 bits, so multiplying the result by 500 should fit 32 bits, so the unwrap is already rejecting some of the cases where the delta is too big.
let delta_tick = (frequency / 2).checked_mul(milliseconds).unwrap() / 500;
```

src/timer.rs
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    fn has_elapsed(self) -> Option<Self> {
+        let cur_tick =
+            syscalls::command(DRIVER_NUMBER, command_nr::GET_CLOCK_VALUE, 0, 0).unwrap_or(0);
+
+        // TODO: handle 24 bits magic
+        let delta_tick = wrapping_sub_u24(cur_tick, self.start_tick);
+        if (delta_tick * 1000) / *CLOCK_FREQUENCY > self.milliseconds as usize {
```
ia0: This would just become `wrapping_sub_u24(cur_tick, end_tick) < 0x800000`.

src/timer.rs
kaczmarczyck: We put some thought into how to not overflow in this calculation [here](https://github.com/google/OpenSK/blob/f2496a8e6d71a4e838884996a1c9b62121f87df2/third_party/libtock-drivers/src/timer.rs#L191). Might be cleaner?

src/timer.rs
kaczmarczyck: Interesting, but isn't asking just once better (in terms of optimization)? We have to assume that the frequency doesn't change over time anyway, right?

src/timer.rs
ia0: I thought I wrote about the `u64` computation but probably forgot. Yes that would be an option also (except that `u64` divisions might be very costly on embedded, to be checked).

Note however that the alternative in comments still overflows.

src/timer.rs
ia0: Yes, I don't think we should plan for cases where the frequency would change over time. The tradeoff is between one syscall per timer creation (very small) and one extra dependency plus the associated code complexity of understanding when that "static syscall" is actually run. I don't think the "optimization" is worth it.

In particular, once this get in `Env`, the frequency would be stored in the environment, so no need for `lazy_static` and the optimization comes for free.

kaczmarczyck: Ah right, if we agree on that endgame, I'm happy with that!

src/timer.rs
shilingwangggg: Done with attempt to avoid division.

src/timer.rs
shilingwangggg: Done.

src/timer.rs
shilingwangggg: Done with attempt to avoid division. PTAL.

src/timer.rs
shilingwangggg: Done.

src/timer.rs
shilingwangggg: Removed lazy_static.

src/timer.rs
shilingwangggg: Done.

src/timer.rs
```diff
@@ -0,0 +1,70 @@
+trait Timer: Sized {
+    fn start(milliseconds: u32) -> Self;
+    fn has_elapsed(self) -> Option<Self>;
```
ia0, line 21: Let's add some documentation for those 2 functions.

src/timer.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+use libtock_core::syscalls;
```
ia0: nit: This should be first in the file. (Given that this file is going to be split into `api::timer` and `env::tock::timer` it's not really an issue.)

src/timer.rs
```diff
@@ -0,0 +1,70 @@
+trait Timer: Sized {
```
ia0: The file should contain the copyright and license notice.

src/timer.rs
```diff
@@ -0,0 +1,70 @@
...
+    lhs.wrapping_add(rhs) & 0xffffff
+}
+fn wrapping_sub_u24(lhs: usize, rhs: usize) -> usize {
+    lhs.wrapping_sub(rhs) & 0xffffff
+}
+
+impl Timer for LibtockAlarmTimer {
+    fn start(milliseconds: u32) -> Self {
+        let clock_frequency =
+            syscalls::command(DRIVER_NUMBER, command_nr::GET_CLOCK_FREQUENCY, 0, 0).unwrap_or(0);
```
ia0: Mmmm... I'm not sure we should `unwrap_or(0)`. We should either `unwrap()` (or `flex_unwrap()` in that case), or propagate the error. Ideally we should propagate the error, but given that we will probably just unwrap it at the call site, it's probably alright to unwrap here (and improve later if needed).

Same below.

src/timer.rs
```diff
@@ -0,0 +1,70 @@
...
+        let start_tick =
+            syscalls::command(DRIVER_NUMBER, command_nr::GET_CLOCK_VALUE, 0, 0).unwrap_or(0);
+        // 32 bits inverted divisor for 1/1000 ( ceil((2^32) / 1000) ), so that (x * INV_DIV) >> 32 ≈ x / 1000
+        const INV_DIV: u64 = 4294968;
+        let delta_tick = ((clock_frequency as u64 * milliseconds as u64 * INV_DIV) >> 32) as usize;
```
ia0: There's 2 issues here:
- The second multiplication can overflow. And it may overflow in cases where the division would have worked.
- The `INV_DIV` doesn't always work, e.g. for 6100999 it gives one more tick than the division. Such imprecision is probably acceptable given that time is approximative in OpenSK anyway.

I think assuming the frequency is a multiple of 2 and the ticks are on 24 bits is the simplest solution (and this assumption is actually true for all our boards).

shilingwangggg: Indeed there is a slight loss of precision, but I believe it is very limited.

If the second multiplication would overflow, then the next line `assert!(delta_tick < 0x800000);` would as well, as we are checking that the result is within 23 bits, and therefore the result of both multiplications should be within 53 bits (23+32). An easy way to prevent the overflow would be to assert the range after the first multiplication. WDYT?

ia0: Yes, one solution would be to assert the range after the first multiplication. But then I think this solution is really the simplest:

```rust
// - Assumes frequency is a multiple of 2 (otherwise under-approximates by 1 tick per second).
// - Only unwraps if frequency * milliseconds / 1000 doesn't fit in 23 bits.
let delta_tick = (frequency / 2).checked_mul(milliseconds).unwrap() / 500;
assert!(delta_tick < 0x800000);
```

src/timer.rs
shilingwangggg: Done. I copied it from the main.rs, the year is 2019, I hope it is fine.

src/timer.rs
shilingwangggg: Done.

src/timer.rs
shilingwangggg: Done.

src/timer.rs
shilingwangggg: Done.

src/timer.rs
shilingwangggg: I cannot use flex_unwrap because syscalls::command uses CommandError rather than TockError. I did the same thing as what is done in flex_unwrap, which is ok().unwrap()

src/timer.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright 2019 Google LLC
```
ia0, line 1: You can change the year to 2022. This is all new code.

src/timer.rs
ia0: Documentation in Rust use three slashes:

```rust
/// Creates and starts a timer.
fn start(milliseconds: u32) -> Self;

/// Returns the remaining timer of a timer.
///
/// In particular, if the timer has elapsed, `None` is returned.
fn has_elapsed(self) -> Option<Self>;
```

And actually I would rename `has_elapsed` to `is_running`, because `Some` usually means `true`.

ia0 (commented):
Looks good. I guess we should try using it in this PR directly. Otherwise we don't know if it works.

src/ctap/main_hid.rs
```diff
@@ -115,13 +112,13 @@ impl MainHid {
     }
 
     /// Returns whether a wink permission is currently granted.
-    pub fn should_wink(&self, now: CtapInstant) -> bool {
-        self.wink_permission.is_granted(now)
+    pub fn should_wink(&self) -> bool {
+        self.wink_permission.is_some() && self.wink_permission.unwrap().has_elapsed().is_some()
```
ia0, line 122: Interesting. First I'm surprised this compiles. Second, the signature of `should_wink` might be problematic with the `has_elapsed` API. I see at least 2 options:
1. `should_wink` only returns the status since last update, i.e. `self.wink_permission.is_some()` (my preference)
2. `should_wink` is `&mut self` and also updates the timer.

src/ctap/main_hid.rs
```diff
@@ -115,13 +112,13 @@ impl MainHid {
...
     }
 
     /// Updates the timeout for the wink permission.
-    pub fn update_wink_timeout(&mut self, now: CtapInstant) {
-        self.wink_permission = self.wink_permission.check_expiration(now);
+    pub fn update_wink_timeout(&mut self) {
+        self.wink_permission = Some(LibtockAlarmTimer::start(Self::WINK_TIMEOUT_DURATION));
```
ia0, line 127: This looks wrong to me. I would expect this to update the timer, not reset it.

```rust
self.wink_permission = self.wink_permission.and_then(LibtockAlarmTimer::has_expired);
```

src/ctap/mod.rs
```diff
@@ -280,17 +277,17 @@ impl StatefulPermission {
     ///
     /// Resets are only possible after a power cycle. Therefore, initialization
     /// means allowing Reset, and Reset cannot be granted later.
-    pub fn new_reset(now: CtapInstant) -> StatefulPermission {
+    pub fn new_reset() -> StatefulPermission {
         StatefulPermission {
-            permission: TimedPermission::granted(now, RESET_TIMEOUT_DURATION),
+            timer: Timer::start(30000),
```
ia0: Why not use the constant?

src/ctap/mod.rs
```diff
@@ -315,8 +312,8 @@ impl StatefulPermission {
     }
 
     /// Clears all state if the permission timed out.
-    pub fn clear_timer(&mut self, now: CtapInstant) {
-        if !self.permission.is_granted(now) {
+    pub fn clear_timer(&mut self) {
+        if self.timer.is_some() && self.timer.unwrap().has_elapsed() {
```
ia0, line 445: We should first update the timer, then check if elapsed.

```rust
self.timer = self.timer.and_then(has_expired);
if self.timer.is_none() { self.clear(); }
```

src/ctap/token_state.rs
```diff
@@ -14,20 +14,18 @@
...
 
-use crate::clock::{ClockInt, CtapInstant};
-
 /// Timeout for auth tokens.
 ///
 /// This usage time limit is correct for USB, BLE, and internal.
 /// NFC only allows 19.8 seconds.
 /// TODO(#15) multiplex over transports, add NFC
-const INITIAL_USAGE_TIME_LIMIT: Milliseconds<ClockInt> = Milliseconds(30000 as ClockInt);
+const INITIAL_USAGE_TIME_LIMIT:u32 = 30000_u32;
```
ia0, line 29: What's the unit?

src/ctap/token_state.rs
```diff
@@ -113,19 +111,18 @@ impl PinUvAuthTokenState {
...
 
     /// Updates the usage timer, and disables the pinUvAuthToken on timeout.
-    pub fn pin_uv_auth_token_usage_timer_observer(&mut self, now: CtapInstant) {
+    pub fn pin_uv_auth_token_usage_timer_observer(&mut self) {
         if !self.in_use {
             return;
         }
-        self.usage_timer = self.usage_timer.check_expiration(now);
-        if !self.usage_timer.is_granted(now) {
+        if self.usage_timer.is_none() || self.usage_timer.unwrap().has_elapsed().is_none() {
```
ia0: This should also update the timer. We could actually provide a function for convenience:

```rust
/// Updates the expiration of a timer.
pub fn update_timer<T: Timer>(timer: &mut Option<T>) {
    *timer = core::mem::replace(timer, None).and_then(T::has_expired);
}
```

kaczmarczyck:
Picked up by #596

PR #470, opened 2022-04-27
Separate RNG library

kaczmarczyck:
This PR moves the RNG implementation of our `crypto` library out. In the future, if `crypto` wants to import any crate that depends on our RNG, we don't get a ciruclar dependency.
No changes in functionality.

### DIFF
--- a/run_desktop_tests.sh
+++ b/run_desktop_tests.sh
@@ -23,6 +23,9 @@ cd ../..
 cd libraries/crypto
 cargo fmt --all -- --check
 cd ../..
+cd libraries/rng256
+cargo fmt --all -- --check
+cd ../..
 cd libraries/persistent_store
 cargo fmt --all -- --check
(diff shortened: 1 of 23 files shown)


run_desktop_tests.sh
```diff
@@ -23,6 +23,9 @@ cd ../..
 cd libraries/crypto
 cargo fmt --all -- --check
 cd ../..
+cd libraries/rng256
+cargo fmt --all -- --check
+cd ../..
```
ia0, line 28: What about the github workflows?

ia0 (dismissed):

run_desktop_tests.sh
kaczmarczyck: I added one that is almost a copy of the crypto workflow!

ia0 (approved):

PR #471, opened 2022-04-28
Add configure-only mode to deploy.py script

liamjm:
This flag allows the deploy.py script to just run the phase of configuring the device. This will benefit the work on Vendor HID, which can cause troubles with configuring.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -760,7 +760,7 @@ class OpenSKInstaller:
     self.update_rustc_if_needed()
 
     if not (self.args.tockos or self.args.application or
-            self.args.clear_storage):
+            self.args.clear_storage or self.args.configure):
       info("Nothing to do.")
       return 0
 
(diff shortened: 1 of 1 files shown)


deploy.py
```diff
@@ -1057,6 +1060,13 @@ def main(args):
       dest="features",
       help=("Compiles the OpenSK application with support for nfc."),
   )
+  main_parser.add_argument(
+      "--configure",
+      action="store_true",
+      default=False,
```
kaczmarczyck: If this default is `False` and the target is opensk, can you please print the warning from above?
```
      info("Your device is not yet configured, and lacks some functionality. "
           "If you run into issues, this command might help:\n\n"
           "./tools/configure.py \\\n"
           "    --certificate=crypto_data/opensk_cert.pem \\\n"
           "    --private-key=crypto_data/opensk.key\n\n"
           "Please read the Certificate considerations in docs/customization.md"
           " to understand the privacy trade-off.")
```

Also we'd have to update the `docs/` to add --configure to the deploy command, or explain that you might want to configure.

liamjm: Actually, let's keep configure defaulted to true to keep existing behavior.

Defaulting to False was a mistake.

kaczmarczyck (approved):

PR #472, opened 2022-04-28
Create a 2nd USB interface for the Vendor HID

liamjm:
This new interface is enumerated by the host, but the device transmits
all packets on the existing interface, so the device won't function
correct in this mode until this this fixed.

The changes are in tock, and so appear as a patch here. These are:
* supporting multiple HIDs in the USB configuration
* adding the HID descriptors for the new inteface
* supporting the vendor_hid feature in relevant Cargo.toml files.
NOTE: only boards/nordic/nrf52840dk_opensk has been updated.

As the changes are in tockos, deploy script needed to pass vendor_hid
feature to tockos build stage.

Demo of output:
lsusb -v -d 1915:521f | grep "NumInterfaces"
    bNumInterfaces          2

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -371,6 +371,8 @@ class OpenSKInstaller:
     env = os.environ.copy()
     if self.args.verbose_build:
       env["V"] = "1"
+    if "vendor_hid" in self.args.features:
+      env["CARGO_FLAGS"] = "--features=vendor_hid"
     self.checked_command(["make"], cwd=props.path, env=env)
 
   def build_example(self):
@@ -864,6 +866,11 @@ class OpenSKInstaller:
         fatal("Unexpected arguments to configure your device. Since you "
               "selected the programmer \"none\", the device is not ready to be "
               "configured yet.")
+      return 0
+
+    if "vendor_hid" in self.args.features:
+      # vendor_hid as a work in progress and is not compatible with configure
+      # mode.
       return 0
 
     # Perform checks if OpenSK was flashed.
--- a/patches/tock/09-add-vendor-hid-usb-interface.patch
+++ b/patches/tock/09-add-vendor-hid-usb-interface.patch
@@ -21,7 +21,7 @@ index 680fe32b2..4f757b93d 100644
 +[features]
 +vendor_hid = []
 diff --git a/capsules/src/usb/descriptors.rs b/capsules/src/usb/descriptors.rs
-index dea2dfed9..8aac509c8 100644
+index dea2dfed9..5dc6b81a6 100644
 --- a/capsules/src/usb/descriptors.rs
 +++ b/capsules/src/usb/descriptors.rs
 @@ -420,7 +420,7 @@ pub fn create_descriptor_buffers(
@@ -45,7 +45,7 @@ index dea2dfed9..8aac509c8 100644
              + cdc_descriptor.map_or(0, |ds| ds.iter().map(|d| d.size()).sum::<usize>());
  
      // Set the number of endpoints for each interface descriptor.
-@@ -521,13 +524,13 @@ pub fn create_descriptor_buffers(
+@@ -521,12 +524,12 @@ pub fn create_descriptor_buffers(
          // Add the interface descriptor.
          len += d.write_to(&other_buf.buf[len..]);
  
@@ -57,16 +57,15 @@ index dea2dfed9..8aac509c8 100644
 -                len += dh.write_to(&other_buf.buf[len..]);
 +        // HID descriptor, if any.
 +        if let Some(dh) = hid_descriptor {
-+          if let Some(hids) = dh.get(0) {
-+            if let Some(d) = hids.get(i) {
-+              len += d.write_to(&other_buf.buf[len..]);
++            if let Some(hids) = dh.get(0) {
++                if let Some(d) = hids.get(i) {
++                    len += d.write_to(&other_buf.buf[len..]);
++                }
              }
-+          }
          }
  
-         // If there is a CDC descriptor array, we include
 diff --git a/capsules/src/usb/usbc_client_ctrl.rs b/capsules/src/usb/usbc_client_ctrl.rs
-index f7899d8c5..d66e88f08 100644
+index f7899d8c5..6956523c6 100644
 --- a/capsules/src/usb/usbc_client_ctrl.rs
 +++ b/capsules/src/usb/usbc_client_ctrl.rs
 @@ -38,6 +38,12 @@ const DESCRIPTOR_BUFLEN: usize = 128;
@@ -97,7 +96,7 @@ index f7899d8c5..d66e88f08 100644
  
      /// Supported language (only one for now).
      language: &'b [u16; 1],
-@@ -104,13 +110,13 @@ impl<'a, 'b, U: hil::usb::UsbController<'a>> ClientCtrl<'a, 'b, U> {
+@@ -104,8 +110,8 @@ impl<'a, 'b, U: hil::usb::UsbController<'a>> ClientCtrl<'a, 'b, U> {
          controller: &'a U,
          device_descriptor_buffer: DeviceBuffer,
          other_descriptor_buffer: DescriptorBuffer,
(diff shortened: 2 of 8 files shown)


deploy.py
```diff
@@ -866,10 +868,18 @@ def run(self) -> int:
...
 
-    # Perform checks if OpenSK was flashed.
-    if self.args.application != "ctap2":
+    if "vendor_hid" in self.args.features:
+      # vendor_hid as a work in progress and is not compatible with configure
+      # mode.
       return 0
 
+    # Perform checks if OpenSK was flashed.
+    if self.args.application == "ctap2" or self.args.configure:
```
kaczmarczyck: Do we still need the configure flag when configuration is guarded by this?
`if "vendor_hid" in self.args.features:`

I think giving users the choice is not bad, but its default should then be `true`. If we introduce it, we should mention it in `docs/customization.md` where users read about the privacy trade-offs of configuration.

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
+diff --git a/boards/nordic/nrf52840dk_opensk/Cargo.toml b/boards/nordic/nrf52840dk_opensk/Cargo.toml
```
kaczmarczyck: The `boards` don't need to be part of the patch. We copy them into Tock during setup, and you can apply and commit your changes to the top level `boards` directory.

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
...
+         // Add the interface descriptor.
+         len += d.write_to(&other_buf.buf[len..]);
+ 
+-        // If there is a HID descriptor, we include
+-        // it with the first interface descriptor.
+-        if i == 0 {
+-            // HID descriptor, if any.
+-            if let Some(dh) = hid_descriptor {
+-                len += dh.write_to(&other_buf.buf[len..]);
++        // HID descriptor, if any.
```
kaczmarczyck: Indentation here is wrong. We use the Rust standard (4 spaces), not the Google standard (2 spaces).

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
...
+-        hid_descriptor: Option<&'b HIDDescriptor<'b>>,
+-        report_descriptor: Option<&'b ReportDescriptor<'b>>,
++        hid_descriptor: Option<[&'b HIDDescriptor<'b>; N_HID_INTERFACES]>,
++        report_descriptor: Option<[&'b ReportDescriptor<'b>; N_HID_INTERFACES]>,
+         language: &'b [u16; 1],
+         strings: &'b [&'b str],
+     ) -> Self {
+         ClientCtrl {
+-            controller: controller,
++          controller: controller,
```
kaczmarczyck: Same indentation comment.

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
...
++};
++
+ pub struct ClientCtapHID<'a, 'b, C: 'a> {
+     client_ctrl: ClientCtrl<'a, 'static, C>,
+ 
+@@ -82,6 +120,10 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtapHID<'a, 'b, C> {
+         product_id: u16,
+         strings: &'static [&'static str],
+     ) -> Self {
++        if cfg!(feature = "vendor_hid") {
```
kaczmarczyck: Why use the other cfg style here, not 
`#[cfg(feature = "vendor_hid")]`
?

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
...
+@@ -130,17 +204,28 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtapHID<'a, 'b, C> {
+                 },
+                 interfaces,
+                 endpoints,
+-                Some(&HID),
++                Some(&[&[
++                    &HID,
++                    #[cfg(feature = "vendor_hid")]
++                    &HID,
++                ]]),
```
ia0: I don't understand why we have 2 layers of descriptors, i.e. `&[&[&X, &X]]` instead of `&[&X, &X]`? In particular you always only look at the first element of the outer layer, which suggests the outer layer is a singleton and could be omitted. Or do you foresee usage in the future? Also, it's probably worth documenting the meaning of those layers. I'll add a comment at the related locations with the marker LAYERS.

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
...
+ 
+-        // If there is a HID descriptor, we include
+-        // it with the first interface descriptor.
+-        if i == 0 {
+-            // HID descriptor, if any.
+-            if let Some(dh) = hid_descriptor {
+-                len += dh.write_to(&other_buf.buf[len..]);
++        // HID descriptor, if any.
++        if let Some(dh) = hid_descriptor {
++          if let Some(hids) = dh.get(0) {
```
ia0: LAYERS: only first element

patches/tock/09-add-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,331 @@
...
+diff --git a/capsules/src/usb/descriptors.rs b/capsules/src/usb/descriptors.rs
+index dea2dfed9..8aac509c8 100644
+--- a/capsules/src/usb/descriptors.rs
++++ b/capsules/src/usb/descriptors.rs
+@@ -420,7 +420,7 @@ pub fn create_descriptor_buffers(
+     mut configuration_descriptor: ConfigurationDescriptor,
+     interface_descriptor: &mut [InterfaceDescriptor],
+     endpoint_descriptors: &[&[EndpointDescriptor]],
+-    hid_descriptor: Option<&HIDDescriptor>,
++    hid_descriptor: Option<&[&[&HIDDescriptor<'static>]]>,
```
ia0: LAYERS: Missing documentation. What does the outer layer mean? (The inner layer is interface.)

deploy.py
liamjm: This PR was stacked on top of #471.
I've separated these, as this check means these features are unrelated.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: Thanks.
Done.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: Fixed.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: Most likely because the else branch was printing something else during development. That is not needed now, so changed to other style.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: Fixed.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: This approach was used to match the approach used for the endpoints:
https://github.com/tock/tock/blob/master/capsules/src/usb/descriptors.rs#L423

I'm happy to take on other  suggestions.

liamjm: :facepalm: 
The endpoints are nested because there are multiple endpoints per interface.
For the descriptors, only one is needed per interfaces, so the extra layer can be removed.


I'll fix that up and update the PR.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: Out of date. Please ignore.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: Out of date, please ignore.

patches/tock/09-add-vendor-hid-usb-interface.patch
liamjm: OK, new commits have removed this. PTAL.

ia0 (approved):

deploy.py
```diff
@@ -866,6 +868,11 @@ def run(self) -> int:
               "configured yet.")
       return 0
 
+    if "vendor_hid" in self.args.features:
+      # vendor_hid as a work in progress and is not compatible with configure
```
kaczmarczyck, line 872: nit: as -> is
but feel free to ignore and we'll fix later

kaczmarczyck (approved):

deploy.py
liamjm: ACK.

PR #474, opened 2022-04-28
Add test_helpers

hcyang-google:
Add set_enterprise_attestation method and used it for unittests and fuzzer.


Cargo.toml
```diff
@@ -27,6 +27,7 @@ arbitrary = { version = "0.4.7", features = ["derive"], optional = true }
 rand = { version = "0.8.4", optional = true }
 
 [features]
+# default = ["std"]
```
ia0: Left over?

src/ctap/mod.rs
```diff
@@ -2067,12 +2067,7 @@ mod test {
         env.customization_mut().enterprise_rp_id_list = vec!["example.com".to_string()];
         let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0));
 
-        let mut key_bytes = [0; 32];
-        let private_key = crypto::ecdsa::SecKey::gensk(env.rng());
-        private_key.to_bytes(array_mut_ref!(key_bytes, 0, 32));
-        storage::set_attestation_certificate(&mut env, &[0xCC]).unwrap();
-        storage::set_attestation_private_key(&mut env, &key_bytes).unwrap();
-        storage::enable_enterprise_attestation(&mut env).unwrap();
+        env.set_enterprise_attestation();
```
ia0: I would argue that this should be done before calling `CtapState::new`. That would be a change with current code (so not related to this PR), but I think it makes more sense for tests with non-empty storage to not let `CtapState::new` initialize the storage. Since this cannot happen in practice that `CtapState` is created and the storage is modified under its feet.

At least something like that would make sense:

```rust
let _ = CtapState::new(&mut env, ...);  // initial boot of device
env.foo();  // simulate some modification without calling the actual functions
let mut ctap_state = CtapState::new(&mut env, ...);  // boot device for test
```

But ideally we want (in particular for fuzzing sequences):

```rust
let mut ctap = CtapState::new(&mut env, ...);  // first boot
ctap.foo(); // some commands (including configure)
let mut ctap = CtapState::new(&mut env, ...);  // second boot
ctap.bar(); // some other commands
```

Modifying directly the environment without going through the ctap library, should only be done for things that may be modified outside the library, e.g. customization (before any calls to the library), rng (the library can't guess the future), user presence (the library can't guess the user behavior), etc. The storage is not supposed to be modified outside the library.

src/ctap/mod.rs
```diff
@@ -2164,12 +2154,7 @@ mod test {
             Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)
         );
 
-        let mut key_bytes = [0; 32];
-        let private_key = crypto::ecdsa::SecKey::gensk(env.rng());
-        private_key.to_bytes(array_mut_ref!(key_bytes, 0, 32));
-        storage::set_attestation_certificate(&mut env, &[0xCC]).unwrap();
-        storage::set_attestation_private_key(&mut env, &key_bytes).unwrap();
-        storage::enable_enterprise_attestation(&mut env).unwrap();
+        env.set_enterprise_attestation();
```
ia0: Theoretically, we should recreate the ctap state after this line:

```rust
let ctap_state = CtapState::new(&mut env, CtapInstant::new(1));
```

src/ctap/mod.rs
```diff
@@ -28,7 +28,7 @@ pub mod main_hid;
 mod pin_protocol;
 pub mod response;
 pub mod status_code;
-mod storage;
+pub mod storage;
```
ia0: I think this should only be `pub` for tests.

```rust
#[cfg(feature = "std")]
pub mod storage;
#[cfg(not(feature = "std"))]
mod storage;
```

src/env/test/mod.rs
```diff
@@ -98,6 +98,16 @@ impl TestEnv {
     pub fn rng(&mut self) -> &mut TestRng256 {
         &mut self.rng
     }
+
+    pub fn set_enterprise_attestation(&mut self) {
```
kaczmarczyck: I wonder if this is the right place to add this function. It only affects the storage, not customization, so it's not really something you do to env, but that uses env to affect storage.

We could end up with a lot of these helper functions that take an `&mut Env` for test purposes and do some setup with it. Maybe a helper mod might be appropriate, in anticipation?

kaczmarczyck: For a `set_*` function, I'd expect an argument that is written into some property. This is more of an "enable" I think?

Cargo.toml
kaczmarczyck: Probably unintentional.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -146,6 +147,22 @@ pub fn process_ctap_any_type(data: &[u8]) -> arbitrary::Result<()> {
     Ok(())
 }
 
+fn setup_storage(unstructured: &mut Unstructured, env: &mut TestEnv) -> arbitrary::Result<()> {
+    if bool::arbitrary(unstructured)? {
+        env.set_enterprise_attestation();
```
kaczmarczyck: This is closer to how I perceive this change. A storage setup inside a helper library. My enterprise attestation tests are basically integration tests more than unit tests, and they almost call externally available APIs. (They call `process_make_credential` directly, but that's easy to move up one more indirection to something public.) So maybe `set_enterprise_attestation` could move to `tests/helper` or something?

kaczmarczyck (commented):
I would like to get @ia0 's opinion on this. I agree with this change in general, just want to properly set the stage for more to come.

ia0 (commented):
How hard would it be to actually configure the ctap library through its API for enterprise attestation?

I agree though that an escape hatch to modify the storage bypassing the library is useful for testing backward compatibility (e.g. running the new version of the library with an old storage).

src/env/test/mod.rs
ia0: Yes, I had a similar concern that `env` should probably not be used for what can be done through the library API.

I don't think a helper mod is needed for tests, they already take `TestEnv`. But it could be used to organize testing code, for example `env::test::storage::enable_enterprise_attestation(&mut TestEnv)` instead of `env::test::TestEnv::enable_enterprise_attestation()`. I don't have strong opinion about that. For now I think we don't need it, but we might indeed need it in the future if we end up with more such functions. Ideally I would hope we actually don't need this function and can directly call into the library API.

src/env/test/mod.rs
```diff
@@ -98,6 +98,16 @@ impl TestEnv {
     pub fn rng(&mut self) -> &mut TestRng256 {
```
ia0: Oops, I just realized this function should be deleted. It's a duplicate of the one in `Env`. I missed that in a prior review I guess.

ia0: And also rename `seed_rng_from_u64` into `seed_from_u64` as suggested here: https://github.com/google/OpenSK/pull/461/files#r853920289.

src/ctap/mod.rs
kaczmarczyck: That would not be necessary if we go through the ctap API to change the storage.

src/ctap/mod.rs
kaczmarczyck: I think not changing `storage` under `CtapState`'s feet makes sense. The tests right now are a bit more concise than what you'd need to do ctap API only. Since we are building a helper function for that, it wouldn't be a readability advantage anymore, of course.

What we'd need to do is:
1. Prepare a ECDH handshake.
2. Use that to setup a PIN.
3. Use the PIN to get an AuthToken.
4. Call the custom `VendorConfig` command to set a certificate and attestation key.
5. Use the AuthToken to send an `AuthenticatorConfig` command.

Sounds like a lot, but steps 1-3 at least will be reusable for other tests. What we currently use is [mostly this](https://github.com/google/OpenSK/blob/a0e11bd5aa01bf73d9ccd4336524f96316962475/src/ctap/client_pin.rs#L559):

ia0: Why do we need a PIN? The deploy script doesn't need it and simply calls the configure vendor command. Why wouldn't the following work?

```rust
ctap_state.process_vendor_configure(&mut env, params, channel)
```

But regardless of the answer, I think we agree that having a helper function for that is useful. The question is where should we put those helper functions. I guess a `#[cfg(feature = "std")]` module at the crate root: `test_helpers`, `test`, `test_helper`?

kaczmarczyck: 1. `process_vendor_configure` is a private function. We have to call [process_parsed_command](https://github.com/google/OpenSK/blob/a0e11bd5aa01bf73d9ccd4336524f96316962475/src/ctap/mod.rs#L530). But that's a small difference.
2. Enabling attestation is not a feature of our own configuration command `VendorConfig`, but CTAP's [`AuthenticatorConfig`](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#authenticatorConfig). I agree the naming is unfortunate. And this command must in almost all cases be protected by user verification.

And I agree it should be a `std` only module! I'd have liked to but it in [`tests`](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory), but I guess the fuzzer can't access it there? So your ideas sound reasonable.

ia0: I see. Then if I understand correctly, we would like the following (in possibly separate PRs or all in this PR):
- Create a `test` module at the root gated by the `std` feature.
- Add a `configure(&mut TestEnv, &mut CtapState, certificate, private_key)` command (bypassing HID and CBOR parsing, we might want wrapper functions for those alternatives) for `VendorConfig`.
- Add a `enable_enterprise_attestation(&mut TestEnv, &mut CtapState)` command for `AuthenticatorConfig`. This might need to deal with PIN and UV.
- Use those in tests and fuzzing.

From a distance, this looks like similar helpers one would need to write a blackbox CTAP test-suite like https://github.com/google/CTAP2-test-tool . Maybe we can have that in mind to help design.

hcyang-google:
Haven't check if fuzz coverage becomes better because I encountered some problems when using cargo fuzz coverage command.

hcyang-google:
`vendor_command` only works when `not(feature = "vendor_hid")`, but we need it for test_helpers setup. What should we do? Only enable those tests when the feature is not enabled? Or use some workarounds in test_helpers when the feature is enabled?

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -146,15 +146,39 @@ pub fn process_ctap_any_type(data: &[u8]) -> arbitrary::Result<()> {
     Ok(())
 }
 
+fn setup_env(
+    unstructured: &mut Unstructured,
+    state: &mut CtapState,
+    env: &mut TestEnv,
+) -> arbitrary::Result<()> {
+    if bool::arbitrary(unstructured)? {
+        test_helpers::env::setup_enterprise_attestation(state, env).ok();
```
ia0: I'm not sure `.ok()` is the right choice. It can hide actual problems. In which circumstances would this return an error that's not a bug? If we don't know I would start with `.unwrap()` and fix/adapt as we get bug reports from the fuzzer.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -146,15 +146,39 @@ pub fn process_ctap_any_type(data: &[u8]) -> arbitrary::Result<()> {
...
+    }
+    Ok(())
+}
+
+fn setup_customization(
```
ia0, line 164: nit: I would swap the order of those 2 functions because they must be called in a specific order and if that order matches the file order it would help readability.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -146,15 +146,39 @@ pub fn process_ctap_any_type(data: &[u8]) -> arbitrary::Result<()> {
...
+    unstructured: &mut Unstructured,
+    env: &mut TestEnv,
```
ia0: nit: If this function is really about customization, why not take `env.customization_mut()` directly instead of `env`? We might need to change that in the future if some weird case happens, but then we would know why.

src/ctap/mod.rs
```diff
@@ -29,6 +29,8 @@ mod pin_protocol;
 pub mod response;
 pub mod status_code;
 mod storage;
+#[cfg(feature = "std")]
+pub mod test_helpers;
```
ia0: Do we need this to be in `ctap`? Or can it be moved at top-level? Ideally the test helpers shouldn't be tied to our implementation.

src/ctap/test_helpers/customization.rs
```diff
@@ -0,0 +1,13 @@
+use crate::ctap::data_formats::EnterpriseAttestationMode;
+use crate::env::test::customization::TestCustomization;
+
+pub fn setup_enterprise_attestation(
```
ia0: I'm wondering if we don't want this to be a method of `TestCustomization` instead. Ideally we would not expose the internal fields and only expose high-level functions that preserve any necessary customization invariant.

In particular, shouldn't we check here that `mode` and `rp_id_list` are correlated?

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+) -> Result<SetupEnterpriseAttestationResponse, Ctap2StatusCode> {
+    let config_params = AuthenticatorConfigParameters {
+        sub_command: ConfigSubCommand::EnableEnterpriseAttestation,
+        sub_command_params: None,
+        pin_uv_auth_param: None,
+        pin_uv_auth_protocol: None,
+    };
+
+    let dummy_key = [0x41u8; key_material::ATTESTATION_PRIVATE_KEY_LENGTH];
+    let dummy_cert = [0xddu8; 20];
```
ia0: Why not directly `vec![0xdd; 20]`? (or `0xddu8` if needed but I don't think so)

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+    env: &mut impl Env,
+) -> Result<SetupEnterpriseAttestationResponse, Ctap2StatusCode> {
+    let config_params = AuthenticatorConfigParameters {
+        sub_command: ConfigSubCommand::EnableEnterpriseAttestation,
+        sub_command_params: None,
+        pin_uv_auth_param: None,
+        pin_uv_auth_protocol: None,
+    };
+
+    let dummy_key = [0x41u8; key_material::ATTESTATION_PRIVATE_KEY_LENGTH];
```
ia0: Do we really need the `u8` suffix? It's used right below so I would assume inference is able to do this.

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub struct SetupEnterpriseAttestationResponse {
+    pub attestation_material: AuthenticatorAttestationMaterial,
+}
+
+pub fn setup_enterprise_attestation(
+    state: &mut CtapState,
+    env: &mut impl Env,
+) -> Result<SetupEnterpriseAttestationResponse, Ctap2StatusCode> {
+    let config_params = AuthenticatorConfigParameters {
```
ia0: Let's move this right before it's used, to not extend scope unnecessarily and improve readability.

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+    AuthenticatorVendorConfigureParameters, Command,
+};
+use crate::ctap::data_formats::ConfigSubCommand;
+use crate::ctap::status_code::Ctap2StatusCode;
+use crate::ctap::{key_material, Channel, CtapState};
+use crate::env::Env;
+
+// In tests where we define a dummy user-presence check that immediately returns, the channel
+// ID is irrelevant, so we pass this (dummy but valid) value.
+const DUMMY_CHANNEL: Channel = Channel::MainHid([0x12, 0x34, 0x56, 0x78]);
```
ia0: Regarding the `vendor_hid` question, I guess you want to have 2 dummy channels, one main and one vendor. Depending whether `vendor_hid` feature is enabled, you want to use the vendor dummy channel for vendor commands.

src/ctap/test_helpers/mod.rs
```diff
@@ -0,0 +1,2 @@
+pub mod customization;
+pub mod env;
```
kaczmarczyck: I think `env` is a misleading name:

- `customization` is a part of `Env`.
- It doesn't convey that we want to call the CTAP API.

Maybe `ctap` is also not useful, since the crate name is `ctap`. Some variant of `state`?

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub fn setup_enterprise_attestation(
+    state: &mut CtapState,
+    env: &mut impl Env,
+) -> Result<SetupEnterpriseAttestationResponse, Ctap2StatusCode> {
+    let config_params = AuthenticatorConfigParameters {
+        sub_command: ConfigSubCommand::EnableEnterpriseAttestation,
+        sub_command_params: None,
+        pin_uv_auth_param: None,
+        pin_uv_auth_protocol: None,
```
kaczmarczyck: I'm surprised this works without `pin_uv_auth_param` and `pin_uv_auth_protocol`?

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+    let attestation_material = AuthenticatorAttestationMaterial {
+        certificate: dummy_cert.to_vec(),
+        private_key: dummy_key,
+    };
+    let configure_params = AuthenticatorVendorConfigureParameters {
+        lockdown: false,
+        attestation_material: Some(attestation_material.clone()),
+    };
+    let vendor_command = Command::AuthenticatorVendorConfigure(configure_params);
+    state.process_parsed_command(env, vendor_command, DUMMY_CHANNEL, CtapInstant::new(0))?;
```
kaczmarczyck: This is why for `cfg(vendor_hid)` we fail. In that case, you can't use
`const DUMMY_CHANNEL: Channel = Channel::MainHid([0x12, 0x34, 0x56, 0x78]);`
but you instead need
`const VENDOR_CHANNEL: Channel = Channel::VendorHid([0x12, 0x34, 0x56, 0x78]);`
That should fix it! Make sure to only use that channel on `AuthenticatorVendorConfigure` (and anything that is upgradability related, but I don't think we'll touch that).

src/ctap/test_helpers/env.rs
```diff
@@ -0,0 +1,49 @@
...
+use crate::ctap::data_formats::ConfigSubCommand;
+use crate::ctap::status_code::Ctap2StatusCode;
+use crate::ctap::{key_material, Channel, CtapState};
+use crate::env::Env;
+
+// In tests where we define a dummy user-presence check that immediately returns, the channel
+// ID is irrelevant, so we pass this (dummy but valid) value.
+const DUMMY_CHANNEL: Channel = Channel::MainHid([0x12, 0x34, 0x56, 0x78]);
+
+pub struct SetupEnterpriseAttestationResponse {
```
kaczmarczyck: Why do we want a single use struct here?

hcyang-google: I'm not sure I understand this question. This struct is mainly for grouping all outputs of the setup so that unittests/fuzzers could access?

hcyang-google: Oh I think I get your question. It only has one field now, why don't just return AttestationMaterial? I think we can do that too, but we might want return type naming to be consistent when we add more setup methods in the future?

kaczmarczyck: Why not directly return `AuthenticatorAttestationMaterial`? Do you intend to extend `setup_enterprise_attestation` and this struct helps with that? Are there other functions that will reuse `SetupEnterpriseAttestationResponse`? Or is this data type going to implement any helpful functions for fuzzing?

src/ctap/test_helpers/env.rs
hcyang-google: No pin was setup so it won't check those params. But we might need to change that after we add setup_pin helpers.

kaczmarczyck: Ah sorry, I misread the [specification](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#authenticatorConfig) when double checking.

The way it's written is quite non-intuitive, it's basic structure is:
> 3. if not alwaysUv etc: goto 5
> 4. if PIN is set or alwaysUv: check PIN
> 5. process command

So what you do works until we also customize or configure alwaysUv!

src/ctap/test_helpers/env.rs
kaczmarczyck: Ah we responded at the same time. :) I'd say if consistent naming is the only reason, remove it.

src/ctap/test_helpers/env.rs
hcyang-google: Then the current logic needs to change? We have
```
if storage::pin_hash(env)?.is_some() || enforce_uv {
```
meaning that we'll check it if pin is set or alwaysUv.

fuzz/fuzz_helper/src/lib.rs
hcyang-google: But this is for fuzzing setup. We arbitrarily setup some customization, then some initial state. I want to decouple the fuzzing of these two, so it's expected that the initial state setup might fail because the customization isn't correct, but it's fine? Otherwise we have to check whether the customization value allows us to do this specific state setup before each setup call and I think that might be too much logic in the fuzzer.

fuzz/fuzz_helper/src/lib.rs
hcyang-google: done

src/ctap/mod.rs
hcyang-google: Resolved in doc.

fuzz/fuzz_helper/src/lib.rs
hcyang-google: done

src/ctap/mod.rs
hcyang-google: moved out

src/ctap/test_helpers/customization.rs
hcyang-google: moved to TestCustomization's method.

hcyang-google: But I made the caller check whether the customization is valid instead as we have is_valid. Or do you think it's better to check in the API, and return some error if the customization doesn't make sense?

src/ctap/test_helpers/env.rs
hcyang-google: I copied it from the original unittest code, fixed.

src/ctap/test_helpers/env.rs
hcyang-google: done

src/ctap/test_helpers/env.rs
hcyang-google: fixed

src/ctap/test_helpers/env.rs
hcyang-google: done

src/ctap/test_helpers/mod.rs
hcyang-google: I moved out the customization test_helper. So the original env helpers are now just under the test_helper module, and we don't need to think of naming issue anymore.

src/ctap/test_helpers/env.rs
hcyang-google: done

src/ctap/test_helpers/env.rs
hcyang-google: done

src/ctap/test_helpers/env.rs
kaczmarczyck: No I think it's fine, my list was shortened for the explanation. You can double check the specification though if you feel like it.

src/ctap/test_helpers/mod.rs
kaczmarczyck: Works for me.

kaczmarczyck (commented):
LGTM for my comments, I'll leave approval to @ia0 .

fuzz/fuzz_helper/src/lib.rs
ia0: I think the problem is that we currently only fuzz for panics and not for correctness, so I see why we don't care about errors. Let's keep it like that for now and hope we rewrite everything when/if we start checking for correctness, because for sure we will miss those silent errors.

src/ctap/test_helpers/customization.rs
ia0: I would have checked in the API to have the customization valid by invariant. But given that it's anyway not the case in prod and not enforced by the trait, I think the current solution is fine. If we ever want something better, we probably need to introduce something like:

```rust
struct ValidCustomization<T: Customization>(T);
impl<T: Customization> ValidCustomization {
    pub fn new(x: T) -> Option<Self> { is_valid(&x).then(|| x) }
}
```

ia0 (approved):

Issue #335, opened 2021-06-26
Cannot generate ssh keys with ed25519-sk


gelintonx:
I'm trying to create ssh keys with ed25519-sk instead of ecdsa-sk but I receive this error: **Key enrollment failed: requested feature not supported** .
This is my Openssh version: **OpenSSH_8.4p1 Debian-5, OpenSSL 1.1.1k **

I'm using a nrf52840 dongle with the last updates of the stable branch.

Any idea why is this happening? Thanks in advance.

jmichelp:
I'll let Fabian comment on this but I don't think we implemented this algorithm. It's not part of the mandatory ones and would take more space on the flash.

gelintonx:
I'm newbie with Rust, do you think I can implement this algorithm if I fork the project? It's pretty difficult?
Thanks for the fast answer.

kaczmarczyck:
While binary size is a concern, having it as optional for devices with enough space (like the one you listed) would be great. I've seen this come up as a feature request in other security keys.

You'd have to:
1. implement a new curve for our crypto library,
2. add it to our CTAP implementation.

I'd say (1) is doable if you are not concerned with side channel resistance (like OpenSK's crypto right now). There should be test vectors out there to check your implementation. Trying Rust is surely a great project if you want to spend the time! Please check the [existing code](https://github.com/google/OpenSK/blob/stable/libraries/crypto/src/ecdsa.rs).
For (2), I can help writing it or give you pointers. Digging through CTAP can be tricky, you'd have to add a new algorithm to `CoseKey` and `match` the algorithm in the commands that use it.

gelintonx:
Hi @kaczmarczyck, I started implementing the ed25519 curve, should I create a function to check that keys are valid? 
Thanks in advance.

kaczmarczyck:
Not sure if this answers your question, but in Rust, you add a module `test` to the end of your implementation like:

```
#[cfg(test)]
mod test {
  fn test_something() {
    assert!(false);
  }
}
```

I'd expect the crypto code to be similarly tested like the [existing crypto library](https://github.com/google/OpenSK/blob/7e5e2a665ff4f9ba68b7e8d098ee2248a3a85bb8/libraries/crypto/src/ecdsa.rs#L353). Does that answer your question?

kaczmarczyck:
Hi @gelintonx ! I'm closing this issue due to inactivity, feel free to re-open if you have any updates!

egor-duda:
Hello!
[...]

PR #476, opened 2022-05-04
Make private keys extensible

kaczmarczyck:
This PR prepares addition of new signature algorithms. It is a requirement for #335.

Resident keys store the private key as a CBOR array, with the first element being the algorithm identifier, and the second the encoded private key. Server-side credentials now have an additional block to encode the used algorithm. The 12 remaining bytes are reserved for future use.

We can still parse old credentials, both resident and server-side. From the ~1kB increased binary size of this change, the extra code required to support legacy is ~300B. Legacy credentials can not be created anymore. If we'd release on different hardware, we can remove the legacy code there, as it could only encountered legitimate legacy credentials by master key transfer shenanigans.

This PR moves credential encryption and decryption to `crypto_wrapper.rs`, as it seemed closely related to the new `PrivateKey`. In a previous design, we had a generic private key in `library/crypto`. However, I believe that this should align better with our CTAP library goals. The generic `PrivateKey` should be inside the library, and eventually call crypto primitives from `Env`. These crypto primitives may or may not be `library/crypto`. By doing so, the PR also removes a few unnecessary `CtapState` dependencies in CTAP1. @ia0 Let me know if you rather have it split for submission, then this PR will act as the reference.

Unit tests for legacy transfer exist, but this change should be tested thoroughly before merging.

### DIFF
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -16,7 +16,7 @@ use super::super::clock::CtapInstant;
 use super::client_pin::{ClientPin, PinPermission};
 use super::command::AuthenticatorCredentialManagementParameters;
 use super::data_formats::{
-    CoseKey, CredentialManagementSubCommand, CredentialManagementSubCommandParameters,
+    CredentialManagementSubCommand, CredentialManagementSubCommandParameters,
     PublicKeyCredentialDescriptor, PublicKeyCredentialRpEntity, PublicKeyCredentialSource,
     PublicKeyCredentialUserEntity,
 };
@@ -92,7 +92,7 @@ fn enumerate_credentials_response(
         key_id: credential_id,
         transports: None, // You can set USB as a hint here.
     };
-    let public_key = CoseKey::from(private_key.genpk());
+    let public_key = private_key.get_pub_key();
     Ok(AuthenticatorCredentialManagementResponse {
         user: Some(user),
         credential_id: Some(credential_id),
@@ -359,6 +359,7 @@ pub fn process_credential_management(
 
 #[cfg(test)]
 mod test {
+    use super::super::crypto_wrapper::PrivateKey;
     use super::super::data_formats::{PinUvAuthProtocol, PublicKeyCredentialType};
     use super::super::pin_protocol::authenticate_pin_uv_auth_token;
     use super::super::CtapState;
@@ -373,7 +374,7 @@ mod test {
         PublicKeyCredentialSource {
             key_type: PublicKeyCredentialType::PublicKey,
             credential_id: rng.gen_uniform_u8x32().to_vec(),
-            private_key,
+            private_key: PrivateKey::from(private_key),
             rp_id: String::from("example.com"),
             user_handle: vec![0x01],
             user_display_name: Some("display_name".to_string()),
--- a/src/ctap/crypto_wrapper.rs
+++ b/src/ctap/crypto_wrapper.rs
@@ -551,4 +551,15 @@ mod test {
 
         assert_eq!(private_key, decrypted_source.private_key);
     }
+
+    #[test]
+    fn test_encrypt_credential_size() {
+        let mut env = TestEnv::new();
+        storage::init(&mut env).ok().unwrap();
+        let private_key = PrivateKey::new(env.rng(), SignatureAlgorithm::ES256);
+
+        let rp_id_hash = [0x55; 32];
+        let encrypted_id = encrypt_key_handle(&mut env, &private_key, &rp_id_hash).unwrap();
+        assert_eq!(encrypted_id.len(), ECDSA_CREDENTIAL_ID_SIZE);
+    }
 }
--- a/src/ctap/crypto_wrapper.rs
+++ b/src/ctap/crypto_wrapper.rs
@@ -39,8 +39,9 @@ use sk_cbor::{cbor_array, cbor_bytes, cbor_int};
 // - 32 bytes: HMAC-SHA256 over everything else.
 pub const LEGACY_CREDENTIAL_ID_SIZE: usize = 112;
 #[cfg(test)]
-pub const ECDSA_CREDENTIAL_ID_SIZE: usize = 128;
-pub const MAX_CREDENTIAL_ID_SIZE: usize = 128;
+pub const ECDSA_CREDENTIAL_ID_SIZE: usize = 129;
+// See encrypt_key_handle v1 documentation.
+pub const MAX_CREDENTIAL_ID_SIZE: usize = 129;
 
 /// Wraps the AES256-CBC encryption to match what we need in CTAP.
 pub fn aes256_cbc_encrypt(
@@ -52,7 +53,8 @@ pub fn aes256_cbc_encrypt(
     if plaintext.len() % 16 != 0 {
         return Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER);
     }
-    let mut ciphertext = Vec::with_capacity(plaintext.len() + 16 * embeds_iv as usize);
+    // The extra 1 capacity is because encrypt_key_handle adds a version number.
+    let mut ciphertext = Vec::with_capacity(plaintext.len() + 16 * embeds_iv as usize + 1);
     let iv = if embeds_iv {
         let random_bytes = rng.gen_uniform_u8x32();
         ciphertext.extend_from_slice(&random_bytes[..16]);
@@ -190,10 +192,10 @@ impl From<ecdsa::SecKey> for PrivateKey {
 /// ECDSA private keys.
 ///
 /// This is v1, and we write the following data for ECDSA (algorithm -7):
+/// -  1 byte : version number
 /// - 16 bytes: initialization vector for AES-256,
 /// -  4 bytes: algorithm in big endian encoding,
-/// -  1 byte : version number
-/// - 11 bytes: reserved,
+/// - 12 bytes: reserved,
 /// - 32 bytes: ECDSA private key for the credential,
 /// - 32 bytes: relying party ID hashed with SHA256,
 /// - 32 bytes: HMAC-SHA256 over everything else.
@@ -202,24 +204,25 @@ pub fn encrypt_key_handle(
     private_key: &PrivateKey,
     application: &[u8; 32],
 ) -> Result<Vec<u8>, Ctap2StatusCode> {
-    match private_key {
+    let master_keys = storage::master_keys(env)?;
+    let aes_enc_key = crypto::aes256::EncryptionKey::new(&master_keys.encryption);
+
+    let mut encrypted_id = match private_key {
         PrivateKey::Ecdsa(ecdsa_key) => {
-            let master_keys = storage::master_keys(env)?;
-            let aes_enc_key = crypto::aes256::EncryptionKey::new(&master_keys.encryption);
             let mut plaintext = [0; 80];
             // Algorithm
             BigEndian::write_i32(&mut plaintext, SignatureAlgorithm::ES256 as i32);
-            // Version number
-            plaintext[4] = 0x01;
             ecdsa_key.to_bytes(array_mut_ref!(plaintext, 16, 32));
             plaintext[48..80].copy_from_slice(application);
-
-            let mut encrypted_id = aes256_cbc_encrypt(env.rng(), &aes_enc_key, &plaintext, true)?;
-            let id_hmac = hmac_256::<Sha256>(&master_keys.hmac, &encrypted_id[..]);
-            encrypted_id.extend(&id_hmac);
-            Ok(encrypted_id)
+            aes256_cbc_encrypt(env.rng(), &aes_enc_key, &plaintext, true)?
         }
-    }
+    };
+
+    // Version number
+    encrypted_id.insert(0, 0x01);
+    let id_hmac = hmac_256::<Sha256>(&master_keys.hmac, &encrypted_id[..]);
+    encrypted_id.extend(&id_hmac);
+    Ok(encrypted_id)
 }
 
 /// Decrypts a credential ID and writes the private key into a PublicKeyCredentialSource.
@@ -242,28 +245,35 @@ pub fn decrypt_credential_source(
         return Ok(None);
     }
     let master_keys = storage::master_keys(env)?;
-    let payload_size = credential_id.len() - 32;
+    let hmac_message_size = credential_id.len() - 32;
     if !verify_hmac_256::<Sha256>(
         &master_keys.hmac,
-        &credential_id[..payload_size],
-        array_ref![credential_id, payload_size, 32],
+        &credential_id[..hmac_message_size],
+        array_ref![credential_id, hmac_message_size, 32],
     ) {
         return Ok(None);
     }
 
+    let is_legacy = credential_id.len() == LEGACY_CREDENTIAL_ID_SIZE;
+    let payload = if is_legacy {
+        &credential_id[..hmac_message_size]
+    } else {
+        // Version number check
+        if credential_id[0] != 1 {
+            return Ok(None);
+        }
+        &credential_id[1..hmac_message_size]
+    };
+
     let aes_enc_key = crypto::aes256::EncryptionKey::new(&master_keys.encryption);
-    let decrypted_id = aes256_cbc_decrypt(&aes_enc_key, &credential_id[..payload_size], true)?;
+    let decrypted_id = aes256_cbc_decrypt(&aes_enc_key, payload, true)?;
 
     if rp_id_hash != &decrypted_id[decrypted_id.len() - 32..] {
         return Ok(None);
     }
-    let sk_option = if credential_id.len() == LEGACY_CREDENTIAL_ID_SIZE {
+    let sk_option = if is_legacy {
         PrivateKey::new_ecdsa_from_bytes(&decrypted_id[..32])
     } else {
-        // Version number check.
-        if decrypted_id[4] == 0 {
-            return Ok(None);
-        }
         let algorithm_int = BigEndian::read_i32(&decrypted_id[..4]);
         match SignatureAlgorithm::from(algorithm_int as i64) {
             SignatureAlgorithm::ES256 => PrivateKey::new_ecdsa_from_bytes(&decrypted_id[16..48]),
--- a/src/ctap/crypto_wrapper.rs
+++ b/src/ctap/crypto_wrapper.rs
@@ -12,10 +12,42 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::ctap::data_formats::{
+    extract_array, extract_byte_string, CoseKey, PublicKeyCredentialSource,
+    PublicKeyCredentialType, SignatureAlgorithm,
+};
 use crate::ctap::status_code::Ctap2StatusCode;
+use crate::ctap::storage;
+use crate::env::Env;
+use alloc::string::String;
+use alloc::vec;
 use alloc::vec::Vec;
+use byteorder::{BigEndian, ByteOrder};
+use core::convert::TryFrom;
 use crypto::cbc::{cbc_decrypt, cbc_encrypt};
+use crypto::ecdsa;
+use crypto::hmac::{hmac_256, verify_hmac_256};
+use crypto::sha256::Sha256;
 use rng256::Rng256;
+use sk_cbor as cbor;
+use sk_cbor::{cbor_array, cbor_bytes, cbor_int};
+
+// Legacy credential IDs consist of
+// - 16 byte initialization vector for AES-256,
+// - 32 byte ECDSA private key for the credential,
+// - 32 byte relying party ID hashed with SHA256,
+// - 32 byte HMAC-SHA256 over everything else.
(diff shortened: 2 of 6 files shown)


src/ctap/credential_management.rs
```diff
@@ -373,7 +374,7 @@ mod test {
         PublicKeyCredentialSource {
             key_type: PublicKeyCredentialType::PublicKey,
             credential_id: rng.gen_uniform_u8x32().to_vec(),
-            private_key,
+            private_key: PrivateKey::from(private_key),
```
ia0, line 377: fyi: could be `private_key.into()`

src/ctap/crypto_wrapper.rs
```diff
@@ -12,10 +12,42 @@
...
+use crypto::sha256::Sha256;
 use rng256::Rng256;
+use sk_cbor as cbor;
+use sk_cbor::{cbor_array, cbor_bytes, cbor_int};
+
+// Legacy credential IDs consist of
+// - 16 byte initialization vector for AES-256,
+// - 32 byte ECDSA private key for the credential,
+// - 32 byte relying party ID hashed with SHA256,
+// - 32 byte HMAC-SHA256 over everything else.
```
ia0: nit: I think the plural of byte is bytes. Also I would put some colons to help parsing:

```
Legacy credential IDs consist of:
- 16 bytes: initialization vector for AES-256
- 32 bytes: ECDSA private key of the credential
- 32 bytes: relying party ID ...
```

src/ctap/crypto_wrapper.rs
```diff
@@ -12,10 +12,42 @@
...
+pub const LEGACY_CREDENTIAL_ID_SIZE: usize = 112;
+// New credential IDs are still ECDSA only, and consist of
+// - 16 byte initialization vector for AES-256,
+// -  4 byte algorithm,
+// - 12 byte reserved,
+// - 32 byte ECDSA private key for the credential,
+// - 32 byte relying party ID hashed with SHA256,
+// - 32 byte HMAC-SHA256 over everything else.
```
ia0: Shouldn't we instead start using forward-compatible format? Either something prefixed with a version number (if we can't do anything polymorphic over multiple versions) or something more flexible like a CBOR map with stable keys (e.g. algo, IV, priv key, RPID, hash algo, etc).

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +94,185 @@ pub fn aes256_cbc_decrypt(
     Ok(plaintext)
 }
 
+/// An asymmetric private key that can sign messages.
+#[derive(Clone, Debug, PartialEq)]
+pub enum PrivateKey {
+    EcdsaKey(ecdsa::SecKey),
```
ia0: nit: Is the `Key` suffix adding information? Why not simply `Ecdsa`?

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +94,185 @@ pub fn aes256_cbc_decrypt(
...
+}
+
+impl PrivateKey {
+    /// Creates a new private key for the given algorithm.
+    ///
+    /// Calling new with the unknown is invalid.
```
ia0: Let's put this in a `Panics` section ([API guidelines](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure)):

```rust
/// # Panics
///
/// Panics if the algorithm is [`SignatureAlgorithm::Unknown`].
```

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +94,185 @@ pub fn aes256_cbc_decrypt(
...
+    }
+}
+
+impl From<ecdsa::SecKey> for PrivateKey {
+    fn from(ecdsa_key: ecdsa::SecKey) -> Self {
+        PrivateKey::EcdsaKey(ecdsa_key)
+    }
+}
+
+/// Encrypts the given private key and relying part ID hash into a credential ID.
```
ia0: typo: party?

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +94,185 @@ pub fn aes256_cbc_decrypt(
...
+///
+/// Other information, such as a user name, are not stored. Since encrypted credential IDs are
+/// stored server-side, this information is already available (unecrypted).
```
ia0: typo: unencrypted?

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +94,185 @@ pub fn aes256_cbc_decrypt(
...
+pub fn encrypt_key_handle(
+    env: &mut impl Env,
+    private_key: &PrivateKey,
+    application: &[u8; 32],
+) -> Result<Vec<u8>, Ctap2StatusCode> {
+    match private_key {
+        PrivateKey::EcdsaKey(ecdsa_key) => {
+            let master_keys = storage::master_keys(env)?;
+            let aes_enc_key = crypto::aes256::EncryptionKey::new(&master_keys.encryption);
+            let mut plaintext = [0; 80];
```
ia0: Is the format specified? Why a hole from byte 4 to 15?

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +94,185 @@ pub fn aes256_cbc_decrypt(
...
+}
+
+/// Decrypts a credential ID and writes the private key into a PublicKeyCredentialSource.
+///
+/// Returns None if
+/// - the HMAC test fails or
+/// - the relying party does not match the decrypted relying party ID hash.
+pub fn decrypt_credential_source(
+    env: &mut impl Env,
+    credential_id: Vec<u8>,
```
ia0, line 236: Why not `&[u8]`? Doesn't look like the function needs ownership.

src/ctap/ctap1.rs
```diff
@@ -311,10 +310,14 @@ impl Ctap1Command {
...
     ) -> Result<Vec<u8>, Ctap1StatusCode> {
-        let credential_source = ctap_state
-            .decrypt_credential_source(env, key_handle, &application)
+        let credential_source = decrypt_credential_source(env, key_handle, &application)
             .map_err(|_| Ctap1StatusCode::SW_WRONG_DATA)?;
         if let Some(credential_source) = credential_source {
+            let ecdsa_key = match credential_source.private_key {
+                PrivateKey::EcdsaKey(k) => k,
+                // With the introduction of different algorithms, add:
+                // _ => return Err(Ctap1StatusCode::SW_WRONG_DATA),
```
ia0: I wonder if adding `#[non_exhaustive]` on `PrivateKey` definition would permit to already add this branch.

src/ctap/data_formats.rs
```diff
@@ -497,7 +498,8 @@ impl From<PackedAttestationStatement> for cbor::Value {
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+/// Signature algorithm identifier, as specified for COSE.
+#[derive(Clone, Copy, Debug, PartialEq)]
```
ia0: It's weird to derive `PartialEq` and not `Eq`, given that the `PartialEq` derived implementation is symmetric.

kaczmarczyck: The whole file only derives the minimum requirement. Most data structures have `PartialEq` and not `Eq`. Should I rather do a follow-up PR that add `Eq` to most, if that makes more sense?

src/ctap/crypto_wrapper.rs
kaczmarczyck: The format is not specified, I made it up. My train of thought was as follows:

1. We have to use multiples of 16 bytes for AES encryption.
2. Credential IDs should look like random data.
3. Everything should be encrypted.
4. The credential ID length must be a multiple of 16.
5. Since we need 1 new block anyway, maybe reserve it for future occasions like this, to make backwards compatibility easier.

src/ctap/crypto_wrapper.rs
kaczmarczyck: See my comment below for some constraints. I'm open to suggestions, a version number make sense. A CBOR map is the most flexible, but also the most expensive:
- keys have to be encoded
- maps don't align with AES block size => padding
- more AES blocks => more compute cost => more latency
- CBOR map parsing costs more binary size

So if you think we gain a lot of forward compatibility, we can do it. Depends on how much we gain compared to i.e. adding a version number only?

src/ctap/data_formats.rs
ia0: Yes, another PR sounds good given it's a generalized issue. But we should merge it before that one. This is a bug to not derive `Eq` when the code relies on equality being symmetric. I just looked at the first instance of `#[derive(PartialEq)]` in `src/ctap/data_formats.rs`, namely `PublicKeyCredentialRpEntity`, and we use `assert_eq!` in tests which means we rely on symmetry at least for that element. If we really don't want to derive `Eq`, we should at least document for which elements equality is supposed to be symmetric.

src/ctap/crypto_wrapper.rs
ia0: Sounds good, but then we should document the rationale for this format either here or where we do the opposite operation, given it's not standard.

src/ctap/crypto_wrapper.rs
ia0: I don't know enough about the usage to know if polymorphic operations across versions make sense. If they don't or if we don't know, then a version number is good.

src/ctap/data_formats.rs
kaczmarczyck: See #477.

src/ctap/crypto_wrapper.rs
kaczmarczyck: It finally moves the credential ID into the created data structure, if successful. And looking at the call sites, this seems to make sense.

ia0: Oh I missed that part. Sounds good.

src/ctap/data_formats.rs
kaczmarczyck: And `Eq` is also added to `PrivateKey` now.

src/ctap/ctap1.rs
kaczmarczyck: It still shows a warning. I agree that it's nicer to have the line active so we don't have to touch the file later. So I ignore the warning now.

src/ctap/crypto_wrapper.rs
kaczmarczyck: The encryption now explains its own version (moving the format description to its docstring), and the decryption explains what versions it can parse. FYI the legacy encryption still exists in code for tests, and likely all old versions will move there.

src/ctap/crypto_wrapper.rs
kaczmarczyck: Done

src/ctap/crypto_wrapper.rs
kaczmarczyck: Done

src/ctap/crypto_wrapper.rs
kaczmarczyck: Done

src/ctap/crypto_wrapper.rs
kaczmarczyck: Renamed

src/ctap/crypto_wrapper.rs
kaczmarczyck: I added a version number byte. For all use cases I see, this should be good enough.

src/ctap/crypto_wrapper.rs
kaczmarczyck: Fixed

src/ctap/crypto_wrapper.rs
```diff
@@ -184,28 +177,40 @@ impl TryFrom<cbor::Value> for PrivateKey {
...
-/// stored server-side, this information is already available (unecrypted).
+/// stored server-side, this information is already available (unencrypted).
 ///
 /// Also, by limiting ourselves to private key and RP ID hash, we are compatible with U2F for
 /// ECDSA private keys.
+///
+/// This is v1, and we write the following data for ECDSA (algorithm -7):
+/// - 16 bytes: initialization vector for AES-256,
+/// -  4 bytes: algorithm in big endian encoding,
+/// -  1 byte : version number
```
ia0, line 194: I was expecting the version number to not be encrypted. Also this documentation is confusing. The way I understand the code, the format is:
- IV (16B)
- Payload encrypted by the IV above (80B):
    - Algo (4B)
    - Version (1B)
    - Padding (11B)
    - Private key (32B)
    - RP id hash (32B)
- Hmac of everything above (32B)

In other words I was expecting something like:
- Version (1B)
- IV (16B)
- Payload encrypted by IV (80B): same as above but without version
- Hmac of everything (32B)

This is one more byte in total.

src/ctap/crypto_wrapper.rs
```diff
@@ -249,6 +260,10 @@ pub fn decrypt_credential_source(
     let sk_option = if credential_id.len() == LEGACY_CREDENTIAL_ID_SIZE {
         PrivateKey::new_ecdsa_from_bytes(&decrypted_id[..32])
     } else {
+        // Version number check.
+        if decrypted_id[4] == 0 {
```
ia0: Shouldn't this be `!= 1`?

src/ctap/ctap1.rs
```diff
@@ -309,10 +309,11 @@ impl Ctap1Command {
         let credential_source = decrypt_credential_source(env, key_handle, &application)
             .map_err(|_| Ctap1StatusCode::SW_WRONG_DATA)?;
         if let Some(credential_source) = credential_source {
+            #[allow(unreachable_patterns)]
+            // CTAP1 only supports ECDSA, the default case applies if CTAP2 adds more algorithms.
```
ia0, line 312: I would put the comment before the `#[allow]` otherwise it's a bit confusing. In particular given that this should apply to the match not the let. But expression attribute are not well supported yet (I think).

src/ctap/crypto_wrapper.rs
kaczmarczyck: Ah right, having the version outside helps if we want to replace AES-CBC or HMAC. The 2 problems I saw with that:

1. I wanted credential IDs to be as indistinguishable from the outside as possible, so it doesn't help with fingerprinting. As an example, if you are using a different version as everyone else, relying parties could link your accounts together. You are probably already identifyable as an OpenSK user.
2. Someone could take a credential ID and downgrade the cryptography, without us noticing. I don't have a concrete attack in mind, but I thought it could be best practise to not allow that.

Might make sense to look at related work for best practises.

src/ctap/crypto_wrapper.rs
kaczmarczyck: I wasn't sure if having a best effort forward compatibility is desireable.

src/ctap/crypto_wrapper.rs
ia0: 1. I see. I think it's a weak argument but still true. It's weak for 2 reasons. First, it's anyway gradual, because just looking at the length of the credential id already gives information, and right now version number is equivalent to length of credential id. The second reason being that the smallest cluster of users with a given version shouldn't be small enough to be a privacy concern, but that's obviously impossible to prove since very subjective and a statement about the future.
2. I'm not sure how it's possible. How would they decrypt the payload? (If the encryption is the same for different versions, then we could duplicate the version in the payload too.)

Now to give the counter-part, the problem I see with the current version is that:
3. We can't change AES-CBC (we should still be able to change HMAC)
4. The format is a bit constrained, but given that AES-CBC is already constrained, it's not really an issue.

So we essentially compare 1 and 3. And I would say 3 is more important since it's about upgradability which comes before most other properties like security or privacy (since a non-upgradable key with security or privacy defects is essentially unusable while a privacy defect cannot affect security). But if you believe AES-CBC is not going to change, then 3 is useless and I'm fine to address 1 first.

src/ctap/crypto_wrapper.rs
ia0: For me, the fact that we go with version numbers, mean we don't expect any kind of polymorphism across versions, otherwise we would use CBOR maps. So I would say versions have to be check exactly.

kaczmarczyck: Changed. Also now that the version can be read before HMAC, this makes even more sense. Future versions might want stronger authenticity checks.
(Note that as an implementation detail, the HMAC is checked before the version, because all known versions need it. We are free to reorder in the future.)

src/ctap/ctap1.rs
kaczmarczyck: Done

src/ctap/crypto_wrapper.rs
kaczmarczyck: The version number is now part of the HMAC, but not part of the encryption. This should leave the information leak as the only problem. At the same time, this is could be used to our advantage. Technically, clients could, whenever they forward credential IDs, warn you to re-create them because of outdated crypto.

ia0: Indeed that could also be used as our advantage.

src/ctap/crypto_wrapper.rs
```diff
@@ -62,6 +89,215 @@ pub fn aes256_cbc_decrypt(
...
+/// Other information, such as a user name, are not stored. Since encrypted credential IDs are
+/// stored server-side, this information is already available (unencrypted).
+///
+/// Also, by limiting ourselves to private key and RP ID hash, we are compatible with U2F for
+/// ECDSA private keys.
+///
+/// This is v1, and we write the following data for ECDSA (algorithm -7):
+/// -  1 byte : version number
+/// - 16 bytes: initialization vector for AES-256,
+/// -  4 bytes: algorithm in big endian encoding,
```
ia0: Out of curiosity, is the algorithm sensitive information? I'm wondering if it could be part of the version number (the algorithm would be a function of the version number). This would make sense if we want the client to notify when using old algorithms (the idea you mentioned in another comment) modulo the increased privacy concerns of version number identification.

ia0 (dismissed):

src/ctap/crypto_wrapper.rs
kaczmarczyck: The relying party knows the algorithm, since they store the public key to the encrypted private key. It might also leak due to credential ID size changing. So we could decide to make these bytes "public" as well. Since we'd save the extra AES block, it would make logic a bit easier (good for maintenance) and assertions a bit faster.

By the way, the above mentioned advantage is very theoretical, to be honest. For that to happen, the client needs to have special logic for any OpenSK-based authenticators. The credential wrapping does not have any recommended defaults (part of why we design this ourselves).

Is there anything you had in mind when you said "function of the version number"? I guess just store both like we do now, but we can have a ECDSA v2 and a EdDSA v1 and a COOL_CRYPTO v3 as the respective latest versions?

ia0: I was thinking if algorithm are changed to better or added as alternative. Because if it's the first, then algorithms are ordered within a version, and I'm assuming versions are also changed for the better, so we have a natural order which can just be a number (we might want to use LEB128 i.e. start using 2 bytes when reaching version 128). That number would fully describe the format that follows (including presence of hmac and its algorithm).

But if algorithms are not ordered within version, then yes, a pair sounds good.

For example, with a single number, decryption would look like:

```rust
match version {
    // Legacy versions
    5 => parse_v5()
    6 => parse_v6()
    7 => parse_v7()
    // Current version (handles are created in this format)
    8 => parse_v8()
    // Unsupported versions (either too old, too new, or invalid)
    _ => error()
}
```

kaczmarczyck: I see. So if we want to i.e. upgrade AES and keep all algorithms as is, we'd issue a new version per algorithm. I think that should be good enough.

ia0: So it would be a real scenario that multiple algorithms live simultaneously while none is better than all others? If yes, I think the pair is alright, and moving out of the encrypted area would just be to save some padding.

kaczmarczyck: I think that is a possible scenario. The current EdDSA PR is an example, I don't think EdDSA is strictly better than ECDSA. Also new algorithms could be different key lengths of old versions, and you just hope that the server signals support for the longer key, but allow the older version. But per variant of `PrivateKey`, there will be only 1 active version.

I think I still like the version only encoding, WDYT?

kaczmarczyck: (The latest commit implements that.)

kaczmarczyck:
Tested and confirmed working on webauthn.io by

1. flashing develop
2. register
3. flashing this branch
4. login

src/ctap/crypto_wrapper.rs
ia0: Sounds good. I'm just worried that version number implies ordering. Maybe renaming to "credential format" or "credential type" or something that doesn't suggest linear ordering. But for now the ordering is meaningful, so fine to keep the name for now.

ia0 (approved):

kaczmarczyck:
Renaming later sounds good. It's purely internal, so that's free.

Issue #485, opened 2022-05-29
Implementing OpenSK-based FIDO2 emulator for QEMU


egor-duda:
I've made a quick-and-dirty prototype of virtual FIDO2 authenticator for QEMU using OpenSK application. The purpose is similar to #288, but instead of creating a CTAP library and linking it into QEMU, I've patched QEMU to add a fido2 device, which uses socket-connected chardev. Then I've modified main OpenSK application to listen on a unix socket and, after connection from QEMU, to process all CTAP requests/responses via that socket connection.

To go beyond prototype, I suppose I have to:

1. Create separate NonEmbeddedEnv (naming suggestions are welcome) for running on non-embedded OS
2. Implement file-based storage for persistent authenticator data. Currently I've used TestEnv with its BufferStorage, which means that master key is being lost after OpenSK application restart, making all previously created credentials invalid
3. Move main protocol I/O function from loop in src/main.rs to Env, and implement socket- or stdio- based I/O in NonEmbeddedEnv
4. Implement user-presence checking in new NonEmbeddedEnv

Obviously, all this new NonEmbeddedEnv stuff should not be compiled in when building token-targeted version of an application.

Is this a viable plan?

kaczmarczyck:
May I ask if you have a specific goal in mind? Is QEMU the purpose like in #288, or do you want a desktop OpenSK?

Let me elaborate a bit on our current work in progress. We noticed that OpenSK is more testable and more broadly usable if its logic was separate from the platform it runs on. With that in mind, we started splitting everything TockOS-related into `TockEnv`. (It looks like you found that, and noticed that it's infinished.)  Some functions in `TockEnv` still contain CTAP logic, that needs a cleaner sparation. Generally, we haven't converged on the best API and `Env` trait yet.
All of this is tracked under a [project](https://github.com/google/OpenSK/projects/7).

Once that is finished, you will be able to support a new platform by adding a new `Env`, i.e. what you call `NonEmbeddedEnv` (I agree we'd need a different name ^^). That would hopefully be easier than it is right now. So your plan is pretty viable!

Since you were asking for "beyond prototype", I assume you mean that you want to create a PR that we would eventually merge. Cleanest and easiest would be for us to first finish the CTAP library, and then merging your new `NonEmbeddedEnv`. Else we'd have to port yet another platform to the new API. Afterwards, you could use that desktop version of OpenSK for whatever you want, for example QEMU. Is this a helpful answer for what you have in mind?

[...]

PR #486, opened 2022-06-02
Make Storage::read_slice take a mutable slice

ia0:
As mentioned in [this comment](https://github.com/google/OpenSK/issues/485#issuecomment-1144024122), the `Storage::read_slice` API is too strong because some implementations cannot return a slice. This PR changes the API to take a mutable slice instead. This essentially enables 2 use-cases:
- File-backed storage as mentioned in #485
- Implementations using [`embedded_storage::nor_flash::NorFlash`](https://docs.rs/embedded-storage/0.3.0/embedded_storage/nor_flash/trait.NorFlash.html)

### DIFF
--- a/libraries/persistent_store/src/store.rs
+++ b/libraries/persistent_store/src/store.rs
@@ -16,7 +16,7 @@
 
 use crate::format::{
     is_erased, CompactInfo, Format, Header, InitInfo, InternalEntry, Padding, ParsedWord, Position,
-    Word, WordState,
+    Word, WordSlice, WordState,
 };
 #[cfg(feature = "std")]
 pub use crate::model::{StoreModel, StoreOperation};
@@ -490,7 +490,8 @@ impl<S: Storage> Store<S> {
     fn recover_initialize(&mut self) -> StoreResult<()> {
         let word_size = self.format.word_size();
         for page in 0..self.format.num_pages() {
-            let (init, rest) = self.read_page(page).split_at(word_size as usize);
+            let page_content = self.read_page(page);
+            let (init, rest) = page_content.split_at(word_size as usize);
             if (page > 0 && !is_erased(init)) || !is_erased(rest) {
                 return Ok(());
             }
@@ -890,7 +891,7 @@ impl<S: Storage> Store<S> {
 
     /// Sets the padding bit of a user header.
     fn set_padding(&mut self, pos: Position) -> StoreResult<()> {
-        let mut word = Word::from_slice(self.read_word(pos));
+        let mut word = Word::from_slice(&self.read_word(pos));
         self.format.set_padding(&mut word)?;
         self.write_slice(pos, &word.as_slice())?;
         Ok(())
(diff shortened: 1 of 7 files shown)


libraries/persistent_store/src/store.rs
```diff
@@ -1044,57 +1045,63 @@ impl<S: Storage> Store<S> {
...
+        let mut result = vec![0; length as usize];
         let index = pos.index(&self.format);
         let max_length = self.format.page_size() - usize_to_nat(index.byte);
-        result.extend_from_slice(self.storage_read_slice(index, min(length, max_length)));
+        self.storage_read_slice(index, &mut result[..min(length, max_length) as usize]);
         if length > max_length {
             // The slice spans the next page.
             let index = pos.next_page(&self.format).index(&self.format);
-            result.extend_from_slice(self.storage_read_slice(index, length - max_length));
+            self.storage_read_slice(index, &mut result[max_length as usize..]);
```
kaczmarczyck, line 1078: Double checking: I was wondering how much more copying we will have to do with this PR. Looks like we have to copy up to 2 pages now, instead of just the final slice that has always been a copy?

ia0: Yes, there will be more allocations and more copying with this PR.

However, for the specific code on which this comment is attached, there's actually no additional allocation or copy. That's probably one of the rare exceptions. For me the most problematic place would be when reading a page just to check that it's erased (this doesn't happen often, at boot and compaction) and when writing a slice to check that the write is valid (this happens at store mutation). When reading, we don't have additional penalty, which is actually ironic since the PR is only changing the reading API. This comes from the fact that the store API is already allocating and copying, so we don't add more. The allocation and copying during store mutation (insert, remove, compaction, etc) is acceptable because it's dominated by the cost of the actual flash write. However the allocation and copying for booting the store is more problematic. But since it happens at boot time, the extra cost is less sensitive.

kaczmarczyck: Do you have a store benchmark you can quickly run, out of curiousity?

kaczmarczyck (dismissed):

libraries/persistent_store/src/store.rs
ia0: Good point, I forgot about it. I've updated the numbers. Here's a summary:
- Boot is 35% slower
- Compaction is 5% slower
- Insert is 30% slower
- Remove is 35% slower

kaczmarczyck: While that may not be problematic, how hard would it be to have an API that is still general over different use cases (files and embedded), but less performance issues on embedded? Just to understand the alternative.

ia0: Indeed, there's an alternative that we may prefer: #487 .

I believe there would be an ideal alternative. I don't think it's hard to implement, so I might just try.

ia0:
Closing in favor of #487

PR #487, opened 2022-06-03
Allow read_slice to return a Vec

ia0:
Similar to #486 but takes the opposite approach:
- Memory-mapped storage can continue to return the slice directly
- Storage that require a mutable slice must allocate themselves

This is preferable to #486 because:
- Nordic has similar performance
- File-backed storage is not for embedded so allocating is ok

If we ever need to support `embedded-storage` trait, then we could add an `Option<&mut [u8]>` argument to be used and returned in the `Cow<[u8]>`. This would avoid allocation and copy. But since we don't have this need, we don't do it.

### DIFF
--- a/examples/store_latency.rs
+++ b/examples/store_latency.rs
@@ -194,10 +194,10 @@ fn main() {
     write_matrix(matrix);
 
     // Results for nrf52840dk_opensk:
-    // StorageConfig { page_size: 4096, num_pages: 20 }
+    // StorageConfig { num_pages: 20 }
     // Overwrite    Length      Boot  Compaction   Insert  Remove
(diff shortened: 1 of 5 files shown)


kaczmarczyck (commented):
I think I like this a bit better than #486. Might be a personal preference, but a quick comment on the non-intuitive use of `Cow` could be nice (we don't actually care about the writing part).

I find the binary size increases of both PRs surprising (in relation, at least). Would be great if we could check somehow if it's variance, but I don't see how.

ia0:
Given that in #488 the binary size didn't increase while adding code, I'm assuming this PR is rather at the bottom of the current slot of added size.

kaczmarczyck (approved):

PR #488, opened 2022-06-03
Test whether supporting passing an optional mutable buffer to Storage::read_slice has drawbacks (it does)

ia0:
Adds optional buffer on top of `Cow` trick.


ia0:
We don't need this. It's the same binary size but it's slower. So closing the PR.

PR #489, opened 2022-06-05
Hostenv file storage

egor-duda:
Part of implementation for #485

This PR implements a storage, backed by file on host OS, for persisting authenticator data between restarts of OpenSK application used for FIDO2 emulation for QEMU.

After changes to storage API there's no more need for double buffering of storage data.

Current `read_slice()` method signature declares storage object as immutable. This conflicts with File object being mutated by read operations (current file position is changed for and after read). To keep `read_slice()` signature unchanged, interior mutability via `RefCell` is used.

### DIFF
--- a/libraries/persistent_store/Cargo.toml
+++ b/libraries/persistent_store/Cargo.toml
@@ -12,4 +12,3 @@ tempfile = "3"
 
 [features]
 std = []
-hostenv = ["std"]
--- a/libraries/persistent_store/Cargo.toml
+++ b/libraries/persistent_store/Cargo.toml
@@ -7,5 +7,9 @@ edition = "2018"
 
 [dependencies]
 
+[dev-dependencies]
+tempfile = "3"
+
 [features]
 std = []
+hostenv = ["std"]
--- a/libraries/persistent_store/src/file.rs
+++ b/libraries/persistent_store/src/file.rs
@@ -98,11 +98,15 @@ impl Storage for FileStorage {
     }
 
     fn max_word_writes(&self) -> usize {
-        usize::MAX
+        // We can write an unlimited amount of times in a file, but the store arithmetic
+        // uses `Nat` so the value should fit in a `Nat`.
+        u32::MAX as usize
     }
 
     fn max_page_erases(&self) -> usize {
-        usize::MAX
+        // We can "erase" an unlimited amount of times in a file, but the store format
+        // encodes the number of erase cycles on 16 bits.
+        u16::MAX as usize
     }
 
     fn read_slice(&self, index: StorageIndex, length: usize) -> StorageResult<Cow<[u8]>> {
--- a/libraries/persistent_store/src/file.rs
+++ b/libraries/persistent_store/src/file.rs
@@ -68,7 +68,7 @@ impl FileStorage {
 
         if file_len == 0 {
             file.seek(SeekFrom::Start(0))?;
-            let buf = vec![0xffu8; options.page_size];
+            let buf = vec![0xff; options.page_size];
             for _ in 0..options.num_pages {
                 file.write(&buf)?;
             }
@@ -128,7 +128,7 @@ impl Storage for FileStorage {
         file.seek(SeekFrom::Start(
             index.range(self.page_size(), self)?.start as u64,
         ))?;
-        file.write_all(&vec![0xffu8; self.page_size()][..])?;
+        file.write_all(&vec![0xff; self.page_size()][..])?;
         Ok(())
     }
 }
--- a/libraries/persistent_store/src/file.rs
+++ b/libraries/persistent_store/src/file.rs
@@ -17,23 +17,26 @@
 //! [`FileStorage`] implements the flash [`Storage`] interface but doesn't interface with an
 //! actual flash storage. Instead it uses a host-based file to persist the storage state.
 
-use crate::{BufferOptions, BufferStorage, Storage, StorageIndex, StorageResult};
+use crate::{BufferOptions, Storage, StorageIndex, StorageResult};
+use alloc::borrow::Cow;
+use core::cell::RefCell;
 use std::fs::{File, OpenOptions};
 use std::io::{Read, Seek, SeekFrom, Write};
 use std::path::Path;
 
 /// Simulates a flash storage using a host-based file.
 ///
-/// It provides same functions as BufferStorage for testing, but also saves stored
-/// data between application restarts.
-///
-/// Metadata, such as word write and page erase counters are not saved between restarts.
-///
+/// This is usable for emulating authenticator hardware on VM hypervisor's host OS
 pub struct FileStorage {
-    /// Content of the storage.
-    storage: BufferStorage,
-    /// File to persist contents of the storage.
-    backing_file: File,
+    // Options of the storage
+    buffer_options: BufferOptions,
+
+    /// File for persisting contents of the storage
+    /// Reading data from File requires mutable reference, as seeking and reading data
+    /// changes file's current position.
+    /// All operations on backing file internally always first seek to needed position,
+    /// so it's safe to borrow mutable reference to backing file for the time of operation.
+    backing_file_ref: RefCell<File>,
 }
 
 const PAGE_SIZE: usize = 0x1000;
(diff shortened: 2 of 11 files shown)


libraries/persistent_store/Cargo.toml
```diff
@@ -7,5 +7,9 @@ edition = "2018"
 
 [dependencies]
 
+[dev-dependencies]
+tempfile = "3"
+
 [features]
 std = []
+hostenv = ["std"]
```
ia0: Let's use `std` directly. We shouldn't worry about compile-time or binary size for `std`. Let's have all features.

libraries/persistent_store/src/storage.rs
```diff
@@ -34,6 +34,18 @@ pub enum StorageError {
 
     /// Implementation-specific error.
     CustomError,
+
+    // I/O error
+    #[cfg(feature = "hostenv")]
+    IOError,
```
ia0: Let's use `CustomError` instead. It's the error for implementation-specific errors, like IO errors in this case.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,179 @@
...
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Seek, SeekFrom, Write};
+use std::path::Path;
+
+/// Simulates a flash storage using a host-based file.
+///
+/// This is usable for emulating authenticator hardware on VM hypervisor's host OS
+pub struct FileStorage {
+    // Options of the storage
+    buffer_options: BufferOptions,
```
ia0: Let's define our own `FileOptions` with only `word_size` and `page_size`. The rest is only useful for tests. Let's also add `num_pages`.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// File for persisting contents of the storage
+    /// Reading data from File requires mutable reference, as seeking and reading data
+    /// changes file's current position.
+    /// All operations on backing file internally always first seek to needed position,
+    /// so it's safe to borrow mutable reference to backing file for the time of operation.
+    backing_file_ref: RefCell<File>,
+}
+
+const PAGE_SIZE: usize = 0x1000;
+const NUM_PAGES: usize = 20;
```
ia0: Those should not be constants. The `FileStorage::new` function should take a `FileOptions`.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,179 @@
...
+    fn page_size(&self) -> usize {
+        self.buffer_options.page_size
+    }
+
+    fn num_pages(&self) -> usize {
+        NUM_PAGES
+    }
+
+    fn max_word_writes(&self) -> usize {
+        self.buffer_options.max_word_writes
```
ia0: Let's use usize::MAX.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    fn max_page_erases(&self) -> usize {
+        self.buffer_options.max_page_erases
```
ia0: Let's use usize::MAX.

ia0 (commented):
This overall looks good to me. I made a first pass of high-level comments. I left the more style-related comments for later.

libraries/persistent_store/Cargo.toml
egor-duda: Done

libraries/persistent_store/src/storage.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/lib.rs
```diff
@@ -381,6 +383,8 @@ pub use self::driver::{
     StoreDriver, StoreDriverOff, StoreDriverOn, StoreInterruption, StoreInvariant,
 };
 #[cfg(feature = "std")]
+pub use self::file::FileStorage;
```
ia0: We should also add FileOptions here.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+use crate::{Storage, StorageIndex, StorageResult};
+use alloc::borrow::Cow;
+use core::cell::RefCell;
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Seek, SeekFrom, Write};
+use std::path::Path;
+
+/// Simulates a flash storage using a host-based file.
+///
+/// This is usable for emulating authenticator hardware on VM hypervisor's host OS
```
ia0: nit: dot at the end of the sentence

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+pub struct FileStorage {
+    // Options of the storage
```
ia0: nit: dot at the end of the documentation title

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+    options: FileOptions,
+
+    /// File for persisting contents of the storage
+    /// Reading data from File requires mutable reference, as seeking and reading data
+    /// changes file's current position.
+    /// All operations on backing file internally always first seek to needed position,
+    /// so it's safe to borrow mutable reference to backing file for the time of operation.
```
ia0, line 40: Reformat the documentation as:

    /// File for persisting contents of the storage.
    ///
    /// Reading data from File requires mutable reference, as seeking and reading data
    /// changes file's current position.
    ///
    /// All operations on backing file internally always first seek to needed position,
    /// so it's safe to borrow mutable reference to backing file for the time of operation.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+    backing_file_ref: RefCell<File>,
```
ia0: nit: This could simply be named backing_file or even just file.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Options for file-backed storage
+pub struct FileOptions {
+    /// Size of a word in bytes.
+    pub word_size: usize,
+
+    /// Size of a page in bytes.
+    pub page_size: usize,
+
+    /// Number of pages in storage
```
ia0: nit: dot at the end of the title

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    /// File for persisting contents of the storage
+    /// Reading data from File requires mutable reference, as seeking and reading data
+    /// changes file's current position.
+    /// All operations on backing file internally always first seek to needed position,
+    /// so it's safe to borrow mutable reference to backing file for the time of operation.
+    backing_file_ref: RefCell<File>,
+}
+
+/// Options for file-backed storage
```
ia0: nit: dot at the end of the title

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    fn max_page_erases(&self) -> usize {
+        usize::MAX
+    }
+
+    fn read_slice(&self, index: StorageIndex, length: usize) -> StorageResult<Cow<[u8]>> {
+        let mut backing_file = self.backing_file_ref.borrow_mut();
+        backing_file.seek(SeekFrom::Start(
+            (index.page * self.page_size() + index.byte) as u64,
```
ia0: We should call index.range(length, self)?.start to test the slice is valid.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+        ))?;
+        let mut buf = vec![0u8; length];
+        backing_file.read_exact(&mut buf)?;
+        Ok(Cow::Owned(buf))
+    }
+
+    fn write_slice(&mut self, index: StorageIndex, value: &[u8]) -> StorageResult<()> {
+        let mut backing_file = self.backing_file_ref.borrow_mut();
```
ia0: Same here, we should use index.range.

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,186 @@
...
+        backing_file.seek(SeekFrom::Start(
+            (index.page * self.page_size() + index.byte) as u64,
+        ))?;
+        backing_file.write_all(value)?;
+        Ok(())
+    }
+
+    fn erase_page(&mut self, page: usize) -> StorageResult<()> {
+        let mut backing_file = self.backing_file_ref.borrow_mut();
+        backing_file.seek(SeekFrom::Start((page * self.page_size()) as u64))?;
```
ia0: We should check page < num_pages.

libraries/persistent_store/src/lib.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

libraries/persistent_store/src/file.rs
egor-duda: Done

ia0 (approved):
Thanks for the PR!

libraries/persistent_store/src/file.rs
```diff
@@ -0,0 +1,189 @@
...
+                .create(true)
+                .open(path)?,
+        );
+        let file = file_ref.get_mut();
+        let file_len = file.metadata()?.len();
+        let store_len: u64 = (options.page_size * options.num_pages) as u64;
+
+        if file_len == 0 {
+            file.seek(SeekFrom::Start(0))?;
+            let buf = vec![0xffu8; options.page_size];
```
kaczmarczyck: We have generally used `0xFF`. Is the `u8` annotation is required?

Other instances of this exist below.

libraries/persistent_store/src/lib.rs
```diff
@@ -365,6 +365,8 @@ extern crate alloc;
 mod buffer;
 #[cfg(feature = "std")]
 mod driver;
+#[cfg(feature = "std")]
```
kaczmarczyck, line 368: Does this `std` condition correctly interact with the rest of the code? We use `std` for tests and fuzzing only so far. Therefore, I was wondering if you can just compile the OpenSK library with `std` later for non-testing purposes?

libraries/persistent_store/src/storage.rs
```diff
@@ -36,6 +36,13 @@ pub enum StorageError {
     CustomError,
 }
 
+#[cfg(feature = "std")]
+impl From<std::io::Error> for StorageError {
```
kaczmarczyck, line 40: Same here, another use of `std` that is actually necessary outside of tests.

kaczmarczyck:
I resolved all of @ia0 's comments since approved. I don't mind merging, just had a nit and an open question about `std``, and if you already tried using this PR in your QEMU project and it all worked.

libraries/persistent_store/src/file.rs
ia0: Right, for consistency with the rest of the library let's use `0xff` if Rust is able to infer the type.

libraries/persistent_store/src/lib.rs
ia0: Yes, `std` means that we compile with the standard library (and thus an OS). We use `std` for tests and fuzzing because that's the only reason we need to compile on an OS. And this example is another reason now.

Note that once CTAP is a library we will compile it with `std` for non-testing purposes too. This library is just one step ahead because it's already a library and doesn't depend on Tock.

ia0 (dismissed):
Not merging to first get a confirmation that this PR solves the problem of running with QEMU.

egor-duda:
Regarding std, I've tested my prototype code:

I run `cargo run --features=std` on linux host system and `ssh-keygen -t ecdsa-sk` on QEMU guest, then restarted qemu virtual fido2 device and opensk application and connected via ssh using generated ecdsa-sk key. Everything worked fine.

However, I've encountered two problems: 
1. Current implementations of Store tries to convert `max_page_erases` and `max_word_writes` to `Nat` type, which fails in current implementation of this PR on 64-bit platform, trying to convert usize::MAX to u32
2. `Format::is_storage_supported` function disallows storages with max_page_erases > 65535

Should I change FileStorage::max_word_writes and max_page_erases to return Nat::MAX and 65535 respectively, or change functions in Format to accommodate current FileStorage max values?

libraries/persistent_store/src/file.rs
egor-duda: Done

ia0:
> Should I change FileStorage::max_word_writes and max_page_erases to return Nat::MAX and 65535 respectively, or change functions in Format to accommodate current FileStorage max values?

Oops my bad, I forgot about that. Yes, your suggestion sounds good. Let's also document for each of them why we use those values because I'm pretty I'm gonna forget again.

```rust
fn max_word_writes(&self) -> usize {
    // We can write an unlimited amount of times in a file, but the store arithmetic
    // uses `Nat` so the value should fit in a `Nat`.
    Nat::MAX as usize
}

fn max_page_erases(&self) -> usize {
    // We can "erase" an unlimited amount of times in a file, but the store format
    // encodes the number of erase cycles on 16 bits.
    u16::MAX as usize
}
```

ia0 (approved):

PR #490, opened 2022-06-06
Connect Vendor HID interface between USB driver and CTAP app

liamjm:
This follows on from work in #446 (introduce vendor HID in app) and #472 (creating the 2nd USB interface) to connect the changes together.
Summary:
* Tock USB capsule (capsules/src/usb/usb_ctap.rs) now passes the endpoint number from the driver to the syscall interface.
* Tock Ctap capsule (capsules/src/usb/usbc_ctap_hid.rs) has in and out buffers for each endpoint and configures the enables the endpoints in the USB driver for the 2nd interface
* Libtock driver now sends / receives the endpoint number through the syscall interface and maps it to the UsbInterface enum.


patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,304 @@
...
++    fn command(&self, cmd_num: usize, endpoint: usize, _arg2: usize, appid: AppId) -> ReturnCode {
+         match cmd_num {
+             CTAP_CMD_CHECK => ReturnCode::SUCCESS,
+             CTAP_CMD_CONNECT => {
+@@ -258,7 +258,7 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> Driver for CtapUsbSyscallDriver<'a,
+                                 ReturnCode::EALREADY
+                             } else if self
+                                 .usb_client
+-                                .transmit_packet(app.buffer.as_ref().unwrap().as_ref())
++                                .transmit_packet(app.buffer.as_ref().unwrap().as_ref(), endpoint)
```
ia0: Shouldn't we check that the endpoint is within bounds?

patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,304 @@
...
+                             {
+                                 app.waiting = true;
+                                 ReturnCode::SUCCESS
+@@ -282,7 +282,7 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> Driver for CtapUsbSyscallDriver<'a,
+                                 ReturnCode::EALREADY
+                             } else {
+                                 app.waiting = true;
+-                                self.usb_client.receive_packet();
++                                self.usb_client.receive_packet(endpoint);
```
ia0: Same here (endpoint validation).

patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,304 @@
...
+     // Send an OUT packet available in the controller back to the client.
+     // This returns false if the client is not ready to receive a packet, and true if the client
+     // successfully accepted the packet.
+-    fn send_packet_to_client(&'a self) -> bool {
++    fn send_packet_to_client(&'a self, endpoint: usize) -> bool {
+         // Copy the packet into a buffer to send to the client.
+         let mut buf: [u8; 64] = [0; 64];
+-        for (i, x) in self.out_buffer.buf.iter().enumerate() {
++        if endpoint == 0 || endpoint > NUM_ENDPOINTS {
++            return false
```
ia0: nit: let's put a semi-colon here

patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,304 @@
...
++        // Set up the vendor in-out endpoint
++        #[cfg(feature = "vendor_hid")]
++        {
++            self.controller()
++                .endpoint_set_in_buffer(VENDOR_ENDPOINT_NUM, &self.in_buffer[1].buf);
++            self.controller()
++                .endpoint_set_out_buffer(VENDOR_ENDPOINT_NUM, &self.out_buffer[1].buf);
++            self.controller()
++                .endpoint_in_out_enable(TransferType::Interrupt, VENDOR_ENDPOINT_NUM);
++        }
```
ia0: Shouldn't this be a loop from 1 to NUM_ENDPOINTS?

patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,304 @@
...
+ 
+     fn attach(&'a self) {
+@@ -383,12 +414,11 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> hil::usb::Client<'a> for ClientCtap
+         match transfer_type {
+             TransferType::Bulk => hil::usb::InResult::Error,
+             TransferType::Interrupt => {
+-                if endpoint != 1 {
+-                    return hil::usb::InResult::Error;
++                if endpoint == 0 || endpoint > NUM_ENDPOINTS {
++                  return hil::usb::InResult::Error
```
ia0: nit: semi-colon (we don't need the fact that return has type `!` and should be explicit about `()`)

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -302,9 +311,15 @@ fn send_or_recv_with_timeout_detail(
         Ok(x) => x,
         Err(_) => return Some(SendOrRecvStatus::Error),
     };
+    let endpoint: usize = match interface {
+        UsbInterface::MainHid => 1,
+        #[cfg(feature = "vendor_hid")]
+        UsbInterface::VendorHid => 2,
```
ia0: To avoid code duplication (and in particular having too many invariants), we should probably have functions to convert from interface to endpoint:

```rust
impl UsbInterface {
    fn from_endpoint(x: usize) -> Option<Self> { match ... }
    fn to_endpoint(self) -> usize { match ... }
```

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Done.

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Done

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Indeed. Fixed.

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Done.

third_party/libtock-drivers/src/usb_ctap_hid.rs
liamjm: Good idea. Done.

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Done.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -71,6 +71,24 @@ pub enum UsbInterface {
     VendorHid = 1,
```
kaczmarczyck: I set the wrong numbers here. They are supposed to be 1 and 2, respectively. Maybe we should even rename the enum to `UsbEndpoint`.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -71,6 +71,24 @@ pub enum UsbInterface {
...
+impl UsbInterface {
+  fn from_endpoint(endpoint: usize) -> Option<Self> {
+    match endpoint {
+      1 => Some(UsbInterface::MainHid),
+      #[cfg(feature = "vendor_hid")]
+      2 => Some(UsbInterface::VendorHid),
+      _ => None,
+    }
+  }
+  fn to_endpoint(self) -> usize {
```
kaczmarczyck: With the above numbers correctly set, you can remove this function and instead just call:
`as usize`

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -71,6 +71,24 @@ pub enum UsbInterface {
     VendorHid = 1,
 }
 
+impl UsbInterface {
+  fn from_endpoint(endpoint: usize) -> Option<Self> {
```
kaczmarczyck: Likewise, this could be standard `try_from(usize)` if `SendOrRecvStatus` was an actual `Result`. But that looks like a bigger change and not sure if it's clearer, so this is fine with me.

patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,349 @@
...
++
++// Converts from an endpoint to the index into the in and out buffers.
++fn endpoint_to_buffer(endpoint: usize) -> Option<usize> {
++  for (i, ep) in ENDPOINTS.iter().enumerate() {
++    if endpoint == *ep {
++      return Some(i);
++    }
++  }
++  None
++}
```
ia0: This looks quite complicated. Why don't we have just 2 functions:

```
fn endpoint_number_to_index(x: usize) -> Option<usize> {
    Some(match x { 1 => 0, #[cfg] 2 => 1, _ => return None, })
}
fn endpoint_index_to_number(x: usize) -> Option<usize> { ... }
```

patches/tock/10-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,349 @@
...
+ 
+-    pub fn transmit_packet(&'a self, packet: &[u8]) -> bool {
++    pub fn transmit_packet(&'a self, packet: &[u8], endpoint: usize) -> bool {
+         if self.pending_in.get() {
+             // The previous packet has not yet been transmitted, reject the new one.
+             false
+         } else {
++            if endpoint_to_buffer(endpoint).is_none() {
++                // Unsupported endpoint
++                return false;
```
ia0: This is not the best error. The user will get `EBUSY` instead of `EINVAL`. I would have just checked `endpoint` directly in the command (which means exposing a function to know the number of endpoints). But if that's too complicated, I guess this version is acceptable.

third_party/libtock-drivers/src/usb_ctap_hid.rs
ia0: I would not be against keeping the function and just using `as usize` in its implementation. I think the name of the function is good for documentation (unless `UsbInterface` is really just `UsbEndpoint`).

third_party/libtock-drivers/src/usb_ctap_hid.rs
liamjm: I've updated the enum numbers.

So USB has 1:many for interfaces: endpoints. So we could keep the endpoint terminology for more of the code. However, at some point, the code makes the 1:1 relationship between interfaces:endpoints. So may as well do it now.
So the name is OK at the moment.

third_party/libtock-drivers/src/usb_ctap_hid.rs
liamjm: Ignored this as it is a bigger changed.

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Changed the return code to be the status code we want, rather than a bool.

patches/tock/10-connect-vendor-hid-usb-interface.patch
liamjm: Out of date with new code, so not relevant.

kaczmarczyck:
Thanks for the work on `third_party/libtock-drivers/src/usb_ctap_hid.rs`, I took that and changed it a bit to match #497. I tested my changes and they worked together with your `patches/tock/10-connect-vendor-hid-usb-interface.patch` (before #494, at least). Now this PR can be submitted once the USB driver is reliable enough. I can test more after the merge/rebase.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -329,7 +330,7 @@ fn send_or_recv_with_timeout_detail(
         #[cfg(feature = "verbose_usb")]
         writeln!(Console::new(), "Cancelling USB transaction due to timeout").unwrap();
         let result_code =
-            unsafe { syscalls::raw::command(DRIVER_NUMBER, command_nr::CANCEL, 0, 0) };
+            unsafe { syscalls::raw::command(DRIVER_NUMBER, command_nr::CANCEL, endpoint as usize, 0) };
```
kaczmarczyck: Why does `CANCEL` need an endpoint? There are 2 invocations in this file, and one gets an endpoint, the other doesn't. So I guess we don't need that here.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -231,8 +231,9 @@ fn recv_with_timeout_detail(
     if matches!(status, Ok(SendOrRecvStatus::Timeout)) {
         #[cfg(feature = "verbose_usb")]
         writeln!(Console::new(), "Cancelling USB receive due to timeout").unwrap();
-        let result_code =
-            unsafe { syscalls::raw::command(DRIVER_NUMBER, command_nr::CANCEL, 0, 0) };
+        let result_code = unsafe {
+            syscalls::raw::command(DRIVER_NUMBER, command_nr::CANCEL, 0, 0)
+        };
```
kaczmarczyck: Do you have a different version of `cargo fmt`? We shouldn't have a diff here?

Also, this is the endpoint-free call to `CANCEL`, see the other comment.

patches/tock/12-connect-vendor-hid-usb-interface.patch
```diff
@@ -0,0 +1,516 @@
...
+-    fn app_packet_received(&self, packet: &[u8; 64], app: &mut App) {
++    fn app_packet_received(&self, packet: &[u8; 64], endpoint: usize, app: &mut App) {
+         if app.connected && app.waiting && app.side.map_or(false, |side| side.can_receive()) {
+             if let Some(buf) = &mut app.buffer {
+                 // Copy the packet to the app's allowed buffer.
+@@ -57,7 +56,7 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> CtapUsbSyscallDriver<'a, 'b, C> {
+                 app.waiting = false;
+                 // Signal to the app that a packet is ready.
+                 app.callback
+-                    .map(|mut cb| cb.schedule(CTAP_CALLBACK_RECEIVED, 1, 0));
```
kaczmarczyck: You can delete patch 11, it was supposed to be superseded by this PR, and its only change is overwritten here.

kaczmarczyck (commented):
Tagging @ia0 for re-review, in case you want to before we submit?

Otherwise minor comments only, I tested this PR and it works!

ia0 (dismissed):

third_party/libtock-drivers/src/usb_ctap_hid.rs
liamjm: In this flow we have the endpoint that is being used so I thought it best to only cancel that one. Without sending the endpoint number, all endpoints have to be cancelled.
But if you think it is fine to always cancel I can do this. LMK

third_party/libtock-drivers/src/usb_ctap_hid.rs
liamjm: My mistake, I manually changed the wrong call to cancel.

patches/tock/12-connect-vendor-hid-usb-interface.patch
liamjm: Done.

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: Cancel per endpoint is fine. I thought this must be unintentional that the two `CANCEL` behave differently, but makes sense. We shouldn't need to cancel the other endpoint really.

kaczmarczyck (approved):

PR #491, opened 2022-06-07
Fix pylint configuration, script and matcher

jmichelp:
I also fixed the few warnings that we couldn't see before because the matcher was obsolete...

### DIFF
--- a/fuzz/make_corpus.py
+++ b/fuzz/make_corpus.py
@@ -19,6 +19,7 @@ import argparse
 import json
 import os.path
 
+
 # Creates a corpus directory to the given path from the given json file.
 def make_corpus(corpus_dir, corpus_json):
   if not os.path.exists(corpus_dir):
--- a/fuzz/make_corpus.py
+++ b/fuzz/make_corpus.py
@@ -1,24 +1,23 @@
+"""Creates a directory containing seed inputs from a json file having
+   the following structure:
+   [
+     {
+       "hex": "a901a1182a182a02a3626964781a6d616b655f6261645f7...",
+       "cbor": "{1: h'42', 2: {\"id\": \"make.example.com\", ...",
+       "description": "make credential parameters"
+     },
+     ...
+   ]
+
+  Usage:
+    - pass the resulting corpus directory path as the first argument
+    - pass the json file path to make the corpus from as the second argument
+  Example:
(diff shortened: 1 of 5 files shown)


tools/configure.py
```diff
@@ -14,6 +14,8 @@
 # limitations under the License.
 # Lint as: python3
 
+"""Tools that implements vendor-specific CTAP2 commands to configure OpenSK.
+"""
```
kaczmarczyck: Is this one-line doc supposed to have it's closing """ on the second line?

Also: Tools that implements -> Tools that implement

tools/deploy_partition.py
```diff
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # Lint as: python3
+"""Tool that is part of firmware upgrabability in OpenSK.
+"""
```
kaczmarczyck: Same as before, shoudln't this be all one one line?

kaczmarczyck (dismissed):

tools/configure.py
jmichelp: this would exceed the 80 column limit in this case.
At least neither yapf nor pylint are complaining about it.

tools/deploy_partition.py
jmichelp: Fixed.

tools/configure.py
kaczmarczyck: I thought that might be the case, but the limit must be slightly smaller, because:

```
$ len('"""Tools that implement vendor-specific CTAP2 commands to configure OpenSK."""')
78
```

fuzz/make_corpus.py
```diff
@@ -19,6 +19,7 @@
 import json
 import os.path
 
+
```
kaczmarczyck, line 22: What did we change that this was not linted before?

kaczmarczyck (dismissed):

fuzz/make_corpus.py
jmichelp: yapf was complaining about a missing line (code style requires 2 empty lines at the end of the import block)

tools/configure.py
jmichelp: Fixed. I was visually evaluating, not actually counting :)

kaczmarczyck (approved):

PR #495, opened 2022-06-10
Use VENDOR_HID when creating USB config when Vendor HID enabled

liamjm:
The correct config was being used in the handling of get_descriptor requests, but not in the initial config descriptor.


liamjm:
Blah... This should be a 1 line diff, but some noise got introduced. 
Let me fix that.

kaczmarczyck (approved):

PR #497, opened 2022-06-10
[RFC] Moving I/O logic to Env

egor-duda:
First draft for implementing part of #485 -- moving I/O logic to Env, and moving CTAP Keepalive handler to ctap library.

Tested with both nrf52840_dongle_dfu and running QEMU from https://github.com/egor-duda/qemu/tree/fido2
with connection to socket created with `socat EXEC:'"cargo run --features=std"' UNIX-LISTEN:$HOME/tmp/fido2.sock`

This draft is definitely not ready for merge, and need more work fixing FIXMEs, but I would like to get comments if this approach is right in general.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -51,7 +51,7 @@ use self::data_formats::{
     PublicKeyCredentialSource, PublicKeyCredentialType, PublicKeyCredentialUserEntity,
     SignatureAlgorithm,
 };
-use self::hid::ChannelID;
+use self::hid::{ChannelID, CtapHid, CtapHidCommand, KeepaliveStatus, ProcessedPacket};
 use self::large_blobs::LargeBlobs;
 use self::response::{
     AuthenticatorGetAssertionResponse, AuthenticatorGetInfoResponse,
@@ -66,7 +66,7 @@ use crate::api::customization::Customization;
 use crate::api::firmware_protection::FirmwareProtection;
 use crate::api::upgrade_storage::UpgradeStorage;
 use crate::clock::{ClockInt, CtapInstant};
-use crate::env::{Env, UserPresence};
+use crate::env::{Env, IOChannel, SendOrRecvStatus, UserPresence};
 use alloc::boxed::Box;
 use alloc::string::{String, ToString};
 use alloc::vec;
@@ -240,6 +240,69 @@ fn verify_signature(
     let public_key = ecdsa::PubKey::try_from(cose_key)?;
     if !public_key.verify_hash_vartime(signed_hash, &signature) {
         return Err(Ctap2StatusCode::CTAP2_ERR_INTEGRITY_FAILURE);
+    }
+    Ok(())
+}
+
+// Returns whether the keepalive was sent, or false if cancelled.
+pub fn send_keepalive_up_needed(
+    env: &mut impl Env,
+    channel: Channel,
+    timeout: isize,
+) -> Result<(), Ctap2StatusCode> {
+    let (transport, cid) = match channel {
+        Channel::MainHid(cid) => (Transport::MainHid, cid),
+        #[cfg(feature = "vendor_hid")]
+        Channel::VendorHid(cid) => (Transport::VendorHid, cid),
+    };
+    let keepalive_msg = CtapHid::keepalive(cid, KeepaliveStatus::UpNeeded);
+    for mut pkt in keepalive_msg {
+        let status = env
+            .io_channel()
+            .send_or_recv_with_timeout(&mut pkt, timeout, transport);
+        match status {
+            None => {
+                debug_ctap!(env, "Sending a KEEPALIVE packet timed out");
+                // TODO: abort user presence test?
+            }
+            Some(SendOrRecvStatus::Error) => panic!("Error sending KEEPALIVE packet"),
+            Some(SendOrRecvStatus::Sent) => {
+                debug_ctap!(env, "Sent KEEPALIVE packet");
+            }
+            Some(SendOrRecvStatus::Received(received_transport)) => {
+                // We only parse one packet, because we only care about CANCEL.
+                let (received_cid, processed_packet) = CtapHid::process_single_packet(&pkt);
+                if received_transport != transport || received_cid != &cid {
+                    debug_ctap!(
+                        env,
+                        "Received a packet on channel ID {:?} while sending a KEEPALIVE packet",
+                        received_cid,
+                    );
+                    return Ok(());
+                }
+                match processed_packet {
+                    ProcessedPacket::InitPacket { cmd, .. } => {
+                        if cmd == CtapHidCommand::Cancel as u8 {
+                            // We ignore the payload, we can't answer with an error code anyway.
+                            debug_ctap!(env, "User presence check cancelled");
+                            return Err(Ctap2StatusCode::CTAP2_ERR_KEEPALIVE_CANCEL);
+                        } else {
+                            debug_ctap!(
+                                env,
+                                "Discarded packet with command {} received while sending a KEEPALIVE packet",
+                                cmd,
+                            );
+                        }
+                    }
+                    ProcessedPacket::ContinuationPacket { .. } => {
+                        debug_ctap!(
+                            env,
+                            "Discarded continuation packet received while sending a KEEPALIVE packet",
+                        );
+                    }
+                }
+            }
+        }
     }
     Ok(())
 }
--- a/src/env/mod.rs
+++ b/src/env/mod.rs
@@ -2,13 +2,34 @@ use crate::api::customization::Customization;
 use crate::api::firmware_protection::FirmwareProtection;
 use crate::api::upgrade_storage::UpgradeStorage;
 use crate::ctap::status_code::Ctap2StatusCode;
-use crate::ctap::Channel;
+use crate::ctap::{Channel, Transport};
 use persistent_store::{Storage, Store};
 use rng256::Rng256;
 
 #[cfg(feature = "std")]
+pub mod host;
+#[cfg(feature = "std")]
 pub mod test;
 pub mod tock;
+
+pub enum SendOrRecvStatus {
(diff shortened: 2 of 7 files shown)


src/env/mod.rs
```diff
@@ -48,4 +70,6 @@ pub trait Env {
     fn write(&mut self) -> Self::Write;
 
     fn customization(&self) -> &Self::Customization;
+
+    fn io_channel(&mut self) -> &mut Self::IOChannel;
```
kaczmarczyck, line 74: Can we have 2 `CtapHidChannel`s here?
```
    fn main_hid_channel(&mut self) -> &mut Self::CtapHidChannel;
    #[cfg(feature = "vendor_hid")]
    fn vendor_hid_channel(&mut self) -> &mut Self::CtapHidChannel;
```

src/env/mod.rs
```diff
@@ -26,6 +47,7 @@ pub trait Env {
     type FirmwareProtection: FirmwareProtection;
     type Write: core::fmt::Write;
     type Customization: Customization;
+    type IOChannel: IOChannel;
```
kaczmarczyck, line 50: Can we rename it from `IOChannel` to `CtapHidChannel`?

src/env/mod.rs
```diff
@@ -2,14 +2,35 @@ use crate::api::customization::Customization;
...
 use persistent_store::{Storage, Store};
 use rng256::Rng256;
 
+#[cfg(feature = "std")]
+pub mod host;
 #[cfg(feature = "std")]
 pub mod test;
 pub mod tock;
 
+pub enum SendOrRecvStatus {
```
kaczmarczyck, line 15: I know this is our own code, but since you touch it anyway, it might make sense to improve! Having a `Result` here could be more idiomatic.

src/env/mod.rs
```diff
@@ -2,14 +2,35 @@ use crate::api::customization::Customization;
...
+    // FIXME: Replace Option<SendOrRecvStatus>::None with SendOrRecvStatus::TimedOut?
+    Error,
+    Sent,
+    Received(Transport),
```
kaczmarczyck, line 19: If we have one channel per HID usage page (Main vs. Vendor), we don't need to use `Transport`.

src/env/mod.rs
```diff
@@ -2,14 +2,35 @@ use crate::api::customization::Customization;
...
+
+pub trait IOChannel {
+    // FIXME: Separate type for timeout.
+    fn recv_with_timeout(&mut self, buf: &mut [u8; 64], timeout: isize)
+        -> Option<SendOrRecvStatus>;
+    fn send_or_recv_with_timeout(
+        &mut self,
+        buf: &mut [u8; 64],
+        timeout: isize,
+        transport: Transport,
```
kaczmarczyck, line 30: Then we also don't need a transport here.

src/env/tock/mod.rs
```diff
@@ -22,6 +21,52 @@ pub struct TockEnv {
...
+            usb_interface,
+        )))
+    }
+    fn send_or_recv_with_timeout(
+        &mut self,
+        buf: &mut [u8; 64],
+        timeout: isize,
+        transport: Transport,
+    ) -> Option<SendOrRecvStatus> {
+        let interface = match transport {
```
kaczmarczyck, line 54: This is why I propose not to use transport in a proper API: We will want to extend it with NFC eventually, and then repurposing the data type for switching between HID channels might be awkward.

src/env/mod.rs
```diff
@@ -2,14 +2,35 @@ use crate::api::customization::Customization;
...
+    Received(Transport),
+}
+
+pub trait IOChannel {
+    // FIXME: Separate type for timeout.
+    fn recv_with_timeout(&mut self, buf: &mut [u8; 64], timeout: isize)
+        -> Option<SendOrRecvStatus>;
+    fn send_or_recv_with_timeout(
+        &mut self,
+        buf: &mut [u8; 64],
```
kaczmarczyck, line 28: I inferred that you want all traffic for your Host env to go through [CTAPHID](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#usb-commands). Other transports would have different packet sizes. But since there is only one channel, it seems you want to reuse it.

src/ctap/mod.rs
```diff
@@ -244,6 +244,69 @@ fn verify_signature(
     Ok(())
 }
 
+// Returns whether the keepalive was sent, or false if cancelled.
+pub fn send_keepalive_up_needed(
```
kaczmarczyck, line 248: Not sure how useful it is to move this function when `check_user_presence` is still in `Env`.

src/env/mod.rs
```diff
@@ -2,14 +2,35 @@ use crate::api::customization::Customization;
...
+    // FIXME: Replace Option<SendOrRecvStatus>::None with SendOrRecvStatus::TimedOut?
+    Error,
+    Sent,
+    Received(Transport),
+}
+
+pub trait IOChannel {
+    // FIXME: Separate type for timeout.
+    fn recv_with_timeout(&mut self, buf: &mut [u8; 64], timeout: isize)
+        -> Option<SendOrRecvStatus>;
```
kaczmarczyck, line 25: We can remove receive for now, since it's not needed for `check_user_presence`. Note that by removing this function, we open ourselves to having 2 channels and not passing a `Transport`.

HID receive is special in a sense that you can NOT choose the channel you are receiving on. Maybe we will eventually have a function that receives on any channel and we can then multiplex the packets, but we don't need to make that decision yet.

kaczmarczyck (commented):
There are 2 directions this PR can go, and I'm not sure which one you intended. All comments refer to option (A), which is also my preferred:

A) IOChannel is HID specific. In this case, we don't need to make bigger design decisions, and refactor the few parts that are needed. Your trait takes 64 byte packets, so you don't consider i.e. NFC or the "internal" transport. I assume intention :)

B) IOChannel is generic over the transport protocol. In that case, We will need bigger changes or we'd have to refactor everything again next time we touch transports. For example, NFC doesn't have `check_user_presence`. Buffer sizes are wrong.

kaczmarczyck:
Also, I haven't looked at Test and Host env changes too closely (and they seem work in progress). I will do so once we converged on the purpose and API of this PR.

egor-duda:
> There are 2 directions this PR can go, and I'm not sure which one you intended. All comments refer to option (A), which is also my preferred:
> 
> A) IOChannel is HID specific. In this case, we don't need to make bigger design decisions, and refactor the few parts that are needed. Your trait takes 64 byte packets, so you don't consider i.e. NFC or the "internal" transport. I assume intention :)

I was planning to go with variant A. Potentially, it can be extended later to handle other transports as well, but for now it's for HID only.
I'm conceptualizing an IOChannel as a kind of "multiplexer", which produces a stream of received data packets along with "metadata" tags pointing to the transport the packet is coming from. Then it sends out responses, routing them via appropriate transports.

Does receiving and demultiplexing logic belong to IOChannel, or should be done outside of it, I'm not sure.

> B) IOChannel is generic over the transport protocol. In that case, We will need bigger changes or we'd have to refactor everything again next time we touch transports. For example, NFC doesn't have `check_user_presence`. Buffer sizes are wrong.

src/env/mod.rs
egor-duda: Something like this?
```
pub enum SendOrRecvStatus {
    TimedOut,
    Sent,
    Received,
}

pub struct SendOrRecvError;

pub type SendOrRecvResult = Result<SendOrRecvStatus, SendOrRecvError>;
```
Or timeout is better viewed as some kind of error?

src/ctap/mod.rs
egor-duda: Moving `check_user_presence` to ctap is supposed to be next step. Then Env would provide implement some kind of `indicate_up_needed()` and `wait_for_up_with_timeout()` functions to interface with actual hardware.
`check_user_presence` function will then call those functions in its main loop.

src/env/mod.rs
ia0: What you described looks good to me (seeing timeout as a success result).

kaczmarczyck:
I think it's cleaner if we consider main HID and vendor HID as two different transports entirely (that by accident happen to both be HID related). This holds for sending, at least. The story might be different for receiving, but we don't need that yet. That's why I found having 2 channels is simpler.

Else, we'd have too move all multiplexing logic inside the channel, and it becomes bigger in scope. And we'd have to consider all future transports in the design as well. Which includes different package sizes etc.

egor-duda:
> I think it's cleaner if we consider main HID and vendor HID as two different transports entirely (that by accident happen to both be HID related). This holds for sending, at least. The story might be different for receiving, but we don't need that yet. That's why I found having 2 channels is simpler.

I've made a branch https://github.com/egor-duda/OpenSK/tree/hostenv-iochannel-v2 
It replaces single multiplexing IOChannel with separate CtapHidChannels for each potential packet source, and changes return type for send and recv functions to Result. Tested it with both hardware dongle and virtual QEMU device. 

If this new version looks conceptually good to go, I'll apply it to this PR and start tidying up the code, tying loose ends, etc.

src/env/mod.rs
kaczmarczyck: I sent out #500 with this type:
`Result<SendOrRecvStatus, TockError>`
Timeout is added like you proposed. Compared to your idea, you need to replace the error (this or simply `()`) of course, since this trait is supposed to be independent of Tock.

And sorry for pushing the merge with #500 on you :)

src/ctap/mod.rs
kaczmarczyck: Sounds good to me!

kaczmarczyck (commented):
I checked your branch and it looks like what I wanted. So go ahead!

egor-duda:
Closed in favor of #501

PR #499, opened 2022-06-13
Fix libfido in configure

kaczmarczyck:
As reported [here](https://github.com/google/OpenSK/issues/335#issuecomment-1153642499) by @gelintonx , libfido2 renamed the `CTAP2` devices to `Ctap2`. If configure worked for you before, you only notice this bug when you call:
```
pip install fido2 --upgrade
```

I wonder if we should leave a hint for users who don't have the upgraded libfido2 yet, but pull the latest develop?

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -46,4 +46,4 @@ rustup install stable
 cargo +stable install elf2tab --version 0.7.0 --root elf2tab/
 
 # Install python dependencies to factory configure OpenSK (crypto, JTAG lockdown)
-pip3 install --user --upgrade colorama tqdm cryptography "fido2>=0.9.1"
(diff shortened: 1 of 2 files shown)


ia0:
We should change the `fido2>=0.9.1` in the `pip3 install` line in `setup.sh` accordingly.

It's sad our `setup.sh` script is not idem-potent (because of the `setup-submodules` is not), otherwise we could just tell users to rerun `setup.sh`. Also I don't know where we could show this message.

kaczmarczyck:
I think the patching makes it hard to have a nice idem-potent setup. We could split it into an install and a setup step, where install can be called as often as you like, and doesn't need to be rerun after reset.

But for this PR, I kept the changes minimal.

ia0 (approved):

PR #500, opened 2022-06-16
Adapts libtock drivers for vendor HID

kaczmarczyck:
This PR took inspiration from #490 and #497 to improve our libtock USB drivers. It adds the same functionality as #490, and uses a style that is useful to address comments in #497. This explains my intent better than comments in these PRs could :)

The patch is a dummy to be replaced with the actual kernel logic from #490. It makes packets over `MainHid` get accepted, but `VendorHid` still doesn't work end-to-end.

Instead of returning an `Option<SendOrRecvStatus>`, where `Error` was one variant, we now move to `Result<Option<SendOrRecvStatus>, TockError>`. This distinguishes the error state better from success states, and we can start using the `?` operator.

I renamed all related `interface`s to consistently be called `endpoint`s. If you like `interface` better, let me know.

The `drop` is an explicit way to make sure the return values are not dropped before we can indirectly use them. Until now, that was an implicit property of Rust, but undocumented.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -42,7 +42,6 @@ use embedded_time::duration::Milliseconds;
 use libtock_drivers::buttons::{self, ButtonState};
 #[cfg(feature = "debug_ctap")]
 use libtock_drivers::console::Console;
-#[cfg(feature = "with_ctap1")]
 use libtock_drivers::result::FlexUnwrap;
 use libtock_drivers::timer::Duration;
 use libtock_drivers::usb_ctap_hid;
@@ -89,14 +88,21 @@ fn main() {
         }
 
         let mut pkt_request = [0; 64];
-        let usb_interface =
-            match usb_ctap_hid::recv_with_timeout(&mut pkt_request, KEEPALIVE_DELAY_TOCK) {
-                Some(usb_ctap_hid::SendOrRecvStatus::Received(interface)) => {
+        let usb_endpoint =
+            match usb_ctap_hid::recv_with_timeout(&mut pkt_request, KEEPALIVE_DELAY_TOCK)
+                .flex_unwrap()
+            {
+                Some(usb_ctap_hid::SendOrRecvStatus::Received(endpoint)) => {
                     #[cfg(feature = "debug_ctap")]
                     print_packet_notice("Received packet", &clock);
-                    Some(interface)
+                    Some(endpoint)
                 }
-                Some(_) => panic!("Error receiving packet"),
+                #[cfg(feature = "debug_ctap")]
+                Some(usb_ctap_hid::SendOrRecvStatus::Sent) => {
+                    panic!("Returned transmit status on receive")
+                }
+                #[cfg(not(feature = "debug_ctap"))]
+                Some(usb_ctap_hid::SendOrRecvStatus::Sent) => panic!(),
                 None => None,
             };
 
@@ -120,20 +126,18 @@ fn main() {
         // don't cause problems with timers.
         ctap.update_timeouts(now);
 
-        if let Some(interface) = usb_interface {
-            let transport = match interface {
-                usb_ctap_hid::UsbInterface::MainHid => Transport::MainHid,
+        if let Some(endpoint) = usb_endpoint {
+            let transport = match endpoint {
+                usb_ctap_hid::UsbEndpoint::MainHid => Transport::MainHid,
                 #[cfg(feature = "vendor_hid")]
-                usb_ctap_hid::UsbInterface::VendorHid => Transport::VendorHid,
+                usb_ctap_hid::UsbEndpoint::VendorHid => Transport::VendorHid,
             };
             let reply = ctap.process_hid_packet(&pkt_request, transport, now);
             // This block handles sending packets.
             for mut pkt_reply in reply {
-                let status = usb_ctap_hid::send_or_recv_with_timeout(
-                    &mut pkt_reply,
-                    SEND_TIMEOUT,
-                    interface,
-                );
+                let status =
+                    usb_ctap_hid::send_or_recv_with_timeout(&mut pkt_reply, SEND_TIMEOUT, endpoint)
+                        .flex_unwrap();
                 match status {
                     None => {
                         #[cfg(feature = "debug_ctap")]
@@ -142,7 +146,6 @@ fn main() {
                         // Since sending the packet timed out, we cancel this reply.
                         break;
                     }
-                    Some(usb_ctap_hid::SendOrRecvStatus::Error) => panic!("Error sending packet"),
                     Some(usb_ctap_hid::SendOrRecvStatus::Sent) => {
                         #[cfg(feature = "debug_ctap")]
                         print_packet_notice("Sent packet", &clock);
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,18 +92,15 @@ fn main() {
             match usb_ctap_hid::recv_with_timeout(&mut pkt_request, KEEPALIVE_DELAY_TOCK)
                 .flex_unwrap()
             {
-                Some(usb_ctap_hid::SendOrRecvStatus::Received(endpoint)) => {
+                usb_ctap_hid::SendOrRecvStatus::Received(endpoint) => {
                     #[cfg(feature = "debug_ctap")]
                     print_packet_notice("Received packet", &clock);
                     Some(endpoint)
                 }
-                #[cfg(feature = "debug_ctap")]
-                Some(usb_ctap_hid::SendOrRecvStatus::Sent) => {
+                usb_ctap_hid::SendOrRecvStatus::Sent => {
                     panic!("Returned transmit status on receive")
                 }
-                #[cfg(not(feature = "debug_ctap"))]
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -89,14 +88,21 @@ fn main() {
...
-                    Some(interface)
+                    Some(endpoint)
+                }
+                #[cfg(feature = "debug_ctap")]
+                Some(usb_ctap_hid::SendOrRecvStatus::Sent) => {
+                    panic!("Returned transmit status on receive")
                 }
-                Some(_) => panic!("Error receiving packet"),
+                #[cfg(not(feature = "debug_ctap"))]
+                Some(usb_ctap_hid::SendOrRecvStatus::Sent) => panic!(),
```
ia0: I don't think we should address binary size at call-site. We should fix this at a higher-level. I would just pay the cost of binary size and keep the code readable. (A simple fix would be to compile with `panic-immediate-abort` to fix all panics.)

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -65,27 +66,38 @@ pub fn setup() -> bool {
...
 
 #[derive(Clone, Copy, PartialEq, Eq)]
 pub enum SendOrRecvStatus {
-    Error,
     Sent,
-    Received(UsbInterface),
+    Received(UsbEndpoint),
 }
 
+type SendOrRecvResult = Result<Option<SendOrRecvStatus>, TockError>;
```
ia0: Since `SendOrRecvStatus` is always used with `Option`, why not put the `Timeout` inside it?

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -253,61 +249,52 @@ fn recv_with_timeout_detail(
         }
     }
 
-    status.get()
+    core::mem::drop(result);
+    core::mem::drop(subscription);
+    core::mem::drop(result_code);
+    status.get().transpose()
```
ia0: We could actually only use `status: Cell<Option<SendOrRecvStatus>> = None` and set to `Some` on the first event (either recv/send or timeout). The second event shouldn't overwrite unless it's recv/send. We yield until `status.get().is_some()`. At the end we can `status.get().unwrap()`.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -361,5 +348,8 @@ fn send_or_recv_with_timeout_detail(
         writeln!(Console::new(), "Cancelled USB transaction!").unwrap();
     }
 
-    status.get()
+    core::mem::drop(result);
+    core::mem::drop(subscription);
+    core::mem::drop(result_code);
+    status.get().transpose()
```
ia0: Same here.

src/main.rs
kaczmarczyck: Done

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: Done

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: No strong preference, changed.

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: Done

ia0 (approved):

egor-duda:
After trying to flash latest develop branch after merging this PR to nrf52840_dongle, device panics after receiving first packet:

```
info: Updating rust toolchain to nightly-2021-03-25
info: syncing channel updates for 'nightly-2021-03-25-x86_64-unknown-linux-gnu'
info: checking for self-updates
info: component 'rust-std' for target 'thumbv7em-none-eabi' is up to date
info: Rust toolchain up-to-date
info: Building Tock OS for board nrf52840_dongle_dfu
    Finished release [optimized + debuginfo] target(s) in 0.02s
info: Building OpenSK application
Testing invariants in customization.rs...
   Compiling libtock_drivers v0.1.0 (/fast/deo/prog/src/OpenSK/third_party/libtock-drivers)
   Compiling rng256 v0.1.0 (/fast/deo/prog/src/OpenSK/libraries/rng256)
   Compiling lang_items v0.1.0 (/fast/deo/prog/src/OpenSK/third_party/lang-items)
   Compiling crypto v0.1.0 (/fast/deo/prog/src/OpenSK/libraries/crypto)
   Compiling ctap2 v1.0.0 (/fast/deo/prog/src/OpenSK)
    Finished test [unoptimized + debuginfo] target(s) in 2.01s
     Running unittests (target/debug/deps/ctap2-abb161d83d2e21f2)
   Compiling libtock_drivers v0.1.0 (/fast/deo/prog/src/OpenSK/third_party/libtock-drivers)
   Compiling rng256 v0.1.0 (/fast/deo/prog/src/OpenSK/libraries/rng256)
   Compiling lang_items v0.1.0 (/fast/deo/prog/src/OpenSK/third_party/lang-items)
   Compiling crypto v0.1.0 (/fast/deo/prog/src/OpenSK/libraries/crypto)
   Compiling ctap2 v1.0.0 (/fast/deo/prog/src/OpenSK)
    Finished release [optimized] target(s) in 7.45s
info: Generating Tock TAB file for application/example ctap2
info: Generating all-merged HEX file: target/nrf52840_dongle_dfu_merged.hex
info: Creating DFU package
info: Please insert the dongle and switch it to DFU mode by keeping the button pressed while inserting...
info: Press [ENTER] when ready.
info: Flashing device using DFU...
  [####################################]  100%          
Device programmed.
```
[device blinks with all LEDs]
[wait for 10 seconds]
[device continues blinking with all LEDs]
[pull out the device]
```
Traceback (most recent call last):
  File "/.../OpenSK/./deploy.py", line 1177, in <module>
    main(main_parser.parse_args())
  File "/.../OpenSK/./deploy.py", line 926, in main
    OpenSKInstaller(args).run()
  File "/.../OpenSK/./deploy.py", line 884, in run
    devices = tools.configure.get_opensk_devices(False)
  File "/.../OpenSK/tools/configure.py", line 60, in get_opensk_devices
    for dev in hid.CtapHidDevice.list_devices():
  File "/home/.../.local/lib/python3.9/site-packages/fido2/hid/__init__.py", line 258, in list_devices
    yield cls(d, open_connection(d))
  File "/home/.../.local/lib/python3.9/site-packages/fido2/hid/__init__.py", line 109, in __init__
    response = self.call(CTAPHID.INIT, nonce)
  File "/home/.../.local/lib/python3.9/site-packages/fido2/hid/__init__.py", line 185, in call
    recv = self._connection.read_packet()
  File "/home/.../.local/lib/python3.9/site-packages/fido2/hid/base.py", line 86, in read_packet
    return os.read(self.handle, self.descriptor.report_size_in)
OSError: [Errno 5] Input/output error
```

1. Clean working tree, git checkout 1d53f3c
2. Switch the board to DFU mode
3. ./deploy.py --board=nrf52840_dongle_dfu --opensk --programmer=nordicdfu

When using previous version (`git checkout 1d53f3c~`), everything works fine.

Is there any way to debug it further on the dongle version of the board, without JTAG console access?

kaczmarczyck:
I can't reproduce. I just tried both the dev board and the dongle. I ran your steps, and was capable of both calling configure and registration on webauthn.io. No blinking outside of user presence checks. Can you flash the application that resets the storage (assuming you don't have any important credentials) and the flash OpenSK again?

kaczmarczyck:
What hardware do you have, only a dongle? If you have a dev board, does it work for you?

ia0:
For reference, this is the command to erase the storage:

```
./deploy.py --board=nrf52840_dongle_dfu --erase_storage --programmer=nordicdfu
```

egor-duda:
>  Can you flash the application that resets the storage (assuming you don't have any important credentials) and the flash OpenSK again?

Erasing storage didn't help
```
./deploy.py --board=nrf52840_dongle_dfu --programmer=nordicdfu --erase_storage
```
[red LED turned off, green LED turned on]
```
./deploy.py --board=nrf52840_dongle_dfu --programmer=nordicdfu --opensk
```
[both LEDs start to flash after receiving first packet, same as before]

checked out 7e0c093, erased storage again, reflashed opensk, everything works

egor-duda:
> What hardware do you have, only a dongle? If you have a dev board, does it work for you?

Only a dongle, unfortunately

egor-duda:
Will try to add some kind of "debug LED blinking" to localize the error

egor-duda:
Found the cause, a classic case of PEBKAC
I hadn't applied 11-usb-endpoint-receive.patch, which caused Endpoint::try_from to return Err, which caused panic in main after recv_with_timeout()
Sorry for false alarm

kaczmarczyck:
The patches are not an ideal solution. There is no easy way to find out if you are up to date after a git pull, and it's not intuitive that you can't just compile. I'll think about if there is a good way to prevent this.

PR #502, opened 2022-06-21
Increase CTAP timeouts while debugging enabled

liamjm:
When debugging was enabled, the CTAP parsing often returned CtapHidError::MsgTimeout, from ctap::hid::MessageAssembler::parse_packet().

This PR makes the timeout dramatically longer when debugging is enabled. It is likely this is caused by enabling the vendor_hid interface. When packets are interleaved between these interfaces, the CTAP timeout can be expired quite easily.

### DIFF
--- a/src/ctap/hid/mod.rs
+++ b/src/ctap/hid/mod.rs
@@ -197,7 +197,10 @@ impl CtapHid {
     pub const CAPABILITY_NMSG: u8 = 0x08;
 
     // TODO: Is this timeout duration specified?
+    #[cfg(not(feature = "debug_ctap"))]
     const TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(100 as ClockInt);
+    #[cfg(feature = "debug_ctap")]
+    const TIMEOUT_DURATION: Milliseconds<ClockInt> = Milliseconds(1000 as ClockInt);
 
     /// Creates a new CTAP HID packet parser.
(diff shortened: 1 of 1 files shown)


kaczmarczyck:
This change has the implication that make debug mode less realistic. Are we working around the fact that OpenSK is slower in debug mode? Then we wouldn't need a whole second? I assume I can reproduce this with #503?

liamjm:
Correctly, without this, #503 will have some failures.

Happy to set this value to a lot lower. I didn't experiment with what value allows the tests to pass.

kaczmarczyck:
If #503 works in release mode, but fails in debug mode, then this fix makes sense! But I think it should be about making behavior similar to release, not too dramatic of a change. If you could scale it down to something roughly equivalent, that would be great.

ia0:
I agree this is not great to change behavior between debug and release, but given that timing is a behavior that is already changing, I think it makes sense. Regarding the actual timeout value, it would be nice to try to preserve behavior, i.e. stuff that doesn't timeout in release shouldn't in debug, and stuff that times out in release should also in debug. (This doesn't need to be very exact, but only mostly true.) (Note also that we shouldn't increase the duration higher than the half overflow period of the clock. But I think we're still far from it.)

liamjm:
This PR is not needed now, with more suitable fixes such as #515 and #514.

Closing.

PR #503, opened 2022-06-21
Add test script to verify Vendor HID interface

liamjm:
This script tests the vendor HID interface, ensuring that ping packets can be sent and received. Importantly it also tests this behavior while interleaving sends & receives with the main HID interface.

h/t to @kaczmarczyck for the bulk of this script.

### DIFF
--- a/tools/vendor_hid_test.py
+++ b/tools/vendor_hid_test.py
@@ -1,0 +1,191 @@
+"""These tests verify the functionality of the VendorHID interface."""
+import hid
+import time
+import unittest
+
+_OPENSK_VID = 0x1915
+_OPENSK_PID = 0x521F
+_FIDO_USAGE_PAGE = 0xF1D0
+_VENDOR_USAGE_PAGE = 0xFF00
+_PACKETS = 4
+_PACKET_SIZE = 64
+_SEND_DATA_SIZE = _PACKET_SIZE + 1
+_DEFAULT_CID = bytes([0xFF, 0xFF, 0xFF, 0xFF])
+
+def sleep():
+    time.sleep(.01)
+
+def ping_data_size(packets):
+  return 57 + 59 * (packets - 1)
+
+class HidDevice(object):
+    def __init__(self, device):
+        self.device = device
+        self.dev = None
+        self.cid = None
+        self.rx_packets = []
+        self.create_and_init()
+
+    def __del__(self):
+        if self.dev:
+          self.dev.close()
+
+
+    def create_and_init(self) -> None:
+        self.dev = hid.Device(path=self.device['path'])
+        # Nonce is all zeros, because we don't care.
+        init_packet = [0] + list(_DEFAULT_CID) + [0x86, 0x00, 0x08] + [0x00] * 57
+        if len(init_packet) != _SEND_DATA_SIZE:
+          raise Exception("Expected packet to be %d but was %d" % (_SEND_DATA_SIZE, len(init_packet)))
+        self.dev.write(bytes(init_packet))
+        self.cid = self.dev.read(_PACKET_SIZE, 2000)[15:19]
+        sleep()
+
+    def ping_init(self, packets=1, byte=0x88) -> int:
+        size = ping_data_size(packets)
+        ping_packet = [0] + list(self.cid) + [0x81, size // 256, size % 256] + [byte] * 57
+        if len(ping_packet) != _SEND_DATA_SIZE:
+          raise Exception("Expected packet to be %d but was %d" % (_PACKET_SIZE, len(ping_packet)))
+        r = self.dev.write(bytes(ping_packet))
+        sleep()
+        return r
+
+
+    def ping_continue(self, num, byte=0x88) -> int:
+        continue_packet = [0] + list(self.cid) + [num] + [byte] * 59
+        if len(continue_packet) != _SEND_DATA_SIZE:
+          raise Exception("Expected packet to be %d but was %d" % (_PACKET_SIZE, len(continue_packet)))
+        r = self.dev.write(bytes(continue_packet))
+        sleep()
+        return r
+
+    def read_and_print(self) -> int:
+        d = self.dev.read(_PACKET_SIZE, 2000)
+        self.rx_packets.append(d)
+        sleep()
+        return len(d)
+
+    def get_received_data(self):
+      """This combines the data from the received packets, to match the ping
+      packets sent."""
+      d = b""
+      if len(self.rx_packets) < _PACKETS:
+        raise Exception("Insufficent packets received - want %d, got %d" % (_PACKETS, len(self.rx_packets)))
+      d += self.rx_packets[-_PACKETS][7:]
+      for p in self.rx_packets[-_PACKETS+1:]:
+        d += p[5:]
+      return d
+
+
+def get_devices(usage_page):
+    for device in hid.enumerate(_OPENSK_VID, _OPENSK_PID):
+        if device['usage_page'] == usage_page:
+            yield device
+
+
+class VendorHid(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.fido_hid = cls.get_device(_FIDO_USAGE_PAGE)
+        cls.vendor_hid = cls.get_device(_VENDOR_USAGE_PAGE)
+
+    @classmethod
+    def get_device(cls, usage_page):
+      devices = list(get_devices(usage_page))
+      if len(devices) != 1:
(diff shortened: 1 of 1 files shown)


tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+    if self.dev:
+      self.dev.close()
+
+  def create_and_init(self) -> None:
+    self.dev = hid.Device(path=self.device['path'])
+    # Nonce is all zeros, because we don't care.
+    init_packet = [0] + list(_DEFAULT_CID) + [0x86, 0x00, 0x08] + [0x00] * 57
+    if len(init_packet) != _SEND_DATA_SIZE:
+      raise Exception(
+          f'Expected packet to be {_SEND_DATA_SIZE} but was {len(init_packet)}')
```
kaczmarczyck: All inputs here are static, right? So this should be an assert?

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+import unittest
+
+_OPENSK_VID = 0x1915
+_OPENSK_PID = 0x521F
+_FIDO_USAGE_PAGE = 0xF1D0
+_VENDOR_USAGE_PAGE = 0xFF00
+_PACKETS = 4
+_PACKET_SIZE = 64
+_SEND_DATA_SIZE = _PACKET_SIZE + 1
+_DEFAULT_CID = bytes([0xFF, 0xFF, 0xFF, 0xFF])
```
kaczmarczyck: Can we call it `_BROADCAST_CID`?

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+  return 57 + 59 * (packets - 1)
+
+
+_BYTE_VAL = 0x10
+
+
+def get_byte():
+  global _BYTE_VAL
+  _BYTE_VAL = _BYTE_VAL + 1
+  return _BYTE_VAL
```
kaczmarczyck: Since `_BYTE_VAL` is technically local to the function, we could also do this:
```
def get_byte():
  get_byte.byte_val += 1
  return get_byte.byte_val
get_byte.byte_val = 0x10
```

I wonder if that's worth the trouble. Sending a const per endpoint would also work for me.

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+    self.cid = self.dev.read(_PACKET_SIZE, 2000)[15:19]
+    sleep()
+
+  def ping_init(self, packets=1, byte=0x88) -> int:
+    size = ping_data_size(packets)
+    ping_packet = [0] + list(self.cid) + [0x81, size // 256, size % 256
+                                         ] + [byte] * 57
+    if len(ping_packet) != _SEND_DATA_SIZE:
+      raise Exception(
+          f'Expected packet to be {_SEND_DATA_SIZE} but was {len(ping_packet)}')
```
kaczmarczyck: Same question about asserts.

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+_VENDOR_USAGE_PAGE = 0xFF00
+_PACKETS = 4
+_PACKET_SIZE = 64
+_SEND_DATA_SIZE = _PACKET_SIZE + 1
+_DEFAULT_CID = bytes([0xFF, 0xFF, 0xFF, 0xFF])
+
+
+def sleep():
+  #time.sleep(.01)
+  pass
```
kaczmarczyck: If we're robust enough to get away without sleeps, feel free to remove it.

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+    ping_packet = [0] + list(self.cid) + [0x81, size // 256, size % 256
+                                         ] + [byte] * 57
+    if len(ping_packet) != _SEND_DATA_SIZE:
+      raise Exception(
+          f'Expected packet to be {_SEND_DATA_SIZE} but was {len(ping_packet)}')
+    r = self.dev.write(bytes(ping_packet))
+    sleep()
+    return r
+
+  def ping_continue(self, num, byte=0x88) -> int:
```
kaczmarczyck, line 74: Type hinting is not used consistently across the file?

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+    d = self.dev.read(_PACKET_SIZE, 2000)
+    self.rx_packets.append(d)
+    sleep()
+    return len(d)
+
+  def get_received_data(self):
+    """This combines the data from the received packets, to match the ping
+packets sent."""
+    d = b''
+    if len(self.rx_packets) < _PACKETS:
```
kaczmarczyck: This looks like test functionality inside `HidDevice`? Simply returning the received message has the advantage that tests are more flexible and can potentially test different lengths in inputs.

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+      raise Exception(f'Insufficent packets received - want {_PACKETS}'
+                      ', got {len(self.rx_packets)}')
+    d += self.rx_packets[-_PACKETS][7:]
```
kaczmarczyck: `-_PACKETS` could simply be index 0 if we consumed the returned packages. They are not accessed afterwards, so should be safe to delete after returning the message. Also means we could check if we have any unexpected extra packets. And move the number of expected packets to the outside, as indicated above.

tools/vendor_hid_test.py
```diff
@@ -0,0 +1,259 @@
...
+    return d
+
+
+def get_devices(usage_page):
+  for device in hid.enumerate(_OPENSK_VID, _OPENSK_PID):
+    if device['usage_page'] == usage_page:
+      yield device
+
+
+class VendorHid(unittest.TestCase):
```
kaczmarczyck: Why `VendorHid`? Seems to test both?

tools/vendor_hid_test.py
liamjm: Done.

tools/vendor_hid_test.py
liamjm: Update to use your suggested code.

I still see value in having this code. Using unique values per test (without having to manage this) allows for checking that the returned data is as expected.

tools/vendor_hid_test.py
liamjm: Done.

tools/vendor_hid_test.py
liamjm: Done.

tools/vendor_hid_test.py
liamjm: Done.

tools/vendor_hid_test.py
liamjm: Not ready to get rid of this just yet.

Subsequent PRs will finish this up.

tools/vendor_hid_test.py
liamjm: Fixed.

tools/vendor_hid_test.py
liamjm: Indeed. Cleaned this up to just return the packets.

tools/vendor_hid_test.py
liamjm: Renamed.

tools/vendor_hid_test.py
```diff
@@ -82,42 +86,46 @@ def read_and_print(self) -> int:
...
+  def get_received_data(self) -> bytes:
     """This combines the data from the received packets, to match the ping
 packets sent."""
     d = b''
-    if len(self.rx_packets) < _PACKETS:
-      raise Exception(f'Insufficent packets received - want {_PACKETS}'
-                      ', got {len(self.rx_packets)}')
-    d += self.rx_packets[-_PACKETS][7:]
-    for p in self.rx_packets[-_PACKETS + 1:]:
+    d += self.rx_packets.pop(0)[7:]
```
kaczmarczyck, line 93: This line implies that you can only call the function once, and it fails in unexpected ways afterwards. Either we reset here, or keep it open to be called multiple times.

kaczmarczyck (dismissed):

tools/vendor_hid_test.py
kaczmarczyck: Okay sounds good!

tools/vendor_hid_test.py
liamjm: Good point.

The code now resets after each call of this function.

kaczmarczyck (approved):

PR #504, opened 2022-06-21
Add check command to patches script

ia0:
After discussion in https://github.com/google/OpenSK/pull/500#issuecomment-1159240079 we realized that it would be nice to detect when the submodules and patches are not in sync. This PR always prints one of the following line during deploy:
- `Done: The submodules and patches are in sync.` (with green)
- `Error: The tock submodule differs from its patches.` (with red)
- same for libtock

Note that this doesn't prevent compilation because the user might be editing the submodules and it would be inconvenient to force saving the patches to compile.

### DIFF
--- a/maintainers/patches
+++ b/maintainers/patches
@@ -7,12 +7,12 @@ PROGRAM="$0"
 MODULES=(tock libtock-rs)
 
 success() {
-  echo -e "\e[1;32mDone:\e[m $1"
+  echo -e "\r\e[1;32mDone:\e[m $1"
   exit 0
 }
 
 fail() {
-  echo -e "\e[1;31mError:\e[m $1"
+  echo -e "\r\e[1;31mError:\e[m $1"
   exit 1
 }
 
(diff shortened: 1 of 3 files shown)


maintainers/patches
```diff
@@ -135,6 +146,29 @@ save() {
...
+  # Overwrite the commit function to do nothing.
+  commit() { true; }
+  for module in "${MODULES[@]}"; do
+    local root="$(get_root "${module}")"
+    ( set -e
+      cd third_party/"${module}"
+      git add .
+      git commit --allow-empty -qmx
+      local top="$(get_head)"
+      apply_module "${root}" "${module}"
```
kaczmarczyck, line 159: I wonder if we clutter our submodule's git that way? If I deploy a few times, is my git history of tock spammed with these commits?

kaczmarczyck (approved):

maintainers/patches
ia0: No, you get back to the exact state you were. However, the reflog will be cluttered. But this is usually only used to recover from an immediate mistake in which case the clutter is not visible.

I agree the design is not great, but that's essentially because we are not using submodules properly. We should have our own private forks.

kaczmarczyck: That's what I thought. reflog should be okay for now.

ia0: Actually it's not exactly the same state. I don't restore the index (just the head and the working tree). Let me do it. It can be useful if someone is adding a new patch by incrementally adding to the index between deploy.

PR #509, opened 2022-06-29
Add a key store to avoid storing secrets in the store

ia0:


src/api/key_store.rs
```diff
@@ -0,0 +1,138 @@
...
+use rng256::Rng256;
+
+use crate::env::Env;
+
+/// Provides storage for secret keys.
+///
+/// Implementations may use the environment store: [`STORE_KEY`] is reserved for this usage.
+pub trait KeyStore {
+    /// Returns the AES key for key handles encryption.
+    fn kh_encryption(&mut self) -> Result<[u8; 32], Error>;
```
kaczmarczyck: I find the `kh_*` names unintuitive. It sounds like these functions encrypt the key handle, but they instead return a key (that should be used for encrypting key handles).

Maybe add a `_key`? And remove `kh_`, but rename the trait to make it more obvious?
`credential_store.encryption_key()`
vs
`key_store.kh_encryption()`

src/api/key_store.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Generates a seed to derive an ECDSA private key.
+    fn generate_ecdsa_seed(&mut self) -> Result<[u8; 32], Error>;
+
+    /// Resets the key store.
+    fn reset(&mut self) -> Result<(), Error>;
+}
+
+/// Key store errors.
+///
+/// They are deliberately indistinguishable from each other to avoid leaking information.
```
kaczmarczyck: nit: "indistinguishable" implies "from each other" I think?

src/api/key_store.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn derive_ecdsa(&mut self, seed: &[u8; 32]) -> Result<[u8; 32], Error> {
+        match SecKey::from_bytes(seed) {
+            None => Err(Error),
+            Some(_) => Ok(*seed),
+        }
+    }
+
+    fn generate_ecdsa_seed(&mut self) -> Result<[u8; 32], Error> {
+        let mut seed = [0; 32];
+        SecKey::gensk(self.rng()).to_bytes(&mut seed);
```
kaczmarczyck, line 74: Curious about your choice of calling `SecKey::gensk` here and not just storing the seed. The function is not used here yet, but I assume we pay `derive_ecdsa` on assertion and `generate_ecdsa_seed` on make credential, and this way, the assertion is cheaper? Or is it because we don't have the API to generate from seed, and instead we need to pass `self.rng()` in?

src/api/key_store.rs
ia0: Yeah, I initially had a `_key` suffix and moved to `kh_` prefix because `_key` is very vague, it's not clear it's for key handle wrapping.

I don't like `credential_store` because that's false. This doesn't store credentials only (and it actually doesn't store any credential since the derive function is a function, it doesn't store the keys, it stores a secret to derive the keys from a seed).

Here are some naming suggestions:
- `KeyStore` -> `SecureStore` or `SecretStore` (although right now it's only use to provide keys)
- `kh_xxx` -> `key_handle_xxx_key` it's very explicit but maybe better? (I think we can drop the `_key` if we don't rename `KeyStore` since it's redundant).
- `kh_xxx` -> `xxx_key` (dropping the suffix if still using `KeyStore`) but this assumes that we only encrypt and authenticate key handles. It's maybe true? Since the rest is about signing AFAIU
- `kh_xxx` -> `master_xxx` with the implicit convention that master means key handle, which AFAICT is currently true (maybe the best option so far IMO)

kaczmarczyck: Thinking about it, I don't find anything much better than the alternatives, so I don't mind keeping it as is. The `KeyStore` alternatives are not better, and I see your point with `CredentialStore`.
`key_handle_` instead of `kh_` is more coherent with our style of being very sparse with acronyms. Your choice.

src/api/key_store.rs
ia0: Done.

src/api/key_store.rs
ia0: It's the second. What to do if the seed is not a valid ecdsa key? We could use some seeded DRBG until we find one. But since we have a RNG, we can instead only generate valid seeds. And as you say, this is also more performant (favoring assertion over registration).

src/api/key_store.rs
ia0: Yes, I also prefer `key_handle_` more. Wasn't a big fan of `kh_` but `key_handle_` seemed quite long. But since it's confusing, it's still better to use the full name.

src/api/key_store.rs
ia0: I also added the usage of derive and generate (not sure if better in this PR or separate).

src/api/key_store.rs
```diff
@@ -120,10 +120,13 @@ fn test_key_store() {
...
-    let authentication_key = key_store.kh_authentication().unwrap();
-    assert_eq!(key_store.kh_encryption(), Ok(encryption_key));
-    assert_eq!(key_store.kh_authentication(), Ok(authentication_key));
+    let encryption_key = key_store.key_handle_encryption().unwrap();
+    let authentication_key = key_store.key_handle_authentication().unwrap();
+    assert_eq!(key_store.key_handle_encryption(), Ok(encryption_key));
+    assert_eq!(
+        key_store.key_handle_authentication(),
+        Ok(authentication_key)
+    );
```
kaczmarczyck: Please move it to a test module like in other files.

src/ctap/crypto_wrapper.rs
```diff
@@ -173,17 +193,18 @@ impl PrivateKey {
...
-                key_bytes
-            }
+            PrivateKey::Ecdsa(ecdsa_seed) => ecdsa_seed.to_vec(),
             #[cfg(feature = "ed25519")]
             PrivateKey::Ed25519(ed25519_key) => ed25519_key.seed().to_vec(),
         }
     }
 }
 
+fn ecdsa_key(env: &mut impl Env, seed: &[u8; 32]) -> Option<ecdsa::SecKey> {
```
kaczmarczyck: I find the name of this function confusing, as it takes some mental effort to understand that there are 2 different `ecdsa_key` functions (one inside `PrivateKey`).

src/ctap/crypto_wrapper.rs
```diff
@@ -101,7 +101,9 @@ pub fn aes256_cbc_decrypt(
 // We shouldn't compare private keys in prod without constant-time operations.
 #[cfg_attr(test, derive(PartialEq, Eq))]
 pub enum PrivateKey {
-    Ecdsa(ecdsa::SecKey),
+    // We store the seed instead of the key since we can't get the seed back from the key. We could
+    // store both if we believe deriving the key is done more than once and costly.
+    Ecdsa([u8; 32]),
```
kaczmarczyck, line 106: Not sure I like the design now. Some of the variants are actual private keys, while others are not. And since it's an enum, this detail is not hidden from the outside. It would probably still work, but it makes it a bit more error-prone to do something like, say, cache the actual ECDSA private key. But I don't have a good idea how to do it better, so I guess it's okay?

src/ctap/crypto_wrapper.rs
```diff
@@ -173,17 +193,18 @@ impl PrivateKey {
     /// Writes the key bytes.
     pub fn to_bytes(&self) -> Vec<u8> {
         match self {
-            PrivateKey::Ecdsa(ecdsa_key) => {
-                let mut key_bytes = vec![0u8; 32];
-                ecdsa_key.to_bytes(array_mut_ref!(key_bytes, 0, 32));
-                key_bytes
-            }
+            PrivateKey::Ecdsa(ecdsa_seed) => ecdsa_seed.to_vec(),
```
kaczmarczyck, line 202: Does this affect the stored bytes, and is it a breaking change? Or is every transformation afterwards the identity?

src/ctap/crypto_wrapper.rs
```diff
@@ -143,22 +152,33 @@ impl PrivateKey {
...
         match self {
-            PrivateKey::Ecdsa(ecdsa_key) => CoseKey::from(ecdsa_key.genpk()),
+            PrivateKey::Ecdsa(seed) => ecdsa_key(env, seed),
+            #[allow(unreachable_patterns)]
+            _ => None,
+        }
+    }
+
+    /// Returns the corresponding public key.
+    pub fn get_pub_key(&self, env: &mut impl Env) -> Option<CoseKey> {
```
kaczmarczyck: I wonder if this should rather be an internal condition, that all private keys are always valid.

src/ctap/crypto_wrapper.rs
ia0: Renamed to `ecdsa_key_from_seed`. I think the public one doesn't need to be renamed because it doesn't take any argument.

src/ctap/crypto_wrapper.rs
ia0: > Some of the variants are actual private keys, while others are not.

This is all relative. There's no such notion of private key or seed. There's just some amount of information that can be more or less pre-processed. And actually, the Ed25519 private key is currently a seed. Only the name is private key. The first thing the sign function does is compute a sha512 hash out of this seed. So I'd say this is not a concern. Private key just means the secret information needed to decrypt/sign regardless of pre-computation and in particular without loss of information. I think this is why Ed25519 keeps the seed and why I need to keep the seed here. Because the derivation done on the seed is not easily inversible.

The fact that the PrivateKey has public implementation is not related to this PR. I agree it would be better if it wasn't.

src/ctap/crypto_wrapper.rs
ia0: This is not a breaking change because `TockEnv` uses the seed as the key. Tested on webauthn.io by flashing develop and this PR successively.

src/ctap/crypto_wrapper.rs
ia0: The `Option` is not because the private key would be invalid. It's because we need to use the environment, which may fail for arbitrary reasons (e.g. kernel fails in a syscall for `TockEnv`).

kaczmarczyck: Ah right, we call `ok()` here. Why not make it `Result` then?

src/ctap/crypto_wrapper.rs
kaczmarczyck: What they had in common before was that they implemented `sign` I think, but I agree to your statement. In the end, the purpose `PrivateKey` is to have something you can simply call `sign` on, and store it etc. And storing seeds should be the implementation detail.

I agree that's not related to this PR. Maybe if `PrivateKey` becomes more complicated, it will eventually need to be better encapsulated.

src/ctap/crypto_wrapper.rs
kaczmarczyck: For both resident keys and server-side?

src/ctap/crypto_wrapper.rs
ia0: But which error should we use?

ia0: But which error should we use?

src/ctap/crypto_wrapper.rs
ia0: Yes.

src/ctap/crypto_wrapper.rs
kaczmarczyck: Don't we usually use VENDOR_INTERNAL for that?

kaczmarczyck: I guess it depends on the exact problem, but one of the 2 VENDOR_ errors?

ia0: Yeah we don't know the problem. I agree Result is better, changed. (The crypto library uses Option everywhere so there's always gonna be some mismatch somewhere.)

src/api/key_store.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+/// Key store errors.
+///
+/// They are deliberately indistinguishable to avoid leaking information.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct Error;
+
+/// Key of the environment store reserved for the key store.
+pub const STORE_KEY: usize = 2046;
```
kaczmarczyck: I wonder if that's a good name. Outside of `storage/key.rs`, it's not immediately clear that this is for storage. Which is, because coincidentally, the trait above is called `KeyStore` and this is `STORE_KEY`. I think `STORAGE_KEY` is already better.

src/api/key_store.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Implements a default key store using the environment rng and store.
+pub trait Helper: Env {}
```
kaczmarczyck, line 54: Is this generally how we intend to implement common behavior between different `Env`s? Basically a mor eexplicit default next to the definition of the trait?
What if we eventually want 2 of these helpers? Can't we give it a better name, like `StorageKeyStore` (not sure if that actually is a good name though)?

kaczmarczyck (dismissed):
Generally looks good, just a few naming questions / suggestions.

src/api/key_store.rs
ia0: Done.

src/api/key_store.rs
ia0: > Is this generally how we intend to implement common behavior between different Envs?

I think that's one way to provide helpers. Another would be to just provide simpler functions to compose. I think this is to be decided on a case by case basis.

> What if we eventually want 2 of these helpers?

We would have 2 helpers. But until it happens we have no idea how they would differ, so I would not try to over-specify.

kaczmarczyck (approved):

src/api/key_store.rs
kaczmarczyck: Since it's public, I thought a more telling name would be appropriate, but no super strong opinion, so your call. Approved.

ia0: Yes, but I can't think of a better name. So I prefer to wait and see if we ever need something else, and do a breaking change (just a rename) at that point.

PR #510, opened 2022-06-30
Configure the flash as read-only at boot

ia0:
This makes code run faster without having to wait for the first mutable flash operation.


kaczmarczyck (approved):
@jmichelp This is the solution for the bug I described.

PR #511, opened 2022-07-05
Abstract attestation support

ia0:


src/api/attestation_store.rs
```diff
@@ -0,0 +1,110 @@
...
+pub struct Attestation {
+    /// ECDSA private key (big-endian).
+    pub private_key: [u8; 32],
+    pub certificate: Vec<u8>,
+}
+
+/// Stores enterprise or batch attestations.
+///
+/// Implementations don't need to distinguish different attestations. In particular, setting one
+/// attestation may set other ones.
```
ia0: We can actually remove this if for the default implementation we decide to only support batch attestation. I think it makes sense because if someone wants to customize with enterprise attestation, they would have a custom implementation anyway. What do you think?

ia0: Actually, I think we want to keep the comment. Because enterprise attestation may be shared between different RP ids. If we want to improve we could add a function to know when this happens:

```rust
/// Returns whether 2 attestation ids share the same attestation.
fn shared_attestation(&self, a: &Id, b: &Id) -> Result<bool, Error>;
```

But since we won't use it in the CTAP library, I'm not sure it's worth doing it now.

src/api/attestation_store.rs
```diff
@@ -0,0 +1,110 @@
+use alloc::string::String;
+use alloc::vec::Vec;
+use persistent_store::{StoreError, StoreUpdate};
+
+use crate::env::Env;
+
+/// Identifies an attestation.
+pub enum Id {
+    Batch,
+    Enterprise { rp_id: String },
```
kaczmarczyck: Let's remove the `rp_id`, I don't think we have use cases for different enterprise attestations at the same time.

src/ctap/ctap1.rs
```diff
@@ -257,12 +257,14 @@ impl Ctap1Command {
...
+        let Attestation {
+            private_key,
+            certificate,
+        } = env
+            .attestation_store()
+            .get(&attestation_store::Id::Batch)
             .map_err(|_| Ctap1StatusCode::SW_MEMERR)?
             .ok_or(Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?;
-        let private_key = storage::attestation_private_key(env)
-            .map_err(|_| Ctap1StatusCode::SW_INTERNAL_EXCEPTION)?
```
kaczmarczyck, line 265: Reason we move to `SW_INTERNAL_EXCEPTION` instead of `SW_MEMERR`? We could match over the actual error, if we want to preserve it?

src/ctap/mod.rs
```diff
@@ -860,7 +861,7 @@ impl CtapState {
                 key_type: PublicKeyCredentialType::PublicKey,
                 credential_id: random_id.clone(),
                 private_key: private_key.clone(),
-                rp_id,
+                rp_id: rp_id.clone(),
```
kaczmarczyck: Probably not necessary if we remove RP ID from the attestation ID.

src/ctap/status_code.rs
```diff
@@ -100,3 +100,14 @@ impl From<key_store::Error> for Ctap2StatusCode {
         Self::CTAP2_ERR_VENDOR_INTERNAL_ERROR
     }
 }
+
+impl From<attestation_store::Error> for Ctap2StatusCode {
```
kaczmarczyck, line 104: We could have a `From` for the CTAP1 errors as well, to resolve the above comment?

src/ctap/storage.rs
```diff
@@ -545,10 +493,9 @@ pub fn enterprise_attestation(env: &mut impl Env) -> Result<bool, Ctap2StatusCod
 }
 
 /// Marks enterprise attestation as enabled.
+///
+/// Doesn't check whether an attestation is setup because it depends on the RP id.
```
kaczmarczyck: Can be reverted? Would be great if this can only be enabled when the key material actually exists.

src/api/attestation_store.rs
ia0: Done.

src/ctap/mod.rs
ia0: Done.

src/ctap/storage.rs
ia0: Done.

src/ctap/status_code.rs
ia0: Good point. I did not see those errors were different.

src/ctap/ctap1.rs
ia0: Done.

src/api/attestation_store.rs
```diff
@@ -48,11 +45,17 @@ pub const STORAGE_KEYS: &[usize] = &[1, 2];
 
 /// Implements a default attestation store using the environment store.
 ///
-/// The same attestation is used for batch and enterprise.
-pub trait Helper: Env {}
+/// Supports only one attestation at a time.
+pub trait Helper: Env {
+    /// Returns the current attestation id.
+    fn attestation_id(&self) -> Id;
```
kaczmarczyck: What the idea behind this function? That's only useful if an `Env` will have exactly one attestation type? And if you enable enterprise attestation, you have to remember to also call this function here to keep the state consistent? Doesn't that make the Helper unusable when we have a setup where the incoming MakeCredential parameters decide whether we use Enterprise or Batch?

ia0: I think we decided that we don't support both batch and enterprise attestation simultaneously. If you want that, you need to implement the real trait yourself. This is just a helper. That's for simple usages, namely U2F and testing (i.e. stay backward compatible while enabling new usages).

src/api/attestation_store.rs
kaczmarczyck: This is resolved by removing the RP ID string I think.

src/ctap/mod.rs
```diff
@@ -918,21 +919,31 @@ impl CtapState {
...
-                Some(vec![attestation_certificate]),
-            )
+        let attestation_id = if env.customization().use_batch_attestation() {
+            Some(attestation_store::Id::Batch)
+        } else if ep_att {
+            Some(attestation_store::Id::Enterprise)
         } else {
-            (private_key.sign_and_encode(env, &signature_data)?, None)
+            None
+        };
```
kaczmarczyck: Just noticed that the order is wrong. Enterprise attestation has priority. For all other occasions, it depends on the security key's customization.

src/api/attestation_store.rs
kaczmarczyck: You can't use them at the same MakeCredential. But generally, the existance of enterprise attestation shouldn't affect attestation for all RP IDs that don't use enterprise attestation.

We can decide to not support batch attestation for enterprise enabled devices. That should then be a tested invariant in customization. Since individual attestation is preferred by OpenSK over batch attestation as the default customization setting, I think that's an okay shortcut to make.

src/ctap/mod.rs
ia0: Done.

src/api/attestation_store.rs
ia0: I'm not sure what you're talking about. There's 2 different things:
1. The `AttestationStore` API. This one supports batch and enterprise attestation simultaneously. That's the real API and that's what the CTAP library uses.
2. The `Helper` API. This one is just a helper to preserve the legacy behavior where only batch is supported (for U2F) except for tests where enterprise is also tested (although back then it was using the batch infrastructure since there was no separate notion of enterprise attestation).

If you're talking about 1, then I don't understand which issue you're referring about. If you're talking about 2, then I don't think we should support both simultaneously. That's just a legacy helper. Users who need fancy stuff should implement the real API.

kaczmarczyck: Talking about (2). And I generally agree, but that "user" you talk about will probably be ourselves. So I wanted to check how useful this is for future us. And suggest we include the customization invariant, because that's what our code currently can do? And a user who implements more than the existing helper can then remove that invariant check since they probably know what they are doing?

ia0: > So I wanted to check how useful this is for future us.

I don't think this is important. By definition of legacy, the `Helper` trait is for current users (or us, it's essentially the same thing). While the `AttestationStore` is for future users (or us).

> And suggest we include the customization invariant

What is this invariant?

Invariants should be specific to the CTAP library implementation, i.e. the assumptions that the CTAP library does over the environment. It shouldn't be invariants about the environnement implementations. Those are part of the tests of that environment implementation.

kaczmarczyck: Do we have a test that fails currently if we switch on EP and batch attestation?

ia0: Not sure what that means. We currently switch between EP and batch in tests. This is a necessary feature. It's not part of the `Helper` API so it has to be done by the implementer. The one for `TockEnv` doesn't while the one for `TestEnv` does.

ia0: I inlined the `Helper` trait in `TockEnv` and `TestEnv`. Resolving.

kaczmarczyck (approved):

PR #512, opened 2022-07-08
Add CBOR credential ID type

hcyang-google:
Update the format of the credential ID we generated to extend the
encrypted portion from only private_key + rp_id_hash to a flexible CBOR
map. This way we can persist more data into the key handle in the future
without need of a new version.

We add credProtectPolicy to the persisted data in this commit too, so we
can correctly check the credProtectPolicy for non-discoverable
credentials in follow-up commits.

TODOs:
- Check credProtectPolicy in next PR (or follow-up commits in this PR)
- To preserve structure and don't modify too much code, I kept the encrypt/decrypt wrapper plus the unit tests in the original place (crypto_wrapper.rs). As the implementation is mainly in credential_id.rs now I want to eventually remove the wrappers, modify the callers, and move the tests to credential_id.rs too. But we can do that afterwards.

### DIFF
--- a/src/ctap/credential_id.rs
+++ b/src/ctap/credential_id.rs
@@ -39,6 +39,7 @@ pub const MAX_CREDENTIAL_ID_SIZE: usize = 241;
 pub const MIN_CREDENTIAL_ID_SIZE: usize = 112;
 
 pub const ECDSA_CREDENTIAL_ID_VERSION: u8 = 0x01;
+#[cfg(feature = "ed25519")]
 pub const ED25519_CREDENTIAL_ID_VERSION: u8 = 0x02;
 pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x03;
 
--- a/src/ctap/credential_id.rs
+++ b/src/ctap/credential_id.rs
@@ -1,0 +1,254 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::crypto_wrapper::{aes256_cbc_decrypt, aes256_cbc_encrypt, PrivateKey};
+use super::data_formats::CredentialProtectionPolicy;
+use super::status_code::Ctap2StatusCode;
+use super::{cbor_read, cbor_write};
+use crate::api::key_store::KeyStore;
+#[cfg(feature = "ed25519")]
+use crate::ctap::data_formats::EDDSA_ALGORITHM;
+use crate::ctap::data_formats::{extract_byte_string, extract_map, ES256_ALGORITHM};
+use crate::env::Env;
+use alloc::vec::Vec;
+use core::convert::{TryFrom, TryInto};
+use crypto::hmac::{hmac_256, verify_hmac_256};
+use crypto::sha256::Sha256;
+use sk_cbor::{cbor_map_options, destructure_cbor_map};
+
+pub const LEGACY_CREDENTIAL_ID_SIZE: usize = 112;
+// CBOR credential IDs consist of
+// - 1   byte : version number
+// - 16  bytes: initialization vector for AES-256,
+// - 240 bytes: encrypted block of the key handle cbor,
+// - 32  bytes: HMAC-SHA256 over everything else.
+#[cfg(test)]
+pub const CBOR_CREDENTIAL_ID_SIZE: usize = 241;
+pub const MAX_CREDENTIAL_ID_SIZE: usize = 241;
+pub const MIN_CREDENTIAL_ID_SIZE: usize = 112;
+
+pub const ECDSA_CREDENTIAL_ID_VERSION: u8 = 0x01;
+pub const ED25519_CREDENTIAL_ID_VERSION: u8 = 0x02;
+pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x03;
+
+pub const MAX_PADDING_LENGTH: u8 = 0xBF;
+
+pub(crate) struct CredentialId {
+    pub(crate) private_key: PrivateKey,
+    pub(crate) rp_id_hash: Vec<u8>,
+    pub(crate) cred_protect_policy: Option<CredentialProtectionPolicy>,
+}
+
+fn decrypt_legacy_key_handle(
+    env: &mut impl Env,
+    bytes: &[u8],
+    algorithm: i64,
+) -> Result<Option<CredentialId>, Ctap2StatusCode> {
+    let aes_enc_key = crypto::aes256::EncryptionKey::new(&env.key_store().key_handle_encryption()?);
+    let plaintext = aes256_cbc_decrypt(&aes_enc_key, bytes, true)?;
+    if plaintext.len() != 64 {
+        return Ok(None);
+    }
+    let private_key = if let Some(key) = match algorithm {
+        ES256_ALGORITHM => PrivateKey::new_ecdsa_from_bytes(&plaintext[..32]),
+        #[cfg(feature = "ed25519")]
+        EDDSA_ALGORITHM => PrivateKey::new_ed25519_from_bytes(&plaintext[..32]),
+        _ => None,
+    } {
+        key
+    } else {
+        return Ok(None);
+    };
+    Ok(Some(CredentialId {
+        private_key,
+        rp_id_hash: plaintext[32..64].try_into().unwrap(),
+        cred_protect_policy: None,
+    }))
+}
+
+// We serialize key handles using CBOR maps. Each field of a key handle
+// is associated with a unique tag, implemented with a CBOR unsigned key.
+enum CborKeyHandleField {
+    PrivateKey = 0,
+    RpIdHash = 1,
+    CredProtectPolicy = 2,
+}
+
+impl From<CborKeyHandleField> for sk_cbor::Value {
+    fn from(field: CborKeyHandleField) -> sk_cbor::Value {
+        (field as u64).into()
+    }
+}
+
+fn decrypt_cbor_key_handle(
+    env: &mut impl Env,
+    bytes: &[u8],
+) -> Result<Option<CredentialId>, Ctap2StatusCode> {
+    let aes_enc_key = crypto::aes256::EncryptionKey::new(&env.key_store().key_handle_encryption()?);
+    let mut plaintext = aes256_cbc_decrypt(&aes_enc_key, bytes, true)?;
+    remove_padding(&mut plaintext)?;
+
+    let decoded_cbor = cbor_read(plaintext.as_slice())?;
+    destructure_cbor_map! {
+      let {
+          CborKeyHandleField::PrivateKey => private_key,
+          CborKeyHandleField::RpIdHash=> rp_id_hash,
+          CborKeyHandleField::CredProtectPolicy => cred_protect_policy,
+      } = extract_map(decoded_cbor)?;
+    }
+    Ok(match (private_key, rp_id_hash, cred_protect_policy) {
+        (Some(private_key), Some(rp_id_hash), cred_protect_policy) => {
+            let private_key = PrivateKey::try_from(private_key)?;
+            let rp_id_hash = extract_byte_string(rp_id_hash)?;
+            let cred_protect_policy = if let Some(policy) = cred_protect_policy {
+                Some(CredentialProtectionPolicy::try_from(policy)?)
+            } else {
+                None
+            };
+            Some(CredentialId {
+                private_key,
+                rp_id_hash,
+                cred_protect_policy,
+            })
+        }
+        _ => None,
+    })
+}
+
+fn is_legacy_key_handle_version(version: u8) -> bool {
+    version == ECDSA_CREDENTIAL_ID_VERSION || version == ED25519_CREDENTIAL_ID_VERSION
+}
+
+/// Pad data to MAX_PADDING_LENGTH+1 (192) bytes using PKCS padding scheme.
+/// Let N = 192 - data.len(), the PKCS padding scheme would pad N bytes of N after the data.
+fn add_padding(data: &mut Vec<u8>) -> Result<(), Ctap2StatusCode> {
+    // The data should be between 1 to MAX_PADDING_LENGTH bytes for the padding scheme to be valid.
+    if data.is_empty() || data.len() > MAX_PADDING_LENGTH as usize {
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+    }
+    let pad_length = MAX_PADDING_LENGTH - (data.len() as u8 - 1);
+    data.extend(core::iter::repeat(pad_length).take(pad_length as usize));
+    Ok(())
+}
+
+fn remove_padding(data: &mut Vec<u8>) -> Result<(), Ctap2StatusCode> {
+    if data.len() != MAX_PADDING_LENGTH as usize + 1 {
+        // This is an internal error instead of corrupted credential ID which we should just ignore because
+        // we've already checked that the HMAC matched.
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+    }
+    let pad_length = *data.last().unwrap();
+    if pad_length == 0 || pad_length > MAX_PADDING_LENGTH {
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+    }
+    data.truncate(data.len() - pad_length as usize);
+    Ok(())
+}
+
+impl CredentialId {
+    /// Encrypts the given private key, relying party ID hash, and cred protect policy into a credential ID.
+    ///
+    /// Other information, such as a user name, are not stored. Since encrypted credential IDs are
+    /// stored server-side, this information is already available (unencrypted).
+    pub(super) fn encrypt_to_bytes(
+        env: &mut impl Env,
+        private_key: &PrivateKey,
+        rp_id_hash: &[u8; 32],
+        cred_protect_policy: Option<CredentialProtectionPolicy>,
+    ) -> Result<Vec<u8>, Ctap2StatusCode> {
+        let mut payload = Vec::new();
+        let cbor = cbor_map_options! {
+          CborKeyHandleField::PrivateKey => private_key,
+          CborKeyHandleField::RpIdHash=> rp_id_hash,
+          CborKeyHandleField::CredProtectPolicy => cred_protect_policy,
+        };
+        cbor_write(cbor, &mut payload)?;
+        add_padding(&mut payload)?;
+
+        let aes_enc_key =
+            crypto::aes256::EncryptionKey::new(&env.key_store().key_handle_encryption()?);
+        let encrypted_payload = aes256_cbc_encrypt(env.rng(), &aes_enc_key, &payload, true)?;
+        let mut credential_id = encrypted_payload;
+        credential_id.insert(0, CBOR_CREDENTIAL_ID_VERSION);
+
+        let id_hmac = hmac_256::<Sha256>(
+            &env.key_store().key_handle_authentication()?,
+            &credential_id[..],
+        );
+        credential_id.extend(&id_hmac);
+        Ok(credential_id)
+    }
+
+    /// Returns None if
+    /// - the format does not match any known versions, or
+    /// - the HMAC test fails.
+    ///
+    /// For v0 (legacy U2F) the credential ID consists of:
+    /// - 16 bytes: initialization vector for AES-256,
+    /// - 32 bytes: encrypted ECDSA private key for the credential,
+    /// - 32 bytes: encrypted relying party ID hashed with SHA256,
+    /// - 32 bytes: HMAC-SHA256 over everything else.
+    ///
+    /// For v1 (ECDSA (algorithm -7)) the credential ID consists of:
+    /// -  1 byte : version number
+    /// - 16 bytes: initialization vector for AES-256,
+    /// - 32 bytes: encrypted ECDSA private key for the credential,
+    /// - 32 bytes: encrypted relying party ID hashed with SHA256,
+    /// - 32 bytes: HMAC-SHA256 over everything else.
+    ///
+    /// For v2 (EdDSA over curve Ed25519 (algorithm -8, curve 6)) the credential ID consists of:
+    /// -  1 byte : version number
+    /// - 16 bytes: initialization vector for AES-256,
+    /// - 32 bytes: encrypted Ed25519 private key for the credential,
+    /// - 32 bytes: encrypted relying party ID hashed with SHA256,
+    /// - 32 bytes: HMAC-SHA256 over everything else.
+    ///
+    /// For v3 (CBOR key handle) the credential ID consists of:
+    /// -  1 byte : version number
+    /// - 16 bytes: initialization vector for AES-256,
+    /// - 64 bytes * N: encrypted CBOR key handle
+    /// - 32 bytes: HMAC-SHA256 over everything else.
+    pub(super) fn decrypt_from_bytes(
+        env: &mut impl Env,
+        credential_id: Vec<u8>,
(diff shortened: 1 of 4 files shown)


src/ctap/crypto_wrapper.rs
```diff
@@ -302,78 +250,44 @@ pub fn encrypt_key_handle(
...
-        user_handle: vec![],
-        user_display_name: None,
-        cred_protect_policy: None,
-        creation_order: 0,
-        user_name: None,
-        user_icon: None,
-        cred_blob: None,
-        large_blob_key: None,
-    }))
+    CredentialId::decrypt_from_bytes(env, credential_id_bytes.clone()).map(|cred_id| {
```
ia0: Why not the following which avoids indentation and is more readable?

```rust
let cred_id = match CredentialId::decrypt_from_bytes(env, credential_id_bytes.clone())? {
    None => return Ok(None),
    Some(x) => x,
};
if ...
Ok(Some(...
```

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,254 @@
...
+pub const ECDSA_CREDENTIAL_ID_VERSION: u8 = 0x01;
+pub const ED25519_CREDENTIAL_ID_VERSION: u8 = 0x02;
+pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x03;
+
+pub const MAX_PADDING_LENGTH: u8 = 0xBF;
+
+pub(crate) struct CredentialId {
+    pub(crate) private_key: PrivateKey,
+    pub(crate) rp_id_hash: Vec<u8>,
+    pub(crate) cred_protect_policy: Option<CredentialProtectionPolicy>,
```
ia0: nit: Do we really need the `pub(crate)` on fields if we already have it on the struct?

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,254 @@
...
+
+pub const ECDSA_CREDENTIAL_ID_VERSION: u8 = 0x01;
+pub const ED25519_CREDENTIAL_ID_VERSION: u8 = 0x02;
+pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x03;
+
+pub const MAX_PADDING_LENGTH: u8 = 0xBF;
+
+pub(crate) struct CredentialId {
+    pub(crate) private_key: PrivateKey,
+    pub(crate) rp_id_hash: Vec<u8>,
```
ia0: To understand the rationale, we don't use `[u8; 32]` because maybe the size will change in the future? Should we use different `CborKeyHandleField`s instead? This would encode the `rp_id_hash` version into the field number. This is particularly useful if the hash technology changes but not the size, because then we don't know what kind of hash it is.

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,254 @@
...
+    if data.len() != MAX_PADDING_LENGTH as usize + 1 {
+        // This is an internal error instead of corrupted credential ID which we should just ignore because
+        // we've already checked that the HMAC matched.
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+    }
+    let pad_length = *data.last().unwrap();
+    if pad_length == 0 || pad_length > MAX_PADDING_LENGTH {
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+    }
+    data.truncate(data.len() - pad_length as usize);
```
ia0: Shouldn't we also check that all padding bytes are equal?

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,254 @@
...
+                #[cfg(feature = "ed25519")]
+                ED25519_CREDENTIAL_ID_VERSION => EDDSA_ALGORITHM,
+                _ => return Ok(None),
+            };
+            decrypt_legacy_key_handle(env, &credential_id[1..hmac_message_size], algorithm)
+        } else if credential_id[0] == CBOR_CREDENTIAL_ID_VERSION {
+            decrypt_cbor_key_handle(env, &credential_id[1..hmac_message_size])
+        } else {
+            Ok(None)
+        }
```
ia0: Looks to me that this whole diff could be rewritten more idiomatically as:

```rust
if len == LEGACY { return decrypt_legacy(); }
match credential_id[0] {
    ECDSA => decrypt_legacy(... ES256),
    ED25519 => decrypt_legacy(... EDDSA),
    CBOR => decrypt_cbor(),
    _ => Ok(None),
}
```

src/ctap/credential_id.rs
kaczmarczyck: I think we use `[u8; 32]` everywhere, so we should make the same assumption here.

src/ctap/credential_id.rs
hcyang-google: Done

src/ctap/crypto_wrapper.rs
hcyang-google: Done

src/ctap/credential_id.rs
hcyang-google: Done.

src/ctap/credential_id.rs
hcyang-google: Done

ia0 (commented):
Regarding code it looks good to me. Binary size increases a bit but not sure it should be a blocker. Regarding whether it's the direction we want to go, I don't see any issue. I'll let @kaczmarczyck take a look too.

hcyang-google:
Added credProtectPolicy check too.

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+pub const LEGACY_CREDENTIAL_ID_SIZE: usize = 112;
+// CBOR credential IDs consist of
+// - 1   byte : version number
+// - 16  bytes: initialization vector for AES-256,
+// - 240 bytes: encrypted block of the key handle cbor,
+// - 32  bytes: HMAC-SHA256 over everything else.
+#[cfg(test)]
+pub const CBOR_CREDENTIAL_ID_SIZE: usize = 241;
+pub const MAX_CREDENTIAL_ID_SIZE: usize = 241;
+pub const MIN_CREDENTIAL_ID_SIZE: usize = 112;
```
kaczmarczyck: Could also be set to `LEGACY_CREDENTIAL_ID_SIZE`, since that's where the number comes from?

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+use core::convert::{TryFrom, TryInto};
+use crypto::hmac::{hmac_256, verify_hmac_256};
+use crypto::sha256::Sha256;
+use sk_cbor::{cbor_map_options, destructure_cbor_map};
+
+pub const LEGACY_CREDENTIAL_ID_SIZE: usize = 112;
+// CBOR credential IDs consist of
+// - 1   byte : version number
+// - 16  bytes: initialization vector for AES-256,
+// - 240 bytes: encrypted block of the key handle cbor,
```
kaczmarczyck: 1 + 16 + 240 + 32 != 241
It should be
// - 192 bytes: encrypted block of the key handle cbor,
?

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+pub const MIN_CREDENTIAL_ID_SIZE: usize = 112;
+
+pub const ECDSA_CREDENTIAL_ID_VERSION: u8 = 0x01;
+#[cfg(feature = "ed25519")]
+pub const ED25519_CREDENTIAL_ID_VERSION: u8 = 0x02;
+pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x03;
+
+pub const MAX_PADDING_LENGTH: u8 = 0xBF;
+
+pub(crate) struct CredentialId {
```
kaczmarczyck: In CTAP, the credential ID is the byte sequence that the RP sees. This is a `ParsedCredentialId` or something similar? We have a data structure with a similar purpose: `PublicKeyCredentialSource`. But it contains more fields that are useful for storage, and also has the RP ID and not the RP ID hash. So probably best to simply rename `CredentialId`.

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+        EDDSA_ALGORITHM => PrivateKey::new_ed25519_from_bytes(&plaintext[..32]),
+        _ => None,
+    } {
+        key
+    } else {
+        return Ok(None);
+    };
+    Ok(Some(CredentialId {
+        private_key,
+        rp_id_hash: plaintext[32..64].try_into().unwrap(),
```
kaczmarczyck, line 82: We usually went with `*array_ref![plaintext, 32, 32]` for these conversions, but we can break our usual style if there is good reason to?

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// -  1 byte : version number
+    /// - 16 bytes: initialization vector for AES-256,
+    /// - 32 bytes: encrypted Ed25519 private key for the credential,
+    /// - 32 bytes: encrypted relying party ID hashed with SHA256,
+    /// - 32 bytes: HMAC-SHA256 over everything else.
+    ///
+    /// For v3 (CBOR key handle) the credential ID consists of:
+    /// -  1 byte : version number
+    /// - 16 bytes: initialization vector for AES-256,
+    /// - 64 bytes * N: encrypted CBOR key handle
```
kaczmarczyck: Why multiples of 64 byte? Block length is 32 bytes. Also I thought we wanted a fixed size of 241 byte for the key handle, so we can reject everything else?

src/ctap/crypto_wrapper.rs
```diff
@@ -214,6 +196,15 @@ fn ecdsa_key_from_seed(
...
+impl From<&PrivateKey> for cbor::Value {
+    fn from(private_key: &PrivateKey) -> Self {
+        cbor_array![
+            cbor_int!(private_key.signature_algorithm() as i64),
+            cbor_bytes!(private_key.to_bytes()),
+        ]
+    }
+}
+
 impl From<PrivateKey> for cbor::Value {
```
kaczmarczyck: Do we still need this version `from`? If we remove it, we don't have to worry about consistency. Doesn't look like a big difference in performance, since `to_bytes` takes a reference. If we want both, we probably want a test to check they are identical. Or one calls the other.

src/ctap/crypto_wrapper.rs
```diff
@@ -302,66 +250,30 @@ pub fn encrypt_key_handle(
...
-///
-/// This functions reads:
-/// - legacy credentials (no version number),
-/// - v1 (ECDSA)
-/// - v2 (EdDSA over curve Ed25519)
+/// - check_cred_protect is set and the credential requires user verification.
 pub fn decrypt_credential_source(
     env: &mut impl Env,
-    credential_id: Vec<u8>,
+    credential_id_bytes: Vec<u8>,
```
kaczmarczyck: This name does not match the specification.

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// - 32 bytes: HMAC-SHA256 over everything else.
+    pub(super) fn decrypt_from_bytes(
+        env: &mut impl Env,
+        credential_id: Vec<u8>,
```
kaczmarczyck: I wonder if this should also be a slice then, like in all other parsing functions?

src/ctap/crypto_wrapper.rs
```diff
@@ -302,66 +250,30 @@ pub fn encrypt_key_handle(
...
-        (&credential_id[1..hmac_message_size], algorithm)
+    let cred_id = match CredentialId::decrypt_from_bytes(env, credential_id_bytes.clone())? {
+        None => return Ok(None),
+        Some(x) => x,
     };
-    if payload.len() != 80 {
-        // We shouldn't have HMAC'ed anything of different length. The check is cheap though.
+    let is_protected =
+        cred_id.cred_protect_policy == Some(CredentialProtectionPolicy::UserVerificationRequired);
+    if rp_id_hash != cred_id.rp_id_hash || (check_cred_protect && is_protected) {
```
kaczmarczyck: I don't think this check should be performed here. Logically, it belongs to the GetAssertion flow that should take the returned credential and check its credProtect policy.

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,268 @@
...
+#[cfg(feature = "ed25519")]
+pub const ED25519_CREDENTIAL_ID_VERSION: u8 = 0x02;
+pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x03;
+
+pub const MAX_PADDING_LENGTH: u8 = 0xBF;
+
+pub(crate) struct CredentialId {
+    pub private_key: PrivateKey,
+    pub rp_id_hash: [u8; 32],
+    pub cred_protect_policy: Option<CredentialProtectionPolicy>,
```
kaczmarczyck: Maybe add credBlob as well, while we're at it, to introduce less legacy code? Could also be a next PR.

kaczmarczyck (commented):
The CBOR map code looks like a clean way to parse credentials. I'm wondering though if real world embedded concerns don't allow us having nice things. Now that we have the numbers, disadvantages compared to a fixed layout (with the same max credential size) seem to be:

- 2 kB binary size from CBOR map destructuring and padding.
- Extra bytes used for padding and CBOR.

Advantages of your implementation:
- Deprecation of fields is possible.
- Clean code.

Since we have a hard limit of 255 bytes for credentials, the lost extra bytes mean that we might not extend the credential ID as often. Now, from the 192 bytes, we used
- 32 (PR ID hash)
- 32 Private Key
- Signature Algorithm, credProtect
Anticipated are:
- credBlob (32 - 64)
- larger private keys of different signature algorithms

Which means we might need that second credential version that can not be parsed by CTAP1 sooner. Maybe that's okay, because these anticipated features can in fact not simply be ignored by CTAP1, but it's a tradeoff. WDYT?

src/ctap/credential_id.rs
hcyang-google: As we won't increase legacy code with this new format (either cbor of fixed) anymore, I think we can add it in next PR.

src/ctap/crypto_wrapper.rs
hcyang-google: discoverable credential's check is performed in find_credential, so I perform this here for consistency. It should also make sense to move both checks outside, should I make this change in this PR too?

hcyang-google:
Acking the style suggestions, will change them soon.

Regarding the most important decision to make: fixed or cbor, I agree your comparison of the pros and cons, all that I've thought of is there. I think you, Julien, or other major owners of this project is more suitable to make the decision. Either way it's easy to modify this PR, it's just changing the parsing & generating.

src/ctap/credential_id.rs
ia0: I think `arrayref` was used because back then it wasn't possible to convert slices to arrays. I think it's better to use what the language provides than rely on a third-party crate.

src/ctap/credential_id.rs
ia0: Yes, if we don't use the fact that we have ownership we shouldn't take it. And it looks like we only read from this. So should be a slice.

src/ctap/crypto_wrapper.rs
kaczmarczyck: Ah right, IIRC this was an optimization to filter inside the storage. Looking at the code now, might make sense to move both to GetAssertion parsing. WDYT?

kaczmarczyck:
We discussed the decision, and our preferred solution is the following: We keep CBOR, we remove the versioned legacy credential support (but not the non-versioned credential format from stable) and call your design v1.

Reasons to remove v1 and v2:
- They only existed inside develop for a brief moment in time, so people shouldn't rely on them.
- They don't handle credProtect properly, so they have a bug, essentially.
- If there are any users who actually use them, they are in a very small anonymity set just by looking at the version.
- Therefore it's probably better to break users and make them recreate their credentials.

Users should be aware that OpenSK develop should never be your only way to access an account. So I believe this is the right thing to do. It does also help with binary size and maintenance in this an all future versions of OpenSK.

src/ctap/credential_id.rs
hcyang-google: done, I thought we need it but we don't.

src/ctap/credential_id.rs
hcyang-google: Done

src/ctap/credential_id.rs
hcyang-google: Done, I keep changing the block size so forgot it.

src/ctap/credential_id.rs
hcyang-google: Done, renamed to ParsedCredentialId

src/ctap/credential_id.rs
hcyang-google: Done, changed to 192

src/ctap/crypto_wrapper.rs
hcyang-google: Done, and changed encrypt_to_bytes/decrypt_from_bytes to just encrypt/decrypt

src/ctap/crypto_wrapper.rs
hcyang-google: After inspecting the locations I need to change for this, I decided it should be in next PR.

src/ctap/credential_id.rs
```diff
@@ -161,12 +149,12 @@ fn remove_padding(data: &mut Vec<u8>) -> Result<(), Ctap2StatusCode> {
...
 }
 
-impl CredentialId {
+impl ParsedCredentialId {
     /// Encrypts the given private key, relying party ID hash, and cred protect policy into a credential ID.
     ///
     /// Other information, such as a user name, are not stored. Since encrypted credential IDs are
     /// stored server-side, this information is already available (unencrypted).
-    pub(super) fn encrypt_to_bytes(
+    pub(super) fn encrypt(
```
kaczmarczyck: Curious about the design choice here. `encrypt` is part of `impl`, but doesn't use `&self`, even though it takes a `&ParsedCredentialId`. Couldn't we call it like
`private_key.encrypt(env, rp_id_hash, cred_protect_policy)`
instead of
`ParsedCredentialId::encrypt(env, private_key, rp_id_hash, cred_protect_policy)`
?

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// For v1 (CBOR key handle) the credential ID consists of:
+    /// -   1 byte : version number
+    /// -  16 bytes: initialization vector for AES-256,
+    /// - 192 bytes: encrypted CBOR key handle
+    /// -  32 bytes: HMAC-SHA256 over everything else.
+    pub(super) fn decrypt(
+        env: &mut impl Env,
+        credential_id: &[u8],
+    ) -> Result<Option<Self>, Ctap2StatusCode> {
+        if credential_id.len() < MIN_CREDENTIAL_ID_SIZE {
```
kaczmarczyck: This function doesn't perform a length check for `decrypt_cbor_key_handle`. That shouldn't be a problem in practise, since we never HMAC anything of a different length, and worst case we fail decryption if it's not a multiple of block size.

I was wondering if we rather want to be more defensive, or keep this function open to accept unexpected credential IDs.

kaczmarczyck (dismissed):
Looks good to me, with 2 open style questions.

src/ctap/credential_id.rs
hcyang-google: Private key is not "more important" than other fields it has, so I don't want it to be a method on private key.

More logically it should be `ParsedCredentialId::new(...).encrypt()`, but this requires ownership and cloning, but in fact only the reference of those fields are ever needed if we are only going to call encrypt() anyway. Any better pattern suggestion?

src/ctap/crypto_wrapper.rs
kaczmarczyck: Sounds good.

src/ctap/credential_id.rs
kaczmarczyck: Ah sorry, the `private_key` is not even a `ParsedCredentialId` anyway. So this function is unrelated to `ParsedCredentialId`. And it is also wrapped by `encrypt_key_handle` as a one-liner. Means we could rename it to `encrypt_key_handle` and then simply import it in `crypto_wrapper.rs`.

This also means that `ParsedCredentialId` is a private type used as an intermediate output of credential parsing, but ultimately converted to a CredentialSource. Maybe we can move pieces around to make everything in `ParsedCredentialId` private (or just delete it)?

hcyang-google: Made a quite big change by directly putting encrypt & decrypt inside credential_id.rs without exposing any other intermediate types anymore. And make some naming changes that is more correct.

src/ctap/credential_id.rs
hcyang-google: Checked the length, I believe we should do so.

Side question, when I was modifying related code I believe we should be stricter on errors after we've checked hmac, or at least more consistent. If HMAC passed then that is something generated by us, so many upcoming errors should actually indicate internal state error.
On the other hand, if we ever produced some invalid key handles and we want to ignore them instead of returning internal errors, we should `.ok()` some results and return Ok(None) in more places. As long as we have a clear rationale it's fine, but the current codebase looks more like using `?` to pass errors whenever we can, and return Ok(None) when we met options that are None. I think we should be more consistent logically, WDYT?

src/ctap/credential_id.rs
hcyang-google: Now that it directly produces a pubkey credential source which owns the credential_id bytes, I changed it back to taking Vec.

src/ctap/crypto_wrapper.rs
hcyang-google: Done

src/ctap/credential_id.rs
ia0: I agree that I was quite confused by the difference between `Ok(None)` and `Err(_)`. It would be nice if we had a clear story. In particular, `Ok(None)` should not be used for errors. So it's all a question of what do we consider as errors. And I guess if something produced by us looks wrong it's an error.

src/ctap/credential_id.rs
```diff
@@ -39,16 +41,31 @@ pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x01;
...
+struct CredentialSource {
+    private_key: PrivateKey,
+    rp_id_hash: [u8; 32],
+    cred_protect_policy: Option<CredentialProtectionPolicy>,
 }
 
-fn decrypt_legacy_key_handle(
+// The data fields contained in the credential ID are serizlied using CBOR maps.
+// Each field is associated with a unique tag, implemented with a CBOR unsigned key.
+enum CredentialSourceField {
```
kaczmarczyck, line 54: I know this is kind of late I'm coming around with this idea, but we could reuse `PublicKeyCredentialSourceField`? It should be a superset of `CredentialSourceField`? It has the same purpose, just for resident keys.

kaczmarczyck (commented):
There seems to be a `String` import problem. I like the cleanup in your last commit. And had that idea to cleanup even more by reusing an enum. I think reuse is nice, it does carry some legacy and resident key only fields though, and goes to 12 already. Note: If the map keys go over 23, we need to use one more byte to encode them. Numbers from -24 to 0 are an option if we ever need that many.

src/ctap/credential_id.rs
hcyang-google: The string problem was fixed.

Probably we can reuse, but we don't have RpIdHash. We can add one entry for it though.

kaczmarczyck: Ah right, we only have a RP ID hash here, whereas we store the full RP ID in resident keys. Maybe that's proof I shouldn't try to entangle things needlessly. :) New enum is good.

kaczmarczyck (approved):

src/ctap/credential_id.rs
```diff
@@ -0,0 +1,475 @@
...
+            } else {
+                None
+            };
+            Some(CredentialSource {
+                private_key,
+                rp_id_hash: rp_id_hash.try_into().unwrap(),
+                cred_protect_policy,
+            })
+        }
+        _ => None,
```
kaczmarczyck, line 121: I double checked the `Ok(None)` vs `Err(_)` question, and this looked interesting, everything is is correct I believe. If the private key or hash are missing, that would be a bad sign indeed, maybe this needs to be an error?

src/ctap/credential_id.rs
kaczmarczyck: I forgot this open conversation, and went back to double check. My new comment is the only piece of code that I'm not sure is correct, WDYT?

ia0 (approved):
I didn't redo a thorough review of the code. It overall looks good to me. Ping me back if I should look more closely.

src/ctap/credential_id.rs
hcyang-google: Let me look into it deeper and open a discussion if needed, let's merge this first.

PR #513, opened 2022-07-14
Fix temporary deadlock in USB pending transaction state

liamjm:
This issue was introduced as part of #490 where the multiple USB endpoints were connected to the multiple interfaces in the app. This moved the pending_out state variable from a singleton to per-endpoint, which causes a deadlock like so:

1. App sends syscall to receive_packet(), setting pending_out to true for both endpoints.
1. USB interrupt happen for endpoint_a and use the pending packet.
1. USB interrupt happen for endpoint_b, the app is busy so the delayed_out is set.
1. Endpoint_a processing completes
1. App sends syscall to receive_packet(), pending_out is already set for endpoint_b, so it never takes delayed_out.

This deadlock eventually is cleared once a timeout is reached in the app and a call to cancel_out_transaction() is made.

This fix reverts the pending_out state variable to a singleton, and therefore fixes this issue. The code currently returns the first packet that is delayed. This  could be changed to a round-robin approach to ensure each endpoint is processed more quickly.

### DIFF
--- a/patches/tock/12-pending-deadlock.patch
+++ b/patches/tock/12-pending-deadlock.patch
@@ -1,0 +1,79 @@
+diff --git a/capsules/src/usb/usbc_ctap_hid.rs b/capsules/src/usb/usbc_ctap_hid.rs
+index 16b80cb10..c71ed639a 100644
+--- a/capsules/src/usb/usbc_ctap_hid.rs
++++ b/capsules/src/usb/usbc_ctap_hid.rs
+@@ -120,7 +120,6 @@ struct EndpointState {
+ 
+     tx_packet: OptionalCell<[u8; 64]>,
+     pending_in: Cell<bool>,
+-    pending_out: Cell<bool>,
+     // Is there a delayed packet?
+     delayed_out: Cell<bool>,
+ }
+@@ -133,7 +132,6 @@ impl EndpointState {
+             out_buffer: Buffer64::default(),
+             tx_packet: OptionalCell::empty(),
+             pending_in: Cell::new(false),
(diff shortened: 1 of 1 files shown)


patches/tock/12-pending-deadlock.patch
```diff
@@ -0,0 +1,79 @@
...
+-                continue;
+-            } else {
+-                s.pending_out.set(true);
++        if self.pending_out.get() {
++            // The previous packet has not yet been received, reject the new one.
++        } else {
++            self.pending_out.set(true);
++            // Process the first endpoint that has a delayed packet.
++            // TODO(liamjm): Determine if a round-robin approach reduces latency
++            // in the worst case.
```
ia0, line 55: This is only an issue in case of long-lasting contention right? I think it's ok if the behavior is unfair. We could however decide which endpoint should be prioritized (by adding a `.rev()` after `.enumerate()`). Maybe we want to prefer the second page. That would be specific to OpenSK, but this driver is already specific to OpenSK.

ia0 (approved):

patches/tock/12-pending-deadlock.patch
kaczmarczyck: I'd say we need it if one endpoint receives for more than 100 ms, and the other CTAPHID times out without fairness here?

kaczmarczyck (approved):

patches/tock/12-pending-deadlock.patch
liamjm: Agreed. I"ll do this in the subsequent PR.

PR #514, opened 2022-07-19
Process incoming packets on different interfaces in a round robin fashion.

liamjm:
The current approach selects the first endpoint  / interface (the main FIDO one) first all the time. This introduces a delay in processing the Vendor HID interface. This can lead to timeouts in processing the packets as each packet must be received within 100ms of the last one. #502 addresses this by increasing this timeout, but this is only a hack.

This PR fixes this problem by always processing the other interface than the previous packet. This is implemented by tracking the last endpoint that returned data, and trying the other one first next time (new next_endpoint_index variable).
This PR is implemented as a change to an existing patch (patches/tock/12-pending-deadlock.patch) as this is this patch has overlapping functionality.

### DIFF
--- a/patches/tock/12-pending-deadlock.patch
+++ b/patches/tock/12-pending-deadlock.patch
@@ -3,7 +3,7 @@ index 16b80cb10..949388b70 100644
 --- a/capsules/src/usb/usbc_ctap_hid.rs
 +++ b/capsules/src/usb/usbc_ctap_hid.rs
 @@ -120,7 +120,6 @@ struct EndpointState {
- 
+
      tx_packet: OptionalCell<[u8; 64]>,
      pending_in: Cell<bool>,
 -    pending_out: Cell<bool>,
@@ -21,13 +21,13 @@ index 16b80cb10..949388b70 100644
 @@ -142,6 +140,10 @@ impl EndpointState {
  pub struct ClientCtapHID<'a, 'b, C: 'a> {
      client_ctrl: ClientCtrl<'a, 'static, C>,
- 
+
 +    // Is there a pending OUT transaction happening?
 +    pending_out: Cell<bool>,
 +    next_endpoint_index: Cell<usize>,
 +
      endpoints: [EndpointState; NUM_ENDPOINTS],
- 
+
      // Interaction with the client
 @@ -264,6 +266,8 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtapHID<'a, 'b, C> {
                  LANGUAGES,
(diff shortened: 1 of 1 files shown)


patches/tock/12-pending-deadlock.patch
```diff
@@ -50,25 +52,38 @@ index 16b80cb10..c71ed639a 100644
...
-     }
+@@ -352,6 +357,13 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> ClientCtapHID<'a, 'b, C> {
+                 self.cancel_in_transaction(endpoint);
  
+                 self.client.map(|client| client.packet_received(&buf, endpoint, app));
++                // Update next packet to send.
++                for (i, ep) in self.endpoints.iter().enumerate() {
++                  if ep.endpoint == endpoint {
++                    self.next_endpoint_index.set((i + 1) % NUM_ENDPOINTS);
++                    break
```
ia0: nit: semi-colon after `break` (and other expressions used as statement that return `!`, i.e. any type, but don't rely on it, i.e. unit `()` would suffice).

ia0 (dismissed):

patches/tock/12-pending-deadlock.patch
```diff
@@ -50,25 +52,38 @@ index 16b80cb10..c71ed639a 100644
 +            // The previous packet has not yet been received, reject the new one.
 +        } else {
 +            self.pending_out.set(true);
-+            // Process the first endpoint that has a delayed packet.
-+            // TODO(liamjm): Determine if a round-robin approach reduces latency
-+            // in the worst case.
-+            for (_, s) in self.endpoints.iter().enumerate() {
++            // Process the next endpoint that has a delayed packet.
++            for i in self.next_endpoint_index.get()..self.next_endpoint_index.get()+NUM_ENDPOINTS {
```
kaczmarczyck: nit: Some formatting is off, like the spaces around + here.

kaczmarczyck (dismissed):

patches/tock/12-pending-deadlock.patch
liamjm: Done.

ia0 (dismissed):

patches/tock/12-pending-deadlock.patch
```diff
@@ -1,9 +1,9 @@
 diff --git a/capsules/src/usb/usbc_ctap_hid.rs b/capsules/src/usb/usbc_ctap_hid.rs
-index 16b80cb10..2b19519e4 100644
+index 16b80cb10..949388b70 100644
 --- a/capsules/src/usb/usbc_ctap_hid.rs
 +++ b/capsules/src/usb/usbc_ctap_hid.rs
 @@ -120,7 +120,6 @@ struct EndpointState {
-
+
```
kaczmarczyck: This commit introduced a bunch of spaces for some reason?

kaczmarczyck (dismissed):

patches/tock/12-pending-deadlock.patch
liamjm: :shrug: 
Perhaps these came from an IDE.

Removed.

kaczmarczyck (approved):

PR #515, opened 2022-07-19
Interleave sending and receiving of packets to reduce rx latency

liamjm:
As discussed in #502 and #514 packet processing latency is too slow when both interfaces are being used. Another cause for this is that when a reply is being sent, it does so in a tight loop (ie inner loop in the main function) without receiving any other packets. This can easily introduce hundreds of milliseconds delay.

This PR removes the innner loop in main and leaves just a single loop. Each iteration it will optionally send a packet and always attempt to receive an incoming packet (#514 ensures suitable routing of the oldest packet). For our use case, the processing of these packets will be interleaved across the FIDO and Vendor HID interfaces (if both are present).

Several changes are required for this:

1. SendOrRecvStatus enum now includes the UsbEndpoint in the Received variant.
1. Code that uses SendOrRecvStatus handles the UsbEndpoint.
1. The HidConnection::send_or_recv_with_timeout now allows any endpoint to be received, not just the one it sent one.
1. HidPacketIterator now supports peeking to see if there is any data.
1. main.rs has new structs to support the state tracking and large changes to the loop.
1. tock CtapUsbSyscallDriver now sends packet before receiving incoming packet. Without this the tx packet is lost.

### DIFF
--- a/patches/tock/13-send-packet-before-receiving.patch
+++ b/patches/tock/13-send-packet-before-receiving.patch
@@ -1,0 +1,32 @@
+diff --git a/capsules/src/usb/usb_ctap.rs b/capsules/src/usb/usb_ctap.rs
+index 2c91c0968..ea8111069 100644
+--- a/capsules/src/usb/usb_ctap.rs
++++ b/capsules/src/usb/usb_ctap.rs
+@@ -253,18 +253,19 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> Driver for CtapUsbSyscallDriver<'a,
+                             if app.waiting {
+                                 ReturnCode::EALREADY
+                             } else {
++                                // Indicates to the driver that we have a packet to send.
++                                let r = self
++                                    .usb_client
++                                    .transmit_packet(app.buffer.as_ref().unwrap().as_ref(), endpoint);
++                                if r != ReturnCode::SUCCESS {
++                                    return r;
++                                }
+                                 // Indicates to the driver that we can receive any pending packet.
+                                 app.waiting = true;
+                                 self.usb_client.receive_packet(app);
+ 
+-                                if !app.waiting {
+-                                    // The call to receive_packet() collected a pending packet.
+-                                    ReturnCode::SUCCESS
+-                                } else {
+-                                    // Indicates to the driver that we have a packet to send.
+-                                    self.usb_client
+-                                        .transmit_packet(app.buffer.as_ref().unwrap().as_ref(), endpoint)
+-                                }
++                                ReturnCode::SUCCESS
++
+                             }
+                         } else {
+                             ReturnCode::EINVAL
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -274,7 +274,7 @@ fn send_keepalive_up_needed(
     let keepalive_msg = CtapHid::keepalive(cid, KeepaliveStatus::UpNeeded);
     for mut pkt in keepalive_msg {
         let ctap_hid_connection = transport.hid_connection(env);
-        match ctap_hid_connection.send_or_recv_with_timeout(&mut pkt, timeout) {
+        match ctap_hid_connection.send_and_maybe_recv(&mut pkt, timeout) {
             Ok(SendOrRecvStatus::Timeout) => {
                 debug_ctap!(env, "Sending a KEEPALIVE packet timed out");
                 // TODO: abort user presence test?
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -284,6 +284,7 @@ fn send_keepalive_up_needed(
                 debug_ctap!(env, "Sent KEEPALIVE packet");
             }
             Ok(SendOrRecvStatus::Received(_)) => {
+                // TODO(liamjm): Support receiving packets on both interfaces.
                 // We only parse one packet, because we only care about CANCEL.
                 let (received_cid, processed_packet) = CtapHid::process_single_packet(&pkt);
                 if received_cid != &cid {
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -83,6 +83,7 @@ use crypto::hmac::hmac_256;
 use crypto::sha256::Sha256;
 use crypto::{ecdsa, Hash256};
 use embedded_time::duration::Milliseconds;
+use libtock_drivers::usb_ctap_hid::UsbEndpoint;
 use rng256::Rng256;
 use sk_cbor as cbor;
 use sk_cbor::cbor_map_options;
@@ -283,7 +284,20 @@ fn send_keepalive_up_needed(
             Ok(SendOrRecvStatus::Sent) => {
                 debug_ctap!(env, "Sent KEEPALIVE packet");
             }
-            Ok(SendOrRecvStatus::Received(_)) => {
+            Ok(SendOrRecvStatus::Received(endpoint)) => {
+                let rx_transport = match endpoint {
+                    UsbEndpoint::MainHid => Transport::MainHid,
+                    #[cfg(feature = "vendor_hid")]
+                    UsbEndpoint::VendorHid => Transport::VendorHid,
+                };
+                if rx_transport != transport {
+                    debug_ctap!(
+                        env,
+                        "Received a packet on transport {:?} while sending a KEEPALIVE packet on transport {:?}",
+                         rx_transport, transport
+                    );
+                }
+
                 // TODO(liamjm): Support receiving packets on both interfaces.
                 // We only parse one packet, because we only care about CANCEL.
                 let (received_cid, processed_packet) = CtapHid::process_single_packet(&pkt);
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -283,7 +283,7 @@ fn send_keepalive_up_needed(
             Ok(SendOrRecvStatus::Sent) => {
                 debug_ctap!(env, "Sent KEEPALIVE packet");
             }
-            Ok(SendOrRecvStatus::Received) => {
+            Ok(SendOrRecvStatus::Received(_)) => {
                 // We only parse one packet, because we only care about CANCEL.
                 let (received_cid, processed_packet) = CtapHid::process_single_packet(&pkt);
                 if received_cid != &cid {
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -296,6 +296,8 @@ fn send_keepalive_up_needed(
                         "Received a packet on transport {:?} while sending a KEEPALIVE packet on transport {:?}",
                          rx_transport, transport
                     );
+                    // Ignore this packet.
+                    continue;
                 }
 
                 // TODO(liamjm): Support receiving packets on both interfaces.
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -297,10 +297,10 @@ fn send_keepalive_up_needed(
                          rx_transport, transport
                     );
                     // Ignore this packet.
+                    // TODO(liamjm): Support receiving packets on both interfaces.
                     continue;
                 }
 
-                // TODO(liamjm): Support receiving packets on both interfaces.
                 // We only parse one packet, because we only care about CANCEL.
                 let (received_cid, processed_packet) = CtapHid::process_single_packet(&pkt);
                 if received_cid != &cid {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,7 +50,7 @@ pub mod clock;
 #[cfg(feature = "std")]
 pub mod ctap;
 #[cfg(not(feature = "std"))]
-mod ctap;
+pub mod ctap;
 pub mod env;
 #[cfg(feature = "std")]
 pub mod test_helpers;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,10 +46,6 @@ macro_rules! debug_ctap {
 
 pub mod api;
 pub mod clock;
-// Implementation details must be public for testing (in particular fuzzing).
-#[cfg(feature = "std")]
-pub mod ctap;
-#[cfg(not(feature = "std"))]
 pub mod ctap;
 pub mod env;
 #[cfg(feature = "std")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,6 +46,7 @@ macro_rules! debug_ctap {
 
 pub mod api;
 pub mod clock;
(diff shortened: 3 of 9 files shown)


patches/tock/13-send-packet-before-receiving.patch
```diff
@@ -0,0 +1,32 @@
...
+--- a/capsules/src/usb/usb_ctap.rs
++++ b/capsules/src/usb/usb_ctap.rs
+@@ -253,18 +253,19 @@ impl<'a, 'b, C: hil::usb::UsbController<'a>> Driver for CtapUsbSyscallDriver<'a,
+                             if app.waiting {
+                                 ReturnCode::EALREADY
+                             } else {
++                                // Indicates to the driver that we have a packet to send.
++                                let r = self
++                                    .usb_client
++                                    .transmit_packet(app.buffer.as_ref().unwrap().as_ref(), endpoint);
```
ia0, line 12: You mention in the PR description that you swap transmit and receive because otherwise the transmit packet would be lost. How is that an issue? And reciprocally, why is the swap not an issue? I don't know the details but one thing that comes to mind is whether the transmitted packet could be misinterpreted (i.e. we would not have sent it if we would have received the packet first).

src/ctap/mod.rs
```diff
@@ -283,7 +283,7 @@ fn send_keepalive_up_needed(
             Ok(SendOrRecvStatus::Sent) => {
                 debug_ctap!(env, "Sent KEEPALIVE packet");
             }
-            Ok(SendOrRecvStatus::Received) => {
+            Ok(SendOrRecvStatus::Received(_)) => {
```
ia0: What if we receive on a different endpoint than the one being kept alive? Does it mean we discard it?

src/lib.rs
```diff
@@ -50,7 +50,7 @@ pub mod clock;
 #[cfg(feature = "std")]
 pub mod ctap;
 #[cfg(not(feature = "std"))]
-mod ctap;
+pub mod ctap;
```
ia0, line 50: Interesting. I'm surprised it used to compile before. I would have expected that the `Ctap::state()` doesn't compile because it is public and its return type is private.

The idea behind a private `ctap` module is that we consider it internal implementation detail. Users should use the `Ctap` struct.

Given that this is already broken, I'd suggest simply using `pub mod ctap` for now (without the `feature = "std"` switch). We could also remove the comment about testing given that it's outdated (although it's the intent).

src/main.rs
```diff
@@ -46,12 +47,72 @@ use libtock_drivers::console::Console;
...
+    }
+}
+
+// A single packet to send.
+struct SendPacket {
+    transport: Transport,
+    packet: [u8; 64],
+}
+
+struct Endpoints {
```
ia0: The name is a bit confusing. It's endpoints but it contains endpoint replies.

src/main.rs
```diff
@@ -46,12 +47,72 @@ use libtock_drivers::console::Console;
...
+            endpoints: [
+                EndpointReply::new(UsbEndpoint::MainHid),
+                #[cfg(feature = "vendor_hid")]
+                EndpointReply::new(UsbEndpoint::VendorHid),
+            ],
+        }
+    }
+
+    pub fn next_packet(&mut self) -> Option<SendPacket> {
+        for ep in self.endpoints.iter_mut() {
```
ia0: I guess round-robin is not necessary here?

src/main.rs
```diff
@@ -127,32 +222,18 @@ fn main() {
...
-                    Ok(SendOrRecvStatus::Timeout) => {
-                        #[cfg(feature = "debug_ctap")]
-                        print_packet_notice("Sending packet timed out", &clock);
-                        // TODO: reset the ctap_hid state.
-                        // Since sending the packet timed out, we cancel this reply.
+            if reply.has_data() {
+                // Update endpoint with the reply.
+                for ep in endpoints.endpoints.iter_mut() {
+                    if ep.endpoint == endpoint {
+                        ep.reply = reply;
```
ia0, line 243: Is it ok to "discard" what has not yet been sent? (I guess yes, but should we print something?)

src/main.rs
liamjm: Not necessary to get the tests to pass. But this doesn't test on host behavior.  It could be done. 
Round robin will:
+ ensure latency between packets for each endpoint is kept small for the host
+ a small reply won't get blocked by a large reply as much
- increase time to send entire reply for the first endpoint

Current approach of Main HID first:
+ prioritises Main HID replies (arguably more important for users)
+ simpler
- delays Vendor HID, even if it small it will be sent after a large Main HID replay.
- may introduce large delays between replies for Vendor HID.

What do you think?

src/lib.rs
liamjm: Yeah, I was a little confused by this also.

As suggested, I've remove the comment and the feature switch.

src/ctap/mod.rs
liamjm: Good find.

At present we discard it. I've added a TODO to address this (for a later PR), but I can address it here is you'd prefer.

src/main.rs
liamjm: Indeed.
Updated names to replies.

patches/tock/13-send-packet-before-receiving.patch
liamjm: Currently the code is like:
```
self.received_packet()
if app.busy {
  return ReturnCode::SUCCESS
}
self.send_packet(buf)
```

So if the receive_packet() call picks up a packet and starts processing it, the app will be busy, this function will return success, but the packet was not sent. From the perspective of the caller, it called send_or_recv() and it returned success, so it has to assume that the packet was sent. Therefore the packet is never sent and next call the next packet will be sent. I think this is clear that this is an issue.

Correct, by sending first, we can't respond differently based on a received packet. However, that is how things currently work, because the sending of replies happens in a tight loop in main.rs without processing any new packets.
I'm not sure the expectations of the host, but expecting a new packet to change the behavior of an already processed packet is inherently racey, so I'm not too concerned about not responded to the received packet promptly. 

But I'm open to other thoughts.

src/main.rs
liamjm: Added a log message.

I wondered if creating a queue of replies, so that multiple replies per endpoint would work better. What do you think?

src/ctap/mod.rs
ia0: I guess it's not really a regression (but I might be wrong). So I would say it's ok to do it later.

src/main.rs
ia0: No, I think it's ok (and actually good) to use bounded channels (here with a capacity of one reply).

src/lib.rs
ia0: This doesn't look resolved. Missing commit?

patches/tock/13-send-packet-before-receiving.patch
ia0: I see, the main problem is that the API is not able to differentiate between:
- successful send and receive
- successful send
- successful receive

I think the previous convention (because of the name "send OR receive") was that "successful send and receive" never happens. The last two are distinguished using some return value? (I don't know)

I agree with you that the current code is wrong because if `send_or_recv` receives, it doesn't try to send again the packet and instead sends the next one (if any). Which suggests that `send_or_recv` is used as if it was `send_and_maybe_recv`. And this is what you're actually doing in this PR (fixing the implementation to reflect the usage and not the name).

I guess renaming is a pain so maybe just documenting `api::HidConnection::send_or_recv_with_timeout` with the expected semantics is already good enough. We could also possibly just rename that function and keep the Tock naming wrong. (I actually prefer that.)

src/main.rs
ia0: I think Vendor HID is more important because you don't want updates to get interrupted. But the correct solution is probably to disable the Main HID during an update.

I'm fine to keep as is if there's no current issue and deal with it later when designing upgradability or when an issue arises.

src/lib.rs
liamjm: Whoops, sorry about that.

Fixed.

patches/tock/13-send-packet-before-receiving.patch
liamjm: yes, exactly.

I've renamed send_or_recv to send_and_maybe_recv.

ia0 (dismissed):
Looks good on my side. I'll be OOO until Aug 3rd so don't wait for me if there are me updates.

src/main.rs
```diff
@@ -32,6 +32,7 @@ use ctap2::api::connection::{HidConnection, SendOrRecvStatus};
 #[cfg(feature = "debug_ctap")]
 use ctap2::clock::CtapClock;
 use ctap2::clock::{new_clock, Clock, ClockInt, KEEPALIVE_DELAY, KEEPALIVE_DELAY_MS};
+use ctap2::ctap::hid::HidPacketIterator;
```
kaczmarczyck, line 35: This is the before mentioned reason we need `ctap` to be public right?

src/ctap/mod.rs
kaczmarczyck: I think it is a regression if we receive, but the channel ID is by chance the same. Then a different transport would interact with this channel.

src/ctap/mod.rs
```diff
@@ -274,7 +274,7 @@ fn send_keepalive_up_needed(
     let keepalive_msg = CtapHid::keepalive(cid, KeepaliveStatus::UpNeeded);
     for mut pkt in keepalive_msg {
         let ctap_hid_connection = transport.hid_connection(env);
-        match ctap_hid_connection.send_or_recv_with_timeout(&mut pkt, timeout) {
+        match ctap_hid_connection.send_and_maybe_recv(&mut pkt, timeout) {
```
kaczmarczyck, line 278: Maybe we have to add more HID tests? The most common situation with unexpected OUT packets (and therefore OUT and IN at the same time) is probably a CANCEL during user presence checks. Should be something like a MakeCredential request, and while user presence is checked, spam it with packets, and see if those are treated correctly.

src/lib.rs
kaczmarczyck: Maybe that's because `Ctap::state()` could have also been behind `#[cfg(feature = "std")]` ? (And we should add that.)

One possible solution to keep it private would be to move `ctap2::ctap::hid` to `ctap2::hid`. I had this option in mind when I last refactored the HID module, and there is not much left that stops us from doing so.

src/main.rs
```diff
@@ -127,32 +222,27 @@ fn main() {
...
-                        // Since sending the packet timed out, we cancel this reply.
+            if reply.has_data() {
+                // Update endpoint with the reply.
+                for ep in replies.replies.iter_mut() {
+                    if ep.endpoint == endpoint {
+                        if ep.reply.has_data() {
+                            #[cfg(feature = "debug_ctap")]
+                            writeln!(
+                                Console::new(),
+                                "Warning overwritting existing reply for endpoint {}",
```
kaczmarczyck: "Warning overwritting" is probably a typo.

src/main.rs
liamjm: Yes.

src/lib.rs
liamjm: Can I leave that refactor for you do to?

src/main.rs
liamjm: Fixed.

src/lib.rs
kaczmarczyck: Yeah why not, please add a TODO for me.

src/ctap/mod.rs
liamjm: Yes, good point.
I can add a check for this.

However, I am wondering about a bigger change. If the code ignores packets on the other interface, then those packets are lost. So perhaps the better approach is to send() but not send_and_recv() at this time. That way the incoming packets are not serviced and will sit there. Looking at the code, it looks challenging to connect incoming packets to the other interface without massive refactoring.
Thoughts?

kaczmarczyck: I generally agree that a send / receive interface would be cleaner. The current solution is there for practical reasons. USB Request Blocks (URB) that need to be answered quickly. And if you only send and have the syscall roundtrip delay from that, you can't answer OUT packets in time.

If you try to go for a send / receive interface, please test for everything that didn't work last time.
@gendx @jmichelp are the original authors who made that decision. Maybe they can help describe a few test cases to look out for?

liamjm: OK.

What I've done in the short term is ignore those packet on the other interface. This stops the regression and leaves room for the above discussed improvements at a later date.
I've add tests to ./tools/vendor_hid_tests.py to verify this behavior.

src/ctap/mod.rs
liamjm: Agreed. Added these tests.

src/lib.rs
liamjm: Done.

kaczmarczyck (commented):
2 high level comments:

1. The new main is different enough (i.e. might discard replies) that we should test OpenSK a bit with it.

2. The test script entangles raw HID and libfido. I don't think there is an elegant solution for this, but I thought it's interesting to mention that there is a way to make libfido talk to the second usage page. You monkey patch it's usage page:
```
import fido2
fido2.hid.base.FIDO_USAGE_PAGE = 0xFF00
```

You have to be careful to do that after GetInfo though, or we patch OpenSK to allow GetInfo on the second usage page. Not saying this approach is better or you should do it, it's just a different tradeoff I think.

liamjm:
> 2 high level comments:
> 
> 1. The new main is different enough (i.e. might discard replies) that we should test OpenSK a bit with it.

Testing with https://github.com/google/CTAP2-test-tool shows no differences.

> 2. The test script entangles raw HID and libfido. I don't think there is an elegant solution for this, but I thought it's interesting to mention that there is a way to make libfido talk to the second usage page. You monkey patch it's usage page:
> 
> ```
> import fido2
> fido2.hid.base.FIDO_USAGE_PAGE = 0xFF00
> ```
> 
> You have to be careful to do that after GetInfo though, or we patch OpenSK to allow GetInfo on the second usage page. Not saying this approach is better or you should do it, it's just a different tradeoff I think.

Good approach. I've used this idea. 
The value is patched only for the period of enumerating the devices for the Vendor HID device, so should be OK.

kaczmarczyck (approved):

kaczmarczyck:
If everything works, then I'm happy with it!
@jmichelp can always complain later if something is wrong that I didn't account for :)
I'll merge #523 first.

PR #516, opened 2022-07-21
Move out check credProtectPolicy logic

hcyang-google:
Move the credProtectPolicy check outside credential ID decryption &
discoverable credential finding. Modify the unit tests, and add unit
tests for credProtectPolicy checking in non resident flows that were
originally missing.

This should have no behavioral differences.

### DIFF
--- a/src/ctap/credential_id.rs
+++ b/src/ctap/credential_id.rs
@@ -353,7 +353,7 @@ mod test {
 
         for length in (1..CBOR_CREDENTIAL_ID_SIZE).step_by(16) {
             assert_eq!(
-                decrypt_credential_id(&mut env, encrypted_id[..length].to_vec(), &rp_id_hash,),
+                decrypt_credential_id(&mut env, encrypted_id[..length].to_vec(), &rp_id_hash),
                 Ok(None)
             );
         }
--- a/src/ctap/credential_id.rs
+++ b/src/ctap/credential_id.rs
@@ -206,7 +206,6 @@ pub fn decrypt_credential_id(
     env: &mut impl Env,
     credential_id: Vec<u8>,
     rp_id_hash: &[u8],
-    check_cred_protect: bool,
 ) -> Result<Option<PublicKeyCredentialSource>, Ctap2StatusCode> {
     if credential_id.len() < MIN_CREDENTIAL_ID_SIZE {
         return Ok(None);
@@ -240,9 +239,7 @@ pub fn decrypt_credential_id(
         return Ok(None);
     };
 
-    let is_protected = credential_source.cred_protect_policy
-        == Some(CredentialProtectionPolicy::UserVerificationRequired);
-    if rp_id_hash != credential_source.rp_id_hash || (check_cred_protect && is_protected) {
+    if rp_id_hash != credential_source.rp_id_hash {
         return Ok(None);
     }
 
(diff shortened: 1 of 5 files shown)


src/ctap/credential_id.rs
```diff
@@ -356,12 +353,7 @@ mod test {
 
         for length in (1..CBOR_CREDENTIAL_ID_SIZE).step_by(16) {
             assert_eq!(
-                decrypt_credential_id(
-                    &mut env,
-                    encrypted_id[..length].to_vec(),
-                    &rp_id_hash,
-                    false
-                ),
+                decrypt_credential_id(&mut env, encrypted_id[..length].to_vec(), &rp_id_hash,),
```
ia0: nit: extra comma

ia0 (commented):
code-wise LGTM

src/ctap/credential_id.rs
hcyang-google: done, I wonder whether this can be caught by any linting check?

ia0: We just need for the [trailing_comma](https://github.com/rust-lang/rustfmt/issues/3379) rustfmt config to stabilize I guess (and possibly use a more recent compiler if needed).

hcyang-google:
I think coverage decreased because the amount of covered code paths along with total code paths decreased. The introduced code paths in this PR should be covered well, and the only deleted unit test corresponds to a deleted path.

src/ctap/mod.rs
```diff
@@ -1802,6 +1820,61 @@ mod test {
...
 
+    #[test]
+    fn test_non_resident_process_make_credential_credential_with_cred_protect() {
+        let mut env = TestEnv::new();
+        let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0));
+
+        let test_policy = CredentialProtectionPolicy::UserVerificationOptionalWithCredentialIdList;
+        let mut make_credential_params =
+            create_make_credential_parameters_with_cred_protect_policy(test_policy);
+        make_credential_params.options.rk = false;
```
kaczmarczyck, line 1831: I think `rk=False` is the default and this is similar to the test above.

src/ctap/mod.rs
```diff
@@ -717,6 +717,22 @@ impl CtapState {
         Ok(())
     }
 
+    fn check_cred_protect(
```
kaczmarczyck: Maybe we can rename this function, to make its purpose clear. This policy check is only correct if performed for a credential coming from an allow list or exclude list. There is also the `is_discoverable` function that does the equivalent check for non-allow-listed credentials (rk=true) during GetAssertion. `check_listed_...` or something similar?

src/ctap/mod.rs
hcyang-google: No, `rk=True` is the default, and if otherwise the original unit test can't have passed before when credProtectPolicy is not persisted.

src/ctap/mod.rs

src/ctap/mod.rs
kaczmarczyck: Ah right, I missed that in `create_make_credential_parameters_with_cred_protect_policy`. Thanks for confirming!

kaczmarczyck (approved):

PR #518, opened 2022-07-23
Support credBlob for non-resident credentials

hcyang-google:
- Add a upper limit of max_cred_blob_length
- Add test cases for cred_blob in non-resident flows
- Modify the test helper functions in ctap/mod.rs a bit

### DIFF
--- a/src/ctap/credential_id.rs
+++ b/src/ctap/credential_id.rs
@@ -50,7 +50,7 @@ struct CredentialSource {
     cred_blob: Option<Vec<u8>>,
 }
 
-// The data fields contained in the credential ID are serizlied using CBOR maps.
+// The data fields contained in the credential ID are serialized using CBOR maps.
 // Each field is associated with a unique tag, implemented with a CBOR unsigned key.
 enum CredentialSourceField {
     PrivateKey = 0,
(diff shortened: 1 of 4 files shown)


src/ctap/credential_id.rs
```diff
@@ -47,6 +47,7 @@ struct CredentialSource {
     private_key: PrivateKey,
     rp_id_hash: [u8; 32],
     cred_protect_policy: Option<CredentialProtectionPolicy>,
+    cred_blob: Option<Vec<u8>>,
 }
 
 // The data fields contained in the credential ID are serizlied using CBOR maps.
```
kaczmarczyck: Just noticed I made a type here. Fix this for me please? :)
"serizlied"

src/ctap/credential_id.rs
```diff
@@ -112,10 +116,16 @@ fn decrypt_cbor_credential_id(
             } else {
                 None
             };
+            let cred_blob = if let Some(blob) = cred_blob {
+                Some(extract_byte_string(blob)?)
+            } else {
+                None
+            };
```
kaczmarczyck: Here and with the above `cred_protect_policy`, a shorter way to write this is:
```
let cred_blob = cred_blob.map(extract_byte_string).transpose()?;
```

src/ctap/credential_id.rs
```diff
@@ -162,12 +172,14 @@ pub fn encrypt_to_credential_id(
     private_key: &PrivateKey,
```
kaczmarczyck, line 166: The description needs to be updated:
`/// Encrypts the given private key, relying party ID hash, and cred protect policy into a credential ID.`

Maybe a more general phrasing is easier to maintain.

kaczmarczyck (commented):
Few nits, otherwise LGTM.

src/ctap/credential_id.rs
hcyang-google: done

src/ctap/credential_id.rs
hcyang-google: done

src/ctap/credential_id.rs
hcyang-google: done

kaczmarczyck (approved):

Issue #519, opened 2022-07-28
User presence touch timeouts get treated as user presence


liamjm:
## Expected Behavior
When waiting for user presence, a timeout should lead to an error, eg CTAP2_ERR_USER_ACTION_TIMEOUT

## Actual Behavior
Successful response is returned to host.

## Steps to Reproduce the Problem

1.  send a AuthenticatorMakeCredential command
1.  don't touch the buttons

## Specifications

- Version: develop branch
- Platform: linux

liamjm:
Issue is caused by the result variable from the presence check being overwritten by the result from sending the keepalive packet:

https://github.com/google/OpenSK/blob/develop/src/ctap/mod.rs#L352-L355

```
result = env.user_presence().wait_with_timeout(...);
if !matches!(result, Err(UserPresenceError::Timeout)) {
   break;
 }   
result = send_keepalive_up_needed(env, channel, KEEPALIVE_DELAY);
if result.is_err() {
  break;
}
```

kaczmarczyck:
I checked on the latest develop. If I wait for touch timeout, it returns `CTAP2_ERR_USER_ACTION_TIMEOUT` as expected. Can be closed?

liamjm:
Yep, SGTM.

PR #520, opened 2022-07-28
Add tests to demonstrate broken user presence timeouts

liamjm:
Demonstrates #519 

Run with command like so `python3 ./tools/vendor_hid_test.py  -v CancelTests`

This shows that when a user touch is not present, it timeouts out and returns successful result.

### DIFF
--- a/tools/vendor_hid_test.py
+++ b/tools/vendor_hid_test.py
@@ -69,7 +69,7 @@ class HidDevice(object):
   def ping_init(self, packets=1, byte=0x88) -> int:
     size = ping_data_size(packets)
     ping_packet = [0] + list(self.cid) + [0x81, size // 256, size % 256
-                                          ] + [byte] * 57
+                                         ] + [byte] * 57
     assert len(ping_packet) == _SEND_DATA_SIZE, (
         f'Expected packet to be {_SEND_DATA_SIZE} but was {len(ping_packet)}')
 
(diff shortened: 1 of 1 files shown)


tools/vendor_hid_test.py
```diff
@@ -265,5 +271,87 @@ def test_11_batch_send_and_interleaved_receive_vendor_first(self):
...
+  def setUpClass(cls):
+    cls.fido = get_fido_device()
+
+  def setUp(self) -> None:
+    super().setUp()
+    server = Fido2Server({
+        'id': 'example.com',
+        'name': 'Example RP'
+    },
+                         attestation='direct')
```
kaczmarczyck, line 316: Indentation here is off? But our Python style checkers don't seem to catch it, curious what's going on here.

kaczmarczyck (dismissed):

tools/vendor_hid_test.py
liamjm: This is the output of yapf. As it does look weird, I ran autopep8 and it changed the indenting a little.

I'd prefer not to worry too much about the output of auto-formatting..

liamjm: OK, that autopep8 formatting doesn't pass linting. :-(

Going back to original formatting that is done by yapf.

kaczmarczyck: Then that's what it is :)

kaczmarczyck (approved):

PR #521, opened 2022-07-28
Fix user presence by not overwriting error with OK()

### LINKED ISSUE #519: User presence touch timeouts get treated as user presence

liamjm:
Fixes #519 

The key change is in check_user_presence(), where the result of the user_presence().wait_with_timeout() call is now not overwritten unless the subsequent call to send_keepalive_up_needed() returns an error.

Tests added in #520 demonstrate the bug.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -349,14 +349,14 @@ fn check_user_presence(env: &mut impl Env, channel: Channel) -> Result<(), Ctap2
         // accordingly, so that all wait_with_timeout invocations are separated by
         // equal time intervals. That way token indicators, such as LEDs, will blink
         // with a consistent pattern.
-        let ka_result = send_keepalive_up_needed(env, channel, KEEPALIVE_DELAY);
-        if ka_result.is_err() {
+        let keepalive_result = send_keepalive_up_needed(env, channel, KEEPALIVE_DELAY);
+        if keepalive_result.is_err() {
             debug_ctap!(
                 env,
                 "Sending keepalive failed with error {:?}",
-                ka_result.as_ref().unwrap_err()
+                keepalive_result.as_ref().unwrap_err()
             );
-            result = ka_result;
+            result = keepalive_result;
             break;
         }
     }
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -1523,6 +1523,7 @@ mod test {
     use super::pin_protocol::{authenticate_pin_uv_auth_token, PinProtocol};
     use super::*;
     use crate::api::customization;
+    use crate::api::user_presence::UserPresenceResult;
     use crate::env::test::TestEnv;
     use crate::test_helpers;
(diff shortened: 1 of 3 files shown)


src/ctap/mod.rs
```diff
@@ -349,8 +349,14 @@ fn check_user_presence(env: &mut impl Env, channel: Channel) -> Result<(), Ctap2
         // accordingly, so that all wait_with_timeout invocations are separated by
         // equal time intervals. That way token indicators, such as LEDs, will blink
         // with a consistent pattern.
-        result = send_keepalive_up_needed(env, channel, KEEPALIVE_DELAY);
-        if result.is_err() {
+        let ka_result = send_keepalive_up_needed(env, channel, KEEPALIVE_DELAY);
```
kaczmarczyck: "ka" is not a common acronym for me at least, so please spell it out.

tools/vendor_hid_test.py
```diff
@@ -1,7 +1,12 @@
 """These tests verify the functionality of the VendorHID interface."""
-from typing import Dict, Iterable
```
kaczmarczyck: I will review this file after #520 landed, since the diff should be small? There is not commit with the exact diff.

liamjm: Yeah, sorry, I didn't base this PR off that one. 

The diff will be small once #520 lands.

kaczmarczyck (commented):
#520 is merged!
I had a high-level comment about this PR, but I can't find it, so maybe it got lost:
The fix is in `src/ctap/mod.rs`, so we should be able to write a unit test that catches this bug? Unit tests have value additional to your script since they can be easily automated.

tools/vendor_hid_test.py
liamjm: Done.

liamjm:
> #520 is merged! I had a high-level comment about this PR, but I can't find it, so maybe it got lost: The fix is in `src/ctap/mod.rs`, so we should be able to write a unit test that catches this bug? Unit tests have value additional to your script since they can be easily automated.

YES!

Let me add tests directly to the source - #522 does this.

src/ctap/mod.rs
liamjm: Done.

liamjm:
> > #520 is merged! I had a high-level comment about this PR, but I can't find it, so maybe it got lost: The fix is in `src/ctap/mod.rs`, so we should be able to write a unit test that catches this bug? Unit tests have value additional to your script since they can be easily automated.
> 
> YES!
> 
> Let me add tests directly to the source - #522 does this.

As suggested there, I've added the tests directly to this PR.

kaczmarczyck (approved):

PR #522, opened 2022-08-02
Add unit tests for User presence bug

liamjm:
Further demonstrating #519.

Tests were added in #520, but this adds them directly to the unit test framework, rather than external tests that need the hardware.

To run just these tests, use
`cargo test  --features std,with_ctap1,verbose test_check_user_presence`

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -69,6 +69,8 @@ use crate::api::connection::{HidConnection, SendOrRecvStatus};
 use crate::api::customization::Customization;
 use crate::api::firmware_protection::FirmwareProtection;
 use crate::api::upgrade_storage::UpgradeStorage;
+#[cfg(test)]
+use crate::api::user_presence::UserPresenceResult;
 use crate::api::user_presence::{UserPresence, UserPresenceError};
 use crate::clock::{ClockInt, CtapInstant, KEEPALIVE_DELAY, KEEPALIVE_DELAY_MS};
 use crate::env::Env;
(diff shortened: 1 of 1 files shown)


src/ctap/mod.rs
```diff
@@ -3746,6 +3748,28 @@ mod test {
...
+        // This will always return timeout.
+        fn user_presence_timeout() -> UserPresenceResult {
+            Err(UserPresenceError::Timeout)
+        }
+
+        let mut env = TestEnv::new();
+        env.user_presence().set(user_presence_timeout);
+        let response = check_user_presence(&mut env, DUMMY_CHANNEL);
+        // TODO(https://github.com/google/OpenSK/issues/519): This should be timeout.
+        assert!(matches!(response, Ok(_)));
```
kaczmarczyck, line 3770: Feel free to add this commit to #521, then we can have the correct assert here right away no?

src/ctap/mod.rs
```diff
@@ -69,6 +69,8 @@ use crate::api::connection::{HidConnection, SendOrRecvStatus};
 use crate::api::customization::Customization;
 use crate::api::firmware_protection::FirmwareProtection;
 use crate::api::upgrade_storage::UpgradeStorage;
+#[cfg(test)]
+use crate::api::user_presence::UserPresenceResult;
```
ia0, line 73: This could go at the beginning of the `mod test`.

src/ctap/mod.rs
liamjm: OK, done, added to #521.

src/ctap/mod.rs
liamjm: Yes!

Done.

kaczmarczyck:
This moved to #521, and all comments resolved there. Closing this PR.

PR #524, opened 2022-08-08
Support configure via the Vendor interface

liamjm:
If the vendor_hid feature is enabled, configure must happen through this interface.

This PR makes that happen by patching fido2 lib to select the Vendor HID interface via its usage page.

To get this to work, the GetInfo command was added to the supported commands on the Vender interface as the fido2 lib calls this command. This seems reasonable change anyway.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -3866,7 +3866,6 @@ mod test {
     }
 
     #[test]
-    #[cfg(feature = "vendor_hid")]
     fn test_main_hid() {
         let mut env = TestEnv::new();
         let mut ctap_state = CtapState::new(&mut env, CtapInstant::new(0));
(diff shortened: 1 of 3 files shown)


src/ctap/mod.rs
```diff
@@ -3880,13 +3880,19 @@ mod test {
             response,
             Ok(ResponseData::AuthenticatorGetInfo(_))
         ));
-        let response = ctap_state.process_parsed_command(
-            &mut env,
-            Command::AuthenticatorGetInfo,
-            VENDOR_CHANNEL,
-            CtapInstant::new(0),
-        );
-        assert_eq!(response, Err(Ctap2StatusCode::CTAP1_ERR_INVALID_COMMAND));
```
kaczmarczyck: I think the intent here was to have a negative test as well, maybe we should use something like Reset here instead to check that some commands are rejected on the vendor HID channels?

kaczmarczyck (commented):
This `patch` is neat, good to know about it :)

src/ctap/mod.rs
liamjm: Good point.

I've renamed this test and add a new test based on the old one that tests for a FIDO command that should not be supported on the Vendor HID.

kaczmarczyck (approved):

PR #529, opened 2022-08-16
Vendor HID for upgrades

kaczmarczyck:
Allows sending upgrade over the Vendor HID transport. Also renames the flag in configure to `--vendor-hid` for consistency.

### DIFF
--- a/tools/deploy_partition.py
+++ b/tools/deploy_partition.py
@@ -257,6 +257,6 @@ if __name__ == "__main__":
       default=False,
       action="store_true",
       dest="use_vendor_hid",
-      help=("Whether to configure the device using the Vendor HID interface"),
+      help=("Whether to upgrade the device using the Vendor HID interface."),
   )
   main(parser.parse_args())
--- a/tools/deploy_partition.py
+++ b/tools/deploy_partition.py
@@ -25,6 +25,7 @@ import hashlib
 import os
 import struct
 from typing import Any
+from unittest.mock import patch
 import uuid
 
 import colorama
@@ -172,7 +173,16 @@ def main(args):
       "signature": sign_firmware(signed_data, priv_key)
   }
 
-  for authenticator in tqdm(get_opensk_devices(args.batch)):
+  if args.use_vendor_hid:
+    patcher = patch.object(hid.base, "FIDO_USAGE_PAGE", 0xFF00)
+    patcher.start()
+    info("Using the Vendor HID interface")
+
+  devices = get_opensk_devices(args.batch)
+
+  if not devices:
+    fatal("No devices found.")
+  for authenticator in tqdm(devices):
     # If the device supports it, wink to show which device we upgrade.
     if authenticator.device.capabilities & hid.CAPABILITY.WINK:
       authenticator.device.wink()
@@ -241,5 +251,12 @@ if __name__ == "__main__":
       default="crypto_data/opensk_upgrade.key",
       dest="priv_key",
       help=("PEM file for signing the firmware."),
+  )
+  parser.add_argument(
+      "--vendor-hid",
+      default=False,
+      action="store_true",
+      dest="use_vendor_hid",
+      help=("Whether to configure the device using the Vendor HID interface"),
   )
   main(parser.parse_args())
--- a/tools/perform_upgrade.sh
+++ b/tools/perform_upgrade.sh
@@ -23,7 +23,7 @@ set -e
 BOARD="$1"
 
 ./deploy.py --board=$BOARD --opensk --programmer=none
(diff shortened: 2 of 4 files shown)


tools/deploy_partition.py
```diff
@@ -242,4 +252,11 @@ def main(args):
       dest="priv_key",
       help=("PEM file for signing the firmware."),
   )
+  parser.add_argument(
+      "--vendor-hid",
+      default=False,
+      action="store_true",
+      dest="use_vendor_hid",
+      help=("Whether to configure the device using the Vendor HID interface"),
```
ia0: Copy paste? s/configure/upgrade/

tools/perform_upgrade.sh
```diff
@@ -23,7 +23,7 @@ set -e
 BOARD="$1"
 
 ./deploy.py --board=$BOARD --opensk --programmer=none
-python3 -m tools.deploy_partition --board=$BOARD
+python3 -m tools.deploy_partition --board=$BOARD "$2"
```
ia0: nit: This is not very clean, but if python accepts empty arguments then why not. I see at least 2 solutions:

```
BOARD="$1"
shift

./deploy.py ...
python3 -m ... --board="${BOARD}" "$@"
```
(note that Google style for bash is `"${var}"`, same for deploy.py line)

```
python3 ... --board="${BOARD}" $2
```
(note the absence of double quotes, this means that if $2 is empty, no extra argument is passed, while double quotes would pass an empty argument)

ia0 (dismissed):

tools/deploy_partition.py
kaczmarczyck: Yes, changed.

tools/perform_upgrade.sh
kaczmarczyck: Seems like I didn't actually test the empty argument case after introducing `"$2". It would indeed not have worked. I agree it doesn't look clean, but seemed pretty flexible to me, since this script doesn't really do more than forward everything to Python.

Also I made it so that you also deploy with Vendor HID if you upgrade over Vendor HID, because that should be the most likely use case.

tools/perform_upgrade.sh
```diff
@@ -20,10 +20,10 @@
 
 set -e
 
-BOARD="$1"
+BOARD="${1}"
 
-./deploy.py --board=$BOARD --opensk --programmer=none
-python3 -m tools.deploy_partition --board=$BOARD "$2"
+./deploy.py --board=$BOARD --opensk --programmer=none $2
```
ia0: nit: Would be nice to also fix the `"${BOARD}"` here too :-)

ia0 (dismissed):

tools/perform_upgrade.sh
```diff
@@ -20,10 +20,10 @@
 
 set -e
 
-BOARD="$1"
+BOARD="${1}"
```
ia0: Actually, the Google style guide says to use "$1" here. Those variables are special.

tools/perform_upgrade.sh
ia0: `$2` is already much cleaner than `"$2"`, the only difference with `"$@"` is that it only supports one optional flag instead of an arbitrary number.

tools/perform_upgrade.sh
kaczmarczyck: Oops, forgot that one.

tools/perform_upgrade.sh
kaczmarczyck: Done

tools/perform_upgrade.sh
```diff
@@ -22,8 +22,8 @@ set -e
 
 BOARD="$1"
 
-./deploy.py --board=$BOARD --opensk --programmer=none
-python3 -m tools.deploy_partition --board=$BOARD
+./deploy.py --board=${BOARD} --opensk --programmer=none $2
```
ia0: The double quotes are necessary because you don't want the variable to expand to multiple words.

```
BOARD='foo bar'
"${BOARD}"  # becomes ["foo bar"] with Rust-like syntax
${BOARD}  # becomes ["foo", "bar"] with Rust-like syntax
```

tools/perform_upgrade.sh
kaczmarczyck: Since we don't actually check the content, this could actually be a feature?

ia0: It's definitely a possible choice. It just has more than 90% chance of being useless compared to the other one. But since it doesn't matter for now we can always fix it in the future (at 90% chance).

ia0 (approved):

PR #530, opened 2022-08-17
Add basic setup for multi-PIN

hcyang-google:
- Reserve the storage keys for maximum of 8 user slots.
- Modify the storage functions to take a slot_id parameter.
- Add the slot_count() customization.
- Assume slot_id as a parameter when needed except these places:
  - Entrance functions of command processing that directly takes the
    command parameter structure. slot_id is set as 0, and will be
    parsed from the parameters when we enable the feature.
  - MakeCredential/GetAssertion/AuthenticatorConfig will take the
    slot_id from active token state when we enable the feature,
    resulting in an `Option<usize>`. Below code will act on the option
    value correctly. When the feature isn't enabled, we're always
    referring to the only PIN slot so set slot_id as Some(0).
  - GetInfo returns verdict of whether PIN is supported and enabled, and
    whether PIN needs to be forced changed. There will be new fields to
    represent those values when the feature is enabled, and the old
    fields will not be populated. So when the feature isn't enabled, we
    can treat slot_id as 0.

Not covered in this commit:
- Unittests for other slots. The existing tests all pass and I plan to
  add unittests for multi-slot case after the codebase allows enabling
  the feature.
- Persisting and checking the slot_id in credentials. This is planned to
  come in the next commit.

### DIFF
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -242,6 +242,14 @@ pub trait Customization {
     /// for 10 years.
     fn max_supported_resident_keys(&self) -> usize;
 
+    /// Sets the slot count of the multi-PIN feature.
+    ///
+    /// # Invariant
+    ///
+    /// - The slot count may not make the storage entries that concatenate data
+    ///   of each slots become larger than the storage page size, or exceed the
+    ///   number of keys we reserve for the storage entries that use unique keys
+    ///   for each slot. The upper bound of this is currently 8.
     fn slot_count(&self) -> usize;
 }
 
@@ -428,6 +436,11 @@ pub fn is_valid(customization: &impl Customization) -> bool {
     if customization.max_rp_ids_length() == 0
         && customization.default_min_pin_length_rp_ids().is_empty()
     {
+        return false;
+    }
+
+    // Slot count should be at most 8.
+    if customization.slot_count() > 8 {
         return false;
     }
 
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -241,6 +241,8 @@ pub trait Customization {
     /// With P=20 and K=150, we have I=2M which is enough for 500 increments per day
     /// for 10 years.
     fn max_supported_resident_keys(&self) -> usize;
+
+    fn slot_count(&self) -> usize;
 }
 
 #[derive(Clone)]
@@ -260,6 +262,7 @@ pub struct CustomizationImpl {
     pub max_large_blob_array_size: usize,
     pub max_rp_ids_length: usize,
     pub max_supported_resident_keys: usize,
+    pub slot_count: usize,
 }
 
 pub const DEFAULT_CUSTOMIZATION: CustomizationImpl = CustomizationImpl {
@@ -278,6 +281,7 @@ pub const DEFAULT_CUSTOMIZATION: CustomizationImpl = CustomizationImpl {
     max_large_blob_array_size: 2048,
     max_rp_ids_length: 8,
     max_supported_resident_keys: 150,
+    slot_count: 1,
 };
 
 impl Customization for CustomizationImpl {
@@ -350,6 +354,10 @@ impl Customization for CustomizationImpl {
 
     fn max_supported_resident_keys(&self) -> usize {
         self.max_supported_resident_keys
+    }
+
+    fn slot_count(&self) -> usize {
+        self.slot_count
     }
 }
 
--- a/src/api/customization.rs
+++ b/src/api/customization.rs
@@ -246,10 +246,15 @@ pub trait Customization {
     ///
     /// # Invariant
     ///
-    /// - The slot count may not make the storage entries that concatenate data
-    ///   of each slots become larger than the storage page size, or exceed the
-    ///   number of keys we reserve for the storage entries that use unique keys
-    ///   for each slot. The upper bound of this is currently 8.
+    /// - The slot count may not:
+    ///   - make the storage entries that concatenate data of each slots
+    ///     become larger than the storage page size,
+    ///   - go over u8, as we only reserve 1 byte for the array index for
+    ///     concatenated entries, or
+    ///   - exceed the number of keys we reserve for the storage entries
+    ///     that use unique keys for each slot.
+    ///
+    /// The upper bound of this is currently 8.
     fn slot_count(&self) -> usize;
 }
 
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -166,9 +166,8 @@ impl ClientPin {
         shared_secret: &dyn SharedSecret,
         pin_hash_enc: Vec<u8>,
     ) -> Result<(), Ctap2StatusCode> {
-        if slot_id >= env.customization().slot_count()
-            || slot_id >= self.consecutive_pin_mismatches.len()
-        {
+        // To prevent out of bounds access in code below.
+        if slot_id >= self.consecutive_pin_mismatches.len() {
             return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
         }
         match storage::pin_hash(env, slot_id)? {
@@ -209,9 +208,7 @@ impl ClientPin {
     ) -> Result<AuthenticatorClientPinResponse, Ctap2StatusCode> {
         // TODO: Parse slot_id from params if multi-PIN feature is enabled.
         let slot_id = 0;
-        if slot_id >= env.customization().slot_count()
-            || slot_id >= self.consecutive_pin_mismatches.len()
-        {
+        if slot_id >= self.consecutive_pin_mismatches.len() {
             return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
         }
         Ok(AuthenticatorClientPinResponse {
--- a/src/ctap/client_pin.rs
+++ b/src/ctap/client_pin.rs
@@ -21,12 +21,13 @@ use super::pin_protocol::{verify_pin_uv_auth_token, PinProtocol, SharedSecret};
 use super::response::{AuthenticatorClientPinResponse, ResponseData};
 use super::status_code::Ctap2StatusCode;
 use super::token_state::PinUvAuthTokenState;
+use crate::api::customization::Customization;
 use crate::ctap::storage;
 use crate::env::Env;
 use alloc::boxed::Box;
-use alloc::str;
 use alloc::string::String;
 use alloc::vec::Vec;
+use alloc::{str, vec};
 use crypto::hmac::hmac_256;
 use crypto::sha256::Sha256;
 use crypto::Hash256;
@@ -78,6 +79,7 @@ fn decrypt_pin(
 /// truncated for persistent storage.
 fn check_and_store_new_pin(
     env: &mut impl Env,
+    slot_id: usize,
     shared_secret: &dyn SharedSecret,
     new_pin_enc: Vec<u8>,
 ) -> Result<(), Ctap2StatusCode> {
@@ -90,7 +92,7 @@ fn check_and_store_new_pin(
     let mut pin_hash = [0u8; PIN_AUTH_LENGTH];
     pin_hash.copy_from_slice(&Sha256::hash(&pin[..])[..PIN_AUTH_LENGTH]);
     // The PIN length is always < PIN_PADDED_LENGTH < 256.
-    storage::set_pin(env, &pin_hash, pin_length as u8)?;
+    storage::set_pin(env, slot_id, &pin_hash, pin_length as u8)?;
     Ok(())
 }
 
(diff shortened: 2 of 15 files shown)


src/api/customization.rs
```diff
@@ -241,6 +241,8 @@ pub trait Customization {
     /// With P=20 and K=150, we have I=2M which is enough for 500 increments per day
     /// for 10 years.
     fn max_supported_resident_keys(&self) -> usize;
+
+    fn slot_count(&self) -> usize;
```
hcyang-google, line 258: TODO: Add doc

src/ctap/client_pin.rs
```diff
@@ -159,28 +161,34 @@ impl ClientPin {
...
         pin_uv_auth_protocol: PinUvAuthProtocol,
         shared_secret: &dyn SharedSecret,
         pin_hash_enc: Vec<u8>,
     ) -> Result<(), Ctap2StatusCode> {
-        match storage::pin_hash(env)? {
+        if slot_id >= env.customization().slot_count()
+            || slot_id >= self.consecutive_pin_mismatches.len()
+        {
+            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        }
```
kaczmarczyck: I assume that this parameter is read and sanitized somewhere else in the future? Since these checks return INTERNAL and not something like INVALID_PARAMETER. In that case, since this is a private function, I'd say we can move the check for `env.customization().slot_count()` up to the caller and express the `len` check through

`self.consecutive_pin_mismatches.get_mut().ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?`

or not check at all, since this should generally hold as a invariant (and panic on out of bounds). What do you think?

src/ctap/storage.rs
```diff
@@ -1088,23 +1170,29 @@ mod test {
...
-        assert_eq!(force_pin_change(&mut env), Ok(()));
-        assert!(has_force_pin_change(&mut env).unwrap());
-        assert_eq!(set_pin(&mut env, &[0x88; 16], 8), Ok(()));
-        assert!(!has_force_pin_change(&mut env).unwrap());
+        assert!(!has_force_pin_change(&mut env, 0).unwrap());
+        assert_eq!(force_pin_change(&mut env, 0), Ok(()));
+        assert!(has_force_pin_change(&mut env, 0).unwrap());
+        assert_eq!(set_pin(&mut env, 0, &[0x88; 16], 8), Ok(()));
+        assert!(!has_force_pin_change(&mut env, 0).unwrap());
     }
```
kaczmarczyck, line 1204: All tests in this file only use slot ID 0?

src/ctap/storage/key.rs
```diff
@@ -73,6 +73,35 @@ make_partition! {
...
+    /// Whether the PIN needs to be changed for each slot, except the first.
+    ///
+    /// If this entry exists and is empty, the PIN needs to be changed.
+    FORCE_PIN_CHANGE = 972..978;
+
+    /// The number of PIN retries for each slot, except the first.
+    ///
+    /// Depending on `Customization::slot_count()`, only a prefix of those keys is used.
+    /// If the entry is absent, the number of PIN retries is `Customization::max_pin_retries()`.
+    PIN_RETRIES = 979..985;
```
kaczmarczyck: These ranges are inclusive for the start and exclusive for the end, right? So they are length 6? Can we:

- Add a test that checks that all of these ranges have the same size.
- Make them consecutive?

@ia0 correct? Also, since this restricts the maximum number of slot IDs, we should be confident that 6 is a good number? (Or 7, since we could make the ranges 1 bigger right now.)

Alternative for all slow-changing values like `FORCE_PIN_CHANGE`: Put them behind a single key, and use an array. And make the others a bit bigger instead.

src/env/test/customization.rs
```diff
@@ -112,6 +113,10 @@ impl Customization for TestCustomization {
     fn max_supported_resident_keys(&self) -> usize {
         self.max_supported_resident_keys
     }
+
+    fn slot_count(&self) -> usize {
+        self.slot_count
```
kaczmarczyck, line 118: There should be a test that this number fits storage constraints.

src/ctap/mod.rs
```diff
@@ -838,13 +849,19 @@ impl CtapState {
...
+                {
                     return Err(Ctap2StatusCode::CTAP2_ERR_PUAT_REQUIRED);
                 }
                 0x00
             }
         };
         flags |= UP_FLAG | AT_FLAG;
+        // By this point, we should have already returned the suitable errors in all situations that
+        // slot_id is none.
+        let slot_id = slot_id.ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
```
kaczmarczyck, line 864: What about the case where the feature is not enabled? Not covered by this PR?

src/ctap/mod.rs
```diff
@@ -1203,8 +1230,12 @@ impl CtapState {
         if extensions.hmac_secret.is_some() || extensions.cred_blob {
             flags |= ED_FLAG;
         }
+        // By this point, we should have already returned the suitable errors in all situations that
+        // slot_id is none.
+        let slot_id = slot_id.ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
```
kaczmarczyck, line 1235: Same question as above.

src/ctap/storage/key.rs
ia0: Yes, the ranges use the same notation as Rust.

I agree we should not be afraid to allocate more than we currently need to anticipate possible increase in the future. You can do the same as for credentials and have 2 consecutive ranges: `XXX: A..B` and `_RESERVED_XXX: B..C`. We could take `C - A = 16`.

src/ctap/client_pin.rs
hcyang-google: Mainly to prevent out of bound access. The parameter should be read and suitable error should be return at functions that parse parameters (like process_XXX), so for the internal private functions it's unexpected to have invalid slot_id.

I think expressing len check through `self.consecutive_pin_mismatches.get_mut()` is a good idea.

src/ctap/storage/key.rs
ia0: Also +1 for using arrays for slow-changing values (in particular when they are small).

src/ctap/storage.rs
hcyang-google: All current tests in all files are to test existing behavior, so assuming only one PIN at slot 0.

src/ctap/storage/key.rs
hcyang-google: I think it's my mistake here when arranging the numbers. I meant to make length 8 and consecutive.

src/ctap/mod.rs
hcyang-google: This PR actually only covers cases when the feature isn't enabled.
When it isn't enabled, slot_id will be `Some(0)`, as requests always meant to access the PIN on slot 0.

kaczmarczyck: Got it. Was hard to infer that, given the incomplete state of the PR.

hcyang-google: I provided the rationale in the commit message, do you think that's enough? Or should I add more comments in the code?

kaczmarczyck: Oh, should have read that more carefully, my bad.

src/ctap/client_pin.rs
hcyang-google: As I used indexing more than one time in one of these functions, I think putting one check at the top is cleaner. And for the other function, even if I only use it once, I kind of think making the check at top is slightly better than coming up with
```
power_cycle_state: Some(
                *(self
                    .consecutive_pin_mismatches
                    .get(slot_id)
                    .ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?)
                    >= 3,
            ),
```
instead of just
```
power_cycle_state: Some(self.consecutive_pin_mismatches[slot_id] >= 3),
```

WDYT?

src/ctap/storage/key.rs
hcyang-google: I largely revised the storage structure to directly concat fields that are just u8 (2 of them) and a u32 field (global signature counter). Dealing with u32 field is more tricky and as @ia0 said maybe a library to more generically handle those concatenated fields are better.
I kept using separated keys for pin properties because I'm not comfortable with "directly" concatenating it without any separator/structure.

Welcome for suggestions though, I'm not really satisfied with the current state of the multi-slot storage and I hope we can come up with something both simple enough and generic, future-proof enough.

src/env/test/customization.rs
hcyang-google: done

src/ctap/storage.rs
```diff
@@ -375,37 +382,60 @@ pub fn set_pin(
...
-    match env.store().find(key)? {
+    match env.store().find(key::PIN_RETRIES)? {
         None => Ok(env.customization().max_pin_retries()),
-        Some(value) if value.len() == 1 => Ok(value[0]),
-        _ => Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR),
+        Some(value) => {
+            if slot_id >= value.len() {
+                Ok(env.customization().max_pin_retries())
+            } else {
+                Ok(value[slot_id])
```
kaczmarczyck: PIN retries change pretty often actually. Every time you enter a PIN, they are decreased, and then reset after the check has successfully completed. They should be treated like counters and have separate storage keys.

src/ctap/storage.rs
```diff
@@ -258,16 +253,21 @@ pub fn global_signature_counter(
     env: &mut impl Env,
     slot_id: usize,
 ) -> Result<u32, Ctap2StatusCode> {
-    let key = check_and_get_key_for_slot(
-        env,
-        slot_id,
-        key::FIRST_GLOBAL_SIGNATURE_COUNTER,
-        key::GLOBAL_SIGNATURE_COUNTER,
```
kaczmarczyck: Signature counters change often and should probably live on multiple storage keys.

src/ctap/storage.rs
hcyang-google: Done

src/ctap/storage.rs
hcyang-google: Done

src/ctap/storage/key.rs
```diff
@@ -114,21 +140,19 @@ make_partition! {
...
     /// If the entry is absent, the number of PIN retries is `Customization::max_pin_retries()`.
-    PIN_RETRIES = 2044;
+    FIRST_PIN_RETRIES = 2044;
 
     /// The PIN hash and length.
     ///
     /// If the entry is absent, there is no PIN set. The first byte represents
     /// the length, the following are an array with the hash.
-    PIN_PROPERTIES = 2045;
+    FIRST_PIN_PROPERTIES = 2045;
```
kaczmarczyck, line 151: Double checking: PIN properties can not be an array in one key, because that could be too long? 8 hashes of 16 byte + a PIN length could fit into one key on Nordic, at least.

hcyang-google: Because it might be harder to migrate when we ever want to change the structure of PIN_PROPERTIES. I can wait until Julien's PR to land though, which provides an elegant solution.

kaczmarczyck: We can submit as is, and then do a new PR that merges #531 and uses it.

hcyang-google: I think I can wait, that CL is quite ready too. I'll prefer not to reserve key fields in a commit and remove them in upcoming commit.

hcyang-google: done

src/ctap/storage.rs
```diff
@@ -359,26 +367,23 @@ pub fn set_pin(
...
-    let properties_key =
-        check_and_get_key_for_slot(env, slot_id, key::FIRST_PIN_PROPERTIES, key::PIN_PROPERTIES)?;
-    let mut value = env.store().find(key::FORCE_PIN_CHANGE)?.unwrap_or_default();
-    value.resize(env.customization().slot_count(), 0);
-    *(value
+    let mut force_pin_change_value = env.store().find(key::FORCE_PIN_CHANGE)?.unwrap_or_default();
+    force_pin_change_value.resize(env.customization().slot_count(), 0);
+    *(force_pin_change_value
         .get_mut(slot_id)
         .ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?) = 0;
```
kaczmarczyck: You are not using the `concat` interface for `FORCE_PIN_CHANGE` because it's simple enough to do it yourself?

src/ctap/client_pin.rs
kaczmarczyck: Not checking at all in private functions is also totally okay! As long as CTAP inputs and the storage is sanitized.

src/ctap/storage/key.rs
kaczmarczyck: Seems to be resolved.

kaczmarczyck (approved):
No strong opinion on my last comment, LGTM.

src/ctap/storage.rs
hcyang-google: Yes, but if you prefer I use concat for consistency that's also ok.

kaczmarczyck: No strong opinion, so feel free to merge.

hcyang-google: I switched to using concatenated value, can you take a last look? Thanks

kaczmarczyck (approved):

kaczmarczyck (approved):

PR #531, opened 2022-08-19
Add support for concatenated values

ia0:
This could be useful for the multi-PIN work. It's an autonomous module like `fragment`. So if it's not used it doesn't cost anything.

### DIFF
--- a/libraries/persistent_store/src/concat.rs
+++ b/libraries/persistent_store/src/concat.rs
@@ -1,0 +1,242 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Support for concatenated entries.
+//!
+//! This module permits to store multiple indexed values under the same key by concatenation. Such
+//! values must be at most 255 bytes and there can't be more than 255 such values under the same
+//! key (they are indexed with a `u8`).
+//!
+//! The rationale for using those particular constraints is that we want the number of bits to store
+//! the index and the number of bits to store the length to fit in an integer number of bytes
+//! (because the values are an integer number of bytes). Using only one byte is too restrictive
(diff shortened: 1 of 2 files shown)


hcyang-google (approved):

libraries/persistent_store/src/concat.rs
```diff
@@ -0,0 +1,242 @@
...
+) -> StoreResult<()> {
+    if value.len() > 255 {
+        return Err(StoreError::InvalidArgument);
+    }
+    let mut values = store.find(key)?.unwrap_or(vec![]);
+    match find(&values, index)? {
+        None => {
+            values.push(index);
+            values.push(value.len() as u8);
+            values.extend_from_slice(value);
```
kaczmarczyck, line 59: Interesting, so it's actually a map. So if we know we will usually use increasing indices, we could also choose to not store the index. But this works for me, just wanted to point out that our use case probably doesn't need it.

libraries/persistent_store/src/concat.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn read_empty_entry() {
+        let store = MINIMAL.new_store();
+        assert_eq!(read(&store, 0, 0), Ok(None));
+        assert_eq!(read(&store, 0, 1), Ok(None));
+    }
+
+    #[test]
+    fn read_missing_value() {
+        let mut store = MINIMAL.new_store();
+        let value = b"\x00\x03foo\x02\x05hello".to_vec();
```
kaczmarczyck, line 133: The tests are arguably more maintainable if they insert into the store through `concat`'s `write`, in case we'd ever change the implementation. But we have to write arrays explicitly for the invalid storage tests, so works for me.

kaczmarczyck (approved):
LGTM, just 2 comments that you may ignore and merge.

libraries/persistent_store/src/concat.rs
ia0: Yes, I thought about this but I actually think that we will eventually need it. Because we can't really ensure we have contiguous slots. What if a user deletes a slot in the middle? We would need to create a transaction with all slot-based entries to shift all slots by one. I think using a map is more robust to future changes.

libraries/persistent_store/src/concat.rs
ia0: Yes I think that's arguable. Those tests are really unit tests (they test a single function) rather than functional tests (they test global behavior). Because the implementation is part of the specification, testing the implementation is actually testing the specification. Also, I don't think we should change the implementation because that would be a breaking change, which is why the implementation is specified.

kaczmarczyck: Okay, SGTM.

PR #532, opened 2022-08-22
Set bumpalo version for fuzzing

kaczmarczyck:
Fixes CI problems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,6 @@ embedded-time = "0.12.1"
 arbitrary = { version = "0.4.7", features = ["derive"], optional = true }
 rand = { version = "0.8.4", optional = true }
 ed25519-compact = { version = "1", default-features = false, optional = true }
-# We import to explicitly lock the version.
-serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 
 [features]
 debug_allocations = ["lang_items/debug_allocations"]
@@ -41,6 +39,8 @@ ed25519 = ["ed25519-compact"]
 
 [dev-dependencies]
 enum-iterator = "0.6.0"
+# We import to explicitly lock the version.
+serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 
 [build-dependencies]
 sk-cbor = { path = "libraries/cbor" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ arbitrary = { version = "0.4.7", features = ["derive"], optional = true }
 rand = { version = "0.8.4", optional = true }
 ed25519-compact = { version = "1", default-features = false, optional = true }
 # We import to explicitly lock the version.
-# Remove these dependencies once CTAP is a library.
 serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,13 @@ persistent_store = { path = "libraries/persistent_store" }
 byteorder = { version = "1", default-features = false }
 arrayref = "0.3.6"
 subtle = { version = "2.2", default-features = false, features = ["nightly"] }
-# This import explicitly locks the version.
-serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 embedded-time = "0.12.1"
 arbitrary = { version = "0.4.7", features = ["derive"], optional = true }
 rand = { version = "0.8.4", optional = true }
 ed25519-compact = { version = "1", default-features = false, optional = true }
+# We import to explicitly lock the version.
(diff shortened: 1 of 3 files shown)


fuzz/Cargo.toml
```diff
@@ -13,6 +13,7 @@ libfuzzer-sys = { version = "0.3" }
 fuzz_helper = { path = "fuzz_helper" }
 # This import explicitly locks the version.
 serde_json = { version = "=1.0.69" }
+bumpalo = "=3.8.0"
```
ia0, line 16: This should ideally be done in the `Cargo.lock` file, but we probably don't have it committed. So I would at least add a comment as to why we add a transitive dependency as a direct dependency, and that we should remove this once CTAP is a library and we can finally run fuzzing with a normal compiler.

kaczmarczyck: Done, in all files that need it.

Cargo.toml
```diff
@@ -20,12 +20,13 @@ persistent_store = { path = "libraries/persistent_store" }
...
 subtle = { version = "2.2", default-features = false, features = ["nightly"] }
-# This import explicitly locks the version.
-serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
 embedded-time = "0.12.1"
 arbitrary = { version = "0.4.7", features = ["derive"], optional = true }
 rand = { version = "0.8.4", optional = true }
 ed25519-compact = { version = "1", default-features = false, optional = true }
+# We import to explicitly lock the version.
+# Remove these dependencies once CTAP is a library.
+serde_json = { version = "=1.0.69", default-features = false, features = ["alloc"] }
```
ia0: Why do we need this? We should not depend on serde unless possibly for "std". But then in that case we should make this dependency optional and add it only for "std".

kaczmarczyck: It's another case were I had to lock the version because of compilation issues. I can probably move it to dev-dependencies, let me try.

ia0: But I don't understand where do we use `serde_json`?

kaczmarczyck: #407 is where I added the serde version. After running `cargo update` as described there, desktop tests still pass.

ia0: Yes, I've seen that PR and would have mentioned that `serde_json` doesn't need to be added there because we shouldn't depend on it.

I just did the following without issues:
```
./reset.sh
# Remove serde_json from Cargo.toml
./setup.sh
./run_desktop_tests.sh
```

So I'm really curious why we need `serde_json` in the ctap2 crate.

kaczmarczyck: In both stable and develop, `serde` is part of the initial commit for `libraries/crypto`. Copied from stable:

```
serde = { version = "1.0", optional = true, features = ["derive"] }
serde_json = { version = "1.0", optional = true }
```

I never checked if we actually need these dependencies, but I removed all of them and it still seems to work. Interesting.

ia0: The crypto library uses them for tests (wycheproof). But ctap2 doesn't.

kaczmarczyck: Oh, we responded at the same time it seems. I had the same conclusion, so let's try to remove serde!

libraries/crypto/Cargo.toml
```diff
@@ -17,10 +17,8 @@ byteorder = { version = "1", default-features = false }
 hex = { version = "0.3.2", default-features = false, optional = true }
 ring = { version = "0.16.11", optional = true }
 untrusted = { version = "0.7.0", optional = true }
-serde = { version = "1.0", optional = true, features = ["derive"] }
-serde_json = { version = "=1.0.69", optional = true }
 regex = { version = "1", optional = true }
 
 [features]
-std = ["hex", "ring", "rng256/std", "untrusted", "serde", "serde_json", "regex"]
```
ia0: I think this file was correct. We use `serde_json` for tests.

kaczmarczyck: Ah, I see. In that case, I think I had random issues with the serde version before I added it to the main `Cargo.toml` as well. Not sure how to reproduce, so maybe we can just remove it outside of crypto and try again until that happens.

kaczmarczyck: Ah, I see. In that case, I think I had random issues with the serde version before I added it to the main `Cargo.toml` as well. Not sure how to reproduce, so maybe we can just remove it outside of crypto and try again until that happens.

ia0 (approved):

libraries/crypto/Cargo.toml
ia0: Yes, now the change is making sense.

I think it was hard to reproduce because of some many states (the `Cargo.lock` file**s** in multiple directories). The best way to test is `./reset.sh` (committing code first otherwise it's lost) but it can be a bit slow for fast iteration.

PR #533, opened 2022-08-23
Set bumpalo version for fuzzing

kaczmarczyck:
Backport of #532

### DIFF
--- a/libraries/crypto/Cargo.toml
+++ b/libraries/crypto/Cargo.toml
@@ -23,6 +23,11 @@ serde = { version = "1.0", optional = true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 regex = { version = "1", optional = true }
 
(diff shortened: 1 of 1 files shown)


fuzz/Cargo.toml
```diff
@@ -11,6 +11,9 @@ cargo-fuzz = true
 [dependencies]
 libfuzzer-sys = { version = "0.3"}
 fuzz_helper = { path = "fuzz_helper" }
+# We import to explicitly lock the version.
+# Remove this dependency once CTAP is a library.
+bumpalo = "=3.8.0"
```
ia0: `bumpalo` is actually only needed by the crypto library (because of `ring`). So let's move this line there.

ia0: Or let's cherry-pick https://github.com/google/OpenSK/pull/534 instead.

ia0 (approved):

PR #536, opened 2022-08-23
Backport bumpalo to stable

kaczmarczyck:
After merging, stable and bugfix should coincide.

### DIFF
--- a/.github/workflows/boards_build.yml
+++ b/.github/workflows/boards_build.yml
@@ -12,7 +12,7 @@ jobs:
   build_boards:
     strategy:
       matrix:
-        os: [ubuntu-18.04, macos-10.15]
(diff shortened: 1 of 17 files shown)


ia0 (dismissed):

jmichelp (dismissed):

kaczmarczyck:
Looks like we also need to backport the workflow fix from develop. Note that I ignored bugfix with that latest commit and went straight onto stable :)

kaczmarczyck:
Looks like we also need to backport the workflow fix from develop. Note that I ignored bugfix with that latest commit and went straight onto stable :)

ia0 (approved):

jmichelp (approved):

PR #537, opened 2022-08-29
Tool fixes

kaczmarczyck:
Fixes some issues in our Python scripts.

`load_priv_key` takes a string to make it easier to import and reuse this function. The argparse `FileType` has a few limitations, i.e. it is harder to make it optional.

### DIFF
--- a/tools/deploy_partition.py
+++ b/tools/deploy_partition.py
@@ -150,8 +150,8 @@ def load_priv_key(priv_key_filename: str) -> Any:
         fatal("Private key must be 256 bits long.")
       info("Private key is valid.")
       return priv_key
-  except IOError:
-    fatal(f"Unable to open file: {priv_key_filename}")
+  except IOError as e:
+    fatal(f"Unable to open file: {priv_key_filename}\n{e}")
 
 
 def sign_firmware(data: bytes, priv_key: Any) -> bytes:
--- a/tools/deploy_partition.py
+++ b/tools/deploy_partition.py
@@ -97,7 +97,7 @@ def check_info(partition_address: int, authenticator: Any):
     if result[0x01] != partition_address:
       fatal("Identifiers do not match.")
   except ctap.CtapError as ex:
-    fatal(f"Failed to read OpenSK upgrade info (error: {ex}")
+    fatal(f"Failed to read OpenSK upgrade info (error: {ex})")
 
 
 def get_kernel(board: str) -> bytes:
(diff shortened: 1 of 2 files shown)


tools/deploy_partition.py
```diff
@@ -97,7 +97,7 @@ def check_info(partition_address: int, authenticator: Any):
     if result[0x01] != partition_address:
       fatal("Identifiers do not match.")
   except ctap.CtapError as ex:
-    error(f"Failed to read OpenSK upgrade info (error: {ex}")
+    fatal(f"Failed to read OpenSK upgrade info (error: {ex}")
```
jmichelp: Missing closing parenthese

tools/deploy_partition.py
```diff
@@ -159,6 +160,8 @@ def sign_firmware(data: bytes, priv_key: Any) -> bytes:
 
 def main(args):
   colorama.init()
+  if not args.priv_key:
```
jmichelp: This test is already done a few lines later, making it redundant.

tools/deploy_partition.py
```diff
@@ -137,17 +137,18 @@ def generate_firmware_image(board: str) -> bytes:
...
-  priv_key = get_private_key(priv_key_file.read())
-  if not isinstance(priv_key, ec.EllipticCurvePrivateKey):
-    fatal("Private key must be an Elliptic Curve one.")
-  if not isinstance(priv_key.curve, ec.SECP256R1):
-    fatal("Private key must use Secp256r1 curve.")
-  if priv_key.key_size != 256:
-    fatal("Private key must be 256 bits long.")
-  info("Private key is valid.")
-  return priv_key
+  with open(priv_key_filename, "rb") as priv_key_file:
```
jmichelp: `argparse.FileType` checks that the file exists and can be opened. Such test must now be backported here before opening the file to avoid exiting by uncaught exceptions.
If you aim at reusability, why not keeping `argparse.FileType` for the argument parsing but changing `load_priv_key` to take `bytes` (i.e. the content of the file)?

kaczmarczyck: I'd prefer that this function is self-sufficient and easy to use. I don't think we will ever want to read a private key from anything but a file.
Therefore, if we want to catch these exceptions, let's catch them.

tools/deploy_partition.py
kaczmarczyck: Unintended artifact from rebasing.

tools/deploy_partition.py
kaczmarczyck: Ah another one :)

tools/deploy_partition.py
```diff
@@ -137,17 +137,21 @@ def generate_firmware_image(board: str) -> bytes:
...
+      if not isinstance(priv_key, ec.EllipticCurvePrivateKey):
+        fatal("Private key must be an Elliptic Curve one.")
+      if not isinstance(priv_key.curve, ec.SECP256R1):
+        fatal("Private key must use Secp256r1 curve.")
+      if priv_key.key_size != 256:
+        fatal("Private key must be 256 bits long.")
+      info("Private key is valid.")
+      return priv_key
+  except IOError:
+    fatal(f"Unable to open file: {priv_key_filename}")
```
jmichelp: nit: add the exception message to the output for an easier debugging.

jmichelp (dismissed):

tools/deploy_partition.py
kaczmarczyck: Done

jmichelp (approved):

PR #539, opened 2022-08-30
New metadata format

kaczmarczyck:
This PR changes to metadata format to include the signature, and a version instead of a timestamp.

tl;dr tell me how to shuffle the format around:
- 32 B upgrade hash (SHA256)
- 64 B signature,
- 32 B padding,
- 20 B version (8 byte used) and
-  4 B partition address in little endian encoding.

There are a few things we can discuss about the format:

1. The version is currently `u64`, since the GetInfo command returns an integer, and our CBOR parser works with up to 63 bit integers. We could make it more sophisticated like SEMVER, but would have to convert it to 63 bit eventually.
2. More padding before the signed part of the metadata starts. Would allow us to write different / bigger / PQC signatures there. I wanted to get your opinions before deciding on a number.

Some things are missing:
- Versions are only checked in the bootloader, but not yet in CTAP. This is consistent with existing behavior, but will probably be stricter soon.
- Code has not moved to TockEnv yet. The command will stay in CTAP, but the Metadata logic will end up there.

### DIFF
--- a/bootloader/src/main.rs
+++ b/bootloader/src/main.rs
@@ -48,12 +48,9 @@ unsafe fn read_page(address: usize) -> Page {
 /// Parsed metadata for a firmware partition.
 struct Metadata {
     checksum: [u8; 32],
-    timestamp: u32,
+    _signature: [u8; 64],
+    version: u64,
     address: u32,
-}
-
-impl Metadata {
-    pub const DATA_LEN: usize = 40;
 }
 
 /// Reads the metadata from a flash page.
@@ -61,8 +58,9 @@ impl From<Page> for Metadata {
     fn from(page: Page) -> Self {
         Metadata {
             checksum: page[0..32].try_into().unwrap(),
-            timestamp: LittleEndian::read_u32(&page[32..36]),
-            address: LittleEndian::read_u32(&page[36..Metadata::DATA_LEN]),
+            _signature: page[32..96].try_into().unwrap(),
+            version: LittleEndian::read_u64(&page[140..][..8]),
+            address: LittleEndian::read_u32(&page[148..][..4]),
         }
     }
 }
@@ -77,14 +75,14 @@ impl BootPartition {
     const FIRMWARE_LENGTH: usize = 0x00040000;
 
     /// Reads the metadata, returns the timestamp if all checks pass.
-    pub fn read_timestamp(&self) -> Result<u32, ()> {
+    pub fn read_version(&self) -> Result<u64, ()> {
         let metadata_page = unsafe { read_page(self.metadata_address) };
         let hash_value = self.compute_upgrade_hash(&metadata_page);
         let metadata = Metadata::from(metadata_page);
         if self.firmware_address != metadata.address as usize {
             #[cfg(debug_assertions)]
             rprintln!(
-                "Firmware address mismatch: expected 0x{:08X}, metadata 0x{:08X}",
+                "Partition address mismatch: expected 0x{:08X}, metadata 0x{:08X}",
                 self.firmware_address,
                 metadata.address as usize
             );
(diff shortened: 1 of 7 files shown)


src/ctap/mod.rs
```diff
@@ -212,45 +211,55 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
...
+    public_key_bytes: &[u8],
     metadata: &[u8],
-) -> Result<[u8; 32], Ctap2StatusCode> {
-    const METADATA_LEN: usize = 40;
+) -> Result<(), Ctap2StatusCode> {
+    const METADATA_LEN: usize = 0x1000;
     if metadata.len() != METADATA_LEN {
         return Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER);
     }
+    let metadata_address = LittleEndian::read_u32(&metadata[148..][..4]);
```
ia0: Shouldn't we ideally use constants for offsets in the metadata? (0, 32, 140, 148)

src/ctap/mod.rs
```diff
@@ -212,45 +211,55 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
...
+    if metadata_address as usize != upgrade_locations.partition_address() {
+        return Err(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER);
+    }
     // The hash implementation handles this in chunks, so no memory issues.
     let partition_slice = upgrade_locations
         .read_partition(0, upgrade_locations.partition_length())
         .map_err(|_| Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
     let mut hasher = Sha256::new();
+    hasher.update(&metadata[128..METADATA_LEN]);
```
ia0: I didn't see where it is documented that the hash starts at offset 128.

ia0 (dismissed):
I didn't look too much in the Python logic.

Regarding how much non-signed padding, I guess it depends how much bigger checksums and signatures can be in the future. I don't know but looks like the metadata allocated size is quite huge (4096 bytes), so there's room. Why not just split in 2 parts? [0-2048) non-signed, and [2048-4096) signed?

bootloader/src/main.rs
```diff
@@ -77,14 +75,14 @@ impl BootPartition {
     const FIRMWARE_LENGTH: usize = 0x00040000;
 
     /// Reads the metadata, returns the timestamp if all checks pass.
```
zhalvorsen:
```suggestion
    /// Reads the metadata, returns the firmware version if all checks pass.
```

zhalvorsen (dismissed):

src/ctap/mod.rs
kaczmarczyck: I added a constant and updated all documentation.

src/ctap/mod.rs
kaczmarczyck: I kept some numbers that only matter inside the function that uses them. I can also make them const, if you like. I went with your suggestion to use half pages.

src/ctap/mod.rs
```diff
@@ -211,37 +211,43 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
 
 /// Parses the metadata of an upgrade, and checks its correctness.
 ///
-/// The metadata consists of:
+/// The metadata is a page starting with:
```
ia0, line 214: By the way, this is specific to Nordic. A page is not necessarily 4kB. It makes sense for the bootloader which is Nordic-specific, but ctap-code should ideally be abstract over a page size. But I guess it's fine to keep it like that or rename "page" to "slice" or something.

ia0 (approved):

src/ctap/mod.rs
kaczmarczyck: Ah yeah, no problem. This is moving to Env in 2 PRs.

PR #540, opened 2022-08-31
Public Key plain byte encoding

kaczmarczyck:
Encoding the public key with it's bytes directly is similar to other hardware platforms. We don't lose flexibilty in practise, since upgrade to a different signature algorithm requires an upgrade that understands it anyway.

Also includes a small fix in `data_formats.rs` I discovered on the way.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -47,7 +47,7 @@ use self::credential_id::{
 use self::credential_management::process_credential_management;
 use self::crypto_wrapper::PrivateKey;
 use self::data_formats::{
-    AuthenticatorTransport, CoseKey, CredentialProtectionPolicy, EnterpriseAttestationMode,
+    AuthenticatorTransport, CredentialProtectionPolicy, EnterpriseAttestationMode,
     GetAssertionExtensions, PackedAttestationStatement, PinUvAuthProtocol,
     PublicKeyCredentialDescriptor, PublicKeyCredentialParameter, PublicKeyCredentialSource,
     PublicKeyCredentialType, PublicKeyCredentialUserEntity, SignatureAlgorithm,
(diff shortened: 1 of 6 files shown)


src/ctap/mod.rs
```diff
@@ -261,14 +261,15 @@ fn parse_metadata(
...
 ) -> Result<(), Ctap2StatusCode> {
     let signature = ecdsa::Signature::from_bytes(signature_bytes)
         .ok_or(Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)?;
-    let cbor_public_key = cbor_read(public_key_bytes)?;
-    let cose_key = CoseKey::try_from(cbor_public_key)?;
-    let public_key = ecdsa::PubKey::try_from(cose_key)?;
+    let x = array_ref!(public_key_bytes, 0, 32);
+    let y = array_ref!(public_key_bytes, 32, 32);
+    let public_key = ecdsa::PubKey::from_coordinates(x, y)
+        .ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
```
ia0, line 270: Looks like we could use `PubKey::from_bytes_uncompressed` and directly the 65 bytes in `build.rs`. No strong preference, but `from_bytes_uncompressed` would make both call sites smaller for just 1 more byte stored in the binary.

ia0 (dismissed):

src/ctap/mod.rs
kaczmarczyck: Good idea, done. Let's check binary size. Currently:
```
OLD 115.2 kiB
NEW 115.0 kiB
```

kaczmarczyck: Still the same now.

ia0 (approved):

PR #542, opened 2022-08-31
Firmware version for upgrades

kaczmarczyck:
Firmware versions are now checked in CTAP and the Python tool, plus shown in GetInfo.

### DIFF
--- a/boards/nordic/nrf52840dk_opensk_a/build.rs
+++ b/boards/nordic/nrf52840dk_opensk_a/build.rs
@@ -11,7 +11,7 @@ fn main() {
     fs::write(
         &dest_path,
         "
-static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 5] = [
+static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 6] = [
     // We implement NUM_PAGES = 20 as 16 + 4 to satisfy the MPU.
     kernel::StorageLocation {
         address: 0xC0000,
@@ -24,6 +24,11 @@ static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 5] = [
         storage_type: kernel::StorageType::Store,
     },
     // Partitions can also be split to maximize MPU happiness.
+    kernel::StorageLocation {
+        address: 0x4000,
+        size: 0x1000,
+        storage_type: kernel::StorageType::Partition,
+    },
     kernel::StorageLocation {
         address: 0x5000,
         size: 0x1000,
--- a/boards/nordic/nrf52840dk_opensk_a/build.rs
+++ b/boards/nordic/nrf52840dk_opensk_a/build.rs
@@ -11,7 +11,7 @@ fn main() {
     fs::write(
         &dest_path,
         "
-static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 6] = [
+static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 5] = [
     // We implement NUM_PAGES = 20 as 16 + 4 to satisfy the MPU.
     kernel::StorageLocation {
         address: 0xC0000,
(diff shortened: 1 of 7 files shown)


src/env/test/upgrade_storage.rs
```diff
@@ -84,6 +84,10 @@ impl UpgradeStorage for BufferUpgradeStorage {
             Err(StorageError::OutOfBounds)
         }
     }
+
+    fn running_firmware_version(&self) -> u64 {
+        0
```
zhalvorsen, line 89: Should this be retrieved from flash or is it going to be a hard-coded version in software?

zhalvorsen: Maybe I got confused. The CtapState.firmware_version will be calling the `running_firmware_version` from storage.rs right?

kaczmarczyck: This is `TestEnv`, meaning this is just a dummy that responds something to allow testing. The interesting implementation is inside `src/env/tock/storage.rs`.

src/env/tock/storage.rs
```diff
@@ -100,6 +101,10 @@ fn block_command(driver: usize, cmd: usize, arg1: usize, arg2: usize) -> Storage
     }
 }
 
+fn read_slice(address: usize, length: usize) -> &'static [u8] {
```
ia0: This function is unsafe without preconditions. We should either mark it `unsafe` to make call-site aware that there is a precondition to check, or document the precondition and hope it's not ignored.

```rust
/// Provides read-access to a slice from flash.
///
/// # Safety
///
/// The slice must not be modified during the lifetime of the program. For example, it should not point to the storage.

File: boards/nordic/nrf52840dk_opensk_a/build.rs [hunk 3]
```diff
@@ -24,6 +24,11 @@ static mut STORAGE_LOCATIONS: [kernel::StorageLocation; 5] = [
         storage_type: kernel::StorageType::Store,
     },
     // Partitions can also be split to maximize MPU happiness.
+    kernel::StorageLocation {
+        address: 0x4000,
+        size: 0x1000,
+        storage_type: kernel::StorageType::Partition,
```
ia0: I guess we can't merge those 2 locations because of MMU? Same for the 2 locations below.

kaczmarczyck: We can actually for metadata, just have to assume some knowledge inside Env. The other partitions not so much, since MPU restrictions.

src/env/tock/storage.rs
kaczmarczyck: I made the function unsafe.

src/env/tock/storage.rs
```diff
@@ -277,19 +277,35 @@ impl TockUpgradeStorage {
...
+                    locations.running_metadata =
+                        ModRange::new(metadata_range.start(), locations.page_size);
+                    locations.metadata = ModRange::new(
+                        metadata_range.start() + locations.page_size,
+                        locations.page_size,
+                    );
+                }
+                _ => (),
+            }
+            match range.start() {
```
ia0, line 281: Why do we have 2 match? We enter the same branch in each. They can be merged.

src/env/tock/storage.rs
```diff
@@ -277,19 +277,35 @@ impl TockUpgradeStorage {
...
+                Self::METADATA_ADDRESS => {
+                    metadata_range = range;
```
ia0: Why not split here and do the swap in partition B as before?

kaczmarczyck: Simplified, PTAL

src/env/tock/storage.rs
kaczmarczyck: Done

ia0 (approved):

PR #543, opened 2022-09-02
New Upgrade Interface

kaczmarczyck:
Concludes the rework of the upgrade command.

Changes to the CTAP API:
- The address is now mandatory, and called offset internally. Before, not offet was meant metadata. Now the metadata is simply part of the partition. In our case, it sits at offset 0.

Changes to the Env API:
- Removed the metadata read and write, as metadata now is part of the writeable partition
- Renamed `partition_address` to `partition_identifier`, as it more clearly explains its role. It doesn't necessarily have to actually be the address, it's just what CTAP forwards as info about the partition you can write to.

The partition logic can now handle ranges that are not connected. The helper struct is in `api/` to be env independent and easily testable.

### DIFF
--- a/src/api/upgrade_storage/helper.rs
+++ b/src/api/upgrade_storage/helper.rs
@@ -204,16 +204,13 @@ impl Partition {
     pub fn ranges_from(&self, start_address: usize) -> Vec<ModRange> {
         let mut result = Vec::new();
         for range in &self.ranges {
-            if start_address >= range.start() && start_address - range.start() >= range.length() {
-                continue;
-            }
-            if start_address > range.start() {
-                if start_address - range.start() < range.length() {
-                    let length = range.length() - (start_address - range.start());
-                    result.push(ModRange::new(start_address, length));
+            match start_address.checked_sub(range.start()) {
+                None | Some(0) => result.push(range.clone()),
+                Some(offset) => {
+                    if range.length() > offset {
+                        result.push(ModRange::new(start_address, range.length() - offset));
+                    }
                 }
-            } else {
-                result.push(range.clone());
             }
         }
         result
--- a/src/api/upgrade_storage/helper.rs
+++ b/src/api/upgrade_storage/helper.rs
@@ -15,6 +15,7 @@
 // For compiling with std outside of tests.
 #![cfg_attr(feature = "std", allow(dead_code))]
 
+use alloc::vec::Vec;
 use core::iter::Iterator;
 use persistent_store::{StorageError, StorageResult};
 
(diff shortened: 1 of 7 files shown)


src/api/upgrade_storage/helper.rs
```diff
@@ -128,6 +150,76 @@ impl ModRange {
...
+    }
+
+    pub fn ranges_from(&self, start_address: usize) -> Vec<ModRange> {
+        let mut result = Vec::new();
+        for range in &self.ranges {
+            if start_address >= range.start() && start_address - range.start() >= range.length() {
+                continue;
+            }
+            if start_address > range.start() {
+                if start_address - range.start() < range.length() {
```
ia0: This condition is always true. Let's just keep the "then" branch.

src/api/upgrade_storage/helper.rs
```diff
@@ -128,6 +150,76 @@ impl ModRange {
...
+                    let length = range.length() - (start_address - range.start());
+                    result.push(ModRange::new(start_address, length));
+                }
+            } else {
+                result.push(range.clone());
+            }
```
ia0, line 214: The loop iteration could be simplified to:

```rust
match start_address.checked_sub(range.start()) {
    // start_address <= range.start()
    None | Some(0) => result.push(range.clone()),
    Some(offset) => match range.length().checked_sub(offset) {
        // start_address >= range.start() + range.length()
        None | Some(0) => (),
        Some(length) => result.push(ModRange::new(start_address, length),
    }
}
```

src/env/tock/storage.rs
```diff
@@ -330,44 +367,39 @@ impl UpgradeStorage for TockUpgradeStorage {
         if data.is_empty() {
             return Err(StorageError::OutOfBounds);
         }
-        let address = self.partition.start() + offset;
-        let write_range = ModRange::new(address, data.len());
-        if self.partition.contains_range(&write_range) {
+        if let Some(address) = self.partition.find_address(offset, data.len()) {
```
ia0: Why not the following which avoids indentation?

```rust
let address = self.partition.find_address(offset, data.len()).ok_or(StorageError::OutOfBounds)?;
```

kaczmarczyck: Done

src/api/upgrade_storage/helper.rs
kaczmarczyck: Solved together with the below.

src/api/upgrade_storage/helper.rs
kaczmarczyck: I like the first `checked_sub`, but replaced the second with a simple `if`, given that we actually ignore one of the match arms.

ia0 (approved):

PR #546, opened 2022-09-21
Parse slot_id in use from token state

hcyang-google:
Support switching to multi-PIN feature and making the default slots 8. But the command to switch to multi-PIN isn't exposed yet because the implementation isn't ready.

Main change of this commit is to cache the slot_id in use inside token state, and retrieve it from token state when needed.

I was thinking about unittest when I added more code and opened up the multi-PIN feature, and I will focus on the following types of unittests:
- storage test cases for multi-PIN: this can already be added, I'll do it in an upcoming PR.
- MakeCredential/GetAssertion test cases for multi-PIN: wait until the PR that enables checking/persisting slot_id in credentials.
- ClientPin test cases for multi-PIN: wait until the PR that parses slot_id from ClientPin command params.
- test cases for commands to be added/updated: should come with the PRs that add those.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,11 @@ sk-cbor = { path = "libraries/cbor" }
 uuid = { version = "0.8", features = ["v4"] }
 openssl = "0.10.36"
 
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = "=1.14" # transitive dependency of openssl
+
 [profile.dev]
 panic = "abort"
 lto = true # Link Time Optimization usually reduces size of binaries and static libraries
--- a/libraries/crypto/Cargo.toml
+++ b/libraries/crypto/Cargo.toml
@@ -25,6 +25,7 @@ regex = { version = "1", optional = true }
 # Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
 # whole block once CTAP is a library.
 bumpalo = "=3.8.0" # transitive dependency of ring
+once_cell = { version = "=1.14", optional = true } # transitive dependency of ring
 
 [features]
 std = ["hex", "ring", "rng256/std", "untrusted", "serde", "serde_json", "regex"]
--- a/src/ctap/config_command.rs
+++ b/src/ctap/config_command.rs
@@ -91,14 +91,10 @@ pub fn process_config(
         pin_uv_auth_protocol,
     } = params;
 
-    let slot_id = if storage::has_multi_pin(env)? {
-        client_pin.get_slot_id_in_use()
-    } else {
-        Some(0)
-    };
+    let slot_id = client_pin.get_slot_id_in_use_or_default(env)?;
     let enforce_uv =
         !matches!(sub_command, ConfigSubCommand::ToggleAlwaysUv) && storage::has_always_uv(env)?;
-    // If multi-PIN feature is enabled, no PIN is in used, and the command is to turn off alwaysUv,
+    // If multi-PIN feature is enabled, no PIN is in use, and the command is to turn off alwaysUv,
     // the PIN check will be skipped here but an OPERATION_DENIED will still be returned later,
     // which is correct behavior.
     if (slot_id.is_some() && storage::pin_hash(env, slot_id.unwrap())?.is_some()) || enforce_uv {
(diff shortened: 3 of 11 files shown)


Cargo.toml
```diff
@@ -45,6 +45,11 @@ sk-cbor = { path = "libraries/cbor" }
 uuid = { version = "0.8", features = ["v4"] }
 openssl = "0.10.36"
 
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = "=1.13.1" # transitive dependency of openssl
```
kaczmarczyck: 1.14 still works for me.

src/ctap/config_command.rs
```diff
@@ -91,10 +91,16 @@ pub fn process_config(
...
-    // TODO: Get the slot id from active token state when multi-PIN is enabled.
-    let slot_id = Some(0);
+    let slot_id = if storage::has_multi_pin(env)? {
+        client_pin.get_slot_id_in_use()
+    } else {
+        Some(0)
+    };
     let enforce_uv =
         !matches!(sub_command, ConfigSubCommand::ToggleAlwaysUv) && storage::has_always_uv(env)?;
+    // If multi-PIN feature is enabled, no PIN is in used, and the command is to turn off alwaysUv,
```
kaczmarczyck: "in use" or "used"

libraries/crypto/Cargo.toml
```diff
@@ -26,6 +26,11 @@ regex = { version = "1", optional = true }
 # whole block once CTAP is a library.
 bumpalo = "=3.8.0" # transitive dependency of ring
 
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = { version = "=1.13.1", optional = true } # transitive dependency of ring
+
 [features]
 std = ["hex", "ring", "rng256/std", "untrusted", "serde", "serde_json", "regex"]
```
ia0, line 31: Not sure how this works, but you don't need to add `once_cell` to the list here (i.e. such that it is enabled when `ring` is enabled).

kaczmarczyck (approved):

src/ctap/mod.rs
```diff
@@ -778,8 +778,11 @@ impl CtapState {
             enterprise_attestation,
         } = make_credential_params;
 
-        // TODO: Get the slot id from active token state when multi-PIN is enabled.
-        let slot_id = Some(0);
+        let slot_id = if storage::has_multi_pin(env)? {
+            self.client_pin.get_slot_id_in_use()
+        } else {
+            Some(0)
+        };
```
ia0: Looks like this pattern could be a helper function in ClientPin, e.g. get_slot_id_in_use_or_default(env) which returns the current slot id if multipin otherwise 0.

src/ctap/storage/key.rs
```diff
@@ -73,6 +73,9 @@ make_partition! {
     // - When adding a (non-persistent) key below this message, make sure its value is bigger or
     //   equal than NUM_PERSISTENT_KEYS.
 
+    /// If this entry exists and is empty, multi-PIN is enabled.
```
ia0: nit: I would be more precise and specify as:

> Whether multi-PIN is enabled.
>
> The value must be empty. Only presence of the value matters.

or something along those lines.

libraries/crypto/Cargo.toml
kaczmarczyck: By the way, the comment is a duplicate, and it says "whole block". So I think we can just list `once_cell` right below `bumpalo`.

kaczmarczyck: Oh, and I realized that it's also a transitive build dependency for `fuzz/`.

kaczmarczyck: Actually, I just didn't properly cargo clean before trying, and it seems to work without it.

src/ctap/storage/key.rs
hcyang-google: done

src/ctap/mod.rs

libraries/crypto/Cargo.toml
hcyang-google: cherry-picked your fix on develop

src/ctap/config_command.rs
hcyang-google: done

kaczmarczyck (approved):

PR #548, opened 2022-09-21
Fix once_cell dependency

kaczmarczyck:
Backporting @hcyang-google 's version fix that should also affect other branches. See #546

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,11 @@ sk-cbor = { path = "libraries/cbor" }
 uuid = { version = "0.8", features = ["v4"] }
 openssl = "0.10.36"
 
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = { version = "=1.14", optional = true } # transitive dependency of ring
+
 [profile.dev]
 panic = "abort"
 lto = true # Link Time Optimization usually reduces size of binaries and static libraries
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ openssl = "0.10.36"
 # We explicitly lock the version of those transitive dependencies because their
 # Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
 # whole block once CTAP is a library.
-once_cell = { version = "=1.14", optional = true } # transitive dependency of ring
+once_cell = { version = "=1.14", optional = true } # transitive dependency of openssl
 
 [profile.dev]
 panic = "abort"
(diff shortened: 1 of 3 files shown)


kaczmarczyck:
Failing bloat is intentional, as it tries to run the code without this PR.

fuzz/Cargo.toml
```diff
@@ -12,6 +12,12 @@ cargo-fuzz = true
 libfuzzer-sys = { version = "0.3" }
 fuzz_helper = { path = "fuzz_helper" }
 
+[build-dependencies]
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = { version = "=1.14", optional = true } # transitive dependency of ring
```
ia0: This shouldn't be needed.

Cargo.toml
```diff
@@ -45,6 +45,11 @@ sk-cbor = { path = "libraries/cbor" }
 uuid = { version = "0.8", features = ["v4"] }
 openssl = "0.10.36"
 
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = { version = "=1.14", optional = true } # transitive dependency of ring
```
ia0: This was openssl in the other PR. Also not optional.

fuzz/Cargo.toml
kaczmarczyck: I checked cargo tree, and once_cell is listed as a dependency of openssl. I noticed that I copy-pasted `ring` in the comment everywhere (about when you wrote your review), does that resolve why I had to add it to fuzz?

Cargo.toml
```diff
@@ -45,6 +45,11 @@ sk-cbor = { path = "libraries/cbor" }
 uuid = { version = "0.8", features = ["v4"] }
 openssl = "0.10.36"
 
+# We explicitly lock the version of those transitive dependencies because their
+# Cargo.toml don't parse with the nightly compiler used by Tock. Remove this
+# whole block once CTAP is a library.
+once_cell = { version = "=1.14", optional = true } # transitive dependency of openssl
```
ia0: This shouldn't be optional. But if it works I guess it does for the same reason as the "not sure it works" comment in the other PR.

fuzz/Cargo.toml
ia0: The comment still applies, I don't see fuzzing depending on openssl.

Cargo.toml
kaczmarczyck: Done

Cargo.toml
kaczmarczyck: Done

fuzz/Cargo.toml
kaczmarczyck: Ah right, it was just just transitive through the ctap module. And it failed for me before probably because I didn't cargo clean after testing something else.

ia0 (approved):

PR #551, opened 2022-09-29
Simplifies Env upgrade API

kaczmarczyck:
The CTAP command logic is now a 1-liner, taking the CBOR-parsed data and calling the corresponding Env API.

### DIFF
--- a/src/api/upgrade_storage/mod.rs
+++ b/src/api/upgrade_storage/mod.rs
@@ -19,22 +19,20 @@ pub(crate) mod helper;
 
 /// Accessors to storage locations used for upgrading from a CTAP command.
 pub trait UpgradeStorage {
-    /// Writes the given data to the given offset address, if within bounds of the partition.
+    /// Writes the given data as part of an upgrade.
     ///
-    /// The offset is relative to the start of the partition, excluding holes. The partition is
-    /// presented as one connected component. Therefore, the offset does not easily translate
-    /// to physical memory address address of the slice.
+    /// The offset indicates the data location inside the bundle.
     ///
     /// The hash is the SHA256 of the data slice. This hash is not a security feature, use it to
     /// check your data integrity.
     ///
     /// # Errors
     ///
-    /// - Returns [`StorageError::OutOfBounds`] if the data does not fit the partition.
+    /// - Returns [`StorageError::OutOfBounds`] if the data does not fit.
     /// - Returns [`StorageError::CustomError`] if any Metadata or hash check fails.
-    fn write_bundle(&mut self, offset: usize, data: Vec<u8>, hash: &[u8; 32]) -> StorageResult<()>;
+    fn write_bundle(&mut self, offset: usize, data: Vec<u8>) -> StorageResult<()>;
 
-    /// Returns an identifier for the partition.
+    /// Returns an identifier for the requested bundle.
     ///
     /// Use this to determine whether you are writing to A or B.
     fn bundle_identifier(&self) -> u32;
--- a/src/api/upgrade_storage/mod.rs
+++ b/src/api/upgrade_storage/mod.rs
@@ -18,35 +18,25 @@ pub(crate) mod helper;
 
 /// Accessors to storage locations used for upgrading from a CTAP command.
 pub trait UpgradeStorage {
-    /// Reads a slice of the partition, if within bounds.
-    ///
-    /// The offset is relative to the start of the partition, excluding holes. The partition is
-    /// presented as one connected component. Therefore, the offset does not easily translate
-    /// to physical memory address address of the slice.
-    ///
-    /// # Errors
-    ///
-    /// Returns [`StorageError::OutOfBounds`] if the requested slice is not inside the partition.
-    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]>;
-
     /// Writes the given data to the given offset address, if within bounds of the partition.
     ///
     /// The offset is relative to the start of the partition, excluding holes.
     /// See `read_partition`.
     ///
+    /// The hash is the SHA256 of the data slice. This hash is not a security feature, use it to
+    /// check your data integrity.
+    ///
     /// # Errors
     ///
     /// - Returns [`StorageError::OutOfBounds`] if the data does not fit the partition.
-    /// - Returns [`StorageError::CustomError`] if any Metadata check fails.
-    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()>;
+    /// - Returns [`StorageError::CustomError`] if any Metadata or hash check fails.
+    fn write_partition(&mut self, offset: usize, data: &[u8], hash: &[u8; 32])
+        -> StorageResult<()>;
 
     /// Returns an identifier for the partition.
     ///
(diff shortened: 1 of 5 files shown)


src/api/upgrade_storage/mod.rs
```diff
@@ -18,36 +18,26 @@ pub(crate) mod helper;
...
-    ///
-    /// Returns [`StorageError::OutOfBounds`] if the requested slice is not inside the partition.
-    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]>;
-
     /// Writes the given data to the given offset address, if within bounds of the partition.
     ///
     /// The offset is relative to the start of the partition, excluding holes.
     /// See `read_partition`.
     ///
+    /// The hash is the SHA256 of the data slice. This hash is not a security feature, use it to
```
ia0: Why isn't the hash checked by CTAP?

src/api/upgrade_storage/mod.rs
```diff
@@ -18,36 +18,26 @@ pub(crate) mod helper;
...
+    /// check your data integrity.
+    ///
     /// # Errors
     ///
     /// - Returns [`StorageError::OutOfBounds`] if the data does not fit the partition.
-    /// - Returns [`StorageError::CustomError`] if any Metadata check fails.
-    fn write_partition(&mut self, offset: usize, data: &[u8]) -> StorageResult<()>;
+    /// - Returns [`StorageError::CustomError`] if any Metadata or hash check fails.
+    fn write_partition(&mut self, offset: usize, data: &[u8], hash: &[u8; 32])
```
ia0: Should we rename partition to bundle or payload?

ia0 (dismissed):

src/api/upgrade_storage/mod.rs
kaczmarczyck: We want to have the option to write lazily. Different scenarios:
- The chunk is too small to be written immediately, so we buffer. Then `read_partition` would need to be more complicated to emulate the buffered state.
- We want to manipulate the input, i.e. mask it before writing. Again, `read_partition` complexity.

We could of course only check the data that is sent before write, not after. This decision seems like an implementation detail to me that we should leave to Env.

ia0: I think we should check before writing (more precisely, as soon as we receive the chunk). The hash is only to check the transfer between client and authenticator. The rest is up to the authenticator (and thus Env).

src/api/upgrade_storage/mod.rs
kaczmarczyck: Done. Also I decided to move the data as a Vec, to give Env the chance to be smart with it's memory management.

src/api/upgrade_storage/mod.rs
kaczmarczyck: Why is that better than checking after writing, if possible?

ia0: Because of separation of concerns. This is the hash of the chunk, so it should be checked when we get the chunk and before we use it, otherwise we use the wrong chunk. If we want to make sure we write what we think, then this should be implemented at the storage layer. And if the upgrade mechanism want another integrity check on top of everything, then it can roll its own hash of what it wants to write and check that in the end.

src/api/upgrade_storage/mod.rs
```diff
@@ -20,8 +21,9 @@ pub(crate) mod helper;
 pub trait UpgradeStorage {
     /// Writes the given data to the given offset address, if within bounds of the partition.
     ///
-    /// The offset is relative to the start of the partition, excluding holes.
-    /// See `read_partition`.
+    /// The offset is relative to the start of the partition, excluding holes. The partition is
```
ia0: This still says partition. It also talks about address which I think is confusing. The environment is fully responsible to interpret a bundle.

ia0 (dismissed):

src/api/upgrade_storage/mod.rs
kaczmarczyck: I removed the hash from the API.

src/api/upgrade_storage/mod.rs
kaczmarczyck: I removed all implementation details from the documentation, and fixed a few typos etc.

src/api/upgrade_storage/mod.rs
```diff
@@ -12,41 +12,30 @@
...
 
+use alloc::vec::Vec;
 use persistent_store::StorageResult;
 
 pub(crate) mod helper;
 
 /// Accessors to storage locations used for upgrading from a CTAP command.
 pub trait UpgradeStorage {
-    /// Reads a slice of the partition, if within bounds.
+    /// Writes the given data as part of an upgrade.
```
ia0: nit: We could say "Processes" instead of "Writes".

src/api/upgrade_storage/mod.rs
```diff
@@ -12,41 +12,30 @@
...
-    ///
-    /// Returns [`StorageError::OutOfBounds`] if the requested slice is not inside the partition.
-    fn read_partition(&self, offset: usize, length: usize) -> StorageResult<&[u8]>;
-
-    /// Writes the given data to the given offset address, if within bounds of the partition.
-    ///
-    /// The offset is relative to the start of the partition, excluding holes.
-    /// See `read_partition`.
+    /// The hash is the SHA256 of the data slice. This hash is not a security feature, use it to
+    /// check your data integrity.
```
ia0: leftover

ia0 (dismissed):

src/ctap/mod.rs
```diff
@@ -1399,19 +1399,14 @@ impl CtapState {
...
-            .map_err(|_| Ctap2StatusCode::CTAP1_ERR_INVALID_PARAMETER)?;
-        let written_hash = Sha256::hash(written_slice);
-        if hash != written_hash {
+        let calculated_hash = Sha256::hash(&data);
+        if hash != calculated_hash {
             return Err(Ctap2StatusCode::CTAP2_ERR_INTEGRITY_FAILURE);
         }
+        env.upgrade_storage()
+            .ok_or(Ctap2StatusCode::CTAP1_ERR_INVALID_COMMAND)?
+            .write_bundle(offset, data)
```
zhalvorsen, line 1408: I still like the idea of checking the hash after it has been written in flash in case corruption happens during the write. What do you think of pulling the hash calculation into `write_bundle`?

kaczmarczyck: That's how it was built before, and we decided to separate the checks instead:
https://github.com/google/OpenSK/pull/551#discussion_r984737093

What we do now is check the hash in CTAP, and compare the slices in Env. Would that work?

src/api/upgrade_storage/mod.rs
kaczmarczyck: Done

src/api/upgrade_storage/mod.rs
kaczmarczyck: Removed

ia0 (approved):

src/ctap/mod.rs
zhalvorsen: Yeah, that works for me

PR #564, opened 2022-10-21
Respect `CARGO_TARGET_DIR` env var

### LINKED ISSUE #563: Respect `CARGO_TARGET_DIR` env var instead of hardcoding target directory
## Expected Behavior

I have set the environment variable `CARGO_TARGET_DIR` to a custom directory.
What is `CARGO_TARGET_DIR`: [Reference](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads)

> Location of where to place all generated artifacts, relative to the current working directory. See [build.target-dir](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir) to set via config.

The deploy script should respect the env var and be able to resolve the path.

## Actual Behavior

The deploy script does not respect the env var and assumes the default hardcoded `target` directory, relative to the current working directory. This causes a failing deployment of the firmware.

1. Set `CARGO_TARGET_DIR` to a custom directory, e.g. `CARGO_TARGET_DIR=custom_dir`
1. Run the deploy script and observe the file not found errors

L0g4n:
This change resolves `CARGO_TARGET_DIR` when set instead of hardcoding Cargos `target/` dir.

Fixes #563 

- [x] Tests pass

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -160,9 +160,7 @@ SUPPORTED_BOARDS = {
 # `src/entry_point.rs`
 APP_HEAP_SIZE = 90000
 
-CARGO_TARGET_DIR = os.environ["CARGO_TARGET_DIR"] \
-  if os.environ.get("CARGO_TARGET_DIR") else "target"
-
+CARGO_TARGET_DIR = os.environ.get("CARGO_TARGET_DIR", "target")
 
 def get_supported_boards() -> Tuple[str]:
   """Returns a tuple all valid supported boards."""
--- a/deploy.py
+++ b/deploy.py
@@ -162,6 +162,7 @@ APP_HEAP_SIZE = 90000
 
 CARGO_TARGET_DIR = os.environ.get("CARGO_TARGET_DIR", "target")
 
+
 def get_supported_boards() -> Tuple[str]:
   """Returns a tuple all valid supported boards."""
   boards = []
@@ -402,9 +403,8 @@ class OpenSKInstaller:
     env["RUSTFLAGS"] = " ".join(rust_flags)
     cargo_command = ["cargo", "build", "--release", f"--target={props.arch}"]
     self.checked_command(cargo_command, cwd="bootloader", env=env)
-    binary_path = os.path.join(
-      CARGO_TARGET_DIR, props.arch, "release", "bootloader"
-    )
+    binary_path = os.path.join(CARGO_TARGET_DIR, props.arch, "release",
+                               "bootloader")
     objcopy_command = [
         "llvm-objcopy", "-O", "binary", binary_path, f"{binary_path}.bin"
     ]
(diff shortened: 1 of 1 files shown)


kaczmarczyck:
Hi, thanks for the pull request. You can locally reproduce the Python lint problems using:
```
./tools/run_pylint.sh
```
Or try something like this, but it is not fully equivalent to pylint:
```
yapf --style=yapf --recursive --exclude third_party -i .
```

deploy.py
```diff
@@ -160,6 +160,9 @@
 # `src/entry_point.rs`
 APP_HEAP_SIZE = 90000
 
+CARGO_TARGET_DIR = os.environ["CARGO_TARGET_DIR"] \
```
kaczmarczyck: `CARGO_TARGET_DIR = os.environ.get("CARGO_TARGET_DIR", "target")`

should also work? The second parameter is the default, if not present.

kaczmarczyck (commented):
One minor style improvement, otherwise looks good!

deploy.py
L0g4n: Ah, yes. It looked kinda weird for being to long in the first place. I did not look at the python docs :smile:

L0g4n: Ah fail, I thought you changed it already ^^

kaczmarczyck:
One of the Python lint workflows was still failing. Not sure if you have access to the error, so I copied it for reference:

```
env["RUSTFLAGS"] = " ".join(rust_flags)
     cargo_command = ["cargo", "build", "--release", f"--target={props.arch}"]
     self.checked_command(cargo_command, cwd="bootloader", env=env)
-    binary_path = os.path.join(
-      CARGO_TARGET_DIR, props.arch, "release", "bootloader"
-    )
+    binary_path = os.path.join(CARGO_TARGET_DIR, props.arch, "release",
+                               "bootloader")
     objcopy_command = [
         "llvm-objcopy", "-O", "binary", binary_path, f"{binary_path}.bin"
     ]
@@ -816,8 +8[15](https://github.com/google/OpenSK/actions/runs/3297670137/jobs/5438836688#step:5:16),7 @@
       if self.args.application:
         self.install_padding()
         self.install_tab_file(
-          f"{CARGO_TARGET_DIR}/tab/{self.args.application}.tab"
-        )
+            f"{CARGO_TARGET_DIR}/tab/{self.args.application}.tab")
         self.install_metadata()
         if not self.verify_flashed_app(self.args.application):
           error(("It seems that something went wrong. App/example not found "
Error: Process completed with exit code 1.
```

L0g4n:
> One of the Python lint workflows was still failing. Not sure if you have access to the error, so I copied it for reference:

Ah yeah sorry, I thought you only needed to run either pyling or yapf but not both at the same time.
Anyway, it should be fixed now.

kaczmarczyck (approved):

kaczmarczyck:
Works, thanks!

PR #550, opened 2022-09-26
Fixed proc-macro2 version

kaczmarczyck:
Fixes #549

Pins the version for our proc-macro2 dependency.

PR #567, opened 2022-10-24
Add storage test cases for multi-PIN

hcyang-google:
I was thinking after adding duplicate test cases for multi-PIN, should we remove the original test case as it is totally covered by mult-PIN case. Removing duplicate code is nice, but I think separating out multi-PIN tests also has advantage as multi-PIN isn't a common use case, and might make the original test case's logic look more complicated. WDYT?


hcyang-google:
@kaczmarczyck ready for review, the second commit might need to be cherry-picked back to develop.

kaczmarczyck (commented):
The feature itself looks good.
For fixing the build, maybe we should do it the way develop does? Reduces maintanance and keeps the code closer to develop. We submitted the Cargo.lock files, see
https://github.com/google/OpenSK/pull/550/files

hcyang-google:
Fixed, I cherry-picked the commit in dev.

kaczmarczyck (approved):

kaczmarczyck:
The binary size report workflow must fail, since it tries to build the code without this PR. Therefore looks good!

PR #569, opened 2022-10-28
Persist/parse slot_id in/from credential

hcyang-google:
Persist slot_id into credential_id or the resident credential record during MakeCredential, and parse it during GetAssertion. Add related unittests.

### DIFF
--- a/src/ctap/credential_management.rs
+++ b/src/ctap/credential_management.rs
@@ -81,6 +81,7 @@ fn enumerate_credentials_response(
         user_icon,
         cred_blob: _,
         large_blob_key,
+        slot_id: _,
     } = credential;
     let user = PublicKeyCredentialUserEntity {
         user_id: user_handle,
@@ -183,12 +184,23 @@ fn process_enumerate_credentials_begin(
         .rp_id_hash
         .ok_or(Ctap2StatusCode::CTAP2_ERR_MISSING_PARAMETER)?;
     client_pin.has_no_or_rp_id_hash_permission(&rp_id_hash[..])?;
+    let slot_id = if let Some(slot_id) = client_pin.get_slot_id_in_use_or_default(env)? {
+        slot_id
+    } else {
+        // enumerateCredentials needs UV, so slot_id must not be None.
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+    };
     let mut iter_result = Ok(());
     let iter = storage::iter_credentials(env, &mut iter_result)?;
     let mut rp_credentials: Vec<usize> = iter
         .filter_map(|(key, credential)| {
             let cred_rp_id_hash = Sha256::hash(credential.rp_id.as_bytes());
-            if cred_rp_id_hash == rp_id_hash.as_slice() {
+            let slot_id_matches = if let Some(cred_slot_id) = credential.slot_id {
+                cred_slot_id == slot_id
+            } else {
+                slot_id == 0
+            };
+            if cred_rp_id_hash == rp_id_hash.as_slice() && slot_id_matches {
                 Some(key)
             } else {
                 None
@@ -385,6 +397,7 @@ mod test {
             user_icon: Some("icon".to_string()),
             cred_blob: None,
             large_blob_key: None,
+            slot_id: None,
         }
     }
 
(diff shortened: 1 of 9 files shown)


src/ctap/credential_management.rs
```diff
@@ -183,12 +184,23 @@ fn process_enumerate_credentials_begin(
         .rp_id_hash
         .ok_or(Ctap2StatusCode::CTAP2_ERR_MISSING_PARAMETER)?;
     client_pin.has_no_or_rp_id_hash_permission(&rp_id_hash[..])?;
+    let slot_id = if let Some(slot_id) = client_pin.get_slot_id_in_use_or_default(env)? {
+        slot_id
+    } else {
+        // enumerateCredentials needs UV, so slot_id must not be None.
+        return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
```
kaczmarczyck: What about credentials without UV that were created before activating multi PIN?

src/ctap/credential_management.rs
```diff
@@ -183,12 +184,23 @@ fn process_enumerate_credentials_begin(
...
+    };
```
kaczmarczyck: Simpler:
```
let slot_id = client_pin.get_slot_id_in_use_or_default(env)?
                        .ok_or(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
```

src/ctap/credential_management.rs
```diff
@@ -183,12 +184,23 @@ fn process_enumerate_credentials_begin(
...
     let iter = storage::iter_credentials(env, &mut iter_result)?;
     let mut rp_credentials: Vec<usize> = iter
         .filter_map(|(key, credential)| {
             let cred_rp_id_hash = Sha256::hash(credential.rp_id.as_bytes());
-            if cred_rp_id_hash == rp_id_hash.as_slice() {
+            let slot_id_matches = if let Some(cred_slot_id) = credential.slot_id {
+                cred_slot_id == slot_id
+            } else {
+                slot_id == 0
+            };
```
kaczmarczyck:
```
let slot_id_matches = credential.slot_id.unwrap_or(0) == slot_id;
```

Or alternatively `unwrap_or_default`. But I like the explicit 0 better.

src/ctap/mod.rs
```diff
@@ -759,6 +759,33 @@ impl CtapState {
...
+    fn check_slot_id_matches(
+        &mut self,
+        credential: &PublicKeyCredentialSource,
+        slot_id: usize,
+    ) -> bool {
+        if let Some(cred_slot_id) = credential.slot_id {
+            cred_slot_id == slot_id
+        } else {
+            slot_id == 0
+        }
```
kaczmarczyck: This is a one-liner that is duplicated in credential management.

src/ctap/mod.rs
```diff
@@ -26,10 +26,10 @@ pub mod hid;
 pub mod key_material;
 mod large_blobs;
 pub mod main_hid;
-mod pin_protocol;
+pub(crate) mod pin_protocol;
 pub mod response;
 pub mod status_code;
-mod storage;
+pub(crate) mod storage;
```
kaczmarczyck: I assume this is for tests? I suspect that there is a design issue if all these crates have to suddenly become public.

src/test_helpers/mod.rs
```diff
@@ -60,3 +64,28 @@ pub fn enable_enterprise_attestation(
...
+        key_agreement_key,
+        pin_uv_auth_token,
+        pin_uv_auth_protocol,
+    );
+    state.set_client_pin_for_test(client_pin);
+    storage::set_pin(env, slot_id, &[0x88; 16], 4)?;
+
+    Ok(authenticate_pin_uv_auth_token(
+        &pin_uv_auth_token,
+        &DUMMY_CLIENT_DATA_HASH,
```
kaczmarczyck: `DUMMY_CLIENT_DATA_HASH` should be passed in, as it seems to be an implementation detail of the caller of `enable_pin_uv`.

src/ctap/mod.rs
hcyang-google: Yes, any better ideas of this? test_helpers isn't under ctap but it needs to access many private things inside.

src/ctap/credential_management.rs
hcyang-google: This is parsing from the ClientPIN state, not from credential.

src/test_helpers/mod.rs
hcyang-google: done

src/ctap/mod.rs

src/ctap/credential_management.rs
hcyang-google: done

src/ctap/credential_management.rs
hcyang-google: done

src/ctap/credential_management.rs
kaczmarczyck: Oops nevermind

src/ctap/mod.rs
kaczmarczyck: We should probably achieve what we want through the public API. While this is more work the first few times we do this, it should become easier the more helpers we have. WDYT?

hcyang-google:
I decided to move that helper back to ctap/mod.rs, such that we don't need to make those changes in this PR. Feels like there many caveats to support ClientPIN related setup in test_helpers, and we should do that separately.

kaczmarczyck (approved):

PR #571, opened 2022-12-06
Add linear view into a storage

ia0:


libraries/persistent_store/src/linear.rs
```diff
@@ -0,0 +1,359 @@
...
+        // We must erase the page, so we save the rest.
+        let prefix = self.save_slice(index.page, 0..index.byte)?;
+        let suffix = self.save_slice(index.page, index.byte + value.len()..page_size)?;
+        self.storage.erase_page(index.page)?;
+        index.byte = 0;
+        self.write_unaligned(index, &prefix)?;
+        index.byte += prefix.len();
+        self.write_unaligned(index, value)?;
+        index.byte += value.len();
+        self.write_unaligned(index, &suffix)
```
kaczmarczyck: This part looks like `erase_within`, I was wondering if we can reuse it here?

kaczmarczyck (dismissed):

libraries/persistent_store/src/linear.rs
ia0: Yes, I created a `Complement` concept.

libraries/persistent_store/src/linear.rs
```diff
@@ -123,23 +123,17 @@ impl<S: Storage> Linear<S> {
     ///
     /// This function will erase the page if needed and restore the non-overwritten part. This is
     /// not power-loss resistant.
-    fn write_within(&mut self, mut index: StorageIndex, value: &[u8]) -> StorageResult<()> {
-        let page_size = self.storage.page_size();
+    fn write_within(&mut self, index: StorageIndex, value: &[u8]) -> StorageResult<()> {
```
kaczmarczyck, line 126: Is this function equivalent to:
```
self.erase_within(index, value.len());
self.write_unaligned(index, value)
```
?

ia0: No because if the slice is already erased, it just writes.

ia0: Oh actually we do this check there, so yes.

kaczmarczyck (approved):

PR #580, opened 2023-01-17
Overview: Add tock v2 support

### LINKED ISSUE #579: Interest in Tock 2.x Port?
Hello there,

as I am working toward getting OpenSK running on the OpenTitan FPGA platform, I successfully ported the existing `develop` branch from the Tock 1.x base to Tock 2.1.x. This includes all the necessary breaking fixes such as the new [SyscallDriver](https://docs.tockos.org/kernel/syscall/trait.syscalldriver) trait with the new `ro_allow` and `rw_allow` as well as the new userspace part.

Of course this has still rough edges, e.g. the bootloader and examples are still untouched and hence do not work, and a fork of the `libtock-rs` to get static heap support (the v2 version of libtock-rs has still a lot of rough edges, unfortunately). However, most rough edges in my fork are due to the fact of the top priority to get it to work on the existing `nrf5280` boards in the first place, which it does, I tested in on both the development board and dongle in practice. 

Those things can be smoothed out, however, this only really makes sense if you, the project owners, are interested in this kind of stuff in the first place. If yes, I could submit a draft pull request in the first place and then we could see where this leads. Otherwise I would leave it in my fork untouched for others to potentially use that version, but of course, contributing back to upstream would be better since it has much better visibility :smile:.

Let me know what you think.

L0g4n:
Fixes #579 

This is the starting point for the maintainers to get an overview of all changes (and can be split up, after further discussion) and hence this PR has not been opened with the intention to merge in this state.  

## Important things to note

> This has still some very rough edges, hence, this is considered a WIP
commit. Next to the missing `firmware_protect` driver for TockOs (for disabling JTAG support and preventing dumping of the firmware), this includes non-working examples. 
> 
> Additionally, it is important to note that OpenSK can not be built anymore on x86 host machines, since `libtock_runtime` (part of libtock-rs 2.0 WIP) is only running on ARM and riscv hosts. 
Hence a normal invocation of `cargo build` or `cargo check` on your average dev machine will fail, but you can still cross-compile for ARM or RISCV and/or run `cargo check --features std` to see if the code compiles fine, but you **need** to apply the *alloc_init* patch for the libtock-rs submodule before running the build commands.

These circumstances probably mean that with the current CI config the pipeline will for sure fail and needs further modifications. Let me know what you think, where problematic areas are, and how we can continue from here.

- [ ] Tests pass


ia0:
Thanks @L0g4n for the work you put in this PR!

@kaczmarczyck is currently on vacation and I'm a currently a bit busy, but I'll take a look next week and provide some comments. Some early comment though that might simplify your workflow: You can use the `./maintainers/patches` script to deal with the tock and libtock-rs patches. Running the command without argument explains how to use it, but the main idea is to use `apply` to get one commit per patch and a clean state, then use `save` once the commits have been modified (including adding new commits) to update the patches.

L0g4n:
> You can use the ./maintainers/patches script to deal with the tock and libtock-rs patches. Running the command without argument explains how to use it, but the main idea is to use apply to get one commit per patch and a clean state, then use save once the commits have been modified (including adding new commits) to update the patches.

Thanks, will try do that. I originally had multiple patches but then just merged it one big file since some of the stuff commits in there revert some earlier changes since this has been a bit of an exploratory journey.

L0g4n:
Quick question regarding the `./maintainers/patches` script: After reading the help text it seems I can add new patches by submitting changes to the previous patches, and then adding commits which reference the previous patch files. What about potentially new patches in this case since this does not seem to be mentioned in the help text?

ia0:
> After reading the help text it seems I can add new patches by submitting changes to the previous patches, and then adding commits which reference the previous patch files. What about potentially new patches in this case since this does not seem to be mentioned in the help text?

I'm not sure I understand the question, but I'll try another explanation which might help.

The logic of the script is to provide you a way to sync between the patches and the chain of commits of the submodule. The `apply` method synchronizes assuming the patches are the source of truth (i.e. it copies the patches to the commits). The `save` method does the opposite (i.e. it copies the commits to the patches). The order of the chain of commits is the alphabetical order of the patches. In particular, this means that if the commits are not ordered alphabetically, then `save` followed by `apply` is not a no-op but we reorder the commits.

Here are a few examples.

```
Patches are: 01-foo, 02-bar, 03-baz.
You apply the patches and get the following commits: 01-foo, 02-bar, 03-baz.
Let's say you add a commit 04-new.
You save the commits and get the following patches: 01-foo, 02-bar, 03-baz, 04-new.
```

```
Same initial patches as above and already applied.
You `git rebase -i` to modify 02-bar to 02-modified-bar.
You save and get: 01-foo, 02-modified-bar, 03-baz.
```

```
Same initial patches as above and already applied.
You `git rebase -i` to delete 02-bar.
You save and get: 01-foo, 03-baz.
```

```
Same initial patches as above and already applied.
You `git rebase -i` to insert 02-new before 02-bar and rename 02-bar and 03-baz to increase their number.
You save and get: 01-foo, 02-new, 03-bar, 04-baz.
```

Note that you can combine any such operations. You just want the chain of commits to be alphabetically ordered and represent the set of patches you want. You can use any git operations you want to modify the chain of commits (including inserting, deleting, modifying, splitting, merging, reordering, renaming commits).

L0g4n:
@ia0 
> The logic of the script is to provide you a way to sync between the patches and the chain of commits of the submodule. The apply method synchronizes assuming the patches are the source of truth (i.e. it copies the patches to the commits). The save method does the opposite (i.e. it copies the commits to the patches). The order of the chain of commits is the alphabetical order of the patches. In particular, this means that if the commits are not ordered alphabetically, then save followed by apply is not a no-op but we reorder the commits

Alright, however, I did not start with the patches as the source of truth, but rather with the tock submodule and then added my commits into the submodule. Can I simply edit the files, add the commits, rebase, and then call `save`? Or do I **have** to start with the patches as a source-of-truth? This is problematic since the patches wont properly apply with the tock submodule updated to 2.1.

P.S: I changed this PR into a draft to avoid getting a load of emails because of failed CI.

ia0:
> Or do I have to start with the patches as a source-of-truth?

No you don't need to start with the patches as source-of-truth. You can just `save` at any time and it will overwrite the patches with whatever chain of commits you have in the submodules. However, it's important that the submodule commit is up-to-date (which is the case in this PR) because that's what defines the beginning of the chain of commits.
It's also better if your chain of commits is linear (no merge) and in alphabetical order (only the short description, i.e. first line).

> Can I simply edit the files, add the commits, rebase, and then call save?

Why do you need to rebase? I'm assuming you start from commit 55c73f1c9f3c0966d975458a9564ae430a5919af in tock, edit files until it works, commit changes as you go, and call `save` once all changes are committed.

> P.S: I changed this PR into a draft to avoid getting a load of emails because of failed CI.

Sounds good.

.gitignore
```diff
@@ -14,3 +14,6 @@ target/
 /reproducible/elf2tab.txt
 /reproducible/reproduced.tar
 __pycache__
+
+# ctags
+**/tags*
```
ia0, line 19: Where do those come from? You're not using rust-analyzer? Or do we have some C files?

Cargo.toml
```diff
@@ -47,10 +64,10 @@ openssl = "0.10.36"
 
 [profile.dev]
 panic = "abort"
-lto = true # Link Time Optimization usually reduces size of binaries and static libraries
+lto = true      # Link Time Optimization usually reduces size of binaries and static libraries
 
 [profile.release]
 panic = "abort"
-lto = true # Link Time Optimization usually reduces size of binaries and static libraries
+lto = true        # Link Time Optimization usually reduces size of binaries and static libraries
```
ia0, line 71: Out of curiosity, what editor do you use? We currently don't enforce any formatting on TOML files, but maybe we should if there are easy solutions. (I just checked and looks like `taplo` would be an LSP server.)

boards/nordic/nrf52840_dongle_opensk/src/main.rs
```diff
@@ -133,49 +146,99 @@ impl kernel::Platform for Platform {
...
-/// Entry point in the vector table called on hard reset.
-#[no_mangle]
-pub unsafe fn reset_handler() {
-    // Loads relocations and clears BSS
-    nrf52840::init();
+impl KernelResources<nrf52840::chip::NRF52<'static, Nrf52840DefaultPeripherals<'static>>>
+    for Platform
+{
+    type SyscallDriverLookup = Self;
+    type SyscallFilter = ();
```
ia0: Shouldn't we use `Self` for `SyscallFilter`? It's implemented just above.

boards/nordic/nrf52840_dongle_opensk/src/main.rs
```diff
@@ -133,49 +146,99 @@ impl kernel::Platform for Platform {
...
+    type ProcessFault = ();
```
ia0, line 183: Same here, shouldn't we use something?

boards/nordic/nrf52840_dongle_opensk/src/main.rs
```diff
@@ -133,49 +146,99 @@ impl kernel::Platform for Platform {
...
+pub unsafe fn main() {
+    nrf52840::init();
+
+    let nrf52840_peripherals = get_peripherals();
+
     // set up circular peripheral dependencies
     nrf52840_peripherals.init();
     let base_peripherals = &nrf52840_peripherals.nrf52;
 
+    // let board_kernel = static_init!(kernel::Kernel, kernel::Kernel::new(&PROCESSES));
```
ia0: Let's remove this if we don't need it?

boards/nordic/nrf52840_dongle_opensk/src/main.rs
```diff
@@ -133,49 +146,99 @@ impl kernel::Platform for Platform {
...
+    }
+    fn context_switch_callback(&self) -> &Self::ContextSwitchCallback {
+        &()
+    }
+}
 
-    let ppi = static_init!(nrf52840::ppi::Ppi, nrf52840::ppi::Ppi::new());
+/// This is in a separate, inline(never) function so that its stack frame is
+/// removed when this function returns. Otherwise, the stack space used for
+/// these static_inits is wasted.
```
ia0, line 214: That's a good point, if `static_init` is taking stack space, then we should have none in `main`. In particular, we might want something like:

```rust
struct Main {
    board_kernel: ...
    ...
}

impl Main {
    fn new() -> Self { ... }
}

#[no_mangle]
pub unsafe fn main() {
    let main = Main::new();
    main.board_kernel.kernel_loop(&main.platform, main.chip, ...)
}
```

examples/console_test.rs
```diff
@@ -12,24 +12,37 @@
...
-    let mut buf = [0; BUFFER_SIZE];
+    let mut buf = [0; 1024];
     loop {
         for i in 1..buf.len() {
             for byte in buf.iter_mut().take(i) {
                 *byte = b'0' + ((i % 10) as u8);
             }
             buf[i] = b'\n';
-            Console::write_unbuffered(&mut buf[..(i + 1)]);
+            Console::<Syscalls>::write(&buf[..(i + 1)]).ok().unwrap();
```
ia0: The `flex_unwrap` trick doesn't work here?

examples/erase_storage.rs
```diff
@@ -36,20 +50,21 @@ fn is_page_erased(storage: &dyn Storage, page: usize) -> bool {
...
-            writeln!(Console::new(), "erased.").unwrap();
+            writeln!(console, "erased.").unwrap();
         }
     }
-    writeln!(Console::new(), "Done.").unwrap();
-    led::get(1).flex_unwrap().off().flex_unwrap();
-    led::get(0).flex_unwrap().on().flex_unwrap(); // green on dongle
+    writeln!(console, "Done.").unwrap();
+    Leds::<Syscalls>::off(1).unwrap();
+    Leds::<Syscalls>::off(0).unwrap(); // green on dongle
```
ia0: This should be `on`.

examples/store_latency.rs
```diff
@@ -22,24 +23,47 @@ use alloc::vec::Vec;
...
+set_main! {main}
 
-fn timestamp(timer: &Timer) -> Timestamp<f64> {
-    Timestamp::<f64>::from_clock_value(timer.get_current_clock().ok().unwrap())
+#[cfg(feature = "std")]
+type Syscalls = fake::Syscalls;
+#[cfg(not(feature = "std"))]
+type Syscalls = TockSyscalls;
+
+fn timestamp<S: libtock_platform::Syscalls>(timer: &Timer<S>) -> Timestamp<f64> {
```
ia0: Since those types are only instantiated with one type for a given compilation, wouldn't it be simpler and more readable to use the instantiated type directly?

```rust
fn timestamp(timer: &Timer<Syscalls>) -> Timestamp<f64> { ... }
```

libraries/persistent_store/src/storage.rs
```diff
@@ -34,6 +34,15 @@ pub enum StorageError {
 
     /// Implementation-specific error.
     CustomError,
+
+    /// libtock-rs Error
+    LibtockRsError(libtock_platform::ErrorCode),
```
ia0: Why do we need to add this? This should fit in the `CustomError` case. However, I can see why the `From` impl can only be defined here. So maybe it would make sense to provide a blanket impl:

```rust
impl From<T> for StorageError {
    fn from(_: T) -> Self { Self::CustomError }
}
```

The idea of this persistent store library is that it should be unrelated to Tock.

libraries/rng256/src/lib.rs
```diff
@@ -41,18 +44,30 @@ fn bytes_to_u32(bytes: [u8; 32]) -> [u32; 8] {
     result
 }
 
-// RNG backed by the TockOS rng driver.
-pub struct TockRng256 {}
+/// RNG backed by the TockOS rng driver.
+pub struct TockRng256<S: Syscalls, C: Config = DefaultConfig> {
+    _marker_s: PhantomData<S>,
+    _marker_c: PhantomData<C>,
```
ia0: Why not call those `_syscalls` and `_config`?

patches/tock/0004-vendor-hid.patch
```diff
@@ -155,163 +153,3 @@ index f7899d8c5..6956523c6 100644
                      } else {
                          hil::usb::CtrlSetupResult::ErrGeneric
                      }
-diff --git a/capsules/src/usb/usbc_ctap_hid.rs b/capsules/src/usb/usbc_ctap_hid.rs
```
ia0: I guess this is removed because work in progress.

ia0 (commented):
I didn't have time to review everything, but went over most of it (at least quickly). I think overall it looks to go in the right direction. I'm wondering though if we need the code to be generic over those `libtock_platform` traits or if we can define the actual types somewhere (gated by appropriate `cfg`) and then use those types directly. Because it doesn't look like we need to instantiate differently in the same binary. This would remove a bit of boilerplate.

.gitignore
L0g4n: Ah, these come from my neovim tooling, more specifically `gutentags`. I do use rust-analyzer, this artefacts are automatically generated when I open sth. nvim. Can probably be removed if annoying.

libraries/persistent_store/src/storage.rs
L0g4n: Yeah, it can be probably removed. I added it for debugging purposes in my process of getting a working MVP to get a more specific error message (i.e. in my case it was some userspace error and in the end `BadRVal`). With simply `CustomError` I could not locate the error.

examples/store_latency.rs
L0g4n: Yup, indeed. This is result of a paste from a WIP to getting the examples to compile in the first place.

Cargo.toml
L0g4n: For rust I typically use vscode with rust analyzer. Sometimes though for quick editing I also use neovim.

boards/nordic/nrf52840_dongle_opensk/src/main.rs
L0g4n: Good question. I think this particular piece of code comes from the upstream tock repo where they do it that way: https://github.com/tock/tock/blob/master/boards/nordic/nrf52840dk/src/main.rs#L244

L0g4n:
Thanks, I added some comments to your remarks, in particular about the platform impl stuff.

**Edit**: BTW, ignore the "ready for review" thing for now, I just changed it from draft to normal PR again as it still informed about failing CI ;)

libraries/persistent_store/src/storage.rs
L0g4n: > Why do we need to add this? This should fit in the `CustomError` case. However, I can see why the `From` impl can only be defined here. So maybe it would make sense to provide a blanket impl:
> 
> ```rust
> impl From<T> for StorageError {
>     fn from(_: T) -> Self { Self::CustomError }
> }
> ```
This won't work as it conflicts with the core implementation:
```
error[E0119]: conflicting implementations of trait `std::convert::From<storage::StorageError>` for type `storage::StorageError`
  --> libraries/persistent_store/src/storage.rs:46:1
   |
46 | impl<T> From<T> for StorageError {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: conflicting implementation in crate `core`:
           - impl<T> From<T> for T;

```

.gitignore
ia0: There should be a way to not generate them. I'm not a partisan of having non-common tooling configuration in the repo, but if others believe that's fine, then fine by me.

Regardless I would just use `tags*`, the `**/` prefix is redundant. Could you provide example filenames to get an idea of the sort of suffixes that gutentags create?

Cargo.toml
ia0: Do you know which one (VSCode or neovim) formats TOML files by trying to "align" line-suffix comments?

boards/nordic/nrf52840_dongle_opensk/src/main.rs
ia0: Yes, it's because they don't want to use a syscall filter. In our case we want to filter syscalls to the flash driver such that the application can only modify its storage locations.

examples/store_latency.rs
ia0: I see. Looking forward for the simplification then.

libraries/persistent_store/src/storage.rs
ia0: I see, then I see at least 2 options (which would be independent PRs):
1. Make `StorageError` generic over the `CustomError` and an associated type of `Storage`.
2. Add a `StorageError::custom(error: impl Debug) -> Self` function which returns `CustomError` by default (ignoring the error), and panics with the error when `cfg(feature = "debug_custom")`. Implementers of `Storage` would use `custom` to wrap their own error.

But if you don't need this debugging feature anymore, let's just keep the code like it was and remove those debugging helpers.

L0g4n: Ok. Will bloat up the error handling in the app then because it can no longer automatically convert the error though.

Cargo.toml
L0g4n: I think this one is caused by the VSCode Plugin "better-toml": https://github.com/bungcip/better-toml
Although I see that is archived now ^^

boards/nordic/nrf52840_dongle_opensk/src/main.rs
L0g4n: Ah, you're right. I completely forgot about that. BTW: Do you know whether OpenSK will continue to be maintained by google or abandoned (like opentitan)?

libraries/persistent_store/src/storage.rs
L0g4n: > But if you don't need this debugging feature anymore, let's just keep the code like it was and remove those debugging helpers.

I still think there should be a way to, at least optionally, have some context to the error. Should it be possible to implement the `Error` [trait](https://doc.rust-lang.org/stable/core/error/trait.Error.html)? This should make it easier for converting errors, also.

ia0: > Should it be possible to implement the Error trait?

Interesting, I wasn't aware Error made it to core. Seems pretty new: https://github.com/rust-lang/rust/issues/103765. We would need to see if we use a recent enough compiler and if it doesn't bloat the binary size too much.

> Will bloat up the error handling in the app then because it can no longer automatically convert the error though.

This is true for solution (2) and status quo. For solution (1) I would expect an `impl<T> From<T> for StorageError<T>` to work. For solution (2) and status quo the bloat is anyway already there. Existing code already produces `CustomError` manually. This should not happen very often because it should only happen in the `impl Storage` methods, so it should be acceptable. (Also, it's only source bloat, not binary bloat.)

boards/nordic/nrf52840_dongle_opensk/src/main.rs
kaczmarczyck: So far, OpenSK is not impacted by recent developments.

boards/nordic/nrf52840dk_opensk/src/main.rs
```diff
@@ -106,7 +108,7 @@ pub mod io;
 // Whether to use UART debugging or Segger RTT (USB) debugging.
 // - Set to false to use UART.
 // - Set to true to use Segger RTT over USB.
-const USB_DEBUGGING: bool = true;
+const USB_DEBUGGING: bool = false;
```
kaczmarczyck, line 111: What's the reason we turn this off?

boards/nordic/nrf52840dk_opensk/src/main.rs
```diff
@@ -121,38 +123,44 @@ static STRINGS: &'static [&'static str] = &[
...
+static mut PROCESS_PRINTER: Option<&'static kernel::process::ProcessPrinterText> = None;
+
+/// Flash buffer for the custom nvmc driver
+static mut APP_FLASH_BUFFER: [u8; 512] = [0; 512];
 
 /// Dummy buffer that causes the linker to reserve enough space for the stack.
 #[no_mangle]
 #[link_section = ".stack_buffer"]
-pub static mut STACK_MEMORY: [u8; 0x1000] = [0; 0x1000];
+pub static mut STACK_MEMORY: [u8; 0x2000] = [0; 0x2000];
```
kaczmarczyck, line 145: FYI: We had to do this sometimes in v1 already. Out of curiosity, did you observe memory problems often before you made this change? I noticed you made comments about stack sizes elsewhere, so I was wondering if you can share some of your gained wisdom :)

deploy.py
```diff
@@ -154,11 +175,13 @@
...
             nordic_dfu=True,
         ),
+    "cw310_opensk":
+        cw310_opensk_board,
 }
 
 # The following value must match the one used in the file
 # `src/entry_point.rs`
-APP_HEAP_SIZE = 90000
+APP_HEAP_SIZE = 90_000
```
kaczmarczyck, line 184: FYI: Not sure if this value is still accurately reflecting how much we should use.

examples/console_test.rs
```diff
@@ -12,24 +12,40 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(feature = "std"), no_main)]
```
kaczmarczyck: Is this temporary or why would we want `std` examples?

patches/libtock-rs/0001-alloc-init-feature.patch
```diff
@@ -0,0 +1,121 @@
...
+diff --git a/rust-toolchain b/rust-toolchain
+index e911a96..d00feab 100644
+--- a/rust-toolchain
++++ b/rust-toolchain
+@@ -1,7 +1,7 @@
+ [toolchain]
+ # See https://rust-lang.github.io/rustup-components-history/ for a list of
+ # recently nightlies and what components are available for them.
+-channel = "nightly-2022-06-10"
++channel = "nightly-2022-10-22"
```
kaczmarczyck: Can we go newer than this?

src/ctap/main_hid.rs
```diff
@@ -30,6 +30,12 @@ pub struct MainHid {
     wink_permission: TimedPermission,
 }
 
+impl Default for MainHid {
```
kaczmarczyck, line 33: Where is this used?

third_party/lang-items/src/panic_handler.rs
```diff
@@ -1,26 +1,25 @@
...
-            core::ptr::read_volatile(0 as *const usize);
-        }
+        let mut writer = Console::<TockSyscalls>::writer();
+        writeln!(writer, "{}", _info).ok();
+        // Exit with a non-zero exit code to indicate failure.
+        TockSyscalls::exit_terminate(ErrorCode::Fail as u32);
     }
-
-    util::flash_all_leds();
+    #[cfg(not(feature = "panic_console"))]
```
kaczmarczyck, line 23: Can we always flash the LEDs?

kaczmarczyck (commented):
Thanks @ia0 for the first review.

Do you have a general opinion on USB? Does Tock v2 have better support and we should follow more closely, rather than maintaining our own stack? Did you already get something to work?

Next steps for you are probably to finish the missing drivers?

I haven't looked into the submodules. But first of all thanks for this big contribution!

boards/nordic/nrf52840dk_opensk/src/main.rs
L0g4n: With Segger USB debugging, I could not see the debug messages at the same time while using the nordic board as a USB device. This is because, after hitting the "reset" button, the USB connection terminates and RTT won't function anymore because USB is already used by CTAP (I guess).

patches/libtock-rs/0001-alloc-init-feature.patch
L0g4n: Probably. This specific pinned version is just because I started around this time. How far ahead do you want to go with the nightlies? Is there some specific feature you want to include?

src/ctap/main_hid.rs
L0g4n: In here, nowhere. This was added (as countless other times) to satisfy the clippy lints.

third_party/lang-items/src/panic_handler.rs
L0g4n: Problem is, when we have the `panic_console` active, both the `TockSyscalls::exit_terminate` and the `flash_all_leds` both never return, as they have a return type of `!`. This will spit out warnings by the rust compiler.

examples/console_test.rs
L0g4n: We don't need `std` example ever. However, this is just to get `rust-analyzer` in vscode to accept the examples without errors.

deploy.py
L0g4n: I see. I wondered how you came up with this specific heap limit, do you have some measurements with the certified versions that give a rough estimate how much heap is needed for allocation? Because this seems to be a rather large value.

boards/nordic/nrf52840dk_opensk/src/main.rs
L0g4n: Nope, this is simply here because it was done in v1. I did no fine tuning with the memory limits, priority was first to get it to work :)

L0g4n:
> Do you have a general opinion on USB? Does Tock v2 have better support and we should follow more closely, rather than maintaining our own stack? Did you already get something to work?

Eh, I don't really have an opinion on this. TBH, I did not ready anything newsworthy in the Tock v2 release notes that mention something in regard to USB. Do you mean with better support the [upstream USB_CTAP capsule](https://github.com/tock/tock/blob/master/capsules/src/usb/ctap.rs)? The code is a lot shorter, and seems a bit more elegant (as they seem not to have handle as much conditionals), fewer limitations (more than one client for example, althought this is not really relevant), but I don't know for certain how it compares in detail.

> Next steps for you are probably to finish the missing drivers?

I planned to work in first all the relevant review comments, and check for review after that. As you are talking about drivers in plural, what else is still missing besides `crp`?

examples/console_test.rs
ia0: You should be able to configure rust-analyzer to not need that. The configuration you're looking for is `rust-analyzer.cargo.target`. I don't know how it's called in VSCode though. You should set it to `thumbv7em-none-eabi`.

patches/libtock-rs/0001-alloc-init-feature.patch
kaczmarczyck: Just wanted to understand if there is any blocker, and how much new freedom we have.

src/ctap/main_hid.rs
kaczmarczyck: We can move the implementation of `new` to `default` then. And call `::default()` instead of `::new()`.

third_party/lang-items/src/panic_handler.rs
kaczmarczyck: Ah right, if the above never returns, no point in having code after.

kaczmarczyck:
Yours seems to work, so we don't have to bother with USB for now it seems. The question is orthogonal then, but I was curious.

I played around with the firmware a bit, and it seems to work. For RTT (as expected), I had to flip the bool in `third_party/tock/boards/nordic/nrf52840dk_opensk/src/main.rs` to see output. So no plural, only `crp` is missing :)

More comments:

- I looked at binary sizes, and both kernel and app are a few kB smaller with your changes. The decrease for the app can be explained by the new compiler.
- Rust compiler version `nightly-2023-02-01` also works. We don't need it, but good to know nothing broke in between.

Which means I am happy. Also I didn't see any substantial or blocking comment. We can go ahead and think about how to submit.

examples/console_test.rs
L0g4n: Hmm, I tried that but then I get failing build scripts errors in vscode, in the "ring" dependency.

kaczmarczyck: What is vscode trying to build, does it give you a command? And do the examples work if you build them manually?
```
./deploy.py --board=nrf52840dk_opensk --console_test
```

L0g4n: The can be built no problem. The "error" was rather that rust-analyzer indicated errors in the examples without the attributes in vscode. Checked just again: Seems to work now without this workaround :)

L0g4n:
> * Rust compiler version `nightly-2023-02-01` also works. We don't need it, but good to know nothing broke in between.

In the last commits I did some work and rebased the tock patches on top of the stable tock 2.1.1 release (despite released later than previous state on master, we're not actually now on an earlier state of the tock repo); I also updated the nightly in this session to the aforementioned one.

I did not have the chance to test it on real hardware though yet, so if you got time you can do that to see if there were hopefully no regressions in Tock :grinning:. As you might have noticed (or not), I am currently not so active working on this PR, as my thesis deadline in Mid-April approaches, I'm currently kinda focusing on getting the OpenTitan port to work (mostly NVMC driver in the flash_ctrl for persistent storage, and a slight variation of the button driver for using the user DIP switches as buttons for the user presence).

kaczmarczyck:
Tested on the development kit, and still works. Thanks for the update on your available time.

kaczmarczyck:
We merged #602 to move all CTAP logic into its own library. You mostly shouldn't need to touch this part of the code anymore. #603 fixes most of the lints in `src/ctap/`, though some of your changes there are missing as clippy hasn't been complaining for me. Otherwise, all you do in `ctap/` is add `UserPresenceError::ButtonListenerFailed`, which you could just call `Error` to make it generally useful, and that should be it!

So when you rebase, all other changes should be in `third_party/`, `src/env/tock/` or `boards/`, respectively. I hope this helps!

kaczmarczyck:
Hi, I hope your thesis is going well? Please let me know if you want to continue to land this PR afterwards.

On CRP: I realized that there are different variants of the nrf52840 with different versions of CRP. We consider removing CRP support until we have a working implementation for the latest chip. I wanted to share since that was the last piece you were missing, and we can agree to ignore it.

L0g4n:
Hey there,

thanks for asking: Yes, I submitted my thesis this week. I did not reach all goals I originally set for myself considering the time available (mostly having to do with OpenTitan blocking things), but overall I'm still positive about it  😀

> On CRP: I realized that there are different variants of the nrf52840 with different versions of CRP. We consider removing CRP support until we have a working implementation for the latest chip. I wanted to share since that was the last piece you were missing, and we can agree to ignore it.

About the question: Yes, I still want to land this PR, since CRP is not anymore required, what is still missing in this PR besides the rebase work with the CTAP library?

kaczmarczyck:
I doubled checked, and this should be it. I expect that your rebase does not touch `libraries/opensk` (which corresponds to `src/ctap`, `src/api` and `src/env/test` here), and you can discard all your changes in these directories when you rebase. The only difference in API you need is addressed in #616 . If you feel like our new API and Env are not good enough, please let me know! The new library (without Tock) is tested on nightly.

Concerning the TockEnv function:
`pub fn lock_firmware_protection(&mut self) -> bool;`
Just let it return false, and I'll deal with the command that calls it.

Thanks for your help!

L0g4n:
You know, I think it might be better if you finish this PR off. That's quite a lot of steps for rebasing (n = 44 and quite some merge conflicts), and I'm not really familiar with all the code changes in the background that you did (removed rng256, crypto changes, ...) and don't want to mess things up, plus it has been a while since I actively worked on this code.

And to be honest, fixing all the merge conflicts is not something I like to do for fun :)

src/env/tock/storage.rs
```diff
@@ -58,85 +61,98 @@ mod memop_nr {
...
-    pub const STORE: usize = 1;
-    pub const PARTITION: usize = 2;
+    pub const STORE: u32 = 1;
+    pub const PARTITION: u32 = 2;
 }
 
-fn get_info(nr: usize, arg: usize) -> StorageResult<usize> {
-    let code = syscalls::command(DRIVER_NUMBER, command_nr::GET_INFO, nr, arg);
-    code.map_err(|_| StorageError::CustomError)
+fn get_info<S: Syscalls>(nr: u32, arg: u32) -> StorageResult<u32> {
```
zhalvorsen, line 68: Should we put these `syscall` functions in a struct similar to those in `third_party/libtock-drivers/src/`?

```Rust
pub trait Config: platform::allow_ro::Config + platform::subscribe::Config
{
}

impl<T: platform::allow_ro::Config + platform::subscribe::Config> Config for T
{
}

pub struct Storage<S: Syscalls + RawSyscalls, C: Config = DefaultConfig>(S, C);

impl<S: Syscalls + RawSyscalls, C: Config> Storage<S, C> {
    pub fn get_info(nr: u32, arg: u32) -> StorageResult<u32> {
        let info = S::command(DRIVER_NUMBER, command_nr::GET_INFO, nr, arg)
            .to_result::<u32, ErrorCode>()
            .map_err(|_| StorageError::CustomError)?;
        Ok(info)
    }

    pub fn memop(nr: u32, arg: u32) -> StorageResult<u32> {
        let registers = unsafe { S::syscall2::<{ syscall_class::MEMOP }>([nr.into(), arg.into()]) };

        let r0 = registers[0].as_u32();
        let r1 = registers[1].as_u32();

        // make sure r0 is the `success with u32` (129) return variant and then return the value in r1 as u32
        // see: https://github.com/tock/tock/blob/master/doc/reference/trd104-syscalls.md#32-return-values
        match (r0, r1) {
            (129, r1) => Ok(r1),
            (_, _) => Err(StorageError::CustomError),
        }
    }

    pub fn write_slice(
        ptr: usize,
        value: &[u8],
    ) -> StorageResult<()> {
        share::scope(|allow_ro| {
            S::allow_ro::<C, DRIVER_NUMBER, { ro_allow_nr::WRITE_SLICE }>(allow_ro, value)
                .map_err(|_| StorageError::CustomError)?;
            block_command::<S, C>(
                DRIVER_NUMBER,
                command_nr::WRITE_SLICE,
                ptr as u32,
                value.len() as u32,
            )
        })
    }

    pub fn erase_page(
        ptr: usize,
        page_length: usize,
    ) -> StorageResult<()> {
        block_command::<S, C>(
            DRIVER_NUMBER,
            command_nr::ERASE_PAGE,
            ptr as u32,
            page_length as u32,
        )
    }

    fn block_command(
        driver: u32,
        cmd: u32,
        arg1: u32,
        arg2: u32,
    ) -> StorageResult<()> {
        let called: Cell<Option<(u32,)>> = Cell::new(None);

        share::scope(|subscribe| {
            S::subscribe::<_, _, C, DRIVER_NUMBER, { subscribe_nr::DONE }>(subscribe, &called)
                .map_err(|_| StorageError::CustomError)?;
            S::command(driver, cmd, arg1, arg2)
                .to_result::<(), ErrorCode>()
                .map_err(|_| StorageError::CustomError)?;
            libtock_drivers::util::Util::<S>::yieldk_for(|| called.get().is_some());
            if called.get().unwrap().0 == 0 {
                Ok(())
            } else {
                Err(StorageError::CustomError)
            }
        })
    }
}

unsafe fn read_slice(address: usize, length: usize) -> &'static [u8] {
    core::slice::from_raw_parts(address as *const u8, length)
}
```

kaczmarczyck: What's the benefit? The calls to these functions are not getting shorter, and the functions are private, so it's not really an API design advantage?

zhalvorsen: I was picturing we would end up putting this in `third_party/libtock-drivers/src`. Is there a reason it is kept here?

kaczmarczyck:
> You know, I think it might be better if you finish this PR off. That's quite a lot of steps for rebasing (n = 44 and quite some merge conflicts), and I'm not really familiar with all the code changes in the background that you did (removed rng256, crypto changes, ...) and don't want to mess things up, plus it has been a while since I actively worked on this code.
> 
> And to be honest, fixing all the merge conflicts is not something I like to do for fun :)

Understood. Thanks for all your work so far! I'm almost done with the rebased PR, just needs more testing now. It sits here until then, for the curious:
https://github.com/kaczmarczyck/OpenSK/tree/tock-v2

kaczmarczyck:
Merged in #620

PR #594, opened 2023-02-13
Fix AuthenticatorConfig parameters

### LINKED ISSUE #593: AuthenticatorConfigParameters swaps pinUvAuthProtocol and pinUvAuthParam field IDs
AuthenticatorConfigParameters in OpenSK:

https://github.com/google/OpenSK/blob/a2229869956b014dd44f1e6e4924c9d87c32316c/src/ctap/command.rs#L470-L473

[The spec says][0]:

Parameter name | Data type | Required? | Notes
-- | -- | -- | --
subCommand (0x01) | Unsigned Integer | Required | subCommand currently being requested
subCommandParams (0x02) | CBOR Map | Optional | Map of subCommands parameters.
**pinUvAuthProtocol (0x03)** | Unsigned Integer | Optional | PIN/UV protocol version chosen by the platform.
**pinUvAuthParam (0x04)** | Byte String | Optional | First 16 bytes of HMAC-SHA-256 of contents using pinUvAuthToken.

When attempting to configure anything with a `pinUvAuthProtocol` / `pinUvAuthParam` parameter, OpenSK returns "Unexpected Type".

[0]: https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#authenticatorConfig

kaczmarczyck:
Fixes #593 

I noticed that Config we lacked the units tests in command. @kaczmarczyck You deserve this bug for forgetting.

Also I verify against an external source of truth, in this case libfido2. See the script in #593 .

### DIFF
--- a/src/ctap/command.rs
+++ b/src/ctap/command.rs
@@ -457,8 +457,8 @@ impl TryFrom<cbor::Value> for AuthenticatorLargeBlobsParameters {
 pub struct AuthenticatorConfigParameters {
     pub sub_command: ConfigSubCommand,
     pub sub_command_params: Option<ConfigSubCommandParams>,
-    pub pin_uv_auth_param: Option<Vec<u8>>,
     pub pin_uv_auth_protocol: Option<PinUvAuthProtocol>,
+    pub pin_uv_auth_param: Option<Vec<u8>>,
 }
 
 impl TryFrom<cbor::Value> for AuthenticatorConfigParameters {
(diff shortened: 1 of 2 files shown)


ia0 (approved):

kaczmarczyck:
The `nrfutil` in `requirements.txt` breaks fuzzing, will fix that in the next PR.

PR #597, opened 2023-03-01
Removes timer updates from CTAP API

kaczmarczyck:
Follow-up on #596 . Checks timers on demand, not periodically.

### DIFF
--- a/src/ctap/mod.rs
+++ b/src/ctap/mod.rs
@@ -404,11 +404,16 @@ impl<E: Env> StatefulPermission<E> {
         }
     }
 
-    /// Gets a reference to the current command state, if any exists.
-    pub fn get_command(&mut self, env: &mut E) -> Result<&StatefulCommand, Ctap2StatusCode> {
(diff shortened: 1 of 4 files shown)


src/ctap/mod.rs
```diff
@@ -466,7 +469,10 @@ impl<E: Env> StatefulPermission<E> {
     }
 
     /// Returns the next storage credential key for enumeration and advances it.
-    pub fn next_enumerate_credential(&mut self) -> Result<usize, Ctap2StatusCode> {
+    pub fn next_enumerate_credential(&mut self, env: &mut E) -> Result<usize, Ctap2StatusCode> {
+        if env.clock().is_elapsed(&self.permission) {
+            self.clear();
+        }
```
ia0: Look like those 3 lines could be their own function. Namely, making `clear_timer` private instead of deleting it, and calling it instead of inlining it.

ia0 (approved):

PR #601, opened 2023-03-06
Maintenance PR for clippy, license and authors

kaczmarczyck:
This PR does 3 hopefully boring things:

- update license headers in `src/` to go to 2023 (or add them where missing)
- add one author in a few `Cargo.toml` files
- apply the clippy default suggestion by `impl Default for T` instead of having a `new` in `impl T`

### DIFF
--- a/fuzz/fuzz_helper/src/lib.rs
+++ b/fuzz/fuzz_helper/src/lib.rs
@@ -85,7 +85,7 @@ fn initialize(ctap: &mut Ctap<TestEnv>) -> ChannelID {
         cmd: CtapHidCommand::Init,
         payload: nonce,
     };
-    let mut assembler_reply = MessageAssembler::new();
+    let mut assembler_reply = MessageAssembler::default();
     let mut result_cid: ChannelID = Default::default();
     for pkt_request in HidPacketIterator::new(message).unwrap() {
         for pkt_reply in ctap.process_hid_packet(&pkt_request, Transport::MainHid) {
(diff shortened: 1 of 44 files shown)


libraries/crypto/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = [
   "Fabian Kaczmarczyck <kaczmarczyck@google.com>",
   "Guillaume Endignoux <guillaumee@google.com>",
   "Jean-Michel Picod <jmichel@google.com>",
+  "Julien Cretin <cretin@google.com>",
```
ia0: I wouldn't put my name here. I don't think I touched the crypto that much.

libraries/rng256/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = [
   "Fabian Kaczmarczyck <kaczmarczyck@google.com>",
   "Guillaume Endignoux <guillaumee@google.com>",
   "Jean-Michel Picod <jmichel@google.com>",
+  "Julien Cretin <cretin@google.com>",
```
ia0: Same here, I don't think I touched this library except for trivial things.

fuzz/fuzz_helper/src/lib.rs
```diff
@@ -85,7 +85,7 @@ fn initialize(ctap: &mut Ctap<TestEnv>) -> ChannelID {
         cmd: CtapHidCommand::Init,
         payload: nonce,
     };
-    let mut assembler_reply = MessageAssembler::new();
+    let mut assembler_reply = MessageAssembler::default();
```
ia0, line 88: fyi: Note that the idea of "new without default" is not to use "default", but to implement `Default` to let generic code use `T: Default` which is how they can call `::new()` because it's not bound to a trait. Non-generic code can just call `new()`. But given the change is done, no need to go back. Using `default()` is also fine.

ia0 (dismissed):

fuzz/fuzz_helper/src/lib.rs
kaczmarczyck: Ack, thanks.

libraries/crypto/Cargo.toml
kaczmarczyck: As you wish.

libraries/rng256/Cargo.toml
kaczmarczyck: The library is arguably trivial, but your call!

ia0 (approved):

PR #602, opened 2023-03-07
CTAP library move

kaczmarczyck:
Moves all source files into its own library

I removed the `Cargo.lock` files from all libraries. As a consequence of this decision and the CTAP move, I changed the workflows and `run_desktop_tests.sh` a bit to reflect the new reality better.

What might be interesting is that some tests and fuzzing now run on nightly. This is because some of our dependency and `cargo fuzz` don't work with Tock's older nightly, if we don't properly pin to older versions. I'd say we care that the OpenSK binary builds and works with the pinned toolchain, and our libraries build and test correctly with the latest toolchain. This might cause sudden and unrelated problems, but I think we want to know.

Also I hope that the local test script runs on a fresh computer that doesn't have my Rust setup, with the new usage of 2 compiler versions. Otherwise, this PR contains no actual code changes.


.github/workflows/cargo_fmt.yml
```diff
@@ -34,41 +38,54 @@ jobs:
...
+          args: -- --check
 
-      - name: Cargo format fuzz/
+      - name: Cargo format libraries/opensk
         uses: actions-rs/cargo@v1
         with:
+          toolchain: nightly
           command: fmt
-          args: --manifest-path fuzz/Cargo.toml --all -- --check
+          args: --manifest-path libraries/opensk/Cargo.toml --all -- --check
```
ia0: nit: We should also be able to get rid of the `--all` here

Same throughout this file.

kaczmarczyck:
I'll look into Coveralls, but you can probably review everything else.

.github/workflows/cargo_fuzz.yml
```diff
@@ -25,7 +26,7 @@ jobs:
         run: ./fuzzing_setup.sh
 
       - name: Cargo fuzz build
-        run: cargo fuzz build
+        run: cd libraries/opensk && cargo fuzz build && cd ../..
```
ia0: I'm surprised we need `cd ../..` (same in the rest of the file). I would expect `run` to start a subshell and not share the same shell as the rest of the job.

libraries/opensk/fuzz/Cargo.lock
```diff
@@ -96,34 +96,6 @@ dependencies = [
  "untrusted",
 ]
 
-[[package]]
```
ia0: Why do we keep this `Cargo.lock` since it's a library?

run_desktop_tests.sh
```diff
@@ -16,7 +16,10 @@
 set -ex
 
 echo "Checking formatting..."
-cargo fmt --all -- --check
+cargo fmt -- --check
+cd libraries/opensk
+cargo +nightly fmt --all -- --check
```
ia0: nit: We probably also don't need `--all` here.

Same in the rest of the file.

src/lib.rs
```diff
@@ -15,109 +15,5 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
```
ia0, line 17: I think we can remove the library and only make `ctap2` a binary.

ia0 (commented):
Now that we plan to have proper published libraries, I think we should change a few things:
1. We should have one workflow per published library.
2. We should specify the MSRV (minimal supported rust version) in the `Cargo.toml` under [`package.rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field).
3. The workflow of each library should have a strategy matrix to check build with the main toolchains: MSRV, latest stable, latest nightly.

EDIT: This can/should probably be in a follow-up PR. Just mentioning it here because those files were touched.

.github/workflows/cargo_fmt.yml
kaczmarczyck: Done

.github/workflows/cargo_fuzz.yml
kaczmarczyck: We don't need to, removed.

libraries/opensk/fuzz/Cargo.lock
kaczmarczyck: Removed all `Cargo.lock` files in `*/fuzz/` .

run_desktop_tests.sh
kaczmarczyck: Done

src/lib.rs
kaczmarczyck: We need the library for now for our examples.

ia0 (dismissed):

ia0 (approved):

PR #603, opened 2023-03-09
Fixes new clippy lints on the latest nightly

kaczmarczyck:
Now that we can use a later toolchain, we see more lints. Overlaps with #580 , but I didn't reproduce all lints. Missing are all changes like:

```
-                large_blob_data.extend(&[0x0C, 0x00]);
+                large_blob_data.extend([0x0C, 0x00]);
```

Not sure what toolchain version these are coming from.

On a side note, other libraries still have lints that need to be fixed on the current nightly.

### DIFF
--- a/libraries/opensk/src/ctap/client_pin.rs
+++ b/libraries/opensk/src/ctap/client_pin.rs
@@ -1224,7 +1224,7 @@ mod test {
             ),
             // Reject PIN "12'\0'4" (a zero byte at index 2).
             (
-                b"12\x004".to_vec(),
+                [b'1', b'2', 0, b'4'].to_vec(),
                 Err(Ctap2StatusCode::CTAP2_ERR_PIN_POLICY_VIOLATION),
             ),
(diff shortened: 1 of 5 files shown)


libraries/opensk/src/ctap/client_pin.rs
```diff
@@ -1224,7 +1224,7 @@ mod test {
             ),
             // Reject PIN "12'\0'4" (a zero byte at index 2).
             (
-                b"12\04".to_vec(),
+                b"12\x004".to_vec(),
```
ia0: nit: `[b'1', b'2', 0, b'4'].to_vec()` might be more readable in this specific case

ia0 (dismissed):

libraries/opensk/src/ctap/client_pin.rs
kaczmarczyck: Done

ia0 (approved):

PR #605, opened 2023-03-14
CTAP HID Lock

kaczmarczyck:
This is a direct translation of our internal implementation.

### DIFF
--- a/libraries/opensk/src/ctap/hid/mod.rs
+++ b/libraries/opensk/src/ctap/hid/mod.rs
@@ -22,6 +22,7 @@ pub use self::receive::MessageAssembler;
 use self::receive::MessageAssembler;
 pub use self::send::HidPacketIterator;
 use super::status_code::Ctap2StatusCode;
+use crate::api::clock::Clock;
 #[cfg(test)]
 use crate::env::test::TestEnv;
 use crate::env::Env;
@@ -55,7 +56,6 @@ pub type ChannelID = [u8; 4];
 pub enum CtapHidCommand {
     Ping = 0x01,
     Msg = 0x03,
-    // Lock is optional and may be used in the future.
     Lock = 0x04,
     Init = 0x06,
     Wink = 0x08,
@@ -92,7 +92,7 @@ pub enum CtapHidError {
     /// The command in the request is invalid.
     InvalidCmd = 0x01,
     /// A parameter in the request is invalid.
-    _InvalidPar = 0x02,
+    InvalidPar = 0x02,
     /// The length of a message is too big.
     InvalidLen = 0x03,
     /// Expected a continuation packet with a specific sequence number, got another sequence number.
@@ -187,6 +187,8 @@ pub struct CtapHid<E: Env> {
     // TODO(kaczmarczyck) We might want to limit or timeout open channels.
     allocated_cids: usize,
     capabilities: u8,
+    locked_cid: Option<ChannelID>,
+    lock_timer: <E::Clock as Clock>::Timer,
 }
 
 impl<E: Env> CtapHid<E> {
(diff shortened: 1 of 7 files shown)


zhalvorsen (commented):
I noticed the other day that lock isn't listed in the capabilities. Should we add this to `libraries/opensk/src/ctap/hid/mod.rs:196`:

```Rust
pub const CAPABILITY_LOCK: u8 = 0x02;
```

and change `libraries/opensk/src/ctap/main_hid.rs:38` to:

```Rust
        let capabilities = CtapHid::<E>::CAPABILITY_WINK
            | CtapHid::<E>::CAPABILITY_LOCK
            | CtapHid::<E>::CAPABILITY_CBOR;
        #[cfg(not(feature = "with_ctap1"))]
        let capabilities = CtapHid::<E>::CAPABILITY_WINK
            | CtapHid::<E>::CAPABILITY_LOCK
            | CtapHid::<E>::CAPABILITY_CBOR
            | CtapHid::<E>::CAPABILITY_NMSG;
```

kaczmarczyck:
Where did you find the capability bit 0x02? All I see is this in [the spec](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#usb-hid-init):

- CAPABILITY_WINK
- CAPABILITY_CBOR
- CAPABILITY_NMSG

zhalvorsen:
Maybe this is a vendor specific bit. Its listed Yubico's python Fido2 library https://github.com/Yubico/python-fido2/blob/main/fido2/hid/__init__.py#L82.

kaczmarczyck:
Interesting. I'd omit it until I see it in anything official though.

zhalvorsen:
Makes sense to me.

libraries/opensk/src/ctap/hid/mod.rs
```diff
@@ -255,7 +283,7 @@ impl<E: Env> CtapHid<E> {
     /// - ERROR
     /// - Unknown and unexpected commands like KEEPALIVE
     /// - LOCK is not implemented and currently treated like an unknown command
```
jmichelp: This line should be removed as part of this PR :)

libraries/opensk/src/ctap/hid/mod.rs
```diff
@@ -306,6 +334,26 @@ impl<E: Env> CtapHid<E> {
...
             }
             CtapHidCommand::Wink => Some(message),
+            CtapHidCommand::Lock => {
+                if message.payload.len() != 1 {
+                    return Some(Self::error_message(cid, CtapHidError::InvalidLen));
+                }
+                if message.payload[0] > 10 {
+                    return Some(Self::error_message(cid, CtapHidError::InvalidPar));
+                }
+                if message.payload[0] == 0 {
```
jmichelp, line 344: Hard to make sure of the whole flow but I assume that `cid` will be compared before calling this `preprocess_message()` method otherwise any channel could remove the lock with a `0` parameter.

It would be nice to add a test to check that if `cid` lock, `cid+1` doesn't unlock

kaczmarczyck: Yes! The `cid` is checked inside the packet parser. It's even stricter than blocking other commands from unlocking, packets on other channel IDs are entirely blocked. There is a test that checked that sending a ping packet on the wrong channel ID returns an error. I added a more explicit test that checks exactly what you ask.

Also I added lib-level tests, because we should have some.

libraries/opensk/src/ctap/hid/mod.rs
kaczmarczyck: Changed to just list Lock as a preprocessed command.

jmichelp (approved):

PR #606, opened 2023-03-17
Adds a trait for crypto, porting EC first

kaczmarczyck:
This is the the first PR to introduce a Crypto API for our library.
Summary and thoughts:

- The API tries to follow Rust Crypto. Names are similar where reasonable. To proof its compatiblity, you can switch the test env to the `p256` crate (run
  `cargo +nightly test --release --features std,ed25519,with_ctap1,vendor_hid,rust_crypto` in `libraries/opensk/`
  ). I couldn't test it on hardware yet, I didn't find a Rust compiler version that works for both Tock v1 and `p256` with its dependencies. Numbers on binary size will be meaningful only after a full port, since `p256` uses `sha2` for signing, so we currently have overlap.
- The `CoseKey` type doesn't easily translate. I intend to keep all types that are used in parsing commands (rule of thumb: converts to CBOR) free of `Env`. `CoseKey` is one of them. The `From` and `TryFrom` replacements in this PR could be improved in a follow-up PR. Best case the compiler could again ensure that we don't create an ECDH `CoseKey` from a ECDSA public key and vice versa.
- The crypto implementations in `TestEnv` and `TockEnv` are currently identical. I was wondering if they `TockEnv` should just import `TestEnv`, but kept them fully independent for now. This way I don't have to mess with the conditional compilation and it's easier to change only one of them later (e.g. hardware crypto).
- I introduced shorthands like
  `pub type EcdhSk<E> = <<<E as Env>::Crypto as Crypto>::Ecdh as Ecdh>::SecretKey;`
  Seems to me to be the best way to have less noise. The trait imports still look a bit clunky, as `SecretKey` has the same name for ECDH and ECDSA. Which makes sense inside its namespace, but needs extra work when `use`ing them. One option is to split the traits into multiple, like replacing the ECDSA `SecretKey` with a general `SecretKey` that has a `random` and a `public_key` function, together with a `Signer` trait. But I feels like unncecessary complexity to me.

Curious about your thoughts!


libraries/opensk/src/api/crypto/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// The size of field elements in the elliptic curve P256.
+pub const EC_FIELD_BYTE_SIZE: usize = 32;
+
+/// The size of a serialized ECDSA signature.
+pub const EC_SIGNATURE_SIZE: usize = 2 * EC_FIELD_BYTE_SIZE;
+
+/// Necessary cryptographic primitives for CTAP.
+///
+/// This is zero-sized type is a collection of other zero-sized types, one for each required
```
ia0: The trait doesn't enforce zero-sized types and not sure it should. I would just remove this comment, the code is self-descriptive enough.

libraries/opensk/src/api/crypto/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+pub mod ecdsa;
+
+use self::ecdh::Ecdh;
+use self::ecdsa::Ecdsa;
+
+/// The size of field elements in the elliptic curve P256.
+pub const EC_FIELD_BYTE_SIZE: usize = 32;
+
+/// The size of a serialized ECDSA signature.
+pub const EC_SIGNATURE_SIZE: usize = 2 * EC_FIELD_BYTE_SIZE;
```
ia0: Why is one constant saying `BYTE_SIZE` and the other `SIZE` while both seems to be the length in bytes?

libraries/opensk/src/api/key_store.rs
```diff
@@ -12,13 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::api::crypto::ecdsa::SecretKey as EcdsaSecretKey;
```
ia0: You can also `use crate::api::crypto::ecdsa::SecretKey as _` if you don't want to introduce the trait name in scope but want to be able to call its methods.

libraries/opensk/src/ctap/ctap1.rs
```diff
@@ -156,6 +160,22 @@ impl TryFrom<&[u8]> for U2fCommand {
     }
 }
 
+fn to_uncompressed(public_key: &impl EcdsaPublicKey) -> [u8; 1 + 2 * EC_FIELD_BYTE_SIZE] {
```
ia0: Here is a case where you want to refer to the trait by name. There's an alternative option if you prefer:

```rust
use crate::api::crypto::ecdsa::{self, PublicKey as _, ...};

fn to_uncompressed(public_key: &impl ecdsa::PublicKey) -> ...
```

libraries/opensk/src/ctap/data_formats.rs
```diff
@@ -748,6 +750,83 @@ impl CoseKey {
     const P_256_CURVE: i64 = 1;
     #[cfg(feature = "ed25519")]
     const ED25519_CURVE: i64 = 6;
+
+    // Ideally, this would be an impl From.
+    // CoseKey needs to be refactored to allow that.
```
ia0: I'm not sure I follow why the problem is CoseKey. To me it looks like the problem is that we just can't have to `From<T> for CoseKey` implementations depending on which trait `T` implements. I think the current solution is the right one.

libraries/opensk/src/env/test/mod.rs
```diff
@@ -27,6 +28,13 @@ use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
 use rng256::Rng256;
 
+#[cfg(not(feature = "rust_crypto"))]
+pub mod crypto;
+#[cfg(feature = "rust_crypto")]
+pub mod rust_crypto;
+#[cfg(feature = "rust_crypto")]
+pub use rust_crypto as crypto;
```
ia0: I see 2 alternatives:
1. Move both software crypto implementations to `src/api/crypto`, keep the feature flag, and name that single type `SoftwareCrypto` (or something along those lines).
2. Move both software crypto implementation to `src/api/crypto` but give each one its own name (e.g. `OpenskCrypto` and `RustCrypto`), then use the feature flag here (i.e. `src/env/test`) to merge those names into `TestCrypto`.

In both cases, the Tock environment will reuse the implementations in `src/api/crypto`, either `SoftwareCrypto` or `OpenskCrypto`.

libraries/opensk/src/env/test/rust_crypto.rs
```diff
@@ -0,0 +1,254 @@
...
+    type Ecdsa = TestEcdsa;
+}
+
+impl ecdh::Ecdh for TestEcdh {
+    type SecretKey = TestEcdhSecretKey;
+    type PublicKey = TestEcdhPublicKey;
+    type SharedSecret = TestEcdhSharedSecret;
+}
+
+pub struct TestEcdhSecretKey {
```
ia0: For both implementations (`crypto.rs` and `rust_crypto.rs`), why use a wrapper type? Can't we just do the following?

```rust
pub type TestEcdhSecretKey = EphemeralSecret;
```

libraries/opensk/src/env/test/rust_crypto.rs
```diff
@@ -0,0 +1,254 @@
...
+pub struct TestEcdsaSecretKey {
+    signing_key: SigningKey,
+}
+
+impl ecdsa::SecretKey for TestEcdsaSecretKey {
+    type PublicKey = TestEcdsaPublicKey;
+    type Signature = TestEcdsaSignature;
+
+    fn random(_rng: &mut impl Rng256) -> Self {
+        let signing_key = SigningKey::random(&mut OsRng);
```
ia0: Shouldn't we implement `CryptoRngCore` for `Rng256` when `rust_crypto` is set and use the `rng` argument instead of `OsRng` (which is counter-intuitive)? Same for the other usages of `OsRng` in this file.

kaczmarczyck: I will, eventually! This is a 2 line fix before I work on RNG.

libraries/opensk/src/env/test/rust_crypto.rs
kaczmarczyck: Can do. I was wondering about this option, and didn't really see a big difference.

libraries/opensk/src/env/test/mod.rs
kaczmarczyck: I was wondering about moving the Crypto implementation somewhere TockEnv can reuse it. I think we need to keep `RustCrypto` completely behind a feature flag, or we can't use an older compiler. Or is it smart enough to not load optional dependencies if they are unused?
A bit less clean if the Crypto implementation is not in `env/`, but in `api/`, but the reuse is probably worth it.

libraries/opensk/src/ctap/data_formats.rs
kaczmarczyck: I was wondering if Rust has a way to also have different `CoseKey<A: Algorithm>` and you get a `CoseKey<ECDSA>::from(EcdsaPublicKey)` and the compiler checks correctness because you never implemented `CoseKey<ECDSA>::from(EcdhPublicKey)`.

Then CoseKey also wouldn't have to store its other map keys, only `x_bytes` and `y_bytes`. (This optimization can also be achieved indepedently by having an enum property of course.)

But this is good enough for me, if there is nothing Rust-y I'm missing.

libraries/opensk/src/ctap/ctap1.rs
kaczmarczyck: I like the explicit usage. We have too many ecdsa etc. in use, so the longer type is useful to understand the code. In that case, the anonymous trait import is nice.

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: I reused the names from Rust Crypto where I found a direct equivalent:

https://docs.rs/elliptic-curve/0.13.1/elliptic_curve/type.FieldBytesSize.html

The `EC_` prefix is because all constants are in the crypto-level mod, and I wanted to make it more obvious this is not AES etc. They are shared between ECDH and ECDSA.

libraries/opensk/src/env/test/rust_crypto.rs
ia0: Ok I see. Let's add a TODO comment, create an issue, or already send the next PR then to not forget.

libraries/opensk/src/env/test/rust_crypto.rs
ia0: I think it will make the implementation much more obvious/readable by removing the boilerplate.

libraries/opensk/src/env/test/mod.rs
ia0: Yes, `RustCrypto` would be behind the feature flag in both cases.

It's not less clean to have default implementations in `api`, on the contrary. Generic implementations should be next to their trait. Most of them are in `env` just because they are not generic/reusable.

libraries/opensk/src/ctap/data_formats.rs
ia0: Yes exactly, at some point you need to erase the enum variant (ECDSA or ECDH) from the type. It can either be between the crypto API and CoseKey, or between CoseKey and CBOR. The problem with the second one is that it will make parsing CBOR more complicated because you make the CoseKey dependent on the algorithm.

I think erasing the algorithm from the crypto API to the CoseKey makes sense given that as you wrote yourself, the CoseKey is a CBOR-related OpenSK implementation detail, so it should look like CBOR.

libraries/opensk/src/ctap/ctap1.rs
ia0: We ideally shouldn't have too many `ecdsa` because one (`opensk::api::crypto::ecdsa`) is shadowing the other (`crypto::ecsda`). Both are only visible when wiring them together.

libraries/opensk/src/api/crypto/mod.rs
ia0: I think those aliases are much better named: [`SignatureBytes`](https://docs.rs/ecdsa/latest/ecdsa/type.SignatureBytes.html) and [`FieldBytes`](https://docs.rs/elliptic-curve/latest/elliptic_curve/type.FieldBytes.html). They eventually use the underlying `SignatureSize` and `FieldBytesSize`.

libraries/opensk/src/env/test/rust_crypto.rs
kaczmarczyck: There is actually a downside. It makes it harder to call the functions correctly. For example, this doesn't work anymore now:
```
let new_key = TestEcdhSecretKey::random(env.rng())
```

The compiler doesn't know which `random` to call, so we have to be more verbose. I'd say we keep the wrapper type then?

ia0: What about `EcdhSk<E>::random(env.rng())`? Does this work?

ia0: But regardless, I see how it can be problematic to have 2 traits implemented for the same type with the same method names. Fine with the current version.

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: Removed, in fact, it would have been wrong for `Sha256`.

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: But these are arrays, not sizes? Anyway, I we don't have to reuse Rust Crypto names. If you don't like the mix of `BYTE_SIZE` and `SIZE`, we can use one or the other.

libraries/opensk/src/env/test/rust_crypto.rs
kaczmarczyck: Nice, so we don't have to be as careful when calling. FYI your other call idea throws this:

```
error[E0277]: the trait bound `TestRng256: CryptoRng` is not satisfied
    --> src/ctap/mod.rs:3509:59
     |
3509 |         let key_agreement_key = EcdhSk::<TestEnv>::random(env.rng());
     |                                 ------------------------- ^^^^^^^^^ the trait `CryptoRng` is not implemented for `TestRng256`
```

So looks like it is still confused.

libraries/opensk/src/api/key_store.rs
kaczmarczyck: Thanks, done

libraries/opensk/src/ctap/ctap1.rs
kaczmarczyck: I think that's the case. We just have two different `PublicKey` (as expected).

libraries/opensk/src/ctap/data_formats.rs
kaczmarczyck: Agreed to the last point. I think my question is closer to: Should we have a `EcdhCoseKey` and a `EcdsaCoseKey`? But this probably comes down to being an implementation detail or micro optimization that we shouldn't care about too much.

libraries/opensk/src/env/test/mod.rs
kaczmarczyck: Moved

libraries/opensk/src/env/test/rust_crypto.rs
kaczmarczyck: Done

libraries/opensk/src/api/crypto/mod.rs
ia0: But it's the same, it's about coherence. I actually went to see if they have an issue about it and they do: https://github.com/RustCrypto/traits/issues/656. `FieldBytesSize` should be `FieldSize`. This suggests that we should use `EC_FIELD_SIZE` and `EC_SIGNATURE_SIZE`.

Also note that if you really wanted to use the same names it should have been `EC_FIELD_BYTES_SIZE`, because it's the size of `FIELD_BYTES`.

libraries/opensk/src/env/test/rust_crypto.rs
ia0: Yes this means it thought it was the other trait...

libraries/opensk/src/ctap/data_formats.rs
ia0: I see, I don't know enough how those are used in CBOR, but if it's always clear whether we want to parse an `EcdhCoseKey` or an `EcdsaCoseKey` then this would work too indeed. Not sure if worth trying to do though.

libraries/opensk/src/env/test/crypto.rs
```diff
@@ -12,208 +12,12 @@
...
-}
+use crate::api::crypto::software_crypto;
+
+pub use software_crypto::{
+    SoftwareCrypto as TestCrypto, SoftwareEcdh as TestEcdh,
+    SoftwareEcdhPublicKey as TestEcdhPublicKey, SoftwareEcdhSecretKey as TestEcdhSecretKey,
+    SoftwareEcdhSharedSecret as TestEcdhSharedSecret, SoftwareEcdsa as TestEcdsa,
+    SoftwareEcdsaPublicKey as TestEcdsaPublicKey, SoftwareEcdsaSecretKey as TestEcdsaSecretKey,
+    SoftwareEcdsaSignature as TestEcdsaSignature,
+};
```
ia0: This is to avoid renaming those usage?

src/env/tock/crypto.rs
```diff
@@ -12,148 +12,12 @@
...
-}
+use opensk::api::crypto::software_crypto;
+
+pub use software_crypto::{
+    SoftwareCrypto as TockCrypto, SoftwareEcdh as TockEcdh,
+    SoftwareEcdhPublicKey as TockEcdhPublicKey, SoftwareEcdhSecretKey as TockEcdhSecretKey,
+    SoftwareEcdhSharedSecret as TockEcdhSharedSecret, SoftwareEcdsa as TockEcdsa,
+    SoftwareEcdsaPublicKey as TockEcdsaPublicKey, SoftwareEcdsaSecretKey as TockEcdsaSecretKey,
+    SoftwareEcdsaSignature as TockEcdsaSignature,
+};
```
ia0: Same question here. Why not use directly `SoftwareFoo` syntax in the `TockEnv` implementation?

ia0 (dismissed):

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: Renamed to `EC_FIELD_SIZE`.

libraries/opensk/src/env/test/crypto.rs
kaczmarczyck: I used the shorthands `EcdsaSk` etc in all places where used, so the implementations don't have to be accessed outside of `Env` anymore. This makes the whole file obsolete.

src/env/tock/crypto.rs
kaczmarczyck: I removed the file.

ia0 (approved):

kaczmarczyck:
The failing tests are unrelated to this PR and seem to be related to a difference in compilation between debug and release mode. Locally, all tests pass.

PR #608, opened 2023-04-04
LTO fix and toml cleanup

kaczmarczyck:
This fixes a compilation problem we have in debug mode with LTO enabled that was surfaced in workflow runs in #606 .


kaczmarczyck:
I also fixed the other failing workflow that had an unrelated issue. Not sure how it worked before though.

ia0 (approved):

kaczmarczyck:
I reported the Rust problem here:

https://github.com/rust-lang/rust/issues/109934

PR #609, opened 2023-04-04
SHA and HMAC for the Crypto trait

kaczmarczyck:
Continuation of #606 . Same style, and I found a missing ECDSA private key in CTAP1 that I also ported over. I made `SharedSecret` an enum instead of a trait to remove the need to return `Box<dyn SharedSecret>`.

### DIFF
--- a/libraries/opensk/src/api/crypto/hmac256.rs
+++ b/libraries/opensk/src/api/crypto/hmac256.rs
@@ -21,14 +21,14 @@ pub trait Hmac256 {
 
     /// Verifies the HMAC.
     ///
-    /// When implementing, make sure that you don't leak side channel information about the key,
-    /// i.e. by using constant time comparisons of the computed and given MAC.
+    /// This function does best effort to not leak information about the key through side-channels
+    /// (e.g. usage of constant time comparison).
     fn verify(key: &[u8; HMAC_KEY_SIZE], data: &[u8], mac: &[u8; HASH_SIZE]) -> bool;
 
     /// Verifies the first bytes of an HMAC.
     ///
-    /// When implementing, make sure that you don't leak side channel information about the key,
-    /// i.e. by using constant time comparisons of the computed and given truncated MAC.
+    /// This function does best effort to not leak information about the key through side-channels
+    /// (e.g. usage of constant time comparison).
     fn verify_truncated_left(
         key: &[u8; HMAC_KEY_SIZE],
         data: &[u8],
--- a/libraries/opensk/src/api/crypto/hmac256.rs
+++ b/libraries/opensk/src/api/crypto/hmac256.rs
@@ -1,0 +1,31 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::{HASH_SIZE, HMAC_KEY_SIZE, TRUNCATED_HMAC_SIZE};
+
+/// For a given hash function, computes and verifies the HMAC.
+pub trait Hmac256 {
+    /// Computes the HMAC.
+    fn mac(key: &[u8; HMAC_KEY_SIZE], data: &[u8]) -> [u8; HASH_SIZE];
+
+    /// Verifies the HMAC.
+    fn verify(key: &[u8; HMAC_KEY_SIZE], data: &[u8], mac: &[u8; HASH_SIZE]) -> bool;
+
+    /// Verifies the first bytes of an HMAC.
+    fn verify_truncated_left(
+        key: &[u8; HMAC_KEY_SIZE],
+        data: &[u8],
+        mac: &[u8; TRUNCATED_HMAC_SIZE],
+    ) -> bool;
+}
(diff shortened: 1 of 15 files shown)


libraries/opensk/src/api/crypto/mod.rs
```diff
@@ -20,18 +20,33 @@ pub mod rust_crypto;
...
 /// The size of a serialized ECDSA signature.
 pub const EC_SIGNATURE_SIZE: usize = 2 * EC_FIELD_SIZE;
 
+/// The size in bytes of a SHA256.
+pub const HASH_SIZE: usize = 32;
+
+/// The size in bytes of truncated HMAC.
+pub const HMAC_KEY_SIZE: usize = 32;
+
+/// The size in bytes of truncated HMAC.
```
ia0: This is specific to CTAP right? I would extend the documentation to explain why this is needed.

libraries/opensk/src/api/crypto/mod.rs
```diff
@@ -20,18 +20,33 @@ pub mod rust_crypto;
...
 /// The size of field elements in the elliptic curve P256.
 pub const EC_FIELD_SIZE: usize = 32;
 
 /// The size of a serialized ECDSA signature.
 pub const EC_SIGNATURE_SIZE: usize = 2 * EC_FIELD_SIZE;
 
+/// The size in bytes of a SHA256.
+pub const HASH_SIZE: usize = 32;
+
+/// The size in bytes of truncated HMAC.
```
ia0: typo: truncated is extra

libraries/opensk/src/api/crypto/hmac256.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    /// Verifies the HMAC.
+    fn verify(key: &[u8; HMAC_KEY_SIZE], data: &[u8], mac: &[u8; HASH_SIZE]) -> bool;
+
+    /// Verifies the first bytes of an HMAC.
+    fn verify_truncated_left(
+        key: &[u8; HMAC_KEY_SIZE],
+        data: &[u8],
+        mac: &[u8; TRUNCATED_HMAC_SIZE],
+    ) -> bool;
```
ia0, line 36: I'm not sure why we need this in the API. Can't this be a helper function calling verify?

If we want to keep those 2 functions without extra guarantee, then we could at least provide a default implementation.

libraries/opensk/src/api/crypto/hmac256.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use super::{HASH_SIZE, HMAC_KEY_SIZE, TRUNCATED_HMAC_SIZE};
+
+/// For a given hash function, computes and verifies the HMAC.
+pub trait Hmac256 {
+    /// Computes the HMAC.
+    fn mac(key: &[u8; HMAC_KEY_SIZE], data: &[u8]) -> [u8; HASH_SIZE];
+
+    /// Verifies the HMAC.
+    fn verify(key: &[u8; HMAC_KEY_SIZE], data: &[u8], mac: &[u8; HASH_SIZE]) -> bool;
```
ia0, line 26: Why do we want this in the API? Do we have guarantees on checking for equality? Otherwise I would just keep the `mac` function.

libraries/opensk/src/ctap/pin_protocol.rs
```diff
@@ -191,32 +233,33 @@ impl SharedSecret for SharedSecretV1 {
...
+    phantom: PhantomData<E>,
 }
 
-impl SharedSecretV2 {
+impl<E: Env> SharedSecretV2<E> {
     /// Creates a new shared secret from the handshake result.
-    fn new(handshake: [u8; 32]) -> SharedSecretV2 {
-        let aes_key = hkdf_empty_salt_256::<Sha256>(&handshake, b"CTAP2 AES key");
+    fn new(handshake: [u8; 32]) -> Self {
+        let aes_key = hkdf_empty_salt_256::<crypto::sha256::Sha256>(&handshake, b"CTAP2 AES key");
```
ia0, line 255: Will this be migrated later?

ia0 (dismissed):

libraries/opensk/src/ctap/pin_protocol.rs
kaczmarczyck: Yes, next PR together with HKDF!

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: Added a sentence about PIN protocol V1.

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: Fixed

libraries/opensk/src/api/crypto/hmac256.rs
kaczmarczyck: I don't want to take any stance on correct cryptographic implementation of `verify*`. Therefore, I added documentation about what to look out for, instead of implementing a default.

libraries/opensk/src/api/crypto/hmac256.rs
kaczmarczyck: See above, this file doesn't have any opinion on correct implementation, and therefore doesn't want to provide a default (although a constant time comparison would be a decent choice, as documented).

libraries/opensk/src/api/crypto/hmac256.rs
```diff
@@ -20,9 +20,15 @@ pub trait Hmac256 {
     fn mac(key: &[u8; HMAC_KEY_SIZE], data: &[u8]) -> [u8; HASH_SIZE];
 
     /// Verifies the HMAC.
+    ///
+    /// When implementing, make sure that you don't leak side channel information about the key,
+    /// i.e. by using constant time comparisons of the computed and given MAC.
```
ia0: nit: descriptive documentation

```rust
/// This function does best effort to not leak information about the key through side-channels
/// (e.g. usage of constant time comparison).
```

It is implicit that what is guaranteed by the function is what implementations must satisfy.

But no need to change.

ia0 (dismissed):

libraries/opensk/src/api/crypto/hmac256.rs
kaczmarczyck: Thanks, done

ia0 (approved):

PR #610, opened 2023-04-04
Adds HKDF to the Crypto trait

kaczmarczyck:
Continues #609 

I was wondering if I should eventually move all SoftwareCrypto unit tests into `libraries/opensk/src/api/crypto/mod.rs`. If you have an opinion, I can do that next PR.

Next PRs are AES, then RNG.


ia0 (approved):
Tests are fine where they are. Why do you want to move them?

kaczmarczyck:
Because they are identical in both files. If we move them, we move up by one layer of indirection.

ia0:
> Because they are identical in both files. If we move them, we move up by one layer of indirection.

I see. Yes in that case it makes sense. In particular given that we never have both implementations at the same time.

PR #612, opened 2023-04-06
Replaces Rng256 with new Rng API

kaczmarczyck:
The old trait was designed with our software crypto in mind. We should use a more standard API going forward. Binary size is not noticably affected.

- Removes libraries/rng256/.
- Ports libraries/crypto/ to rand_core.
- Creates a new RNG trait in api/.
- Unifies passed around `&mut impl Rng256` and `&mut dyn Rng256` to `&mut E`.

The latter also helps with clippy complaining differently in different Rust versions, about when it can infer types or not. Starting with this PR, we could experiment with RustCrypto in `TockEnv`... if our compiler wasn't too old. Definitely something to try on the Tockv2 branch after rebasing.

For libraries/crypto/, I went with the easiest approach to only require `RngCore` and use `OsRng` for tests. I could also require `CryptoRng` and move away from OsRng (e.g. by wrapping it and implementing `CryptoRng`, or using another RNG from the rand crate). But I figured that we don't plan to touch our software crypto really, and don't need to invest there.


libraries/opensk/src/env/test/mod.rs
```diff
@@ -19,45 +19,61 @@ use crate::api::connection::{HidConnection, SendOrRecvResult, SendOrRecvStatus};
...
+    rng: TestRng,
     user_presence: TestUserPresence,
     store: Store<BufferStorage>,
     upgrade_storage: Option<BufferUpgradeStorage>,
     customization: TestCustomization,
     clock: TestClock,
 }
 
-pub struct TestRng256 {
+pub struct TestRng {
```
ia0: I don't remember if we discussed this. But is there a reason not to do `pub type TestRng = StdRng`?

ia0 (dismissed):

libraries/opensk/src/env/test/mod.rs
kaczmarczyck: I thought I couldn't implement all traits without a wrapper. But it looks like I can!

ia0: Yes, because those additional traits are from the crate itself.

ia0 (approved):

PR #613, opened 2023-04-11
Makes our CredRandom derivation FIPS compliant

kaczmarczyck:
This change breaks existing usage of CredRandom.


libraries/crypto/src/hkdf.rs
```diff
@@ -77,4 +94,39 @@ mod test {
...
+            hex::decode("a0027dcffb27d356317199c6e65f153a9286ba114aee2d3cf45bdba83cb7c065")
+                .unwrap(),
+            hex::decode("786d1f89f54668bac443cc6a8887c95d6fbde07702cb4c16d76c452e87c50f79")
+                .unwrap(),
+            hex::decode("8e9a5bdf362c5aec2c31a742dfebd0b7b56e16ab8408d9d0609a4fad06446875")
+                .unwrap(),
+            hex::decode("4a35d3d7c80ff4fab65f7e30d6b305fc7bb39ffe905aabedd6593354f86177b6")
+                .unwrap(),
+            hex::decode("b1121deabd8b4308f3805cda8af991ee976bd8e413bcb6a8dd3fc26ebe2312d2")
+                .unwrap(),
```
ia0: The `hex::decode()` could be done in the loop to minimize boilerplate.

ia0 (dismissed):

libraries/crypto/src/hkdf.rs
kaczmarczyck: Done. Also I forgot to push my last change. The `rust_crypto` feature is only tested locally, not in workflows, as it is more of an example. In case you are wondering why CI didn't notice on my first commit.

ia0 (approved):

PR #614, opened 2023-04-14
Moves vendor commands into TockEnv

kaczmarczyck:
This move changes the Env trait: It removes all functionality that is used only in vendor commands (`FirmwareProtection`, `UpgradeStorage`) and adds a function to call when parsing CBOR commands, plus a firmware version call.

The abstraction necessary to test these commands is instead realized through compile flags. The mock upgrade storage is active when compiled for std for example.


ia0 (commented):
I went quickly through the code, LGTM. A few questions before I approve:
1. Is there any code being duplicated? (It doesn't look like it, but asking in case I missed it.)
2. What was the testing strategy for normal operation, upgrades, and other impacted parts of the code? (I'm thinking manual webauthn, CTAP2 test suite, local scripts.)

kaczmarczyck:
This function in `src/env/tock/mod.rs` is almost a duplicate, except that it returns a result:
```
#[cfg(feature = "std")]
pub fn take_storage() -> StorageResult<Storage> {
    // Use the Nordic configuration.
    const PAGE_SIZE: usize = 0x1000;
    const NUM_PAGES: usize = 20;
    let store = vec![0xff; NUM_PAGES * PAGE_SIZE].into_boxed_slice();
    let options = BufferOptions {
        word_size: 4,
        page_size: PAGE_SIZE,
        max_word_writes: 2,
        max_page_erases: 10000,
        strict_mode: true,
    };
    Ok(BufferStorage::new(store, options))
}
 ```

The other version is used in `TestEnv`, same functionality. I can put it as a helper function somewhere if you like? Everything else should be purely moved IIRC.

With respect to testing, I tried:

- tools/configure.py
- upgrading
- webauthn.io after each deploy and script

ia0 (approved):
No need to deduplicate the buffer options, they don't need to match.

PR #615, opened 2023-04-19
Cryptographic Secret type

kaczmarczyck:
This PR allows other `Env`s to fully zeroize its secret. For `TockEnv`, it is a best-effort implementation:

- `api/crypto` now writes into mutable slices to return secrets, with SHA256 offering both, as it is often used on non-secrets.
- `libraries/crypto` is adapted to that API change, and now zeroizes its public-facing types, as a first low-effort measure. For full zeroization, we could invest more into the implementation details if we want to use the library in the future.

Limitations:

- Storage is not encrypted, as we don't have a good way to generate an encryption key that does not live in the storage itself. Therefore, whatever we read from storage is not zeroized, as it is readable there anyway.
- The CBOR library is not zeroized, this affects `PrivateKey` in `encrypt_to_credential_id` and `PublicKeyCredentialSource`.

A different `Env` that has a means to derive keys from stored seeds would not be affected by these 2 limitations. It would have to implement a new `AttestationStore` and `KeyStore`.

With this PR, the crypto API shouldn't change too much anymore. The exception is the `ed25519` support behind a compile flag that should move into the API eventually.

### DIFF
--- a/libraries/crypto/src/ec/exponent256.rs
+++ b/libraries/crypto/src/ec/exponent256.rs
@@ -16,9 +16,10 @@ use super::int256::{Digit, Int256};
 use core::ops::Mul;
 use rand_core::RngCore;
 use subtle::{self, Choice, ConditionallySelectable, CtOption};
+use zeroize::Zeroize;
 
 // An exponent on the elliptic curve, that is an element modulo the curve order N.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Zeroize)]
 // TODO: remove this Default once https://github.com/dalek-cryptography/subtle/issues/63 is
 // resolved.
 #[derive(Default)]
@@ -91,7 +92,7 @@ impl Mul for &ExponentP256 {
 }
 
 // A non-zero exponent on the elliptic curve.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Zeroize)]
 // TODO: remove this Default once https://github.com/dalek-cryptography/subtle/issues/63 is
 // resolved.
 #[derive(Default)]
--- a/libraries/crypto/src/ec/exponent256.rs
+++ b/libraries/crypto/src/ec/exponent256.rs
@@ -18,7 +18,9 @@ use rand_core::RngCore;
 use subtle::{self, Choice, ConditionallySelectable, CtOption};
 use zeroize::Zeroize;
 
-// An exponent on the elliptic curve, that is an element modulo the curve order N.
+/// An exponent on the elliptic curve, that is an element modulo the curve order N.
+///
+/// Never call zeroize explicitly, to not invalidate any invariants.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Zeroize)]
 // TODO: remove this Default once https://github.com/dalek-cryptography/subtle/issues/63 is
 // resolved.
@@ -91,7 +93,9 @@ impl Mul for &ExponentP256 {
     }
 }
 
-// A non-zero exponent on the elliptic curve.
+/// A non-zero exponent on the elliptic curve.
+///
+/// Never call zeroize explicitly, to not invalidate any invariants.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Zeroize)]
 // TODO: remove this Default once https://github.com/dalek-cryptography/subtle/issues/63 is
 // resolved.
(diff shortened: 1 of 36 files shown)


libraries/crypto/src/ec/exponent256.rs
```diff
@@ -91,7 +92,7 @@ impl Mul for &ExponentP256 {
 }
 
 // A non-zero exponent on the elliptic curve.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Zeroize)]
```
ia0, line 99: There's a potential risk for bugs here. I think we should document that it is not an invariant of `NonZeroExporentP256` to be non-zero. In particular, unsafe code should not rely on this (false) invariant. For safe code it's fine, it's just gonna misbehave (possibly panic), but the behavior is still defined.

The same applies for other types with such invariants like `ExponentP256` and `GFP256`.

libraries/crypto/src/lib.rs
```diff
@@ -33,16 +33,30 @@ pub mod util;
...
+    fn finalize(self, output: &mut [u8; 32]);
 
     fn hash(contents: &[u8]) -> [u8; 32] {
         let mut h = Self::new();
         h.update(contents);
-        h.finalize()
+        let mut output = [0; 32];
+        h.finalize(&mut output);
+        output
+    }
```
ia0, line 43: This function should be implemented from `hash_mut` to avoid logic duplication.

libraries/crypto/src/lib.rs
```diff
@@ -33,16 +33,30 @@ pub mod util;
...
+        h.update(contents);
+        h.finalize(output)
     }
 
     fn hmac(key: &[u8; 32], contents: &[u8]) -> [u8; 32] {
-        hmac::software_hmac_256::<Self>(key, contents)
+        let mut output = [0; 32];
+        hmac::software_hmac_256::<Self>(key, contents, &mut output);
+        output
+    }
```
ia0, line 55: Same here.

This is particularly important here, because if someone just implements `hash_mut` then `hash` is wrong. We should probably document that the `_mut` variants should be implemented first (and the possibly the non-`_mut` ones). Otherwise we could provide helper functions like `RngCore`.

libraries/opensk/src/api/crypto/sha256.rs
```diff
@@ -16,11 +16,23 @@ use super::HASH_SIZE;
...
+    /// This function does not let you control the memory allocation. It should therefore not be
+    /// used for secrets that need zeroization.
     fn digest(data: &[u8]) -> [u8; HASH_SIZE] {
         let mut hasher = Self::new();
         hasher.update(data);
-        hasher.finalize()
+        let mut output = [0; HASH_SIZE];
+        hasher.finalize(&mut output);
+        output
+    }
```
ia0, line 30: Same here, the non-`_mut` version should be implemented using the `_mut` version.

ia0 (dismissed):

libraries/crypto/src/ec/exponent256.rs
kaczmarczyck: Good point, I updated the documentation to warn about that for all types that derive `Zeroize`.

libraries/crypto/src/lib.rs
kaczmarczyck: Done

libraries/opensk/src/api/crypto/sha256.rs
kaczmarczyck: Done

libraries/crypto/src/lib.rs
kaczmarczyck: I decided to document this requirement on both affected traits. Before I'd implement helpers for the `libraries/crypto` trait, I'd rather remove it. I don't think we are going to add more types to it. But the commit is the shortest fix.

libraries/crypto/src/ec/gfp256.rs
```diff
@@ -17,10 +17,14 @@ use core::ops::Mul;
...
-// This implements enough methods to validate points and to convert them to/from the Montgomery
-// form, which is more convenient to operate on.
+/// A field element on the elliptic curve, that is an element modulo the prime P.
+///
+/// This is the format used to serialize coordinates of points on the curve.
+/// This implements enough methods to validate points and to convert them to/from the Montgomery
+/// form, which is more convenient to operate on.
+///
+/// Never call zeroize explicitly, to not invalidate any invariants.
+
```
ia0: Why is there a space here?

ia0 (dismissed):

libraries/crypto/src/ec/gfp256.rs
kaczmarczyck: Désolé, fixed

ia0 (approved):

PR #618, opened 2023-04-24
Vendor Command + HID fix

kaczmarczyck:
I did all my tests on hardware with this fix, and now I'm surprised that it didn't end up on develop. So should have been part of #614 .


kaczmarczyck:
I decided to test more, and realized that the upgrade script doesn't correctly work with Vendor HID since some time. I added the easiest fix for that, plus a test.

tools/perform_upgrade.sh
```diff
@@ -18,12 +18,12 @@
...
 # To be run from the OpenSK base path.
 
-set -e
+set -ex
 
 BOARD="$1"
 
 ./deploy.py --board="${BOARD}" --opensk --programmer=none $2
-python3 -m tools.deploy_partition --board="${BOARD}" $2
+python3 -m tools.deploy_partition --board="$@"
```
ia0: This is a bit unintuitive. Those 3 lines would be more readable like that:

```sh
BOARD="$1"; shift

./deploy.py --board="${BOARD}" --opensk --programmer=none "$@"
python3 -m tools.deploy_partition --board="${BOARD}" "$@"
```

If the arguments are different between both lines, then this script should get some documentation on how to use it.

kaczmarczyck: I was considering removing the script and just putting both Python calls into the documentation on how to upgrade. It would remove the need to document, as argparse for both of them has some hints how to correctly call them. Since we have issues with the new version of `nrfjprog` anyway, that wouldn't cost us much. WDYT?

ia0: Removing the script sounds good to me.

kaczmarczyck: Done, and the documentation is updated.

ia0 (approved):

PR #620, opened 2023-04-26
Tock V2 port - rebased and updated

kaczmarczyck:
This PR is similar to #580. I made some changes and new observations during testing. This PR is WIP still. Most parts of this PR are ready for review, only the `patches/` need more attention (see observations).

Differences (FYI @L0g4n) :

- I removed all configs for the opentitan board, as I can't test on them. Adding them back shouldn't be hard: Changes were in `boards/` and `deploy.py`.
- I reverted to RTT debugging for the dev board, instead of UART.

Observations:

- Binary size decreases by 6 kB.
- If we use RustCrypto instead of our own implementation, binary size increases by 23 kB (compared to Tock V2 with our own crypto). Also usage feels much slower. It was working otherwise, so it should work for any non-embedded implementations.
- The storage driver implementation needs a page-sized buffer and does potentially unnecessary copies on each write.
- ~~The USB driver makes the kernel panic when canceling UP checks. My best guess is the `CTAP_CMD_TRANSMIT_OR_RECEIVE` command, is this command is the unique source of CTAP HID cancel packets (not to be confused with the USB cancel syscall).~~


deploy.py
```diff
@@ -502,14 +502,16 @@ def create_tab_file(self, binary_names: Dict[str, str]):
     elf2tab_ver = self.checked_command_output(
         ["elf2tab/bin/elf2tab", "--version"]).split(
             "\n", maxsplit=1)[0]
-    if elf2tab_ver != "elf2tab 0.7.0":
-      error(("Detected unsupported elf2tab version {elf2tab_ver!a}. The "
-             "following commands may fail. Please use 0.7.0 instead."))
+    if elf2tab_ver != "elf2tab 0.10.2":
+      error(("Detected unsupported elf2tab version {elf2tab_ver}! The "
```
jmichelp: Are you sure you don't need the formatting parameter `!a` anymore?

boards/nordic/nrf52840_dongle_opensk/src/main.rs
```diff
@@ -91,16 +99,18 @@ pub struct Platform {
...
         'static,
+        VirtualMuxAlarm<'static, nrf52840::rtc::Rtc<'static>>,
         components::process_console::Capability,
     >,
     console: &'static capsules::console::Console<'static>,
     gpio: &'static capsules::gpio::GPIO<'static, nrf52840::gpio::GPIOPin<'static>>,
     led: &'static capsules::led::LedDriver<
         'static,
         LedLow<'static, nrf52840::gpio::GPIOPin<'static>>,
+        4,
```
jmichelp, line 110: What does the 4 here means?
Is it the amount of LEDs (the dongle only has 2 LEDs but 1 is RGB)?

boards/nordic/nrf52840_dongle_opensk/src/main.rs
```diff
@@ -368,15 +444,20 @@ pub unsafe fn reset_handler() {
...
         usb,
-        crp,
-        ipc: kernel::ipc::IPC::new(board_kernel, &memory_allocation_capability),
+        ipc: kernel::ipc::IPC::new(
+            board_kernel,
+            kernel::ipc::DRIVER_NUM,
+            &memory_allocation_capability,
+        ),
+        scheduler,
+        systick: cortexm4::systick::SysTick::new_with_calibration(64000000),
```
jmichelp: Nit: underscores to make the number easier to read

boards/nordic/nrf52840_mdk_dfu/src/main.rs
```diff
@@ -321,14 +401,20 @@ pub unsafe fn reset_handler() {
...
         nvmc,
         usb,
-        ipc: kernel::ipc::IPC::new(board_kernel, &memory_allocation_capability),
+        ipc: kernel::ipc::IPC::new(
+            board_kernel,
+            kernel::ipc::DRIVER_NUM,
+            &memory_allocation_capability,
+        ),
+        scheduler,
+        systick: cortexm4::systick::SysTick::new_with_calibration(64000000),
```
jmichelp, line 410: Nit: underscores for readability on big numbers

patches/tock/06-persistent-storage.patch
```diff
@@ -112,21 +121,17 @@ index adbc2a2b5..4092cf346 100644
...
-+    waiting: OptionalCell<AppId>,
++    apps: NvmcDriverGrant,
++    waiting: OptionalCell<ProcessId>,
 +    deferred_caller: &'static DynamicDeferredCall,
 +    deferred_handle: OptionalCell<DeferredCallHandle>,
++    buffer: TakeCell<'static, [u8]>,
 +}
 +
-+pub const DRIVER_NUM: usize = 0x50003;
++pub const DRIVER_NUM: usize = 0x50_003;
```
jmichelp: Seems some rustfmt configuration issue here. For hex, it would be more natural to split them by group of 4 characters instead of 3 like in base-10

kaczmarczyck:
The USB driver is fixed. I tested a bit more, and OpenSK worked reliably on the second commit. Desription is updated.

third_party/libtock-drivers/src/usb_ctap_hid.rs
```diff
@@ -12,322 +12,510 @@
...
-        timeout_delay.ms(),
-        buf as &[u8]
-    )
-    .unwrap();
-
-    let result = send_or_recv_with_timeout_detail(buf, timeout_delay, endpoint);
-
-    #[cfg(feature = "verbose_usb")]
-    if let Ok(SendOrRecvStatus::Received(received_endpoint)) = result {
+    /// Sends a packet to a give endpoint.
```
jmichelp: typo: given

.github/workflows/cargo_clippy.yml
```diff
@@ -25,10 +25,10 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-targets --features std
```
jmichelp, line 28: Just to make sure: does this mean that the clippy workflow won't test examples but only the main code?

boards/nordic/nrf52840_dongle_opensk/src/main.rs
kaczmarczyck: It is the number of LEDs, and the blinking behavior is the same as it was before. So it probably counts the RGB as 3.

.github/workflows/cargo_clippy.yml
kaczmarczyck: Examples will still be checked with `cargo check`. Since clippy is running with `--features std`, examples don't compile. So yes! To have examples compile with std, we'd need to make them more verbose with lots of compile flag switching.

deploy.py
kaczmarczyck: Reverted

kaczmarczyck: Reverted

boards/nordic/nrf52840_mdk_dfu/src/main.rs
kaczmarczyck: Done

patches/tock/06-persistent-storage.patch
kaczmarczyck: Reverted in both appearances

third_party/libtock-drivers/src/usb_ctap_hid.rs
kaczmarczyck: Done

ia0:
I tested the following on a dev board (clear storage + panic console):
- reset, setup, deploy, configure
- webauthn.io register, authenticate, restart, authenticate
- CTAP2 test tool (see below)

I don't know if it's the test suite or this PR, but some button touch are not taken into account.

The test suite did not finish successfully, I've hit the following error:

```
Test successful: Tests whether the PIN auth token regenerates after replug.
You have 10 seconds for the next touch after pressing enter.
Please replug the device, then hit enter.

F0503 11:57:45.783972 1074220 command_state.cc:64] Check failed: fido2_tests::Status::kErrNone == device_->Init() CTAPHID initialization failed
[...]
    @     0x5619ac5e0e7a  fido2_tests::CommandState::PromptReplugAndInit()
    @     0x5619ac5e0f9e  fido2_tests::CommandState::Reset()
    @     0x5619ac5e1305  fido2_tests::CommandState::Prepare()
    @     0x5619ac5de469  fido2_tests::BaseTest::Setup()
    @     0x5619ac594a99  fido2_tests::runners::RunTests()
[...]
```

I've also hit the following warnings and errors during test suite:

```
[...]
Test successful: Tests if the PIN protocol parameter is checked in MakeCredential.
Expected error code `CTAP2_ERR_PIN_AUTH_INVALID`, got `CTAP1_ERR_INVALID_PARAMETER`.
Test successful: Tests if a PIN auth is rejected without a PIN set in MakeCredential.
[...]
Failed test: Tests if client PIN fails with missing parameters in MakeCredential. (id: make_credential_pin_auth_missing_parameter) - Missing PIN parameters were not rejected when PIN is set.
A prompt was sent unexpectedly.
Expected error code `CTAP2_ERR_PIN_REQUIRED`, got `CTAP2_OK`.
[...]
Test successful: Tests if storing lots of credentials is handled gracefully.
The test for full store errors was aborted after 50 credentials were successfully created.
[...]
Failed test: Tests is user verification set to true is accepted in GetAssertion. (id: get_assertion_option_uv_true) - The user verification option (true) was not accepted.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
The failing error code is `CTAP2_ERR_PIN_AUTH_INVALID`.
[...]
Test successful: Tests if assertions with resident keys work.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
Please touch your security key!
[...]
Test successful: Tests if assertions with non-resident keys work.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
Please touch your security key!
[...]
Test successful: Tests if the PIN protocol parameter is checked in GetAssertion.
Expected error code `CTAP2_ERR_PIN_AUTH_INVALID`, got `CTAP1_ERR_INVALID_PARAMETER`.
Please touch your security key!
[...]
Failed test: Tests if the PIN auth is correctly checked with a PIN set in GetAssertion. (id: get_assertion_pin_auth) - Falsely rejected valid PIN auth.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
The failing error code is `CTAP2_ERR_PIN_AUTH_INVALID`.
[...]
```

Not sure if any of those should be particularly blocking. LGTM to proceed otherwise.

kaczmarczyck:
The first looks like something that happens when the device hasn't booted quickly enough after a reset. The retry loop in the Test Tool doesn't always find the device. Is it reproducible for you? For the warnings and errors, I'll check if any of them is a regression.  They could be outdated tests.

ia0:
The warnings and errors are reproducible, here are the remaining ones:

```
[...]
Test successful: Tests if Reset actually deletes credentials.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
[...]
RESULTS
Failed test: Tests if client PIN fails with missing parameters in MakeCredential. (id: make_credential_pin_auth_missing_parameter) - Missing PIN parameters were not rejected when PIN is set.
A prompt was sent unexpectedly.
Expected error code `CTAP2_ERR_PIN_REQUIRED`, got `CTAP2_OK`.
Failed test: Tests is user verification set to true is accepted in GetAssertion. (id: get_assertion_option_uv_true) - The user verification option (true) was not accepted.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
The failing error code is `CTAP2_ERR_PIN_AUTH_INVALID`.
Failed test: Tests if the PIN auth is correctly checked with a PIN set in GetAssertion. (id: get_assertion_pin_auth) - Falsely rejected valid PIN auth.
A prompt was expected, but not performed. Sometimes it is just not recognized if performed too fast.
The failing error code is `CTAP2_ERR_PIN_AUTH_INVALID`.
Passed 63 out of 66 tests.
To re-run tests that failed, supply the following flag:
--test_ids=make_credential_pin_auth_missing_parameter,get_assertion_option_uv_true,get_assertion_pin_auth
[...]
```

This time I gave plenty of time between resetting the board and hitting enter in the test suite. The tool did not fail.

Regarding the button issue, I'm even more unsure it's the PR. I used to use the next message from the tool as feedback that the button was pressed, but sometimes the tool takes a considerable amount of time before showing the next line. I tried to check whether the LEDs were blinking and could confirm that some of the times when I didn't get feedback in the tool, the LEDs actually stopped blinking which means the button touch worked. However, it's hard to keep an eye on both the tool and the LEDs at the same time, so not sure if all cases can be explained with this theory. Suggestion would be to improve the tool by printing when the button is touched.

I also noticed that when the tool was saying "don't touch although LEDs are blinking", it happened once that they were not blinking (or were blinking after some non-negligible amount of time). Not sure if expected.

Apart from that, no major testing issues identified on my side.

zhalvorsen:
Have you checked boot time differences on the dev board between Tock V1 and Tock V2? It sounds like you are seeing the same thing I am.

I'm about 1/4 the way through reviewing. I'll finish today.

kaczmarczyck:
From looking at the debug output, it looks like Tock boots in under 1 second for me on the dev board. No big difference between V1 and V2.

kaczmarczyck:
Also the Test Tool outputs from before and after this PR are identical. Not sure about the other observations you made, but most tests consist of more than one call to the authenticator. So there can be delay between starting the test and the blinking. In this specific case, there is an easier explanation: After the long warning message that you shouldn't touch, the Tool waits 2 seconds. This is because people tend to just spam press the button to pass all tests, and it's hard to react to the warning message otherwise.

So looks to me like everything is working so far? Also no problems on the dongle for me.

ia0 (dismissed):
I didn't review, only tested as described in https://github.com/google/OpenSK/pull/620#issuecomment-1532763449 and https://github.com/google/OpenSK/pull/620#issuecomment-1532896104. No blocking issues on my side. However, I think we should wait at least @zhalvorsen or @jmichelp to approve before merging.

src/env/tock/storage.rs
```diff
@@ -20,34 +20,37 @@ use super::TockEnv;
...
     include_bytes!(concat!(env!("OUT_DIR"), "/opensk_upgrade_pubkey.bin"));
 
 mod subscribe_nr {
-    pub const DONE: usize = 0;
+    pub const DONE: u32 = 0;
 }
 
 mod command_nr {
-    pub const GET_INFO: usize = 1;
+    pub const GET_INFO: u32 = 1;
```
zhalvorsen, line 41: Bringing a question from the previous PR here.

Is there a specific reason this syscall driver is here instead of in `third_party/libtock-drivers`? I think it is the only one that lives outside

src/env/tock/storage.rs
```diff
@@ -362,23 +394,23 @@ impl TockUpgradeStorage {
         let write_range = ModRange::new(address, data.len());
         if self.contains_metadata(&write_range)? {
             let new_metadata = &data[self.metadata.start() - address..][..self.metadata.length()];
-            check_metadata::<TockEnv>(self, UPGRADE_PUBLIC_KEY, new_metadata)?;
+            check_metadata::<TockEnv<S>, S, C>(self, UPGRADE_PUBLIC_KEY, new_metadata)?;
```
zhalvorsen: It looks like we are taking the default config for TockEnv here. Do we want to pass the config on from above?
```suggestion
            check_metadata::<TockEnv<S, C>, S, C>(self, UPGRADE_PUBLIC_KEY, new_metadata)?;
```

zhalvorsen (dismissed):

zhalvorsen:
LGTM but I didn't look too closely at the patches

src/env/tock/storage.rs
kaczmarczyck: Historic reasons. I could move it, this is just the smallest diff solution. I can see why that could make things cleaner.

zhalvorsen: If we do it, it would probably be better in a different PR anyway. So I'm good to move on.

kaczmarczyck: Sounds good. I'll look into it tomorrow and merge if @jmichelp has no objections.

zhalvorsen (approved):

kaczmarczyck:
MacOS CI didn't run on the last small commit, but they were green on all other commits.

L0g4n:
Thanks for the work of all contributors 👍🏾

kaczmarczyck:
Thanks for your initiative! If you want to add your board, feel free to open another PR. :)

PR #624, opened 2023-05-05
Credential wrapping in Env

kaczmarczyck:
Moves all logic for credential wrapping into the key store. No change in logic. We expose and use more data types from `src/ctap`, but I'm not sure if we should move all of them into `src/api`, or re-export them. Maybe a cleanup for a next PR, when we know a general strategy.


libraries/opensk/src/ctap/mod.rs
```diff
@@ -16,9 +16,8 @@ pub mod apdu;
 mod client_pin;
 pub mod command;
 mod config_command;
-mod credential_id;
 mod credential_management;
-mod crypto_wrapper;
+pub mod crypto_wrapper;
```
ia0, line 20: `PrivateKey` should indeed be moved out to `api`. That's the only type that needs to move.

libraries/opensk/src/api/key_store.rs
```diff
@@ -12,12 +12,60 @@
...
+// - 16  bytes: initialization vector for AES-256,
+// - 192 bytes: encrypted block of the key handle cbor,
+// - 32  bytes: HMAC-SHA256 over everything else.
+pub const CBOR_CREDENTIAL_ID_SIZE: usize = 241;
+pub const MIN_CREDENTIAL_ID_SIZE: usize = LEGACY_CREDENTIAL_ID_SIZE;
+pub const MAX_CREDENTIAL_ID_SIZE: usize = CBOR_CREDENTIAL_ID_SIZE;
+
+pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x01;
+
+pub const MAX_PADDING_LENGTH: u8 = 0xBF;
```
ia0: Why do all those constants need to be public?

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: What about the others we expose and use?

- `Secret`
- `CredentialProtectionPolicy`
- `aes256_cbc_decrypt`, `aes256_cbc_encrypt`
- `cbor_read`, `cbor_write`

And more for vendor commands from the other PR:

- `Ctap2StatusCode`
- `check_user_presence`
- several other types from `data_formats::`

I think this is the downside of more freedom in `Env`: You need to understand some CTAP concepts to implement credential wrapping, vendor commands etc.

libraries/opensk/src/api/key_store.rs
kaczmarczyck: Some don't, fixed. The remaining are exposed in GetInfo or used in CTAP1.

libraries/opensk/src/ctap/mod.rs
ia0: AFAICT, this PR only exposes `PrivateKey`. This is true that we should do the same for `Secret` but that can be another PR. I can't see usages of `Ctap2StatusCode` but yes, if it's part of functions of the API, it should also be in the API, but I think so far we use custom errors in the API.

So for this PR I would just suggest to fix `PrivateKey`. And we can have another PR for `Secret`. The rest so far looks good to me, but I might be missing something. Note that I consider `data_formats` part of the API. But we could split that if some types should not be public.

libraries/opensk/src/api/key_store.rs
ia0: But those are not part of the API. We should use `pub(crate)` then.

libraries/opensk/src/ctap/mod.rs
ia0: Actually the best way to figure out what should be in the API, is just to compile the documentation with `cargo doc` and go through all the API looking at the exposed types and see if they make sense.

kaczmarczyck: Question is if all `data_formats` should split and some of it moves to API?

The other types and functions like `Ctap2StatusCode` and `check_user_presence` are exposed so that we can implement custom vendor functions. We could reimplement all of them, and have trivial conversions into the internal types, but that seems like a bad idea.

If we go the `cargo doc` route and plan our API, I think that's a bigger change than what is proposed here. So just moving `PrivateKey` is one option for now?

libraries/opensk/src/api/key_store.rs
kaczmarczyck: Being strict with `pub` and `pub(crate)` is probably a bigger chance also, that clearly divides what is supposed to be in the API and what isn't. Probably as much effort as hiding `src/ctap` from the outside and re-exporting in `src/lib` only what we need, and doing a proper API design.

libraries/opensk/src/ctap/mod.rs
ia0: > So just moving PrivateKey is one option for now?

Yes, that's exactly what my initial comment was about. It's fine to not fix everything, but PRs should not add new things to fix.

The rest is just the guidelines to decide what should be in the API or not. Such that future PRs follow that guideline.

libraries/opensk/src/api/key_store.rs
ia0: I'm not saying to `pub(crate)` everything that should be `pub(crate)`, I'm just saying we should not make those constant public in this PR while they were not public before. None of those constants should be in the API as far as I can tell.

kaczmarczyck: If you implement your own credential wrapping, but want to add padding to match the standard size, then `CBOR_CREDENTIAL_ID_SIZE` would be useful.

ia0: Yes good point. Let's keep that one and mention it explicitly in the documentation of the wrap function.

kaczmarczyck: Added to the trait description.

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: Right, done!

libraries/opensk/src/api/key_store.rs
```diff
@@ -29,19 +30,18 @@ use rand_core::RngCore;
...
 pub const CBOR_CREDENTIAL_ID_SIZE: usize = 241;
-pub const MIN_CREDENTIAL_ID_SIZE: usize = LEGACY_CREDENTIAL_ID_SIZE;
-pub const MAX_CREDENTIAL_ID_SIZE: usize = CBOR_CREDENTIAL_ID_SIZE;
+const MIN_CREDENTIAL_ID_SIZE: usize = LEGACY_CREDENTIAL_ID_SIZE;
+pub(crate) const MAX_CREDENTIAL_ID_SIZE: usize = CBOR_CREDENTIAL_ID_SIZE;
 
-pub const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x01;
-
-pub const MAX_PADDING_LENGTH: u8 = 0xBF;
+const CBOR_CREDENTIAL_ID_VERSION: u8 = 0x01;
```
ia0: This needs to be public too. It's mentioned in the doc.

libraries/opensk/src/api/key_store.rs
```diff
@@ -81,6 +81,14 @@ pub trait KeyStore {
     /// The output is encrypted and authenticated. Since the wrapped credentials are passed to the
     /// relying party, the choice for credential wrapping impacts privacy. Looking at their size and
     /// structure, a relying party can guess the authenticator model that produced it.
+    ///
+    /// A credential ID that imitates the default needs the following structure:
+    /// - The length is `CBOR_CREDENTIAL_ID_SIZE`.
+    /// - The first byte is the version. The latest version is `CBOR_CREDENTIAL_ID_VERSION`.
```
ia0: nit: You can make those links such that it's easy to navigate the doc.

```rust
    /// - The length is [`CBOR_CREDENTIAL_ID_SIZE`].
    /// - The first byte is the version. The latest version is [`CBOR_CREDENTIAL_ID_VERSION`].
```

(notice the square brackets)

libraries/opensk/src/api/private_key.rs
```diff
@@ -0,0 +1,328 @@
...
+            PrivateKey::Ecdsa(ecdsa_seed) => {
+                CoseKey::from_ecdsa_public_key(ecdsa_key_from_seed(env, ecdsa_seed)?.public_key())
+            }
+            #[cfg(feature = "ed25519")]
+            PrivateKey::Ed25519(ed25519_key) => CoseKey::from(ed25519_key.public_key()),
+        })
+    }
+
+    /// Returns the encoded signature for a given message.
+    pub fn sign_and_encode(
```
ia0: nit: This could be `pub(crate)`.

libraries/opensk/src/api/key_store.rs
kaczmarczyck: True, fixed

libraries/opensk/src/api/key_store.rs
kaczmarczyck: Done

libraries/opensk/src/api/private_key.rs
kaczmarczyck: Done

zhalvorsen:
LGTM

ia0 (approved):

PR #626, opened 2023-05-06
macos-latest for GitHub workflows

kaczmarczyck:
Workflows stopped running since we were on old MacOS versions. Using latest should fix this longer-term.


kaczmarczyck:
Missing workflows should be the old versions from before this PR that are trying to run.

ia0 (approved):

PR #628, opened 2023-05-07
Compile flag for AuthenticatorConfig

kaczmarczyck:
The command can be disabled for authenticators that don't want users to change their configuration.

To try, call the new tool:
```
python3 -m tools.authenticator_config --always-uv
```


libraries/opensk/src/ctap/client_pin.rs
```diff
@@ -119,6 +119,7 @@ pub enum PinPermission {
     CredentialManagement = 0x04,
     _BioEnrollment = 0x08,
     LargeBlobWrite = 0x10,
+    #[cfg(any(feature = "config_command", feature = "std"))]
```
ia0: What's the rational to have `std`? If we want `std` to imply `config_command` we can do it at the feature level by adding `config_command` in the dependencies of `std`.

libraries/opensk/src/ctap/storage.rs
```diff
@@ -39,6 +41,8 @@ struct PinProperties {
     hash: [u8; PIN_AUTH_LENGTH],
 
     /// Length of the current PIN in code points.
+    // This field is unused when compiled without config_command.
+    #[allow(dead_code)]
```
ia0: This could become `[cfg_attr(not(feature = "config_command"), allow(dead_code))]` when `std` is moved out of the equation.

libraries/opensk/src/ctap/client_pin.rs
kaczmarczyck: An integration test for enterprise attestation (with a customization that allows it) would first need to turn EP on through Config, then call MakeCredential. What the feature implies is that if you use it, your Env turns on EP through some other means. We should probably also do that ourselves, now that you say it.

Not allowing the Config command in `std`. would imply that we can't use our current `test_helpers` and would need to add a lot more conditionals throughout tests.

ia0: > Not allowing the Config command in std

What do you mean? I said the opposite: make the `config_command` a dependency of `std`, i.e. the line that currently looks like:

```toml
# libraries/opensk/Cargo.toml
std = ["crypto/std", "persistent_store/std", "rand/std_rng"]
```

would become:

```toml
std = ["crypto/std", "persistent_store/std", "rand/std_rng", "config_command"]
```

kaczmarczyck: I didn't want to imply that this isn't a solution. Just stating the requirements. I agree this is nicer already.

libraries/opensk/src/ctap/storage.rs
kaczmarczyck: Done

libraries/opensk/src/ctap/client_pin.rs
kaczmarczyck: Done

libraries/opensk/src/ctap/storage.rs
```diff
@@ -39,6 +41,8 @@ struct PinProperties {
     hash: [u8; PIN_AUTH_LENGTH],
 
     /// Length of the current PIN in code points.
+    // This field is unused when compiled without config_command.
```
ia0: This comment is now obsolete/redundant.

ia0 (dismissed):

libraries/opensk/src/ctap/storage.rs
kaczmarczyck: True, removed.

ia0 (approved):

PR #636, opened 2023-06-27
Remove duplicated alarm syscall.

zhalvorsen:
The alarm syscall is implemented in libtock-rs, but was duplicated here. This removes the duplicated code and changes the references to point to libtock-rs directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,6 +168,7 @@ dependencies = [
  "ed25519-compact",
  "enum-iterator",
  "lang_items",
+ "libtock_alarm",
  "libtock_buttons",
  "libtock_console",
  "libtock_drivers",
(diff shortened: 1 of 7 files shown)


ia0 (approved):
I tested on webauthn (register, login, reset, login) and it works.

However, I was able to generate a panic too (but unrelated to this PR):

```
panicked at 'assertion failed: self.pending_out.take()', capsules/src/usb/usbc_ctap_hid.rs:364:17
```

I don't really have reproduction steps because it looks like a timing issue. Here are the steps I'm taking:
- Have a nordic dev board with opensk and another security key (U2F only) connected (not sure it matters, but that's the setup)
- Register both.
- Authenticate repeatedly (cancelling the popup and not touching any key)
- The dev board will eventually panic when clicking Authenticate (not when cancelling the popup)

kaczmarczyck (approved):
If @ia0 's panic is unrelated to the PR, feel free to merge. We can always test and fix it more later.

PR #633, opened 2023-06-12
Move storage syscalls to libtock-drivers

kaczmarczyck:
Fixes #627

This PR is purely stylistic, and doesn't change any functionality.

PR #637, opened 2023-06-29
Add ability to get more storage information.

zhalvorsen:
This adds parameters to the storage location syscalls to specify which storage location to get the address and size for.


ia0 (approved):
Thanks! This is actually a fix of a regression introduced by #633.

PR #641, opened 2023-08-08
Add more transparency into some EC structures.

zhalvorsen:
This adds the ability to create ECDH keys from raw bytes and export signatures as raw bytes.

### DIFF
--- a/libraries/crypto/src/ecdh.rs
+++ b/libraries/crypto/src/ecdh.rs
@@ -78,6 +78,17 @@ impl SecKey {
         p.getx().to_int().to_bin(&mut x);
         x
     }
+
+    /// Creates a private key from the exponent's bytes, or None if checks fail.
+    pub fn from_bytes(bytes: &[u8; 32]) -> Option<SecKey> {
+        let a = NonZeroExponentP256::from_int_checked(Int256::from_bin(bytes));
+        // The branching here is fine because all this reveals is whether the key was invalid.
+        if bool::from(a.is_none()) {
+            return None;
+        }
+        let a = a.unwrap();
+        Some(SecKey { a })
+    }
 }
 
 impl PubKey {
--- a/libraries/crypto/src/ecdsa.rs
+++ b/libraries/crypto/src/ecdsa.rs
@@ -19,9 +19,7 @@ use super::ec::point::PointP256;
 use super::Hash256;
 use alloc::vec;
 use alloc::vec::Vec;
-#[cfg(feature = "std")]
-use arrayref::array_mut_ref;
-use arrayref::{array_ref, mut_array_refs};
+use arrayref::{array_mut_ref, array_ref, mut_array_refs};
 use core::marker::PhantomData;
 use rand_core::RngCore;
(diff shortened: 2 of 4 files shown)


libraries/crypto/Cargo.toml
```diff
@@ -25,3 +25,4 @@ zeroize = { version = "1.5.7", features = ["derive"] }
 [features]
 std = ["hex", "ring", "untrusted", "serde", "serde_json", "regex", "rand_core/getrandom"]
 with_ctap1 = []
+acvp = []
```
kaczmarczyck: I don't think we need a feature. Just make anything public without the feature as necessary!

libraries/opensk/src/ctap/hid/receive.rs
```diff
@@ -83,6 +83,7 @@ impl<E: Env> MessageAssembler<E> {
         let (cid, processed_packet) = CtapHid::<E>::process_single_packet(packet);
         if let Some(locked_cid) = locked_cid {
             if locked_cid != cid {
+                #[cfg(not(feature = "acvp"))]
```
kaczmarczyck: Why is this necessary? This should probably not be here.

libraries/crypto/Cargo.toml
zhalvorsen: In both crates?

libraries/opensk/src/ctap/hid/receive.rs
zhalvorsen: Removed

libraries/crypto/Cargo.toml
kaczmarczyck: Yes, go ahead! It is just accessors that, if useful for us, might be useful for others.

zhalvorsen: Done

kaczmarczyck (dismissed):

libraries/opensk/src/api/crypto/rust_crypto.rs
```diff
@@ -85,6 +85,11 @@ impl ecdh::SecretKey for SoftwareEcdhSecretKey {
         let shared_secret = self.ephemeral_secret.diffie_hellman(&public_key.public_key);
         SoftwareEcdhSharedSecret { shared_secret }
     }
+
+    fn from_slice(_bytes: &[u8; EC_FIELD_SIZE]) -> Option<Self> {
+        // Currently (8/8/2023) cannot create EphemeralSecret from a slice. Update this, if needed.
+        unimplemented!();
```
kaczmarczyck: I didn't know that RustCrypto doesn't support this operation. In that case, we probably also shouldn't on this repository.

kaczmarczyck (commented):
I have to retract on one of the new functions, sorry for the extra work.

libraries/opensk/src/api/crypto/rust_crypto.rs
zhalvorsen: Removed

kaczmarczyck (approved):

PR #642, opened 2023-08-09
New key wrapping API

kaczmarczyck:
Allows key wrapping to be different between persistent and server-side storage.

To accomplish this, the PrivateKey now always stores the fully reconstructed key, and has different methods to serialize it for the respective use case.

### DIFF
--- a/libraries/opensk/src/api/key_store.rs
+++ b/libraries/opensk/src/api/key_store.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use crate::api::crypto::aes256::Aes256;
-use crate::api::crypto::ecdsa::SecretKey as _;
 use crate::api::crypto::hmac256::Hmac256;
 use crate::api::crypto::HASH_SIZE;
 use crate::api::private_key::PrivateKey;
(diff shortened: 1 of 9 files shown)


kaczmarczyck:
This PR adds 600 B of binary size.

kaczmarczyck:
This PR is a breaking change for all credentials created with OpenSK before: both server-side and resident keys.

We can retroactively change that, if we want to be compatible with `stable`: just revert the second commit. If we also want to be backwards compatible with a some intermediate `develop` stages, we would have to parse the CBOR in `PrivateKey` differently: If the byte array is longer, assume that it is encrypted and decrypt. Otherwise just use it as is.

As upgradability was only introduced in `develop`, there shouldn't yet be users that don't actively pull `develop` themselves that run into this problem. If you trigger this backwards incompatibility, you notice that your credentials get rejected for server-side, or don't appear to exist for resident keys.

However, the credentials are not actually gone! Flashing a firmware before this PR should solve the problem. You'd lose access to the freshly generated credentials from the new firmware though, as the change is neither backwards not forward compatible.

ia0 (dismissed):

ia0 (approved):

PR #651, opened 2023-09-28
Update install.md to add correct board type

### LINKED ISSUE #650: wrong board type in flashing command in install.md
## Expected Behavior
Board type wrong in install.md

```
If your board is already flashed with Tock OS, you may skip installing it: ./deploy.py --board=nrf52840dk --opensk --no-tockos
```
should be :
```
If your board is already flashed with Tock OS, you may skip installing it: ./deploy.py --board=nrf52840dk_opensk --opensk --no-tockos
```

iayanpahwa:
Change board from nrf52840dk to nrf52840dk_opensk

Fixes #650

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -111,7 +111,7 @@ above. However, we support other programmers:
     left to the user.
 
 If your board is already flashed with Tock OS, you may skip installing it:
-`./deploy.py --board=nrf52840dk --opensk --no-tockos`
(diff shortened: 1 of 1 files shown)


ia0 (commented):
Thanks for the PR but the stable branch does not have an `nrf52840dk_opensk` board. This is specific to the `develop` branch. And the documentation there is up-to-date. By default on Github shows you the stable branch which may be confusing.

PR #653, opened 2023-11-06
venv for Python setup

### LINKED ISSUE #647: Setup script can't install pip packages to the user's environment on Arch Linux
## Expected Behavior
Setup script install pip3 packages into the user's home directory without needing elevated permissions.

## Actual Behavior
The setup script fails with this Output:

```
[-] Copying additional boards to Tock... DONE.
[-] Applying patch "01-persistent-storage.patch"... DONE.
[-] Applying patch "02-usb.patch"... DONE.
[-] Applying patch "03-usb-debugging.patch"... DONE.
[-] Applying patch "04-additional-boards.patch"... DONE.
[-] Applying patch "05-mpu-fix.patch"... DONE.
[-] Applying patch "06-update-uicr.patch"... DONE.
[-] Applying patch "07-firmware-protect.patch"... DONE.
[-] Applying patch "01-remove-test-runner.patch"... DONE.
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try 'pacman -S
    python-xyz', where xyz is the package you are trying to
    install.

    If you wish to install a non-Arch-packaged Python package,
    create a virtual environment using 'python -m venv path/to/venv'.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.

    If you wish to install a non-Arch packaged Python application,
    it may be easiest to use 'pipx install xyz', which will manage a
    virtual environment for you. Make sure you have python-pipx
    installed via pacman.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

1. Run `./setup.sh`.

## Specifications

- Version: commit 5b67f3f9a8920ed1d503a6772a406fa04727d4b7
- Platform: Arch Linux
```
$ uname -a
Linux REDACTED 6.4.12-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 24 Aug 2023 00:38:14 +0000 x86_64 GNU/Linux
```

kaczmarczyck:
Makes the Python part of OpenSK's installer independent of system changes.

Some missing Python packets for testing are added to the requirements.

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env py_virtual_env/bin/python3
+#!py_virtual_env/bin/python3
 # Copyright 2019-2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/deploy.py
+++ b/deploy.py
@@ -1,5 +1,5 @@
-#!/usr/bin/env python3
-# Copyright 2019 Google LLC
+#!/usr/bin/env py_virtual_env/bin/python3
+# Copyright 2019-2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tools/configure.py
+++ b/tools/configure.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env py_virtual_env/bin/python3
+#!py_virtual_env/bin/python3
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/tools/configure.py
+++ b/tools/configure.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env py_virtual_env/bin/python3
 # Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/tools/run_pylint.sh
+++ b/tools/run_pylint.sh
@@ -16,4 +16,10 @@
 # Ensure we are at the project root directory
 cd $(readlink -f $(dirname $0))/..
 
-pylint --score=n `git ls-files --deduplicate --exclude-standard --full-name '*.py'`
+if [ -f "py_virtual_env/bin/pylint" ]; then
+    PYLINT=py_virtual_env/bin/pylint
+else
+    PYLINT=pylint
+fi
+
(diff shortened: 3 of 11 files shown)


deploy.py
```diff
@@ -1,5 +1,5 @@
-#!/usr/bin/env python3
-# Copyright 2019 Google LLC
+#!/usr/bin/env py_virtual_env/bin/python3
```
ia0: No need for `env` now that we know the path.
```suggestion
#!py_virtual_env/bin/python3
```

tools/configure.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env py_virtual_env/bin/python3
```
ia0: Same here.
```suggestion
#!py_virtual_env/bin/python3
```

tools/run_pylint.sh
```diff
@@ -16,4 +16,10 @@
 # Ensure we are at the project root directory
 cd $(readlink -f $(dirname $0))/..
 
-pylint --score=n `git ls-files --deduplicate --exclude-standard --full-name '*.py'`
+if [ -f "py_virtual_env/bin/pylint" ]; then
```
ia0: When may this not exist?

ia0 (commented):
LGTM

tools/run_pylint.sh
kaczmarczyck: In our GitHub workflow for Python style. I was considering rewriting the workflow, but this seemed easier.

ia0: I see. There's a few alternatives:
- Call the `setup` script in the python workflow. This should also simplify other manual steps.
- Use `export PATH="py_virtual_env/bin:$PATH"` in this script instead of the `if`.

deploy.py

tools/configure.py
kaczmarczyck: Done

tools/run_pylint.sh
kaczmarczyck: I go with option 2, more elegant.

I'd have to also install `yapf` into our venv in the workflow if it would use setup / venv. And exclude more temporary directories. And it would take longer with the Rust setup that we don't need for Python style.

ia0 (approved):

PR #654, opened 2023-11-07
Updates the metadata to MDS 3

kaczmarczyck:
This metadata is currently submitted to FIDO's MDS.

Workflows are not yet fixed, so we will force push even when they fail.


ia0 (approved):
Just stamping, didn't really review.

jmichelp (approved):

Issue #507, opened 2022-06-23
cargo_clippy github workflow fails on forked repo


egor-duda:
After pushing changes to fork repo, `cargo_clippy` workflow fails with error "Resource not accessible by integration", see e.g. https://github.com/egor-duda/OpenSK/actions/runs/2547812258
This also hides real clippy errors because next `run: cargo clippy ...` command is not run after `actions-rs/clippy-check` failure.

This is caused by actions-rs/clippy-check#2, due to inability of `clippy-check` to add annotations, while running with read-only token.

Adding check for running `clippy-check` action only in main repo fixes this issue, and starts reporting `cargo clippy` errors in forked repos
```
diff --git a/.github/workflows/cargo_clippy.yml b/.github/workflows/cargo_clippy.yml
index 7d10e1c..ac4e08d 100644
--- a/.github/workflows/cargo_clippy.yml
+++ b/.github/workflows/cargo_clippy.yml
@@ -23,6 +23,7 @@ jobs:
         run: ./setup.sh
 
       - uses: actions-rs/clippy-check@v1
+        if: github.repository == 'google/OpenSK'
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --all-targets --features std

```

kaczmarczyck:
Thanks for reporting. I'll look into it when I have time. Seems to be not too high impact, since you can always test locally with `run_desktop_tests` and it works on our main repository.

kaczmarczyck:
The script should now run clippy, locally or in the repository of your choice. Please reopen if the problem persists.

Issue #171, opened 2020-09-29
Only run full setup.sh on relevant workflows


gendx:
## Expected Behavior

GitHub workflows do only the necessary amount of work and complete as fast as possible.

## Actual Behavior

Most of our workflows run `./setup.sh`, which compiles and installs `elf2tab`, even when not needed by the rest of the workflow. This slows down CI.

kaczmarczyck:
Workflows are simplified, and setup is used correctly.

Issue #168, opened 2020-09-29
Run Clippy with various feature combinations


gendx:
It seems that Clippy checks code only for a specific set of features. In #162, we added support for Clippy with `--features std`, we should also test other features, similarly to unit tests.

kaczmarczyck:
Clippy is now run on all libraries.

PR #661, opened 2023-11-08
Reworks workflows into script

### LINKED ISSUE #50: Ensure that the deploy script is also covered by Travis runs

kaczmarczyck:
Advantages are:
- Only one set of tests needs maintenance.
- Local results match workflows, no surprises.
- Reduced reliance on GitHub actions.

Fixes #50, #168, #169, #171, #507


run_desktop_tests.sh
```diff
@@ -15,124 +15,104 @@
...
+cargo check --release --target=thumbv7em-none-eabi --manifest-path bootloader/Cargo.toml
+cargo check --release --manifest-path tools/heapviz/Cargo.toml
 
-echo "Checking that bootloader builds properly..."
-cd bootloader
-cargo check --release --target=thumbv7em-none-eabi
-cd ..
+echo "Checking Rust formatting..."
+cargo fmt -- --check
+cargo +nightly fmt --manifest-path libraries/opensk/Cargo.toml -- --check
```
ia0, line 42: Here we rely on the nightly toolchain to be installed. This happens to be the case on Github and locally for most, but it's not necessarily true. We would need some `rustup install nightly` in `setup.sh`.

And I guess we need nightly because the old nightly has problems.

libraries/opensk/src/api/crypto/rust_crypto.rs
```diff
@@ -205,7 +205,6 @@ impl ecdsa::Signature for SoftwareEcdsaSignature {
         Some(SoftwareEcdsaSignature { signature })
     }
 
-    #[cfg(feature = "std")]
```
ia0, line 208: I guess this was actually used when combining `std` and `rust-crypto`?

ia0 (approved):

libraries/opensk/src/api/crypto/rust_crypto.rs
kaczmarczyck: We needed this function in another Env (outside of `std`), so it had to become part of the API. That was already done for `software_crypto`, but I forgot for `rust_crypto`. Our new checks cover more, so this was caught.

run_desktop_tests.sh
kaczmarczyck: `rustup install nightly` is in `setup.sh`. Do we also need some components, or just an oversight?

kaczmarczyck: And we don't need nightly anymore outside of testing, I think. It's just that I want to test all libraries with the latest clippy etc, to get early warnings when anything breaks.

ia0: My bad, I just checked for `rustup show` and assumed there wouldn't be more. All good.

ia0 (dismissed):

ia0 (approved):

PR #663, opened 2023-11-09
Latest compiler and package versions

### LINKED ISSUE #125: Add a CI workflow to run cargo-outdated checks
The [cargo-outdated](https://github.com/kbknapp/cargo-outdated) tool checks for outdated dependencies in a Cargo.toml. As a complement to cargo-audit which checks for vulnerable dependencies, we should leverage that to make sure dependencies are up-to-date.

### LINKED ISSUE #261: Do not rely on submodules to specify the rust toolchain
We currently use a symlink to specific the `rust-toolchain` based on what is used by `libtock-rs`. This is problematic for github workflows that don't need to clone the submodules since they would use the stable toolchain instead of nightly (and the same version as libtock).

The current proposal solution is to have a copy of the libtock rust-toolchain file instead of a symlink. The CI check executed when bumping libtock should check that the file is in sync.

Once this is fixed, we should clean-up github workflows that force the toolchain to nightly to workaround this issue (e.g. `.github/workflows/persistent_store_test.yml`).

kaczmarczyck:
Other parts of this PR:

- Adds cargo audit to the desktop tests.
- All tests use the same nightly version.
- toolchains reference the file in the base directory.
- Remove 2 unused `Cargo.lock`.

This PR is tested on hardware.


bootloader/rust-toolchain
ia0, line 1: We shouldn't need this file. `cargo` will look up the tree until it finds a `rust-toolchain` file.

.github/workflows/ci.yml
```diff
@@ -30,4 +30,5 @@ jobs:
       - run: ./setup.sh
       - run: rustup component add rustfmt --toolchain nightly
       - run: rustup component add clippy --toolchain nightly
+      - run: rustup component add audit --toolchain nightly
```
ia0: Do we still use nightly?

rust-toolchain
ia0, line 1: nit: Let's rename to `rust-toolchain.toml` given it's a TOML file.

.github/workflows/ci.yml
```diff
@@ -30,4 +30,5 @@ jobs:
       - run: ./setup.sh
       - run: rustup component add rustfmt --toolchain nightly
       - run: rustup component add clippy --toolchain nightly
+      - run: cargo +stable install cargo-audit
```
ia0: Why `+stable`?

.github/workflows/ci.yml
ia0: We can even remove it from the setup script.

.github/workflows/ci.yml
kaczmarczyck: We always install with +stable , i.e., see:
- `cargo +stable install cargo-fuzz` in `./fuzzing_setup.sh`
- `cargo +stable install elf2tab --version 0.10.2 --root elf2tab/` in `setup.sh`

My personal experience is that sometimes tools don't compile on the nightly versions we are running. But they should always run with stable?

bootloader/rust-toolchain
kaczmarczyck: Removed

.github/workflows/ci.yml
kaczmarczyck: Removed

rust-toolchain
kaczmarczyck: Renamed

.github/workflows/ci.yml
ia0: I'm not sure there's any guarantee a crate compiles on either stable or nightly (it happens that people just do breaking change and something stops compiling on stable). It seems simpler to always use the same toolchain everywhere. Also it's a bit ugly to force-install tools globally for the user, but that would be more work to fix.

No strong opinion, just curious why using `+stable`.

.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
           python-version: "3.10"
       - run: ./setup.sh
-      - run: rustup component add rustfmt --toolchain nightly
-      - run: rustup component add clippy --toolchain nightly
+      - run: rustup component add rustfmt
+      - run: rustup component add clippy
```
ia0: We shouldn't need this if they are in the rust-toolchain file.

kaczmarczyck: Ah right, I only needed it before since I wanted to run these specifically as nightly. Let's try without them.

.github/workflows/ci.yml
kaczmarczyck: In that case, I'll go with `stable` as the status quo that worked for me.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,5 @@ jobs:
         with:
           python-version: "3.10"
       - run: ./setup.sh
-      - run: rustup component add rustfmt --toolchain nightly
-      - run: rustup component add clippy --toolchain nightly
+      - run: cargo +stable install cargo-audit
```
ia0: Shouldn't this be in the setup script? We use `cargo audit` in the `run_desktop_test`, not only in CI.

kaczmarczyck: Moved. I was thinking about being conservative in setup and cargo global installs. But we already do it anyway.

kaczmarczyck:
Btw, this PR reduces binary size from 118.6 KiB to 115.0 KiB.

ia0 (approved):

PR #668, opened 2023-12-15
Change PKI so that attestation certs are fully compliant.

### LINKED ISSUE #457: Create correct certificates
Our certificate is not correctly generated, as surfaced in #455. We have an immediate fix for the reportedly breaking website in #456. This issue tracks the correctness of the certifice according to the specification. Some more information is in #455.

Things to watch out for:
- OIDs values must be generated according to the transports we can talk to.
- Instead of having a file for the extensions only, we need a full openssl.conf file.

jmichelp:
Fixes #457 

Initially we generated the smallest certificate possible.
Unfortunately sometimes attestation certificates are thoroughly checked and the FIDO x509v3 extensions must be present.
This PR now creates a PKI (root CA and signing CA) with corresponding CRLs and also allows to create multiple batch certificates for the keys instead of a single one.
The latest generated batch cert/key is automatically symlinked so that the previous documentation still holds.

-   [x] Local tests pass (running `run_desktop_tests.sh`)
-   [ ] Tested against boards
    -   [ ] Nordic nRF52840 DK
    -   [ ] Nordic nRF52840 Dongle (JTAG programmed)
    -   [ ] Nordic nRF52840 Dongle (DFU programmed)
    -   [ ] Makerdiary nRF52840 MDK USB Dongle
-   [x] Appropriate changes to README are included in PR

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2019 Google LLC
+# Copyright 2019-2023 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


kaczmarczyck:
I tested on the dev board, with batch attestation enabled. No regressions.

kaczmarczyck (approved):

PR #670, opened 2023-12-18
Upgrade setup-python workflows

kaczmarczyck:
Maintenance PR for workflows


jmichelp:
Seems that `actions/checkout@v2` is also issuing a warning.
Mind bumping it to `v4` in this PR?

kaczmarczyck:
Okay, I bumped all actions to the latest major versions, lets see.

jmichelp (commented):
LGTM.
Might be worth noting that cargo-bloat and binary-size-reports are the next 2 workflows which could break. But they require a bit more changes than bumping the version

jmichelp (approved):

kaczmarczyck:
Yeah, I realized, but wanted to keep this low effort.

PR #673, opened 2024-01-09
Allows initialization without Reset permission

kaczmarczyck:
This PR is useful for all implementations that can trigger a reboot without user intervention. In these cases, we don't want to allow the Reset command. It should only be allowed after a user initiated power cycle.

Adds tests to the new functionality and a few other coverage holes.


ia0 (dismissed):
Looks good to me, but why is soft reset an input to `Ctap::new`? Could there be an API function that returns whether there was a soft reset? This way `Ctap::new` doesn't take arguments but instead calls into that API function. I'm afraid that we would start adding too many inputs to `Ctap::new`, making all call-sites more complicated.

kaczmarczyck:
Makes sense, changed to be part of `Env`.

ia0 (approved):

PR #679, opened 2024-04-02
Adds a high level persistent storage API

kaczmarczyck:
Changes the store API to a wider API for data in a persistant way. Joins the attestation store into the regular store, and allows more control over how to special case some of the storage.

The `KeyStore` had some persistent elements that were moved out, and the crypto-related parts were kept separately.

In some cases, we can consider unifying even more, like the way the PIN hash is stored. To keep the PR smaller, it only implements the general API changes, no updates to specific parts of the API when not necessary.

One planned follow-up is LargeBlob (see the TODO comment).


libraries/opensk/src/api/persist.rs
```diff
@@ -0,0 +1,367 @@
...
+///
+/// This trait might get appended to with new versions of CTAP.
+///
+/// To implement this trait, you have 2 options:
+/// - Implement all high level functions with default implementations,
+///   calling `unimplemented!` in the key-value accessors.
+///   When we update this trait in a new version, OpenSK will panic when calling any new functions.
+/// - Implement the key-value accessors, and special case as many default implemented high level
+///   functions as desired.
+///   When the trait gets extended, new features will silently work.
```
ia0, line 47: They will silently work but will need explicit migration if a special case is needed. Not sure it's worth updating the doc. Just raising that point.

libraries/opensk/src/api/persist.rs
```diff
@@ -0,0 +1,367 @@
...
+///   Credentials still need keys to be identified by.
+pub trait Persist {
+    /// Retrieves the value for a given key.
```
ia0, line 50: Since we specify the maximum length below, shouldn't we also specify the maximum key here?

Note that we could also not specify anything, then implementation just return an error if the argument is not supported.

libraries/opensk/src/api/persist.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// Values up to a length of 1023 Byte must be supported.
+    fn insert(&mut self, key: usize, value: &[u8]) -> CtapResult<()>;
+
+    /// Removes a key, if present.
+    fn remove(&mut self, key: usize) -> CtapResult<()>;
+
+    /// Iterator for all present keys.
+    fn iter(&self) -> CtapResult<PersistIter<'_>>;
+
+    /// Checks consistency on boot, and if necessary fixes or initializes problems.
```
ia0: typo: fixes problems or initializes a virgin storage

libraries/opensk/src/api/persist.rs
```diff
@@ -0,0 +1,367 @@
...
+        }
+    }
+
+    /// Sets the minimum PIN length.
+    #[cfg(feature = "config_command")]
+    fn set_min_pin_length(&mut self, min_pin_length: u8) -> CtapResult<()> {
+        self.insert(keys::MIN_PIN_LENGTH, &[min_pin_length])
+    }
+
+    /// Returns the list of RP IDs that may read the minimum PIN length.
```
ia0, line 220: nit: let's specify that it defaults to the empty vector (instead of returning `Option<Vec<u8>>`).

ia0 (commented):
I didn't check in details, but for the change to be backward compatible, we need to use the same keys and encode the values the same way. From what I've seen so far it seems like it, but just making sure it's conscious.

libraries/opensk/src/api/persist.rs
kaczmarczyck: I made sure nothing gets moved, but there are 2 new keys:
- `RESET_COMPLETION`, if present Reset again on next boot (to make Reset atomic)
- `ATTESTATION_ID`, to remember what kind of attestation is stored

We assume `Batch` if no ID is stored, to be backwards compatible. So should be accounted for, reset is good as is.

libraries/opensk/src/api/persist.rs
kaczmarczyck: Documentation improved, more explicit on trait level.

libraries/opensk/src/api/persist.rs
kaczmarczyck: Done

libraries/opensk/src/api/persist.rs
kaczmarczyck: Done

libraries/opensk/src/api/persist.rs
```diff
@@ -0,0 +1,550 @@
...
+pub type PersistCredentialIter<'a> = Box<dyn Iterator<Item = CtapResult<(usize, Vec<u8>)>> + 'a>;
+
+/// Stores data that persists across reboots.
+///
+/// This trait might get appended to with new versions of CTAP.
+///
+/// The default implementations using the key-value store have assumptions on the ranges for key
+/// and value, if you decide to use them:
+/// - Keys must be valid at least in [0, 4095[.
+/// - Values must be byte arrays of size at least 1023.
```
ia0:
```suggestion
/// - Keys within 0 and 4095 are supported.
/// - Values of at most 1023 bytes are supported.
```

ia0 (dismissed):
LGTM
What's the testing strategy?

kaczmarczyck:
I flashed the PR on hardware, then went through the following tests, mostly storage related:

- webauthn.io: register, then login.
- Go to Chrome settings, delete some credentials, change the PIN, reset the security key.
- A script that writes and then reads a LargeBlob.
- A script that toggles alwaysUv.
- Our configure script to write batch attestation key material.
- webauthn.io again, but using attestation this time.

Then deploy `erase_storage`, deploy OpenSK again, and another round of the above. Before, it upgraded from the storage as it was right before this PR.

ia0 (approved):

PR #682, opened 2024-04-05
LargeBlob trait supports buffering choice

kaczmarczyck:
Also makes LargeBlob a proper stateful command that loses state when interleaved.

To make sure the API does indeed support buffering in both memory and flash, see the following diff for reference:

```
diff --git a/libraries/opensk/src/api/persist.rs b/libraries/opensk/src/api/persist.rs
index 6caaec9..b9e3543 100644
--- a/libraries/opensk/src/api/persist.rs
+++ b/libraries/opensk/src/api/persist.rs
@@ -236,7 +236,11 @@ pub trait Persist {
     ///
     /// Returns a buffer that is returned to other API calls for potential usage.
     fn init_large_blob(&mut self, expected_length: usize) -> CtapResult<Vec<u8>> {
-        Ok(Vec::with_capacity(expected_length))
+        debug_assert!(expected_length <= keys::LARGE_BLOB_SHARDS.len() * VALUE_LENGTH);
+        for key in keys::LARGE_BLOB_SHARDS {
+            self.remove(key)?;
+        }
+        Ok(Vec::new())
     }
 
     /// Writes a large blob chunk to the buffer.
@@ -246,13 +250,22 @@ pub trait Persist {
         &mut self,
         offset: usize,
         chunk: &mut Vec<u8>,
-        buffer: &mut Vec<u8>,
+        _buffer: &mut Vec<u8>,
     ) -> CtapResult<()> {
-        if buffer.len() != offset {
-            // This should be caught on CTAP level.
-            return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+        for (i, key) in keys::LARGE_BLOB_SHARDS.enumerate() {
+            let start = cmp::max(i * VALUE_LENGTH, offset);
+            let end = cmp::min((i + 1) * VALUE_LENGTH, offset.saturating_add(chunk.len()));
+            if start >= end {
+                continue;
+            }
+            let mut stored = self.find(key)?.unwrap_or_default();
+            let shard_offset = start - i * VALUE_LENGTH;
+            if stored.len() != shard_offset {
+                return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
+            }
+            stored.extend_from_slice(&chunk[start - offset..][..end - start]);
+            self.insert(key, &stored)?;
         }
-        buffer.append(chunk);
         Ok(())
     }
 
@@ -266,14 +279,8 @@ pub trait Persist {
         &self,
         mut offset: usize,
         byte_count: usize,
-        buffer: Option<&Vec<u8>>,
+        _buffer: Option<&Vec<u8>>,
     ) -> CtapResult<Option<Vec<u8>>> {
-        if let Some(buffer) = buffer {
-            let start = cmp::min(offset, buffer.len());
-            let end = offset.saturating_add(byte_count);
-            let end = cmp::min(end, buffer.len());
-            return Ok(Some(buffer[start..end].to_vec()));
-        }
         let mut result = Vec::with_capacity(byte_count);
         for key in keys::LARGE_BLOB_SHARDS {
             if offset >= VALUE_LENGTH {
@@ -297,19 +304,8 @@ pub trait Persist {
     }
 
     /// Sets a byte vector as the serialized large blobs array.
-    fn commit_large_blob_array(&mut self, buffer: &Vec<u8>) -> CtapResult<()> {
-        debug_assert!(buffer.len() <= keys::LARGE_BLOB_SHARDS.len() * VALUE_LENGTH);
-        let mut offset = 0;
-        for key in keys::LARGE_BLOB_SHARDS {
-            let cur_len = cmp::min(buffer.len().saturating_sub(offset), VALUE_LENGTH);
-            let slice = &buffer[offset..][..cur_len];
-            if slice.is_empty() {
-                self.remove(key)?;
-            } else {
-                self.insert(key, slice)?;
-            }
-            offset += cur_len;
-        }
+    #[allow(clippy::ptr_arg)]
+    fn commit_large_blob_array(&mut self, _buffer: &Vec<u8>) -> CtapResult<()> {
         Ok(())
     }
```

### DIFF
--- a/libraries/opensk/src/api/persist.rs
+++ b/libraries/opensk/src/api/persist.rs
@@ -18,6 +18,7 @@ use crate::api::crypto::EC_FIELD_SIZE;
 use crate::ctap::secret::Secret;
 use crate::ctap::status_code::{Ctap2StatusCode, CtapResult};
 use crate::ctap::PIN_AUTH_LENGTH;
+use alloc::borrow::Cow;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::cmp;
@@ -27,6 +28,7 @@ use enum_iterator::IntoEnumIterator;
 
 pub type PersistIter<'a> = Box<dyn Iterator<Item = CtapResult<usize>> + 'a>;
 pub type PersistCredentialIter<'a> = Box<dyn Iterator<Item = CtapResult<(usize, Vec<u8>)>> + 'a>;
+pub type LargeBlobBuffer = Vec<u8>;
 
 /// Stores data that persists across reboots.
 ///
@@ -235,7 +237,7 @@ pub trait Persist {
     /// Prepares writing a new large blob.
     ///
     /// Returns a buffer that is returned to other API calls for potential usage.
-    fn init_large_blob(&mut self, expected_length: usize) -> CtapResult<Vec<u8>> {
+    fn init_large_blob(&mut self, expected_length: usize) -> CtapResult<LargeBlobBuffer> {
         Ok(Vec::with_capacity(expected_length))
     }
 
@@ -245,14 +247,14 @@ pub trait Persist {
     fn write_large_blob_chunk(
         &mut self,
         offset: usize,
-        chunk: &mut Vec<u8>,
-        buffer: &mut Vec<u8>,
+        chunk: &[u8],
+        buffer: &mut LargeBlobBuffer,
     ) -> CtapResult<()> {
         if buffer.len() != offset {
             // This should be caught on CTAP level.
             return Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR);
         }
-        buffer.append(chunk);
+        buffer.extend_from_slice(chunk);
         Ok(())
     }
 
@@ -262,17 +264,17 @@ pub trait Persist {
     /// available. This includes cases of offset being beyond the stored array.
     ///
     /// The buffer is passed in when writing is in process.
-    fn get_large_blob(
-        &self,
+    fn get_large_blob<'a>(
+        &'a self,
         mut offset: usize,
         byte_count: usize,
-        buffer: Option<&Vec<u8>>,
-    ) -> CtapResult<Option<Vec<u8>>> {
+        buffer: Option<&'a LargeBlobBuffer>,
+    ) -> CtapResult<Option<Cow<[u8]>>> {
         if let Some(buffer) = buffer {
             let start = cmp::min(offset, buffer.len());
             let end = offset.saturating_add(byte_count);
             let end = cmp::min(end, buffer.len());
-            return Ok(Some(buffer[start..end].to_vec()));
+            return Ok(Some(Cow::from(&buffer[start..end])));
         }
         let mut result = Vec::with_capacity(byte_count);
         for key in keys::LARGE_BLOB_SHARDS {
@@ -288,16 +290,16 @@ pub trait Persist {
             }
             let end = cmp::min(end, value.len());
             if end < offset {
-                return Ok(Some(result));
+                return Ok(Some(Cow::from(result)));
             }
             result.extend(&value[offset..end]);
             offset = offset.saturating_sub(VALUE_LENGTH);
         }
-        Ok(Some(result))
+        Ok(Some(Cow::from(result)))
     }
 
     /// Sets a byte vector as the serialized large blobs array.
-    fn commit_large_blob_array(&mut self, buffer: &Vec<u8>) -> CtapResult<()> {
+    fn commit_large_blob_array(&mut self, buffer: &LargeBlobBuffer) -> CtapResult<()> {
         debug_assert!(buffer.len() <= keys::LARGE_BLOB_SHARDS.len() * VALUE_LENGTH);
         let mut offset = 0;
         for key in keys::LARGE_BLOB_SHARDS {
--- a/libraries/opensk/src/api/persist.rs
+++ b/libraries/opensk/src/api/persist.rs
@@ -265,11 +265,11 @@ pub trait Persist {
     ///
     /// The buffer is passed in when writing is in process.
     fn get_large_blob<'a>(
-        &'a self,
+        &self,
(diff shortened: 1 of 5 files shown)


libraries/opensk/src/api/persist.rs
```diff
@@ -232,27 +232,48 @@ pub trait Persist {
...
+        Ok(Vec::with_capacity(expected_length))
+    }
+
+    /// Writes a large blob chunk to the buffer.
+    ///
+    /// This can be the passed in buffer, or a custom solution.
+    fn write_large_blob_chunk(
+        &mut self,
+        offset: usize,
+        chunk: &mut Vec<u8>,
```
ia0: I don't understand why we pass this as `&mut Vec<u8>` instead of `&[u8]`. We don't reuse the allocation and `u8` is copy, so we don't need ownership. The `.append(` below would be `.extend_from_slice(`.

libraries/opensk/src/api/persist.rs
```diff
@@ -232,27 +232,48 @@ pub trait Persist {
...
-    /// SHA256, to a total length of 17 byte (which is the shortest legitimate
-    /// large blob entry possible).
-    fn get_large_blob_array(
+    /// The buffer is passed in when writing is in process.
+    fn get_large_blob(
         &self,
         mut offset: usize,
         byte_count: usize,
+        buffer: Option<&Vec<u8>>,
     ) -> CtapResult<Option<Vec<u8>>> {
```
ia0: We could take `Option<&[u8]>` now that we hard-coded the `Vec<u8>` type for the state.

And we could also return `Option<Cow<[u8]>>` to avoid a copy when the result is just a slice from the input.

libraries/opensk/src/api/persist.rs
```diff
@@ -276,12 +297,12 @@ pub trait Persist {
     }
 
     /// Sets a byte vector as the serialized large blobs array.
-    fn commit_large_blob_array(&mut self, large_blob_array: &[u8]) -> CtapResult<()> {
-        debug_assert!(large_blob_array.len() <= keys::LARGE_BLOB_SHARDS.len() * VALUE_LENGTH);
+    fn commit_large_blob_array(&mut self, buffer: &Vec<u8>) -> CtapResult<()> {
```
ia0: Same here, we can take `&[u8]`.

libraries/opensk/src/ctap/mod.rs
```diff
@@ -548,6 +549,24 @@ impl<E: Env> StatefulPermission<E> {
...
+        self.clear_timer(env);
+        if let Some(StatefulCommand::LargeBlob(_)) = &self.command_type {
+        } else {
+            self.set_command(
+                env,
+                StatefulCommand::LargeBlob(LargeBlobState::default()),
+                channel,
+            );
+        }
+        if let Some(StatefulCommand::LargeBlob(ref mut large_blob_state)) = &mut self.command_type {
```
ia0: I don't think you need the `ref mut` here. My understanding is that it should be automatic.

libraries/opensk/src/api/persist.rs
kaczmarczyck: The idea of `append` instead instead of `extend_from_slice` is to save memory. The former drains the `Vec`, the latter copies. Since this is a potentially memory sensitive operation, I think this optimization of restricting the type could be useful.

ia0: That's what I meant by "we don't reuse the allocation", we don't save memory by using `append` because `u8` is copy and the initial vector won't be deallocated until everything is copied. If the input was `Vec<Box<Foo>>` then yes, `append` would reuse the `Foo` allocations (but still not reuse the `Vec` allocation).

kaczmarczyck: True, there isn't really anything to gain from `append`.

libraries/opensk/src/api/persist.rs
kaczmarczyck: Good point with Cow, let's use it.

libraries/opensk/src/api/persist.rs
kaczmarczyck: `LargeBlobBuffer` type introduced here.

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: Nice, it is indeed unnecessary.

libraries/opensk/src/api/persist.rs
```diff
@@ -262,17 +264,17 @@ pub trait Persist {
     /// available. This includes cases of offset being beyond the stored array.
     ///
     /// The buffer is passed in when writing is in process.
-    fn get_large_blob(
-        &self,
+    fn get_large_blob<'a>(
+        &'a self,
```
ia0: This seems wrong to me. We don't want the result to borrow from self.

```suggestion
        &self,
```

libraries/opensk/src/api/persist.rs
```diff
@@ -262,17 +264,17 @@ pub trait Persist {
...
         mut offset: usize,
         byte_count: usize,
-        buffer: Option<&Vec<u8>>,
-    ) -> CtapResult<Option<Vec<u8>>> {
+        buffer: Option<&'a LargeBlobBuffer>,
+    ) -> CtapResult<Option<Cow<[u8]>>> {
```
ia0: We have to be explicit otherwise it will use the lifetime of self.

```suggestion
    ) -> CtapResult<Option<Cow<'a, [u8]>>> {
```

libraries/opensk/src/api/persist.rs
kaczmarczyck: I blindly followed the compiler suggestion, but I agree it felt wrong. Yours is correct!

libraries/opensk/src/api/persist.rs
kaczmarczyck: See above, done.

ia0 (approved):

PR #683, opened 2024-04-30
Create dependabot.yml

jmichelp:
Enable dependabot on 2.1 branch.

Once approved, I'll copy the file to other branches as well

PR #694, opened 2024-07-23
Update openssl dependency

kaczmarczyck:
Fixes a vulnerable package found by `cargo audit`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -499,9 +499,9 @@ dependencies = [
 
 [[package]]
 name = "openssl"
-version = "0.10.64"
+version = "0.10.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


ia0 (approved):
Why didn't we get a dependabot PR for this? We have a dependabot alert. Is it because of those lines?

https://github.com/google/OpenSK/blob/ef05c7e3e4bc7f1046c1efb4dfac94a543f19980/.github/dependabot.yml#L22-L27

This dependabot configuration applies to both version updates and security updates. Not sure if there's a way to fix this, other than removing those lines.

Issue introduced by #683.

kaczmarczyck:
We got one on 2.1, I think it's on the default branch only.

ia0:
> We got one on 2.1, I think it's on the default branch only.

I'm still failing to find that PR :D but it would make sense that it's only the default branch. Can you give me a link to that dependabot PR for openssl on the 2.1 branch?

PR #695, opened 2024-07-23
Removes the dependabot version update config

kaczmarczyck:
The goal of this commit is to enable security updates. Before, security updates were prevented because they followed the config.
We didn't benefit from version updates so far because of this same file, and we don't want them. While 2.1 is our main branch, version updates make even less sense, since we want to keep it stable.


ia0 (approved):
The dependabot configuration had at least the following issues:
- Using `directory: "/"` only enables version updates for the top-level crate (i.e. not for any of the libraries).
- The configuration applies to both version updates and security updates, so indirect dependencies and patch updates would not create a dependabot PR for both version and security updates.

PR #699, opened 2024-08-30
Refactors HID connections to support busy response

kaczmarczyck:
The main intended difference is that OpenSK now responds with ERR_CHANNEL_BUSY when it receives a packet while waiting for touch in in a CTAP2 command.

To support this more elegantly, we changed `HidConnection` to have a `send` and `recv` instead of `send_and_maybe_recv` for a cleaner API. This also resolves an older TODO to respond to incoming channels on the other USB endpoint.

It doesn't process the incoming traffic correctly, but unblock the host at least, and tells it to come back later. We still only allow one active channel at the same time.

We now don't need the `TRANSMIT_AND_RECEIVE` syscall anymore. I didn't remove it from the Tock pack for simplicity, but cleaned up libtock-drivers.

The Env API changes from having one connection per endpoint, to having one send function that takes an endpoint, and a receive function that receives on all endpoints at the same time.
The `HidConnection` already received on all endpoints before, which was inconsistent with the existance of, e.g., `VendorHidConnection` being able to receive on the main endpoint.

Since we now have a cleaner send and receive API, we use this in `src/main.rs` and simplify it a bit, while making it more consistent with other calls to the HID API.

I found an additional inaccuracy in our implementation while refactoring: We want to send keepalives every 100 ms. To do so, we first wait for a button callback for 100 ms. Then we send the keepalive and check if a cancel packet appeared. What should have happened instead is that we listen for HID packets and button presses at the same time during these 100 ms. If nothing happens that stops the loop, we send the keepalive.

The old implementation would, in some implementations, wait 200 ms for each keepalive: First 100 for touch, then 100 for `send_and_maybe_receive`.

The new implementation can loop faster in case there is a lot of unrelated HID traffic. To make the touch timeout last 30 seconds, we introduce an extra timer and loop until time is up. This might still make us blink the LEDs too fast, but that is already the case in the main loop, and generally would need a bigger refactoring.
The only simple workaround to make the LEDs slightly more precise is to wait for touch and packets until it is time to flip the LEDs, and if we return too early within some thresholds, wait again. If we care enough, we can fix that in a later PR.

Not sure how to make the test work that I removed in ctap/mod.rs. I'd need to advance the time while the loop is running. Setting a user presence function that advances time seems hard with the Rust borrowing rules.

### DIFF
--- a/libraries/opensk/src/api/connection.rs
+++ b/libraries/opensk/src/api/connection.rs
@@ -46,7 +46,8 @@ pub struct SendOrRecvError;
 pub type SendOrRecvResult = Result<SendOrRecvStatus, SendOrRecvError>;
 
 pub trait HidConnection {
-    fn send_and_maybe_recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> SendOrRecvResult;
+    fn send(&mut self, buf: &[u8; 64], endpoint: UsbEndpoint) -> SendOrRecvResult;
+    fn recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> SendOrRecvResult;
 }
 
 #[cfg(test)]
--- a/libraries/opensk/src/api/connection.rs
+++ b/libraries/opensk/src/api/connection.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::ctap::status_code::{Ctap2StatusCode, CtapResult};
 use core::convert::TryFrom;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -22,32 +23,26 @@ pub enum UsbEndpoint {
 }
 
 impl TryFrom<usize> for UsbEndpoint {
-    type Error = SendOrRecvError;
+    type Error = Ctap2StatusCode;
 
-    fn try_from(endpoint_num: usize) -> Result<Self, SendOrRecvError> {
+    fn try_from(endpoint_num: usize) -> CtapResult<Self> {
         match endpoint_num {
             1 => Ok(UsbEndpoint::MainHid),
             #[cfg(feature = "vendor_hid")]
             2 => Ok(UsbEndpoint::VendorHid),
-            _ => Err(SendOrRecvError),
+            _ => Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_HARDWARE_FAILURE),
         }
     }
 }
 
-pub enum SendOrRecvStatus {
+pub enum RecvStatus {
     Timeout,
-    Sent,
     Received(UsbEndpoint),
 }
 
-#[derive(Debug, PartialEq, Eq)]
-pub struct SendOrRecvError;
-
-pub type SendOrRecvResult = Result<SendOrRecvStatus, SendOrRecvError>;
-
 pub trait HidConnection {
-    fn send(&mut self, buf: &[u8; 64], endpoint: UsbEndpoint) -> SendOrRecvResult;
-    fn recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> SendOrRecvResult;
+    fn send(&mut self, buf: &[u8; 64], endpoint: UsbEndpoint) -> CtapResult<()>;
+    fn recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> CtapResult<RecvStatus>;
 }
 
 #[cfg(test)]
@@ -59,6 +54,9 @@ mod test {
         assert_eq!(UsbEndpoint::try_from(1), Ok(UsbEndpoint::MainHid));
         #[cfg(feature = "vendor_hid")]
         assert_eq!(UsbEndpoint::try_from(2), Ok(UsbEndpoint::VendorHid));
-        assert_eq!(UsbEndpoint::try_from(3), Err(SendOrRecvError));
+        assert_eq!(
+            UsbEndpoint::try_from(3),
+            Err(Ctap2StatusCode::CTAP2_ERR_VENDOR_HARDWARE_FAILURE)
+        );
     }
 }
--- a/libraries/opensk/src/api/user_presence.rs
+++ b/libraries/opensk/src/api/user_presence.rs
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::api::connection::UsbEndpoint;
+use crate::api::connection::RecvStatus;
+use crate::ctap::status_code::CtapResult;
 
 #[derive(Debug)]
 pub enum UserPresenceError {
@@ -22,14 +23,10 @@ pub enum UserPresenceError {
     Canceled,
     /// User presence check timed out.
     Timeout,
-    /// Unexpected (e.g., hardware) failures
-    Fail,
 }
+pub type UserPresenceResult = Result<(), UserPresenceError>;
 
-pub type UserPresenceResult = (
-    Result<(), UserPresenceError>,
-    Option<([u8; 64], UsbEndpoint)>,
-);
+pub type UserPresenceWaitResult = CtapResult<(UserPresenceResult, RecvStatus)>;
 
 pub trait UserPresence {
     /// Initializes for a user presence check.
@@ -40,8 +37,13 @@ pub trait UserPresence {
     /// Waits until user presence is confirmed, rejected, or the given timeout expires.
     ///
     /// Must be called between calls to [`Self::check_init`] and [`Self::check_complete`].
-    /// Additionally returns a packet if one was received during the wait.
-    fn wait_with_timeout(&mut self, timeout_ms: usize) -> UserPresenceResult;
+    /// The function may write to the packet buffer, if it receives one during the wait.
+    /// If it does, the returned option contains the endpoint it received the data from.
+    fn wait_with_timeout(
+        &mut self,
+        packet: &mut [u8; 64],
+        timeout_ms: usize,
+    ) -> UserPresenceWaitResult;
 
     /// Finalizes a user presence check.
     ///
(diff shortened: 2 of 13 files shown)


libraries/opensk/src/api/connection.rs
```diff
@@ -46,7 +46,8 @@ pub struct SendOrRecvError;
 pub type SendOrRecvResult = Result<SendOrRecvStatus, SendOrRecvError>;
 
 pub trait HidConnection {
-    fn send_and_maybe_recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> SendOrRecvResult;
+    fn send(&mut self, buf: &[u8; 64], endpoint: UsbEndpoint) -> SendOrRecvResult;
+    fn recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> SendOrRecvResult;
```
ia0: Shouldn't we have different responses now?
- Send returns Sent, Timeout?, or Error.
- Receive returns Received(UsbEndpoint), Timeout, or Error.

libraries/opensk/src/api/user_presence.rs
```diff
@@ -24,7 +26,10 @@ pub enum UserPresenceError {
     Fail,
 }
 
-pub type UserPresenceResult = Result<(), UserPresenceError>;
+pub type UserPresenceResult = (
+    Result<(), UserPresenceError>,
+    Option<([u8; 64], UsbEndpoint)>,
+);
```
ia0: We should use a struct (possibly a second one for the option content). Also I'm a bit surprised it's not `Result<Option<(...)>, Error>`.

libraries/opensk/src/ctap/hid/mod.rs
```diff
@@ -440,6 +440,11 @@ impl<E: Env> CtapHid<E> {
         })
     }
 
+    /// Generates a the HID response packets for a busy error.
```
ia0: typo
```suggestion
    /// Generates the HID response packets for a busy error.
```

libraries/opensk/src/ctap/mod.rs
```diff
@@ -255,120 +257,80 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
...
+        if let Some((packet, rx_endpoint)) = data {
+            let (received_cid, processed_packet) = CtapHid::<E>::process_single_packet(&packet);
+            if rx_endpoint != endpoint || received_cid != cid {
+                debug_ctap!(
+                    env,
+                    "Received an unrelated packet on endpoint {:?} while sending a KEEPALIVE packet",
+                     rx_endpoint,
+                );
+                let busy_error = CtapHid::<E>::busy_error(received_cid);
+                send_packets(env, rx_endpoint, busy_error);
```
ia0: We don't continue here, which means we will still process the packet below. Is that expected?

ia0 (commented):
I didn't see the test for "advance the time", but since tests use std, we should be able to have the time under `Arc<Mutex>` and clone it to the user presence fake.

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: Correct and intended. We want to ignore incoming traffic as before, but send a quick response in addition to what we already do.

libraries/opensk/src/api/user_presence.rs
kaczmarczyck: Both UserPresenceResult and SendOrRecvResult need to be refactored still, I just reused existing types for now. One more consideration is that usually, when we receive, we pass in a buffer to be filled with the packet, like in:
`fn recv(&mut self, buf: &mut [u8; 64], timeout_ms: usize) -> SendOrRecvResult;`

We can do that, and make the type:
`pub type UserPresenceResult = Result<Option<UsbEndpoint>, UserPresenceError>;`

libraries/opensk/src/api/connection.rs
kaczmarczyck: This is currently just a convenience to not have 2 types, and the Send status is unreachable in `recv` and vice versa.
In our implementation, `send` can actually timeout, but that is an error, not a valid status like in the blocking `recv`. So `send` should return a:
`Result<(), SendOrRecvError>` (Do we want to specify the error more? I don't think any caller cares.)
and `recv` can just remove the Send status.

libraries/opensk/src/ctap/mod.rs
```diff
@@ -3396,22 +3360,6 @@ mod test {
...
-        fn user_presence_timeout() -> UserPresenceResult {
-            Err(UserPresenceError::Timeout)
-        }
-
-        let mut env = TestEnv::default();
-        env.user_presence().set(user_presence_timeout);
-        let response = check_user_presence(&mut env, DUMMY_CHANNEL);
-        assert!(matches!(
-            response,
-            Err(Ctap2StatusCode::CTAP2_ERR_USER_ACTION_TIMEOUT)
```
kaczmarczyck: This is the removed test.

libraries/opensk/src/ctap/mod.rs
ia0: But this is not what we do. If we get a CANCEL on another channel/endpoint, we will process it and cancel the current user presence check.

libraries/opensk/src/api/user_presence.rs
ia0: I'm not sure I follow the logic of this function, which is why I think using a struct with fields would help. But I agree that this needs to be refactored, so probably not worth trying to improve it before the refactor.

libraries/opensk/src/ctap/mod.rs
ia0: We need to change `TestClock` to be thread-safe with interior mutability (so with `now_ms: Arc<Mutex<usize>>` for example). And then we can control the `Arc<Mutex<usize>>` from the test function.

libraries/opensk/src/api/connection.rs
ia0: It's not really a convenience to not have 2 types if every usage needs to have a panic branch when matching the wrong enum variant. And yes, having specific errors is probably useless. We should either use the CTAP errors or use a dummy unit type.

kaczmarczyck: Convenience in the sense of keeping this diff smaller for an initial review. But the new error types touched most of the files again, so turned out not too reviewer friendly. I hope its more readable now!

libraries/opensk/src/api/user_presence.rs
kaczmarczyck: Refactored to be a `Result` on the outside. The logic should now be more straightforward: Error if any unexpected failure happens. If `Ok`, then one entry for the touch result, one entry for the USB result.

libraries/opensk/src/ctap/hid/mod.rs
kaczmarczyck: Done

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: Oh yes, I missed that. Function is split up to allow more `?` usage and simplify the flow.

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: Thanks for the pointer, fixed.

kaczmarczyck:
This fix surfaced a bug in the USB stack:

```
panicked at 'assertion failed: self.pending_out.take()', capsules/src/usb/usbc_ctap_hid.rs:366:17
```

Keeping this PR open until I looked into the problem.

ia0:
> Keeping this PR open until I looked into the problem.

Sounds good. Click the `Re-request review` button once it's ready for review.

libraries/opensk/src/env/test/mod.rs
```diff
@@ -50,12 +51,17 @@ pub struct TestTimer {
...
 }
 
 impl TestClock {
     pub fn advance(&mut self, milliseconds: usize) {
-        self.now_ms += milliseconds;
+        let mut locked_now_ms = self.now_ms.lock().unwrap();
+        *locked_now_ms += milliseconds;
+    }
+
+    pub fn access(&self) -> Arc<Mutex<usize>> {
```
ia0: Why not derive `Clone` on `TestClock` and use `advance` in the test?

libraries/opensk/src/env/test/mod.rs
```diff
@@ -64,18 +70,18 @@ impl Clock for TestClock {
 
     fn make_timer(&mut self, milliseconds: usize) -> Self::Timer {
         TestTimer {
-            end_ms: self.now_ms + milliseconds,
+            end_ms: *self.now_ms.lock().unwrap() + milliseconds,
```
ia0: Instead of reading the private field, we could add a method `fn TestClock::now(&self) -> usize`.

ia0 (dismissed):

libraries/opensk/src/env/test/mod.rs
kaczmarczyck: Done

libraries/opensk/src/env/test/mod.rs
kaczmarczyck: `now` added

libraries/opensk/src/env/test/mod.rs
```diff
@@ -48,20 +48,20 @@ pub struct TestTimer {
...
-        *locked_now_ms += milliseconds;
+    pub fn now(&self) -> usize {
+        *self.now_ms.lock().unwrap()
     }
 
-    pub fn access(&self) -> Arc<Mutex<usize>> {
-        self.now_ms.clone()
+    pub fn advance(&self, milliseconds: usize) {
+        let mut locked_now_ms = self.now_ms.lock().unwrap();
+        *locked_now_ms += milliseconds;
```
ia0, line 64: fyi, this could have just been `*self.now_ms.lock().unwrap() += milliseconds`

ia0 (approved):

Issue #714, opened 2025-02-06
Unable to debug OpenSK on nRF52840DK


gnmarco:
I tried many times to enable debugging on OpenSK deployed on the nRF52840DK board without success. However, I managed to get the board up and running without debugging.

# Actual behavior

1. Execute `JLinkExe -device nrf52 -if swd -speed 1000 -autoconnect 1` in a terminal
2. Execute `JLinkRTTClient` in another
3. Connect board JTAG port to the computer
4. Run `./deploy.py --board=nrf52840dk_opensk --opensk --debug` 
5. The configuration of the device fails (as expected), but I can see debug output: 
```
Initialization complete. Entering main loop
NRF52 HW INFO: Variant: Unspecified, Part: N52840, Package: QI, Ram: K256, Flash: K1024
Hello world from OpenSK!
Entering main ctap loop
tock$ ...................................................
```
6. Connect micro USB cable to the USB port of the device (by keeping the JTAG connected)
7. Press "RESET" (debug output stops), this is needed in order for the board to be recognized by the computer.
8. Run 
```
./tools/configure.py \
    --certificate=crypto_data/opensk_cert.pem \
    --private-key=crypto_data/opensk.key
```
9. The command hangs and the board has LED1 blinking with non-regular intervals.

# Expected behavior
In point 7. the output should not stop and point 9. should not happen.

# Additional information

Even without the debugger enabled (i.e. without executing steps 1 and 2, it still does not work and I get the same results).
Am I doing something wrong? Is this a known bug? I tried having a look at #561 but I cannot get it to work.

# UPDATE

I managed to configure the `develop` branch of the firmware, but I am still not able to debug using the aforementioned steps. If I change steps (1) and (2) and put them in place of steps (3) and (4) the board complains that it is locked and must be unprotected (by wiping the internal memory, of course, rendering `deploy.py` useless).

Thanks.

kaczmarczyck:
I tried to reproduce, and what I agree that the blinking patterns in debug mode are irregular. Could be because the timings are off when OpenSK is slowed down by printing debug output.

In my case, I don't have to press "RESET". But if I do, the debug output also doesn't stop, as long as I keep both cables plugged it. Can you explain why rebooting the chip helps the host OS recognize the board?

Step (8) works without problems for me. I do have to touch at some point, when the terminal says

> Please touch the device to confirm...

I'm not sure why your device behaves differently than mine. If you don't mind losing all previously stored data, you can also try if resetting the storage helps:

```
[...]

PR #715, opened 2025-02-12
Warns users when deploying with --debug

kaczmarczyck:
Problems with the `debug` feature were reported in #714 . Since all new development is happening in the `develop` branch, I added an error that redirects people to the other branch.


ia0 (approved):

kaczmarczyck:
CI fails because of `cargo audit`, but that is already fixed in `develop`, so I'll ignore it.

PR #721, opened 2025-03-14
Fixes pylint for later versions

kaczmarczyck:
While not caught by the pylint version in the GitHub workflows yet, this fails locally:

```
py_virtual_env/bin/pylint --score=n `git ls-files --deduplicate --exclude-standard --full-name '*.py'`
```


ia0 (approved):

kaczmarczyck:
Fails on cargo audit, which is to be fixed next.

PR #718, opened 2025-02-28
Add Support for User Verification (UV) and BioEnrollment

bradjc-veridian:
This PR includes our changes to OpenSK to support UV and bioenrollment. Because there is no fingerprint sensor on the nRF52840dk or in Tock, the environment implementation is just a stub.

We are happy to test this PR against our internal hardware platform as updates are made to this PR after feedback.

-   [ ] Local tests pass (running `run_desktop_tests.sh`)
-   [ ] Tested against boards
    -   [ ] Nordic nRF52840 DK
    -   [ ] Nordic nRF52840 Dongle (JTAG programmed)
    -   [ ] Nordic nRF52840 Dongle (DFU programmed)
    -   [ ] Makerdiary nRF52840 MDK USB Dongle
-   [ ] Appropriate changes to README are included in PR

PR #734, opened 2025-06-16
Uv bio enrollment, continued

kaczmarczyck:
Rebases and continues #718 .

The second commits adds the following missing features:

- PIN authentication of BioEnrollment subcommands that require it.
- A compile flag for all new features.
- Cleanup and generalization of the fingerprint API. (Example: `setloglevel` can be part of the `TockEnv` implementation, but shouldn't be included in the API.)
- A more useful fake implementation of the fingerprint API.
- Customization of Veridian's hardcoded numbers.
- Rollback of customizations that are specific to Veridian.
- Whether or not state is held is now left to the implementer of the fingerprint API. BioEnrollment is explicitly not a stateful command according to the specification.
- Don't ignore errors from hardware interaction, but forward them.
- Correct handling of UV retries: Reset everywhere necesary, and don't decrease the retry counter when asking for UV, but only when the user actually presented their finger.
- Removed magic numbers.
- Style matches existing code.
- CBOR parsing is more readable with new structs.
- Unit tests.
- Fixed all comments from the former PR.

Still missing:

- A more complete fake for testing.
- Tests for all subcommands.
- A working TockEnv implementation.

As is, the implementation is not usable on hardware.

Moves a lot of the code from the first commit around, so looking at these commits in isolation is harder than reviewing their sum.

### DIFF
--- a/libraries/opensk/src/api/customization.rs
+++ b/libraries/opensk/src/api/customization.rs
@@ -145,7 +145,7 @@ pub trait Customization {
     /// this value.
     fn max_msg_size(&self) -> usize;
 
-    /// Sets the number of consecutive failed PINs before blocking interaction.
+    /// The number of consecutive failed PINs before blocking interaction.
     ///
     /// # Invariant
     ///
@@ -154,6 +154,42 @@ pub trait Customization {
     ///
     /// The fail retry counter is reset after entering the correct PIN.
     fn max_pin_retries(&self) -> u8;
+
+    /// Maximum number of UV retries performed before returning an error.
+    ///
+    /// Corresponds to maxUvAttemptsForInternalRetries in CTAP 2.1 onwards.
+    /// When internal retries are allowed, the authenticator will retry UV
+    /// before communicating its result through CTAP.
+    ///
+    /// # Invariant
+    ///
+    /// - Must be in the range [1, `max_uv_retries`].
+    /// - If `preferred_platform_uv_attempts` is not 1, must be in [1, 5].
+    #[cfg(feature = "fingerprint")]
+    fn max_uv_attempts_for_internal_retries(&self) -> u8;
+
+    /// The number of consecutive failed UV attempts before blocking built-in UV.
+    ///
+    /// Corresponds to maxUvRetries in CTAP 2.1 onwards.
+    ///
+    /// # Invariant
+    ///
+    /// - Must be in the range [1, 25].
+    #[cfg(feature = "fingerprint")]
+    fn max_uv_retries(&self) -> u8;
+
+    /// Preferred number of UV attempts before falling back to PIN.
+    ///
+    /// Corresponds to preferredPlatformUvAttempts in CTAP 2.1 onwards.
+    /// States the preference how often the platform should invoke
+    /// `getPinUvAuthTokenUsingUvWithPermissions` before falling back to
+    /// `getPinUvAuthTokenUsingPinWithPermissions` or displaying an error.
+    ///
+    /// # Invariant
+    ///
+    /// - Must be greater than 0.
+    #[cfg(feature = "fingerprint")]
+    fn preferred_platform_uv_attempts(&self) -> usize;
 
     /// Enables or disables basic attestation for FIDO2.
     ///
@@ -253,30 +289,11 @@ pub trait Customization {
     /// for 10 years.
     fn max_supported_resident_keys(&self) -> usize;
 
-    /// This specifies the preferred number of invocations of the
-    /// `getPinUvAuthTokenUsingUvWithPermissions` subCommand the platform may
-    /// attempt before falling back to the
-    /// `getPinUvAuthTokenUsingPinWithPermissions` subCommand or displaying an
-    /// error.
+    /// Maximum size of a friendly name for a UV template.
     ///
-    /// MUST be greater than zero. If the value is 1 then all uvRetries are
-    /// internal and the platform MUST only invoke the
-    /// getPinUvAuthTokenUsingUvWithPermissions subCommand a single time. If the
-    /// value is > 1 the authenticator MUST only decrement uvRetries by 1 for
-    /// each iteration.
-    fn preferred_platform_uv_attempts(&self) -> usize;
-
-    /// Sets the number of consecutive failed User Verification attempts before
-    /// blocking built-in UV.
-    ///
-    /// # Invariant
-    ///
-    /// - CTAP2.1: Maximum PIN retries must be 25 at most.
-    fn max_uv_retries(&self) -> u8;
-
-    /// The maximum number of times the authenticator will retry internally when
-    /// internalRetry is true as part of the performBuiltInUv() algorithm.
-    fn max_uv_attempts_for_internal_retries(&self) -> u8;
+    /// This value limits storage requirements for fingerprints.
+    #[cfg(feature = "fingerprint")]
+    fn max_template_friendly_name(&self) -> usize;
 }
 
 #[derive(Clone)]
@@ -291,6 +308,12 @@ pub struct CustomizationImpl {
     pub enterprise_rp_id_list: &'static [&'static str],
     pub max_msg_size: usize,
     pub max_pin_retries: u8,
+    #[cfg(feature = "fingerprint")]
+    pub max_uv_attempts_for_internal_retries: u8,
+    #[cfg(feature = "fingerprint")]
+    pub max_uv_retries: u8,
+    #[cfg(feature = "fingerprint")]
+    pub preferred_platform_uv_attempts: usize,
     pub use_batch_attestation: bool,
     pub use_signature_counter: bool,
     pub max_cred_blob_length: usize,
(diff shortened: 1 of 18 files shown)


libraries/opensk/src/api/persist.rs
```diff
@@ -386,6 +418,102 @@ pub trait Persist {
...
+            if self.find(key)?.is_some() {
+                continue;
+            }
+            let template_info = TemplateInfo {
+                template_id: template_id.to_vec(),
+                template_friendly_name: None,
+            };
+            let mut data = Vec::new();
+            cbor_write(cbor::Value::from(template_info), &mut data)?;
+            self.insert(key, &mut data)?;
```
ia0: The `mut` seems weird.
```suggestion
            self.insert(key, &data)?;
```

libraries/opensk/src/api/persist.rs
```diff
@@ -386,6 +418,102 @@ pub trait Persist {
...
+    #[cfg(feature = "fingerprint")]
+    fn remove_template_id(&mut self, template_id: &[u8]) -> CtapResult<()> {
+        for key in keys::FRIENDLY_NAMES {
+            if let Some(data) = self.find(key)? {
+                let cbor_value = cbor_read(&data)
+                    .map_err(|_| Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
+                let template_info = TemplateInfo::try_from(cbor_value)?;
+                if &template_info.template_id == template_id {
+                    self.remove(key)?;
+                    return Ok(());
```
ia0, line 483: This only removes the first such template, but there could be more. In `store_template_id` we don't seem to check for duplicates.

libraries/opensk/src/api/persist.rs
```diff
@@ -386,6 +418,102 @@ pub trait Persist {
...
+    #[cfg(test)]
+    #[cfg(feature = "fingerprint")]
+    fn get_friendly_name(&self, template_id: &[u8]) -> CtapResult<Option<String>> {
+        for key in keys::FRIENDLY_NAMES {
+            if let Some(data) = self.find(key)? {
+                let cbor_value = cbor_read(&data)
+                    .map_err(|_| Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
+                let template_info = TemplateInfo::try_from(cbor_value)?;
+                if &template_info.template_id == template_id {
+                    return Ok(template_info.template_friendly_name);
```
ia0, line 502: Same here which suggests there should be an invariant that there's at most one template per template id.

libraries/opensk/src/api/persist.rs
```diff
@@ -386,6 +418,102 @@ pub trait Persist {
...
+        for key in keys::FRIENDLY_NAMES {
+            if let Some(bytes) = self.find(key)? {
+                let cbor_value = cbor_read(&bytes)
+                    .map_err(|_| Ctap2StatusCode::CTAP2_ERR_VENDOR_INTERNAL_ERROR)?;
+                let mut template_info = TemplateInfo::try_from(cbor_value)?;
+                if &template_info.template_id == template_id {
+                    template_info.template_friendly_name = Some(String::from(friendly_name));
+                    let mut data = Vec::new();
+                    cbor_write(cbor::Value::from(template_info), &mut data)?;
+                    self.insert(key, &mut data)?;
```
ia0:
```suggestion
                    self.insert(key, &data)?;
```

libraries/opensk/src/api/persist.rs
```diff
@@ -386,6 +418,102 @@ pub trait Persist {
...
+            }
+        }
+        Ok(None)
+    }
+
+    /// Store a BioEnrollment friendly name for a given template ID.
+    ///
+    /// Returns `Err(CTAP2_ERR_INVALID_OPTION)` if the template was not found.
+    #[cfg(feature = "fingerprint")]
+    fn store_friendly_name(&mut self, template_id: &[u8], friendly_name: &str) -> CtapResult<()> {
```
ia0: We could take a `String` as argument since we unconditionally convert it to a `String` on successful operation.

libraries/opensk/src/api/persist.rs
```diff
@@ -386,6 +418,102 @@ pub trait Persist {
...
+            }
+        }
+        Ok(id_list)
+    }
+
+    /// Stores a new template ID.
+    ///
+    /// Returns `Err(CTAP2_ERR_FP_DATABASE_FULL)` if the storage is full.
+    #[cfg(feature = "fingerprint")]
+    fn store_template_id(&mut self, template_id: &[u8]) -> CtapResult<()> {
```
ia0: We could take `Vec<u8>` directly since we unconditionally convert it on success.

libraries/opensk/src/api/persist/keys.rs
```diff
@@ -103,6 +103,16 @@ make_partition! {
     /// The stored large blob can be too big for one key, so it has to be sharded.
     LARGE_BLOB_SHARDS = 2000..2004;
 
+    /// Stored friendly names for enrolled fingerprints.
+    #[cfg(feature = "fingerprint")]
+    FRIENDLY_NAMES = 2031..2036;
```
ia0: Let's not leave an empty key if we don't need to. If we want to reserve it, we can `_RESERVED_FRIENDLY_NAME = 2036` like for credentials.
```suggestion
    FRIENDLY_NAMES = 2031..2037;
```

ia0 (commented):
I only reviewed:
- libraries/opensk/src/api/persist.rs
- libraries/opensk/src/api/persist/key.rs
- libraries/opensk/src/ctap/storage.rs

libraries/opensk/src/api/persist/keys.rs
kaczmarczyck: Makes sense, done.

libraries/opensk/src/api/persist.rs
kaczmarczyck: Done

libraries/opensk/src/api/persist.rs
kaczmarczyck: Done

libraries/opensk/src/api/persist.rs
kaczmarczyck: I was surprised that this wasn't caught by clippy, and indeed, the new feature wasn't covered yet. Fixed all other clippy findings too.

libraries/opensk/src/api/persist.rs
kaczmarczyck: Done, see above wrt clippy

libraries/opensk/src/api/persist.rs
kaczmarczyck: This is supposedly an invariant of the fingerprint module. I made it explicit now, documented it and return an internal error in case of duplicates.
So remove is untouched, and `store_` is adapted.

libraries/opensk/src/api/persist.rs
kaczmarczyck: Won't fix because of above introduced invariant.

run_desktop_tests.sh
```diff
@@ -54,7 +54,7 @@ echo "Running Clippy lints..."
 cargo clippy --lib --tests --bins --benches --features std -- -D warnings
 cargo clippy --lib --tests --bins --benches --features std,"$MOST_FEATURES" -- -D warnings
 (cd libraries/opensk && cargo clippy --features std -- -D warnings)
-(cd libraries/opensk && cargo clippy --features std,config_command,debug_ctap,with_ctap1,vendor_hid,ed25519 -- -D warnings)
+(cd libraries/opensk && cargo clippy --features std,config_command,debug_ctap,with_ctap1,vendor_hid,ed25519,fingerprint -- -D warnings)
```
ia0: What about `std,"$MOST_FEATURES"` like 2 lines above?

ia0 (dismissed):
LGTM for the storage-related part.

run_desktop_tests.sh
kaczmarczyck: That doesn't work since these are features from the base Cargo.toml, and here, clippy is running inside `libraries/opensk`. But I could use `--all-features`, and the only difference would be `fuzz`. I don't remember why I explicitly omitted `fuzz` when I wrote this. But it works either way.

ia0: I see, sounds good.

ia0 (dismissed):

bradjc (approved):
I just reviewed the changes and they all look good to me.

libraries/opensk/src/ctap/fingerprint.rs
```diff
@@ -0,0 +1,407 @@
...
+        let response = process_bio_enrollment(&mut env, &mut client_pin, params.clone());
+        assert_eq!(response, Err(Ctap2StatusCode::CTAP2_ERR_INVALID_OPTION));
+
+        let template_id = create_fingerprint(&mut env);
+        let response = process_bio_enrollment(&mut env, &mut client_pin, params);
+        match response.unwrap() {
+            ResponseData::AuthenticatorBioEnrollment(Some(response)) => {
+                assert!(response.modality.is_none());
+                assert!(response.fingerprint_kind.is_none());
+                assert!(response.max_capture_samples_required_for_enroll.is_none());
```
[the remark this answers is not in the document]
kaczmarczyck: The `internal_retry` variable is used in `libraries/opensk/src/ctap/client_pin.rs:394`.
The marked lines of this comment seem unrelated.

libraries/opensk/src/api/customization.rs
```diff
@@ -446,6 +524,31 @@ pub fn is_valid(customization: &impl Customization) -> bool {
...
+    {
+        // Maximum UV retries must be in range [1, 25].
+        if customization.max_uv_retries() < 1 || customization.max_uv_retries() > 25 {
+            return false;
+        }
+
+        // Maximum internal UV attemps must be in range [1, `max_uv_retries`].
+        if customization.max_uv_attempts_for_internal_retries() < 1
+            || customization.max_uv_attempts_for_internal_retries() > customization.max_uv_retries()
+        {
```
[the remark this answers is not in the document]
kaczmarczyck: That seems to be exactly what the check does, and setting different default values makes the test pass or fail according to expectation.

libraries/opensk/src/api/customization.rs
```diff
@@ -446,6 +524,31 @@
...
+            return false;
+        }
+        if customization.preferred_platform_uv_attempts() != 1
+            && customization.max_uv_attempts_for_internal_retries() > 5
+        {
```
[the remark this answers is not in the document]
kaczmarczyck: Again, seems to be working as intended. I added a comment above the check

In unrelated news, when I changed the default values, I noticed that a different test has a bug for some non-default values, so I split it up and fixed it. The problem was that the UV attempts were not reset between calls.

ia0 (approved):

PR #736, opened 2025-06-24
Patch to support new NRF revision

### LINKED ISSUE #714: Unable to debug OpenSK on nRF52840DK

kaczmarczyck:
This is the patch crafted by @gnmarco (see comment https://github.com/google/OpenSK/issues/714#issuecomment-2651554243), as a PR.
Thanks again for your help!

I only have an older revision dev board, I will merge when we can confirm it works on all hardware.

### DIFF
--- a/patches/tock/08-support-new-nrf52840dk-board.patch
+++ b/patches/tock/08-support-new-nrf52840dk-board.patch
@@ -1,0 +1,336 @@
+diff --git a/boards/nordic/nrf52_components/src/startup.rs b/boards/nordic/nrf52_components/src/startup.rs
+index 7b3ea5a59..87a566550 100644
+--- a/boards/nordic/nrf52_components/src/startup.rs
++++ b/boards/nordic/nrf52_components/src/startup.rs
+@@ -36,6 +36,13 @@ impl<'a> Component for NrfStartupComponent<'a> {
+     type StaticInput = ();
+     type Output = ();
+     unsafe fn finalize(self, _s: Self::StaticInput) -> Self::Output {
++        // Disable APPROTECT in software. This is required as of newer nRF52
++        // hardware revisions. See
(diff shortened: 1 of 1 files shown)


ia0 (approved):

kaczmarczyck:
Tock will be removed. This might be interesting as a code sample for https://github.com/google/wasefire, @ia0

PR #737, opened 2025-07-16
Makes `PrivateKey` take actual private keys

kaczmarczyck:
Before, it would store their seed bytes in a `Secret`. While zeroization helps with some attacks, we didn't leave it to the cryptography implementation to further protect the information.

This introduces a new `Env` type dependency in `PrivateKey`. Instead of forwarding it to both types that store `PrivateKey`, namely `CredentialSource` and `PublicKeyCredentialSource`, we make them hold a wrapped version of that private key. This has several advantages:

- We now iterate over stored credentials and filter them more easily without having to decrypt the private key material.
- Private keys wrapping moved out of high level data types, and closer to the cryptographic implementation.

The latter is useful for removing `to_slice` in `SecretKey`, and replacing it with a more generic `wrap`, so that private key material does not have to leave hardware cryptography.

The second goal this PR prepares for is to move Ed25519 into `Env`'s crypto API.

### DIFF
--- a/libraries/opensk/src/api/private_key.rs
+++ b/libraries/opensk/src/api/private_key.rs
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 use crate::api::crypto::ecdsa::{SecretKey as _, Signature};
+#[cfg(test)]
+use crate::api::crypto::EC_FIELD_SIZE;
+use crate::api::key_store::KeyStore;
 use crate::ctap::crypto_wrapper::{aes256_cbc_decrypt, aes256_cbc_encrypt};
 use crate::ctap::data_formats::{extract_array, extract_byte_string, CoseKey, SignatureAlgorithm};
 use crate::ctap::secret::Secret;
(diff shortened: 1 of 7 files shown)


libraries/opensk/src/api/private_key.rs
```diff
@@ -181,51 +191,45 @@ mod test {
...
     fn test_new_ecdsa_from_bytes() {
         let mut env = TestEnv::default();
-        let private_key = PrivateKey::new(&mut env, SignatureAlgorithm::Es256);
+        let private_key = PrivateKey::<TestEnv>::new(&mut env, SignatureAlgorithm::Es256);
         let key_bytes = private_key.to_bytes();
-        assert_eq!(
-            PrivateKey::new_ecdsa_from_bytes(&key_bytes),
-            Some(private_key)
-        );
+        assert!(PrivateKey::<TestEnv>::new_ecdsa_from_bytes(&key_bytes) == Some(private_key));
```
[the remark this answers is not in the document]
kaczmarczyck: `PrivateKey` does not implement `Debug`. With the `Env` dependency, we would need a custom `Debug` implementation, like we have for `Clone` and `PartialEq`. For simplicity, I instead moved to `==`.

libraries/opensk/src/api/private_key.rs
```diff
@@ -130,48 +142,46 @@ impl PrivateKey {
...
         match self {
-            PrivateKey::Ecdsa(key_bytes) => bytes.copy_from_slice(key_bytes.deref()),
+            PrivateKey::Ecdsa(key) => {
+                let mut array_bytes: Secret<[u8; 32]> = Secret::default();
+                key.to_slice(array_bytes.deref_mut());
+                bytes.copy_from_slice(array_bytes.deref())
+            }
             #[cfg(feature = "ed25519")]
             PrivateKey::Ed25519(ed25519_key) => bytes.copy_from_slice(ed25519_key.seed().deref()),
         }
```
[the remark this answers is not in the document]
kaczmarczyck: This does not compile.
The function is used in 2 places: In `to_cbor` in the same class, and as entropy for `cred_random`. 
This PR is supposed to be non-breaking. In a follow-up, I will touch those again when I implement the aforementioned changes to `SecretKey`. Then, I will retire `Secret` from this function, so I will keep this function as is until then.

libraries/opensk/src/api/private_key.rs
```diff
@@ -21,97 +24,107 @@ use crate::env::{AesKey, EcdsaSk, Env};
...
     /// Helper function that creates a private key of type ECDSA.
-    pub fn new_ecdsa_from_bytes(bytes: &[u8]) -> Option<Self> {
-        if bytes.len() != 32 {
-            return None;
-        }
-        let mut seed: Secret<[u8; 32]> = Secret::default();
-        seed.copy_from_slice(bytes);
-        Some(PrivateKey::Ecdsa(seed))
+    fn new_ecdsa_from_bytes(bytes: &[u8]) -> Option<Self> {
+        let bytes = <&[u8; 32]>::try_from(bytes).ok()?;
```
ia0, line 101: This seems wrong. We should not assume the length is 32.

libraries/opensk/src/api/private_key.rs
```diff
@@ -130,48 +143,46 @@ impl PrivateKey {
     pub fn to_bytes(&self) -> Secret<[u8]> {
         let mut bytes = Secret::new(32);
         match self {
-            PrivateKey::Ecdsa(key_bytes) => bytes.copy_from_slice(key_bytes.deref()),
+            PrivateKey::Ecdsa(key) => {
+                let mut array_bytes: Secret<[u8; 32]> = Secret::default();
```
ia0, line 147: Same here. Or is this test only? I would expect `src/api/crypto/ecdsa.rs` to change, but the file is not modified by this PR.

ia0 (commented):
Am I missing something? I thought the idea was to make `ecdsa::SecretKey::from_slice()` and `to_slice()` take a slice reference instead of an array reference, and have an additional constant or function to know what's the expected length.

libraries/opensk/src/api/private_key.rs
kaczmarczyck: Will happen in the next PR!

ia0 (approved):
Apparently this is not the PR I'm looking for.

libraries/opensk/src/api/private_key.rs
kaczmarczyck: Yes, same here.

PR #740, opened 2025-07-21
Ed25519 in Env

### LINKED ISSUE #731: Ed25519 into Env
The current implementation hasn't been ported to the Crypto API yet.

kaczmarczyck:
Extends the crypto API to include Ed25519, given the corresponding feature is set.

The `ecdsa` trait is renamed to `ec_signing` to indicate that we can implement more than one curve.
Also removes some functions from the EC trait that were used in tests or outside of the library only:

- `PublicKey`: `from_coordinates` (`test` only)
- `PublicKey`: `verify` (`test` only)
- `PublicKey`: `verify_prehash` (`TockEnv` only)
- `Signature`: `from_slice` (`test` only)

TockEnv now calls RustCrypto directly, and the tests are either rewritten or removed.
Removing the `test` only function is seperated into a second commit, so reviewers can choose if the want to review the whole package for its smaller diff, or the commits individually to see the logical steps.

Since we chose the crate for Ed25519 some time ago, I re-evaluated it and compared it to its own version 2 and the RustCrypto-endorsed [ed25519-dalek](https://crates.io/crates/ed25519-dalek).

<details>
<summary>Diff used to make speed tests</summary>
  
```
diff --git a/libraries/opensk/Cargo.toml b/libraries/opensk/Cargo.toml
index 7ad7621..6ab1947 100644
--- a/libraries/opensk/Cargo.toml
+++ b/libraries/opensk/Cargo.toml
@@ -18,7 +18,6 @@ byteorder = { version = "1", default-features = false }
 arrayref = "0.3.6"
-ed25519-compact = { version = "1", default-features = false, optional = true }
 rand_core = "0.6.4"
 rand = { version = "0.8.5", default-features = false, optional = true }
 sha2 = { version = "0.10.6", default-features = false }
@@ -27,6 +26,7 @@ hkdf = { version = "0.12.3", default-features = false }
 aes = { version = "0.8.2", default-features = false }
 cbc = { version = "0.1.2", default-features = false }
 der = { version = "0.7.10", default-features = false }
+ed25519-dalek = { version = "2.2.0", default-features = false, optional = true, features = ["rand_core"] }
 zeroize = { version = "1.5.7", features = ["derive"] }
 
 [dependencies.p256]
@@ -42,7 +42,7 @@ std = ["persistent_store/std", "rand/std_rng", "config_command"]
 with_ctap1 = []
 vendor_hid = []
 fuzz = ["arbitrary", "std"]
-ed25519 = ["ed25519-compact"]
+ed25519 = ["ed25519-dalek"]
 fingerprint = []
 
 [dev-dependencies]
diff --git a/libraries/opensk/src/api/crypto/rust_crypto.rs b/libraries/opensk/src/api/crypto/rust_crypto.rs
index c901ac4..c0182cb 100644
--- a/libraries/opensk/src/api/crypto/rust_crypto.rs
+++ b/libraries/opensk/src/api/crypto/rust_crypto.rs
@@ -26,6 +26,8 @@ use aes::cipher::{
     BlockDecrypt, BlockDecryptMut, BlockEncrypt, BlockEncryptMut, KeyInit, KeyIvInit,
 };
 use alloc::vec::Vec;
+#[cfg(feature = "ed25519")]
+use core::convert::TryFrom;
 #[cfg(test)]
 use core::cell::RefCell;
 #[cfg(feature = "ed25519")]
@@ -40,6 +42,8 @@ use p256::elliptic_curve::sec1::ToEncodedPoint;
 use sha2::Digest;
 #[cfg(test)]
 use std::sync::{Mutex, MutexGuard};
+#[cfg(feature = "ed25519")]
+use ed25519_dalek;
 
 // To be able to support hardware cryptography, we want to make sure we never compute multiple
 // sha256 in parallel. Note that calling `digest` or `digest_mut` is statically correct.
@@ -225,7 +229,7 @@ impl ec_signing::EcSigning for SoftwareEd25519 {
 
 #[cfg(feature = "ed25519")]
 pub struct SoftwareEd25519SecretKey {
-    signing_key: ed25519_compact::SecretKey,
+    signing_key: ed25519_dalek::SigningKey,
 }
 
 #[cfg(feature = "ed25519")]
@@ -234,37 +238,34 @@ impl ec_signing::SecretKey for SoftwareEd25519SecretKey {
     type Signature = SoftwareEd25519Signature;
 
     fn random(rng: &mut impl Rng) -> Self {
-        let mut bytes = [0; 32];
-        rng.fill_bytes(&mut bytes[..]);
-        let seed = ed25519_compact::Seed::from_slice(&bytes).unwrap();
-        let signing_key = ed25519_compact::KeyPair::from_seed(seed).sk;
+        let signing_key = ed25519_dalek::SigningKey::generate(rng);
         SoftwareEd25519SecretKey { signing_key }
     }
 
     fn public_key(&self) -> Self::PublicKey {
-        SoftwareEd25519PublicKey {
-            verifying_key: self.signing_key.public_key(),
-        }
+        let verifying_key = ed25519_dalek::VerifyingKey::from(&self.signing_key);
+        SoftwareEd25519PublicKey { verifying_key }
     }
 
     fn sign(&self, message: &[u8]) -> Self::Signature {
-        let signature = self.signing_key.sign(message, None);
+        let signature = self.signing_key.sign(message);
         SoftwareEd25519Signature { signature }
     }
 
     fn export(&self) -> Vec<u8> {
-        self.signing_key[..].to_vec()
+        self.signing_key.to_keypair_bytes().to_vec()
     }
 
     fn import(bytes: &[u8]) -> Option<Self> {
-        let signing_key = ed25519_compact::SecretKey::from_slice(bytes).ok()?;
+        let bytes = <&[u8; 64]>::try_from(bytes).ok()?;
+        let signing_key = ed25519_dalek::SigningKey::from_keypair_bytes(bytes).ok()?;
         Some(SoftwareEd25519SecretKey { signing_key })
     }
 }
 
 #[cfg(feature = "ed25519")]
 pub struct SoftwareEd25519PublicKey {
-    verifying_key: ed25519_compact::PublicKey,
+    verifying_key: ed25519_dalek::VerifyingKey,
 }
 
 #[cfg(feature = "ed25519")]
@@ -273,8 +274,7 @@ impl ec_signing::PublicKey for SoftwareEd25519PublicKey {
 
     #[cfg(test)]
     fn from_coordinates(x: &[u8; EC_FIELD_SIZE], _y: &[u8; EC_FIELD_SIZE]) -> Option<Self> {
-        let verifying_key = ed25519_compact::PublicKey::from_slice(x).ok()?;
-        // The y coordinate is unused when exporting and importing.
+        let verifying_key = ed25519_dalek::VerifyingKey::from_bytes(x).ok()?;
         Some(SoftwareEd25519PublicKey { verifying_key })
     }
 
@@ -286,15 +286,14 @@ impl ec_signing::PublicKey for SoftwareEd25519PublicKey {
     }
 
     fn to_coordinates(&self, x: &mut [u8; EC_FIELD_SIZE], y: &mut [u8; EC_FIELD_SIZE]) {
-        x.copy_from_slice(&self.verifying_key[..]);
-        // The public key can be reconstructed from the x slice only.
-        y.copy_from_slice(&[0u8; 32]);
+        x.copy_from_slice(self.verifying_key.as_bytes());
+        y.copy_from_slice(&[0u8; EC_FIELD_SIZE]);
     }
 }
 
 #[cfg(feature = "ed25519")]
 pub struct SoftwareEd25519Signature {
-    signature: ed25519_compact::Signature,
+    signature: ed25519_dalek::Signature,
 }
 
 #[cfg(feature = "ed25519")]
@@ -302,17 +301,18 @@ impl ec_signing::Signature for SoftwareEd25519Signature {
     #[cfg(test)]
     fn from_slice(bytes: &[u8; EC_SIGNATURE_SIZE]) -> Option<Self> {
         // Assumes EC_SIGNATURE_SIZE == 64
-        let signature = ed25519_compact::Signature::from_slice(bytes).ok()?;
+        let signature = ed25519_dalek::Signature::from_bytes(bytes);
         Some(SoftwareEd25519Signature { signature })
     }
 
     fn to_slice(&self, bytes: &mut [u8; EC_SIGNATURE_SIZE]) {
-        bytes.copy_from_slice(&self.signature[..]);
+        bytes.copy_from_slice(&self.signature.to_bytes());
     }
 
     fn to_der(&self) -> Vec<u8> {
         // The ECDSA implementation has implicit unwraps, too.
-        let signature_any = Any::new(der::Tag::OctetString, &self.signature[..]).unwrap();
+        let signature_bytes = self.signature.to_bytes();
+        let signature_any = Any::new(der::Tag::OctetString, signature_bytes).unwrap();
         signature_any.to_der().unwrap()
     }
 }
diff --git a/rust-toolchain.toml b/rust-toolchain.toml
index 711b7d7..fddec4c 100644
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2024-05-01"
+channel = "nightly-2024-07-01"
 components = ["clippy", "miri", "rustfmt", "rust-src"]
 targets = ["thumbv7em-none-eabi",
            "riscv32imac-unknown-none-elf",

```
  
</details>

Binary size increases by 30 kB regardless of crate or version choice. Therefore, using software Ed25519 is currently incompatible with upgradability.

Runtimes, measure with `crypto_bench`, are:

| Crate      | `random` | `public_key` | `sign` |
| ---------- | -------: | -----------: | -----: |
| compact v1 |     51.4 |          0.0 |   50.3 |
| compact v2 |     50.4 |          0.0 |   97.7 |
| dalek      |     56.7 |          0.0 |   56.1 |

Decision is to stick with compact v1. Signing is twice as slow in v2. Switching to dalek doesn't seem to have any advantages, and the current Rust toolchain is too old for dalek, so we would need a small jump in minor nightly version.

### DIFF
--- a/examples/crypto_bench.rs
+++ b/examples/crypto_bench.rs
@@ -29,7 +29,9 @@ use libtock_drivers::timer;
 use libtock_drivers::timer::{Timer, Timestamp};
 use libtock_runtime::{set_main, stack_size, TockSyscalls};
 use opensk::api::crypto::aes256::Aes256;
-use opensk::api::crypto::ec_signing::SecretKey as _;
+use opensk::api::crypto::ec_signing::EcSecretKey;
+#[cfg(feature = "ed25519")]
+use opensk::api::crypto::ec_signing::EdSecretKey;
 use opensk::api::crypto::sha256::Sha256;
 #[cfg(feature = "ed25519")]
 use opensk::env::Ed25519Sk;
--- a/examples/crypto_bench.rs
+++ b/examples/crypto_bench.rs
@@ -29,11 +29,13 @@ use libtock_drivers::timer;
 use libtock_drivers::timer::{Timer, Timestamp};
 use libtock_runtime::{set_main, stack_size, TockSyscalls};
 use opensk::api::crypto::aes256::Aes256;
-use opensk::api::crypto::ecdsa::SecretKey as _;
+use opensk::api::crypto::ec_signing::SecretKey as _;
 use opensk::api::crypto::sha256::Sha256;
+#[cfg(feature = "ed25519")]
+use opensk::env::Ed25519Sk;
 use opensk::env::{AesKey, EcdsaSk, Sha};
 
-stack_size! {0x2000}
+stack_size! {0x4000}
 set_main! {main}
 
 type Syscalls = TockSyscalls;
@@ -118,6 +120,26 @@ fn main() {
     bench(&mut console, &timer, "Ecdsa::SecretKey::sign", || {
         sk.sign(&[]);
     });
+
+    // Ed25519
+    #[cfg(feature = "ed25519")]
+    {
+        bench(&mut console, &timer, "Ed25519::SecretKey::random", || {
+            Ed25519Sk::<TockEnv<Syscalls>>::random(&mut rng);
+        });
+        let sk = Ed25519Sk::<TockEnv<Syscalls>>::random(&mut rng);
+        bench(
+            &mut console,
+            &timer,
+            "Ed25519::SecretKey::public_key",
+            || {
+                black_box(sk.public_key());
+            },
+        );
+        bench(&mut console, &timer, "Ed25519::SecretKey::sign", || {
+            sk.sign(&[]);
+        });
+    }
 
     writeln!(console, "****************************************").unwrap();
     writeln!(console, "All the benchmarks are done.\nHave a nice day!").unwrap();
--- a/libraries/opensk/src/api/crypto/ec_signing.rs
+++ b/libraries/opensk/src/api/crypto/ec_signing.rs
@@ -62,24 +62,12 @@ pub trait SecretKey: Sized {
 pub trait PublicKey: Sized {
     type Signature: Signature;
 
-    /// Creates a public key from its coordinates.
-    #[cfg(test)]
-    fn from_coordinates(x: &[u8; EC_FIELD_SIZE], y: &[u8; EC_FIELD_SIZE]) -> Option<Self>;
-
-    /// Verifies if the signature matches the message.
-    #[cfg(test)]
-    fn verify(&self, message: &[u8], signature: &Self::Signature) -> bool;
-
     /// Writes the public key coordinates into the passed in parameters.
     fn to_coordinates(&self, x: &mut [u8; EC_FIELD_SIZE], y: &mut [u8; EC_FIELD_SIZE]);
 }
 
 /// Elliptic curve signature.
 pub trait Signature: Sized {
-    /// Creates a signature from its affine coordinates, represented as concatenated bytes.
-    #[cfg(test)]
-    fn from_slice(bytes: &[u8; EC_SIGNATURE_SIZE]) -> Option<Self>;
-
     /// Writes the signature bytes into the passed in parameter.
     fn to_slice(&self, bytes: &mut [u8; EC_SIGNATURE_SIZE]);
 
--- a/libraries/opensk/src/api/crypto/ec_signing.rs
+++ b/libraries/opensk/src/api/crypto/ec_signing.rs
@@ -12,18 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::{EC_FIELD_SIZE, EC_SIGNATURE_SIZE, HASH_SIZE};
+use super::{EC_FIELD_SIZE, EC_SIGNATURE_SIZE};
 use crate::api::rng::Rng;
 use alloc::vec::Vec;
 
 /// Container for all ECDSA cryptographic material.
-pub trait Ecdsa {
+pub trait EcSigning {
     type SecretKey: SecretKey<PublicKey = Self::PublicKey, Signature = Self::Signature>;
     type PublicKey: PublicKey<Signature = Self::Signature>;
     type Signature: Signature;
 }
 
-/// ECDSA signing key.
+/// Elliptic curve signing key.
 pub trait SecretKey: Sized {
     type PublicKey: PublicKey;
     type Signature: Signature;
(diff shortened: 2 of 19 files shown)


kaczmarczyck:
@gemini-code-assist Retry the review, but note that your summary is already wrong. It assumes that a patch inside my PR description is actually part of the changes under review.

examples/crypto_bench.rs
```diff
@@ -29,11 +29,13 @@ use libtock_drivers::timer;
...
 use opensk::api::crypto::aes256::Aes256;
-use opensk::api::crypto::ecdsa::SecretKey as _;
+use opensk::api::crypto::ec_signing::SecretKey as _;
 use opensk::api::crypto::sha256::Sha256;
+#[cfg(feature = "ed25519")]
+use opensk::env::Ed25519Sk;
 use opensk::env::{AesKey, EcdsaSk, Sha};
 
-stack_size! {0x2000}
+stack_size! {0x4000}
```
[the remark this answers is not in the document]
kaczmarczyck: This is the same stack size as OpenSK's `src/main.rs`, so it isn't particularly large.
And indeed, the change was necessary to allow running ed25519-compact.

libraries/opensk/src/api/crypto/ec_signing.rs
ia0, line 1: The comment line 39 is only true for ECDSA P-256. For Ed25519 it's SHA512. So we can just remove that line.

libraries/opensk/src/api/crypto/mod.rs
```diff
@@ -55,7 +55,9 @@ pub const TRUNCATED_HMAC_SIZE: usize = 16;
 pub trait Crypto {
     type Aes256: Aes256;
     type Ecdh: Ecdh;
-    type Ecdsa: Ecdsa;
+    type Ecdsa: EcSigning;
+    #[cfg(feature = "ed25519")]
+    type Ed25519: EcSigning;
```
ia0: There's a slight problem with using the same trait, is that we can't implement "both" at the same time for the same type, i.e we can't use `Self` for both `Ecdsa` and `Ed25519`. Probably not an issue, but good to know if it was not intentional.

libraries/opensk/src/api/crypto/ec_signing.rs
```diff
@@ -58,32 +58,16 @@ pub trait SecretKey: Sized {
...
-    /// For hashing, SHA256 is used implicitly.
-    fn verify(&self, message: &[u8], signature: &Self::Signature) -> bool;
-
-    /// Verifies if the signature matches the hash of the message.
-    ///
-    /// Prehash is the SHA256 of the signed message.
-    fn verify_prehash(&self, prehash: &[u8; HASH_SIZE], signature: &Self::Signature) -> bool;
-
     /// Writes the public key coordinates into the passed in parameters.
     fn to_coordinates(&self, x: &mut [u8; EC_FIELD_SIZE], y: &mut [u8; EC_FIELD_SIZE]);
```
ia0, line 72: This is not correct for Ed25519. A public key is only 32 bytes.

ia0 (commented):
I didn't check where the public key for Ed25519 is used, but it should be different than ECDSA P-256, because it's only 32 bytes.

libraries/opensk/src/api/crypto/ec_signing.rs
kaczmarczyck: The traits are split now, but the documentation doesn't mention hashing details anymore in both cases.

libraries/opensk/src/api/crypto/mod.rs
kaczmarczyck: Wasn't a problem, but still fixed .

libraries/opensk/src/api/crypto/ec_signing.rs
kaczmarczyck: The only thing we do with public keys in the library is convert them to COSE. In `CoseKey`, we filled y with 0 in the case of Ed25519. This does not conform to the COSE RFC, and the field for y is now an `Option`.

That being said, I found other problems with our implementation that made the stack not work with Ed25519 only. Can be tested on [webauthn.io](https://webauthn.io/) for example using Advanced Settings.

kaczmarczyck:
Tested and working with [libfido2](https://github.com/Yubico/python-fido2/blob/main/examples/resident_key.py) (by disabling support for ES256 in `fido2/cose.py`) and with [webauthn.io](https://webauthn.io/).

ia0 (approved):

PR #745, opened 2025-07-31
Lenient subcommand parameters in BioEnrollment

kaczmarczyck:
The parameter is now optional when the map is empty. Only one subcommand can trigger this case, and it is now fixed.
Chrome triggered this behavior in its standard flow.


ia0 (approved):

ia0:
Tested, it works.

PR #746, opened 2025-08-04
Changes the Env API to late choice of template ID

kaczmarczyck:
Before, a fingerprint sensor had to choose a template ID at the start of the enrollment process. Not all sensors assign one immediately, and it conceptually makes more sense to expect an ID after a successful enrollment.

We introduce a state that tracks whether

- The fingerprint sensor chose a template ID already,
- and we sent a random template ID over CTAP until then.

We also add tests for the introduced state. Since the command is not considered stateful by CTAP, it means that enrollment can interleave with other CTAP commands. And the state is not part of `StatefulCommand`.

### DIFF
--- a/libraries/opensk/src/api/fingerprint.rs
+++ b/libraries/opensk/src/api/fingerprint.rs
@@ -72,10 +72,6 @@ pub trait Fingerprint {
     ) -> CtapResult<(Ctap2EnrollFeedback, usize, Option<Vec<u8>>)>;
 
     /// Cancel a fingerprint enrollment.
-    ///
-    /// Returns the newly assigned hardware template ID. This is not to be
-    /// confused with FIDO's template ID that the CTAP command returns when we
-    /// being enrollment.
     fn cancel_enrollment(&mut self) -> CtapResult<()>;
 
     /// Delete the fingerprint matching the given template ID.
--- a/libraries/opensk/src/api/fingerprint.rs
+++ b/libraries/opensk/src/api/fingerprint.rs
@@ -45,9 +45,7 @@ impl From<Ctap2EnrollFeedback> for cbor::Value {
 
 pub trait Fingerprint {
     /// Starts the fingerprint enrollment process.
-    ///
-    /// Returns the newly assigned template ID.
-    fn prepare_enrollment(&mut self) -> CtapResult<Vec<u8>>;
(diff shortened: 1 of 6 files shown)


libraries/opensk/src/env/test/mod.rs
```diff
@@ -196,6 +199,22 @@ impl TestEnv {
...
+        self.fingerprint().prepare_enrollment()?;
+        let mut template_id = None;
+        for _ in 0..self.fingerprint().max_capture_samples_required_for_enroll() {
+            let (_, remaining_samples, hardware_id) = self.fingerprint().capture_sample(None)?;
+            template_id = template_id.or(hardware_id);
+            if remaining_samples == 0 {
+                break;
+            }
+        }
+        let template_id = template_id.unwrap();
```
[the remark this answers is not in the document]
kaczmarczyck: Comment on the comment: I think the priority rating of this comment is wrong.

About the suggested change: This `unwrap` is different from the forwarded errors, in that it is not something that should ever happen in a test, so we want it to fail as early as possible. This `unwrap` indicates that not the code under test is wrong, but `TestEnv` itself.

In other case, the returned error code might be intentional and correct. For example, if `TestEnv` could simulate running out of fingerprint storage, this should throw `CTAP2_ERR_FP_DATABASE_FULL`.

libraries/opensk/src/api/fingerprint.rs
```diff
@@ -56,16 +54,28 @@ pub trait Fingerprint {
...
     fn capture_sample(
         &mut self,
-        template_id: &[u8],
         timeout_ms: Option<usize>,
-    ) -> CtapResult<(Ctap2EnrollFeedback, usize)>;
+    ) -> CtapResult<(Ctap2EnrollFeedback, usize, Option<Vec<u8>>)>;
 
     /// Cancel a fingerprint enrollment.
+    ///
+    /// Returns the newly assigned hardware template ID. This is not to be
```
ia0: copy paste? This whole paragraph seems out of context.

ia0 (dismissed):
It seems to work. There's an issue, but it's probably on my side (hardware side).

libraries/opensk/src/api/fingerprint.rs
kaczmarczyck: Indeed, removed.

ia0 (approved):

PR #749, opened 2025-08-05
Updates UP and UV requirements

kaczmarczyck:
Fixes how devices with fingerprint recognize the contributions of internal UV to grant UP and UV.


ia0 (approved):
This seems to work, but there's some weird thing going on in webauthn.io.

/gemini review

PR #755, opened 2025-10-20
Fixes the error code for blocked UV

kaczmarczyck:
First, with an answer on the specification question, we change the comment on the error code being a potential typo: https://github.com/fido-alliance/fido-2-specs/issues/1672

Also, I realized that the default was wrong before, so now the error code with `*_UV_*` is used for UV operations as intended.

### DIFF
--- a/libraries/opensk/src/ctap/mod.rs
+++ b/libraries/opensk/src/ctap/mod.rs
@@ -265,7 +265,7 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
 
 // For MakeCredential and GetAssertion, the specification says:
 // If the uvRetries counter is 0, return CTAP2_ERR_PIN_BLOCKED. See:
-// https://github.com/fido-alliance/fido-2-specs/issues/1672
+// https://github.com/fido-alliance/fido-2-specs/issues/1672 (private)
 #[cfg(feature = "fingerprint")]
 fn map_uv_block_error(result: CtapResult<()>) -> CtapResult<()> {
     result.map_err(|e| match e {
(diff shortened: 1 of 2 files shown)


libraries/opensk/src/ctap/mod.rs
```diff
@@ -263,6 +263,20 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
...
+// If the uvRetries counter is 0, return CTAP2_ERR_PIN_BLOCKED. See:
+// https://github.com/fido-alliance/fido-2-specs/issues/1672
+#[cfg(feature = "fingerprint")]
+fn map_uv_block_error(result: CtapResult<()>) -> CtapResult<()> {
+    result.map_err(|e| {
+        if e == Ctap2StatusCode::CTAP2_ERR_UV_BLOCKED {
+            Ctap2StatusCode::CTAP2_ERR_PIN_BLOCKED
+        } else {
+            e
+        }
```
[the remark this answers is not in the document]
kaczmarczyck: Fixed Gemini code after Gemini suggestion.

libraries/opensk/src/ctap/mod.rs
```diff
@@ -263,6 +263,17 @@ fn truncate_to_char_boundary(s: &str, mut max: usize) -> &str {
     }
 }
 
+// For MakeCredential and GetAssertion, the specification says:
+// If the uvRetries counter is 0, return CTAP2_ERR_PIN_BLOCKED. See:
+// https://github.com/fido-alliance/fido-2-specs/issues/1672
```
ia0: FYI: This repository seems private, but maybe it's still useful to keep the link.

ia0 (dismissed):

libraries/opensk/src/ctap/mod.rs
kaczmarczyck: I know it is private, the reference is still useful to me. Added a `(private)` warning. Also removed the now outdated occurances.

kaczmarczyck:
Markdown CI failure seems unrelated, and I don't touch markdown.

ia0 (approved):

ia0:
> Markdown CI failure seems unrelated, and I don't touch markdown.

I'm re-running it.

kaczmarczyck:
I also tried re-running, I think some necessary server is down?

ia0:
Seems like a problem of the docker image... then we should either bypass to merge (if we believe it's transient) or not make markdown required (or completely remove it).

kaczmarczyck:
I will merge for now, and retry later to check if transient.

kaczmarczyck:
Latest run worked.

PR #759, opened 2025-11-13
Fixes an integer overflow in a helper class

kaczmarczyck:
This was not exploitable as a vulnerability as far as I know. All inputs to the function are coming from configs, and are never user inputs.

The new unit test would have panicked without the fix.

### DIFF
--- a/src/env/tock/storage_helper.rs
+++ b/src/env/tock/storage_helper.rs
@@ -188,7 +188,7 @@ impl Partition {
         for range in &self.ranges {
             if offset < range.length() {
                 return if range.length() - offset >= length {
-                    Some(range.start() + offset)
+                    range.start().checked_add(offset)
                 } else {
                     None
                 };
@@ -377,5 +377,12 @@ mod tests {
         assert_eq!(&second_range, &all_ranges[1..]);
         let partial_range = partition.ranges_from(0x30000);
         assert_eq!(partial_range[0], ModRange::new(0x30000, 0x30000));
+    }
+
+    #[test]
+    fn partition_find_address_overflow() {
+        let mut partition = Partition::default();
+        partition.append(ModRange::new(usize::MAX - 100, 200));
+        assert_eq!(partition.find_address(150, 1), None);
     }
 }
--- a/src/env/tock/storage_helper.rs
+++ b/src/env/tock/storage_helper.rs
@@ -188,7 +188,7 @@ impl Partition {
         for range in &self.ranges {
             if offset < range.length() {
                 return if range.length() - offset >= length {
-                    range.start().checked_add(offset)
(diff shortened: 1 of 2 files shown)


src/env/tock/storage_helper.rs
```diff
@@ -378,4 +378,11 @@ mod tests {
         let partial_range = partition.ranges_from(0x30000);
         assert_eq!(partial_range[0], ModRange::new(0x30000, 0x30000));
     }
+
+    #[test]
+    fn partition_find_address_overflow() {
+        let mut partition = Partition::default();
+        partition.append(ModRange::new(usize::MAX - 100, 200));
+        assert_eq!(partition.find_address(150, 1), None);
```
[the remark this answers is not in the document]
kaczmarczyck: I think the name of the test describes the 3 line function well enough.

src/env/tock/storage_helper.rs
```diff
@@ -378,4 +378,12 @@ mod tests {
         let partial_range = partition.ranges_from(0x30000);
         assert_eq!(partial_range[0], ModRange::new(0x30000, 0x30000));
     }
+
+    #[test]
+    #[should_panic]
+    fn partition_find_address_overflow() {
```
ia0, line 384: I don't think this test is useful. We should instead fix `ModRange::new()` to panic if `start + length` overflows (or return an error but that would mean more changes).

kaczmarczyck: Some unit tests for `ModRange` check slices that overflow. I'd have to check if we rely on this behavior somewhere, or it makes something more robust. But the panic in `ModRange` would be a change in working existing behavior.

ia0: This PR is also changing the existing behavior in almost exactly the same way. If `start + length` overflow for a range in `find_address` then it will overflow the same way in `ModRange::new()`. Also the documentation of `ModRange` is clear that it represents `[start, start + length)` which is nonsensical in case of overflow.

kaczmarczyck: `ModRange` is used more widely than `Partition`. So we can fix partition and make it robust, without touching `ModRange`.

`ModRange::new` does not overflow, so I think you mean it should have an end that wouldn't overflow?
```
    pub fn new(start: usize, length: usize) -> ModRange {
        ModRange { start, length }
    }
```

I'm not sure why a `ModRange` that doesn't have an end within `usize` is not acceptable as is. Its public functions seem to not contradict this, e.g. `intersects_range`? Can't it represent `[start, start + length)` as the interval you know from Maths?

kaczmarczyck:
Discussed offline, not exploitable so not worth the change.

Issue #758, opened 2025-11-05
Full CTAPHID_CANCEL support


kaczmarczyck:
Until we implemented fingerprint supported, only commands that check for user presence had a long enough runtime to need keepalives and accept cancellation in practise. Our `Env` therefore had USB packets in its user presence API:
https://github.com/google/OpenSK/blob/016e81ca772b909fb5673f4a5851d02eb9173e9c/libraries/opensk/src/api/user_presence.rs#L42

Now that we start supporting fingerprints, waiting for the user here needs to be reactive the same way. There are two ways to implement this:

A) Add it the same way we did for user presence.
B) Make OpenSK's API more async, and the calls to various parts of the hardware more independent.

The first solution is a direct fix for the problem, and the second one a bigger refactoring of the whole repository. While more work, it has some advantages:

1.
We can correctly implement the CTAPHID_CANCEL not only to cancel commands that wait for user input, but also any that have a longer processing time. The specification requests this [1], our other commands never needed it because they were fast enough. Some software cryptography implementations or PQC might make it a necessity though.

2.
It is a requirement for fix #685. In general, this decision can have implications on what hardware we support.

3.
It makes our implementation cleaner, and it should be easier for users to implement `Env` in the future.

4.
It simplifies some other work that is outstanding:
- We can move the main loop inside the library. This can help with making OpenSK a crate.
- It makes it easier to multiplex between USB and NFC packets, if we get to support that fully in the future.

[1] https://fidoalliance.org/specs/fido-v2.2-ps-20250714/fido-client-to-authenticator-protocol-v2.2-ps-20250714.html#usb-hid-cancel

PR #768, opened 2026-04-24
Moves OpenSK to Wasefire

### LINKED ISSUE #649: Key not supported on ios for Apple Id 2fa
## Expected Behavior

I tried adding my OpenSK NRF52840 dongle key with Apple. Since the implementation is FIDO certified it should work with Apple ID 

## Actual Behavior

IMG_8544

1. Go to settings on ios device
2. Click on your apple account 
3. Tap Sign in & Security
4. Tap on two-factor authentication
5. Tap on security keys > Add new key 
6. Insert and activate OpenSK key 

## Specifications

- Version: CTAP 2.0 release
- Platform: ios 17.0

### LINKED ISSUE #685: `async` API for `Env`?
Hi! I'm working on [a project](https://kb2.antonok.com) using the [RTIC](https://rtic.rs/2/book/en/) framework on an STM32F4 microcontroller, and I'm interested in porting OpenSK to that architecture if at all possible.

RTIC leans heavily on Rust's `async` with interrupts for resource sharing and task scheduling. Right now it looks like the current definition of traits required by `Env` exclusively use blocking function calls. After a first review for my use-case it looks like `HidConnection::send_and_maybe_recv`, `AttestationStore::{get,set}`, `Storage::{read_slice,write_slice,erase_page}`, and `UserPresence` would all need to access some kind of shared I/O peripherals. Under RTIC, non-blocking implementations of these traits would be awkward at best (and maybe highly unsafe at worst).

Before I go down this route, I wanted to check in with the maintainers - is this something you'd be willing to merge if a PR was provided? Any other thoughts?

kaczmarczyck:
OpenSK is now based on Wasefire as the main supported OS instead of TockOS. See Future Work for the problems we want to solve. We also reduce maintenance cost since platforms are now only kept in Wasefire.

### Initial commits

0a2a34adb32a727a23fb24c29a73c857b3a90082
- Deletes all Tock related code.
- Uses uv instead of manual venv setup.
- Moves Wasefire env into src.
- Fixes errors and lints.
    
e167421427ac966b9fc716dc77e251ce9081bae4
- Adds Wasefire and its existing OpenSK Env
- Moves to the new Rust toolchain

a6158efc975a9e8709e7e18f83dd2e8fac0c8efb
- Updates documentation to Wasefire
- Adepts the Customization to untie batch attestation from CTAP1
- Prevents privacy issues by removing per device attestation
- For now, removes batch certificate injection
- The binary size workflow is disabled until fixed

### Planned coming PRs

- Check and reinstate MacOS support
- Fix the binary size workflow
- Rename features in library to match naming convention in Wasefire
- Add method to inject batch attestation key (functionality of old vendor command Configure)
- Remove legacy backwards compatibility code (since this is a breaking change anyway)

### Future work

Replace #767 with dependency bump
Move `rand_core` to 0.10.1, and other dependabot alerts. Requires some changes to the crypto implementations in Env.
We are not stuck with an old compiler version anymore. Before, we would have needed to keep up with TockOS versions.

Fix #649
Since we have to redo our batch attestation, I implemented it so that we fake batch attestation until then. This incidentally let's users login to Apple when they use `use_batch_attestation` to activate the random fake batch key. They expect batch attestation without actually checking it. It does mean that you see error messages on webauthn.io in that case.

Fix #685 and solve #758
We can have an async Env now, which was impossible with TockOS. Sending busy messages is easier in an async Env.

Fingerprint support
Experimental in Wasefire with one sensor, will eventually be tested more

NFC
The experimental code was TockOS based, so this will likely not happen soon. Would need to land in Wasefire first.


docs/install.md
```diff
@@ -4,159 +4,141 @@
...
+### OpenTitan
 
-The storage is backward compatible to non-upgradable boards. Deploying an
-upgradable board automatically installs the bootloader. Please keep in mind that
-you have to safely store your private signing key for upgrades if you want to
-use this feature. For more information on the cryptographic material, see
-[Customization](customization.md).
+The applet feature `ed25519` is supported and needs the `software-ed25519`
+platform feature. Once the OpenTitan crypto library implements Ed25519, it will
+be possible to enable the `ed25519` platform feature instead.
```
ia0:
```suggestion
The applet feature `ed25519` is supported and needs the `ed25519`
platform feature.
```

I forgot to update the opensk example in https://github.com/google/wasefire/pull/1070.

src/lib.rs
```diff
@@ -1,19 +1,67 @@
...
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#![cfg_attr(not(feature = "std"), no_std)]
+//! Example applet running OpenSK.
+//!
+//! See the [README] for more information.
+//!
+//! [README]: https://github.com/google/wasefire/blob/main/examples/rust/opensk/README.md
```
ia0: Do you want to keep this?

rust-toolchain.toml
```diff
@@ -1,6 +1,9 @@
...
+channel = "nightly-2025-11-07"
 components = ["clippy", "miri", "rustfmt", "rust-src"]
-targets = ["thumbv7em-none-eabi",
-           "riscv32imac-unknown-none-elf",
-           "riscv32imc-unknown-none-elf"]
+targets = [
+    "thumbv7em-none-eabi",
+    "wasm32-unknown-unknown",
+    "riscv32imac-unknown-none-elf",
+    "riscv32imc-unknown-none-elf",
```
ia0: Those are probably not needed.

Cargo.toml
```diff
@@ -1,72 +1,52 @@
...
-panic = "abort"
-lto = true # Link Time Optimization usually reduces size of binaries and static libraries
 
-[profile.release]
-panic = "abort"
-lto = true # Link Time Optimization usually reduces size of binaries and static libraries
-opt-level = "z"
-codegen-units = 1
+[lints]
+rust.dead_code = "allow"
```
ia0: Is this needed? We can probably just use `#[allow(dead_code)]` where it's needed in the code. That's a bit risky to have this globally.

ia0 (dismissed):

src/lib.rs
kaczmarczyck: No, would go stale, good catch.

rust-toolchain.toml
kaczmarczyck: Removed then.

Cargo.toml
kaczmarczyck: Yes, I forgot to do it properly after I "temporarily" disabled it. Moved to `src/lib.rs`.

ia0 (dismissed):

ia0 (approved):

PR #770, opened 2026-04-28
Fixes GitHub workflow problems

kaczmarczyck:
- The bloat workflow now uses the Wasefire mechanism.
- The CI workflow has a recognizable name.
- OSS-Fuzz in the CIFuzz workflow failed because it misses uv. The setup script now installs it.

### DIFF
--- a/.github/workflows/bloat_formatter.sh
+++ b/.github/workflows/bloat_formatter.sh
@@ -17,8 +17,9 @@ cd "$(dirname "$0")"
 
 # New output file is $1
 # Old output file is $2
-TMP=comment.md
-WARNING="Note: numbers above are a result of guesswork. They are not 100% correct and never will be."
+# Optional output file is $3
+TMP="${3:-bloat_comment.md}"
+WARNING="never will be"
 NEW_SIZE=$(cat "$1" | sed -nr 's/.*100.0% (.*)KiB .text.*/\1/p')
 OLD_SIZE=$(cat "$2" | sed -nr 's/.*100.0% (.*)KiB .text.*/\1/p')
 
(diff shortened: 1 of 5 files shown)


ia0 (approved):

.github/workflows/bloat_formatter.sh
```diff
@@ -17,8 +17,9 @@ cd "$(dirname "$0")"
 
 # New output file is $1
 # Old output file is $2
-TMP=comment.md
-WARNING="Note: numbers above are a result of guesswork. They are not 100% correct and never will be."
+# Optional output file is $3
+TMP="${3:-bloat_comment.md}"
+WARNING="never will be"
 NEW_SIZE=$(cat "$1" | sed -nr 's/.*100.0% (.*)KiB .text.*/\1/p')
 OLD_SIZE=$(cat "$2" | sed -nr 's/.*100.0% (.*)KiB .text.*/\1/p')
```
ia0, line 24: Actually it's probably better to only measure `opensk` instead of (or in addition to) everything.

It would be something like `s/.* ([0-9.]+)KiB opensk.*/\1/p`.

kaczmarczyck: I was considering it, but I wasn't sure what is more accurate. I have seen the crate numbers change after minor code changes, presumably because some function got inlined or otherwise attributed to a different crate.

Wasefire is a fixed target in all PRs that don't bump its commit. And the number we care about is likely the total, for hardware compatibility questions? We can change it to `opensk` later if we see the numbers are better also.

ia0: Yeah why not. Both make sense. Let's keep that and see if we need the other in practice.

kaczmarczyck: (Feel free to object, I'm merging for now and will handle the comment in a next PR if necessary.)

kaczmarczyck: Oh, race condition. Resolved then.

PR #777, opened 2026-05-27
Fix touch responsiveness and USB HID issues

pwnall:
This PR addresses two responsiveness issues that may cause the device to hang or drop connections:

1. **Touch State Leak**: Implemented `Drop` for `Touch` to correctly clear the `STATE` when a user presence request is interrupted (e.g., by a USB packet or timeout). Added safe extraction in `State::touch` to handle queued hardware button interrupts (bounces) without hitting `unreachable!()` and panicking.
2. **USB HID Communication Fixes**: 
    - Updated `MessageAssembler::parse_packet` to allow *any* valid initialization packet (not just `CTAPHID_INIT`) to abort a hung transaction, in accordance with the CTAP 2.1 specification.
    - Fixed a bug where a new initialization packet arriving immediately after a transaction timeout was incorrectly dropped while returning the `MsgTimeout` error.

These fixes prevent the firmware from entering unrecoverable states where the LED fails to blink or the device stops responding to the host.

### DIFF
--- a/libraries/opensk/src/ctap/hid/receive.rs
+++ b/libraries/opensk/src/ctap/hid/receive.rs
@@ -95,7 +95,9 @@ impl<E: Env> MessageAssembler<E> {
 
             // If the packet is from the timed-out channel, send back a timeout error.
             // Otherwise, proceed with processing the packet.
-            if cid == current_cid {
+            if cid == current_cid
+                && matches!(processed_packet, ProcessedPacket::ContinuationPacket { .. })
+            {
                 return Err((cid, CtapHidError::MsgTimeout));
             }
         }
@@ -595,6 +597,41 @@ mod test {
                 None
             ),
             Err(([0x12, 0x34, 0x56, 0x78], CtapHidError::MsgTimeout))
+        );
+    }
+
+    #[test]
+    fn test_timed_out_new_init() {
+        let mut env = TestEnv::default();
+        let mut assembler = MessageAssembler::default();
+        assert_eq!(
+            assembler.parse_packet(
+                &mut env,
+                &zero_extend(&[0x12, 0x34, 0x56, 0x78, 0x81, 0x00, 0x40]),
+                None,
+            ),
+            Ok(None)
+        );
+        env.clock().advance(TIMEOUT_DURATION_MS);
+        assert_eq!(
+            assembler.parse_packet(
+                &mut env,
+                &zero_extend(&[0x12, 0x34, 0x56, 0x78, 0x81, 0x00, 0x40]),
+                None
+            ),
+            Ok(None)
+        );
+        assert_eq!(
+            assembler.parse_packet(
+                &mut env,
+                &zero_extend(&[0x12, 0x34, 0x56, 0x78, 0x00]),
+                None
+            ),
+            Ok(Some(Message {
+                cid: [0x12, 0x34, 0x56, 0x78],
+                cmd: CtapHidCommand::Ping,
+                payload: vec![0x00; 0x40]
+            }))
         );
     }
 
--- a/src/touch.rs
+++ b/src/touch.rs
@@ -37,6 +37,15 @@ impl Touch {
     }
 }
 
+impl Drop for Touch {
+    fn drop(&mut self) {
+        if Arc::strong_count(&self.touched) == 2 {
+            // We're the last object, so we can stop listening.
+            *STATE.lock() = None;
+        }
+    }
+}
+
 static STATE: Mutex<Option<State>> = Mutex::new(None);
 
 struct State {


src/touch.rs
```diff
@@ -37,6 +37,12 @@ impl Touch {
     }
 }
 
+impl Drop for Touch {
+    fn drop(&mut self) {
+        *STATE.lock() = None;
```
ia0: Indeed, we should implement `Drop` for `Touch` to clear the `STATE` when needed. However this should be done only when this is the last `Touch` object being dropped.

```suggestion
        if Arc::strong_count(&self.touched) == 2 {
            // We're the last object, so we can stop listening.
            *STATE.lock() = None;
        }
```

src/touch.rs
```diff
@@ -67,11 +73,10 @@ impl State {
...
     fn touch(this: &mut Option<State>) {
-        let Some(state) = this.take() else {
-            unreachable!()
-        };
-        state.button.stop();
-        state.touched.store(true, Relaxed);
+        if let Some(state) = this.take() {
+            state.button.stop();
+            state.touched.store(true, Relaxed);
+        }
```
ia0: We should not reach `unreachable!()`. I wonder if you had to add this because you unconditionally reset the `STATE` in `Drop for Touch`. In theory, `state.button.stop()` will start dropping future events and clear any existing pending events. If that's not the case it's a bug in Wasefire.

That said, the applet could be resilient against Wasefire bugs. I don't think it's a good idea because (at least for development) we want to fail as soon as something impossible happens. We could have a `resilient` cargo feature to avoid crashes as much as possible, which would be used for production keys (there could be security concerns though, it's a trade-off with availability). @kaczmarczyck what do you think?

ia0 (commented):
Thanks for the PR! I'll only comment on the Touch issue and let @kaczmarczyck address the second point.

src/touch.rs
pwnall: Thank you very much for the better version!

ia0 (dismissed):
LGTM for `src/touch.rs`. Is there a reason the PR is still a draft?

src/touch.rs
pwnall: Thank you very much for explaining!

I removed this diff. I don't want to make development more difficult 😄

libraries/opensk/src/ctap/hid/receive.rs
```diff
@@ -95,7 +95,9 @@ impl<E: Env> MessageAssembler<E> {
 
             // If the packet is from the timed-out channel, send back a timeout error.
             // Otherwise, proceed with processing the packet.
-            if cid == current_cid {
+            if cid == current_cid
+                && matches!(processed_packet, ProcessedPacket::ContinuationPacket { .. })
```
kaczmarczyck, line 99: Thanks for reporting this bug in a very old part of the code. I wrote a test to check the new behavior, feel free to add it here, or I will later:

```
    #[test]
    fn test_timed_out_new_init() {
        let mut env = TestEnv::default();
        let mut assembler = MessageAssembler::default();
        assert_eq!(
            assembler.parse_packet(
                &mut env,
                &zero_extend(&[0x12, 0x34, 0x56, 0x78, 0x81, 0x00, 0x40]),
                None,
            ),
            Ok(None)
        );
        env.clock().advance(TIMEOUT_DURATION_MS);
        assert_eq!(
            assembler.parse_packet(
                &mut env, 
                &zero_extend(&[0x12, 0x34, 0x56, 0x78, 0x81, 0x00, 0x40]),
                None
            ),
            Ok(None)
        ); 
        assert_eq!(
            assembler.parse_packet(
                &mut env,
                &zero_extend(&[0x12, 0x34, 0x56, 0x78, 0x00]),
                None
            ),
            Ok(Some(Message {
                cid: [0x12, 0x34, 0x56, 0x78],
                cmd: CtapHidCommand::Ping,
                payload: vec![0x00; 0x40]
            }))
        );
    }
```

kaczmarczyck (dismissed):
Approving for HID library code

src/touch.rs
kaczmarczyck: We should double check how we deal with parallel connections, and that we don't miss anything else. We didn't test enough since the Wasefire refactoring.

Touches should never be used twice. CTAP2 library code should already ensure this, but the combination with CTAP1 can work in one direction: First requesting on CTAP1, then going into the CTAP2 UP loop. We should drop CTAP1 then. But doesn't need to be done here.

kaczmarczyck: And to answer the question, once it is impossible, we can fail when the impossible happens.

libraries/opensk/src/ctap/hid/receive.rs
pwnall: Done.

Thank you very much for showing me how to put together this test!

src/touch.rs
pwnall: Thank you very much for your thoughts!

I'll try to address this in a separate PR.

ia0: I don't think there's anything to do here in the short term:
1. We need to check whether OpenSK can create 2 Touch objects. If no, simplify the code, but I can do it. 
2. We need to run enough tests to make sure the unreachable is correct.

pwnall:
@ia0 @kaczmarczyck PTAL?

Apologies if I was supposed to leave the PR as-is rather than add the tests.

ia0 (approved):

ia0:
> Apologies if I was supposed to leave the PR as-is rather than add the tests.

No I think that's fine. I'll let @kaczmarczyck merge.

kaczmarczyck (approved):

PR #778, opened 2026-05-28
Fix U2F UP not timing out

ia0:
There was 2 issues preventing the U2F UP from timing out:
1. There was no timeout waiting for packets while U2F UP is on.
2. When timing out for packets the U2F UP logic was skipped.

The fixes are:
1. A timeout of 1 second is used when U2F UP is on.
2. We execute the rest of the loop body (except for packet processing).

This needs the `impl Drop for Touch` of #777 to actually fix the blinking.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,13 +36,15 @@ fn main() -> ! {
         update_state(&mut wink, opensk_ctap.should_wink(), || {
             blink::Blink::new_ms(100)
         });
-        let mut packet = Some([0; 64]);
-        let mut timeout = None;
-        timeout = timeout.or(wink.is_some().then_some(500));
+        let mut recv_packet = Some([0; 64]);
+        let mut timeout = false;
+        timeout |= wink.is_some();
         #[cfg(feature = "ctap1")]
-        (timeout = timeout.or(u2f.is_some().then_some(1000)));
-        match env::hid_connection::recv(packet.as_mut().unwrap(), timeout).unwrap() {
-            RecvStatus::Timeout => packet = None,
+        (timeout |= u2f.is_some());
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -33,32 +33,45 @@ fn main() -> ! {
...
             opensk_ctap.u2f_grant_user_presence();
         }
-        for packet in opensk_ctap.process_hid_packet(&packet, Transport::MainHid) {
-            opensk_ctap
-                .env()
-                .hid_connection()
-                .send(&packet, UsbEndpoint::MainHid)
-                .unwrap();
+        if let Some(packet) = packet {
+            for packet in opensk_ctap.process_hid_packet(&packet, Transport::MainHid) {
```
kaczmarczyck: Rename the inner packet please, to not reuse the name too often.

src/lib.rs
```diff
@@ -33,32 +33,45 @@ fn main() -> ! {
...
-        match env::hid_connection::recv(&mut packet, timeout).unwrap() {
-            RecvStatus::Timeout => continue,
+        update_state(&mut wink, opensk_ctap.should_wink(), || {
+            blink::Blink::new_ms(100)
+        });
+        let mut packet = Some([0; 64]);
+        let mut timeout = None;
+        timeout = timeout.or(wink.is_some().then_some(500));
+        #[cfg(feature = "ctap1")]
+        (timeout = timeout.or(u2f.is_some().then_some(1000)));
```
kaczmarczyck: I suggest the same timeout for both cases, i.e., make it 500.

kaczmarczyck (dismissed):

kaczmarczyck (approved):

PR #781, opened 2026-06-04
Reintroduces the batch attestation cert command

kaczmarczyck:
We can call `tools/configure.py` again to set the batch attestation private key and certificate.
Also the AAGUID can now be passed in, instead of being part of customization. One only really makes sense with the other anyway.

Documentation is also update to reflect the state of batch attestation support: Fake if not configured, and real if configured. And the manual is written in a way that the privacy warning is hard to miss for anyone who runs the configure command.
Before random batch attestation, we had to send people towards the configure command more aggressively, since U2F only worked after configure was used. Now configure is actively discouraged since U2F just works without it. Only use it if you really want to build your own hardware security key and register it with FIDO.


docs/customization.md
```diff
@@ -27,26 +27,39 @@ If you delete the key material, it will be randomly regenerated by `setup.sh`.
...
 100,000 security keys of the same model. If you build your own OpenSK, your
 private key is unique to you. This makes you identifiable across registrations:
 Two websites could collaborate to track if registrations were attested with the
 same key material. If you use OpenSK beyond experimentation, please consider
-carefully if you want to take this privacy risk.
+carefully if you want to take the privacy risk of using the configure tool.
+
+The default randomly generated, ephemeral batch attestation keys, are helpful in
+practise: Without a key, U2F does not work. Also a few more relying parties
+accept OpenSK responses because of this trick. It is not meant to proof any
```
ia0: typo?
```suggestion
accept OpenSK responses because of this trick. It is not meant to prove any
```

ia0 (dismissed):

ia0 (approved):

